PR #13, opened 2018-08-04
issue_3: read from BQ for the file access pattern

zqian:


dashboard/cron.py
```diff
@@ -90,22 +96,62 @@ def update_with_udw_file(request):
...
+                    if ("enriched_events" == table.table_id):
+                        logger.debug('\t{}'.format("found table"))
+
+                        # query to retrieve all file access events for one course
+                        query = 'select CAST(SUBSTR(JSON_EXTRACT_SCALAR(event, "$.object.id"), 35) AS STRING) AS FILE_ID, ' \
+                                'SUBSTR(JSON_EXTRACT_SCALAR(event, "$.membership.member.id"), 29) AS USER_ID, ' \
+                                'EVENT_TIME as ACCESS_TIME, ' \
+                                'SUBSTR(JSON_EXTRACT_SCALAR(event, "$.group.id"),31) AS COURSE_ID ' \
+                                'FROM learning_datasets.enriched_events ' \
+                                'where JSON_EXTRACT_SCALAR(event, "$.edApp.id") = \'http://umich.instructure.com/\' ' \
```
pushyamig: We are querying off of `http://umich.instructure.com` this edApp but I will make sure they aren't other edApp that we are not missing data from. I know that caliper Live events have machine generated events that we would like to ignore, 
How about this `JSON_EXTRACT_SCALAR(event, "$.edApp.id") != http://cluster32-files.instructure.com/`?

Dockerfile
```diff
@@ -17,6 +17,8 @@ RUN pip install mysql-connector mysqlclient sqlalchemy psycopg2-binary
 
 RUN pip install canvasapi pandas
 
+RUN pip install google-cloud-bigquery
+
 RUN pip install PyMySQL
```
pushyamig: All the install should be happening from `requirments.txt` and we should only be doing all the installs with one command `RUN pip install -r requirements.txt`

dashboard/cron.py
zqian: We are still waiting for the official Canvas live event document. Before that, we are not sure what is involved with the URL "http://cluster32-files.instructure.com". 

We are only dealing with 'http://umich.instructure.com/' edApp right now.

Dockerfile
zqian: fixed!

Issue #17, opened 2018-08-09
integrate student-dashboard tool with shibboleth


zqian:
Please refer to the following documentations:

- https://docs.google.com/document/d/1QJfHBBO-gV5cHJRMUAr0QJNKF5qaznkO9AHqXMQooTo/edit

-https://github.com/knaperek/djangosaml2

PR #18, opened 2018-08-09
Issue 17 - Fixup shibboleth support

jonespm:
Fix for #17


.gitignore
```diff
@@ -1,3 +1,5 @@
 # exclude dashboard/__pycache__ folder
 dashboard/__pycache__/*
 .env
+saml
```
zqian: Add instructions inside README file for creating saml folder with student-dashboard-saml.key file?

jonespm: I put in a request to get the authorization, after that's tested I'll document all of this and use the same pattern you did for the other secrets file.

zqian: I am still in the process of getting the existing json file uploaded as secret file and successfully build in OpenShift.

jonespm: Okay, if you're working on that I'll wait for it!

Probably would need a private git repo somewhere with our secrets in it that openshift can checkout and mount? I'm modifying the README as if

/secrets/saml/ (Directory)
and 
/secrets/bq_cred.json (File)

Are available on the created image on start.sh and can be put in the appropriate place(s).

PR #22, opened 2018-08-10
issue#16 creating/loading data associated with assignment requirment …

pushyamig:
This PR is addressing the issue https://github.com/tl-its-umich-edu/student-dashboard-django/issues/16.

Created mainly 4 tables for this requirement.


dashboard/cron.py
```diff
@@ -196,6 +272,11 @@ def util_function(sql_string, mysql_table):
         dbname=UDW_DATABASE)
 
     df = pd.read_sql(sql_string, udw_conn)
+    # Sql returns boolean value so grouping course info along with it so that this could be stored in the DB table.
```
zqian: not quite sure about the comment line "Sql returns boolean value".

pushyamig: When the `weight_consideration` function run it will return a boolean value (i.e if course consider weights for grade calculation or not). The 3 lines below after that comment is adding course name associated with the boolean return. That's what the comment mean. I couldn't figure out an elegant sql solution to combine boolean value with course_id so I took the help of `dataframes` in doing that and adding it to the stu_dash tables

PR #28, opened 2018-08-14
WIP - Switch build to alpine, upgrade python from 3.5 to 3.7

jonespm:


jonespm:
I was thinking of switching this image over from python:3.5 to the faster alpine build of python 3.7.

Just didn't know if this wasn't done because nobody did it yet or some other reason. ;)

At the moment I'm using [amancevice/pandas:0.23.0-python3-alpine](https://github.com/amancevice/pandas)  which is a docker image with pandas and numpy pre-built because that seems to take about 10 minutes to build.

I was thinking a further improvement would be to fork this image, and rebuild our own with all of the dependencies we use and distribute that. This would probably save a lot of time on the rebuild/deploy process. 

This seems to come up fine as is, but I didn't fully test everything yet.

jonespm:
This is working a little better. Is cutting the image size down from 1.4GB to 900MB and starting up faster. I think once we package this into our own docker this will be really nice and even better.

dashboard/cron.py
```diff
@@ -284,4 +282,4 @@ def util_function(sql_string, mysql_table, table_identifier=None):
     df.drop_duplicates(keep=False, inplace=True)
 
     # write to MySQL
-    df.to_sql(con=engine, name=mysql_table, if_exists='append', index=False)
\ No newline at end of file
+    df.to_sql(con=engine, name=mysql_table, if_exists='append', index=False)
```
zqian: There is no change made to this line, right?

jonespm: Doesn't look like it but that reminds me I need to readd psycopg2 as you were using it. Also I need to either get this image or a fork of this image into the openshift containers before committing this so this is still WIP. I think I'll fork this project so we're not reliant on another users project. The big reason I'm using this instead of a more official image was because it has to build pandas from source which is slow, we can pre-compile all of that in an image and put that up!

jonespm:
I fixed the "non change" with the eol (vim automatically insert eol if the file doesn't have it) and also squashed the commits. Still this is a WIP but closer.

jonespm:
Thanks for the reviews. I'm still wanting to have our own docker image rather than use this before merging, maybe can get it in by tomorrow!

jonespm:
This was a good idea but is out of date and should be done differently. Closing for now to re-do this.

PR #31, opened 2018-08-14
Adding specific names to containers and updating README closes #30

### LINKED ISSUE #30: Give docker containers names to make it easier to reference them in the docs
The images containers aren't getting names that make it easy to reference tasks needed to accomplish in the README. It looks like we can add a name to these to make it easier for developers and others to work with.

jonespm:


jonespm:
This will require updating the env locally for
```
- MYSQL_HOST=student-dashboard-django_mysql_1
+ MYSQL_HOST=student_dashboard_mysql
```

I don't know what impact this will have for Openshift. It looks like no impact and this will really just improve local development.

pushyamig:
It should have no impact as by looking at the Openshift Environmental variable for `MYSQL_HOST` the hostname is different, which is miserver SQL version `dashboard-mysql.miserver.it.umich.edu`

jonespm:
Thanks for the confirmation. It will break your local .env for MySQL but I'm seeing a lot of better advantages of doing this like being able to easily copy files into the container and get the gunicorn to re-read the files to slightly speed up some tests. There probably will be some others.

jonespm:
I'm going to merge this. Remember to update your env MYSQL variable before rebuilding! ;)

PR #36, opened 2018-08-15
issue_2: updated the policy for removing duplicates from dataframe

zqian:
when remove duplicate record from dataframe, keep the first record.

Also added several log message for showing number of rows inserted.


zqian:
easy changes

PR #38, opened 2018-08-15
Fix to configure root logger, fixes #37

### LINKED ISSUE #37: Need to add configuration for root logger
It seems like a log of loggers fall under the root logger and are debug and not configurable, this should be configurable.

jonespm:


jonespm:
I didn't see anyway to do this directly and needed more logging in a few classes.

jonespm:
@zqian I think these loggers will all be revisited later before production but I don't see anyway other than modifying the code to see some of the messages in the views.py?

PR #42, opened 2018-08-15
Move all of the testloader files to a single page so it's easier to l…

### LINKED ISSUE #41: Make it easier to load up sample data
We have to go to a bunch of URL's over and over to get the sample data in dev, it should be pretty straightforward to have a simpler process for now. (That can be iterated and improved later)

jonespm:
…oad the

test data. Fixes #41

### DIFF
--- a/dashboard/urls.py
+++ b/dashboard/urls.py
@@ -16,6 +16,8 @@ Including another URLconf
 from django.conf.urls import url
 from django.contrib import admin
 from django.views.static import serve
+from django.views.generic.base import TemplateView
+
 from django.conf import settings
 from django.conf.urls import include
 from django.conf.urls.static import static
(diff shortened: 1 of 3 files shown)


jonespm:
Just looks like this for now. Just a quick hack that can be quickly run after re building the server/db.

image

dashboard/urls.py
```diff
@@ -45,7 +47,9 @@
     url(r'^update_groups', cron.update_groups, name='update_groups'),
     url(r'^submission', cron.submission, name='submission'),
     url(r'^weight_consideration', cron.weight_consideration, name='weight_consideration'),
-
+    url(r'^testloader', TemplateView.as_view(
```
zqian: This is good. I need to move other templates here following the TemplateView format. Nice work!

jonespm: Yeah this pattern seemed pretty easy to just display templates, we can re-organize the other templates on a future PR.

jonespm: I think there's some other refactoring that would be nice to do too, like breaking up views.py a bit before it gets too large.

PR #43, opened 2018-08-15
Fixes#19 Grade distribution view

pushyamig:
This PR is addressing the [issues](https://github.com/tl-its-umich-edu/student-dashboard-django/issues/19). The sample data/image of the viz is checked in for easier reviewing purposes.


dashboard/views.py
```diff
@@ -110,6 +114,52 @@ def file_access(request):
 
     return HttpResponse(df.to_json(orient='records'))
 
+
+def grade_distribution(request):
+
+    # Later this could be coming from a table specific to the course
+    bins = [0, 50, 65, 78, 89, 100]
```
jonespm: That's a good idea, I'm sure each course is going to want these to be configurable. This will probably have to pull from Canvas as we incorporate more of the canvas API's? 

https://canvas.instructure.com/doc/api/grading_standards.html#method.grading_standards_api.context_index

dashboard/static/js/d3.tip.v0.6.3.js
```diff
@@ -0,0 +1,280 @@
+// d3.tip
```
jonespm: Can we pull this from npm ? https://www.npmjs.com/package/d3-tip 

Maybe need the bower -> npm patch that I have like 90% done ;)

dashboard/static/css/grades.css
```diff
@@ -0,0 +1,119 @@
...
+    fill: none;
+    stroke: #000;
+    shape-rendering: crispEdges;
+}
+
+.filter {
+    color: #fff
+}
+
+.B {
```
jonespm: Do we have any CSS naming rules? Maybe this . should be `grade--B`based on BEM rules?  We also should probably get some SASS library here here to make these easier to configure in the future. Maybe another ticket.

dashboard/views.py
pushyamig: This needs some research, I don't know if courses have the grading schema published in canvas or not. Some might have it in canvas but not sure if they totally follow while grading

dashboard/static/js/d3.tip.v0.6.3.js
pushyamig: Yes... I agree Once you have it running I will add to `package.json`

PR #54, opened 2018-08-21
Add support for protecting URL's being authentication fixes #34

### LINKED ISSUE #34: put student_dashboard URLs behind "/accounts/login" and retrieve current user id via REMOTE_USER
Now that http://dashboard.tl.it.umich.edu/accounts/login is integrated with Shibboleth, need to make other dashboard urls behind this login path, so that login is always enforced.

Once the user logs in, the environment variable "REMOTE_USER" should be populated with the student uniqname info. Need to use that value and for subsequent data visulizations

jonespm:


dashboard/urls.py
```diff
@@ -15,50 +15,61 @@
 """
 from django.conf.urls import url
 from django.contrib import admin
+
+from .common.util import *
+
+if getenv_bool('STUDENT_DASHBOARD_SAML', 'false'):
+    from django.contrib.auth.decorators import login_required
```
zqian: need a bit of doc here -- not sure about the logic...
if "STUDENT_DASHBOARD_SAML" is true, login_required is imported, right?

dashboard/urls.py
```diff
@@ -69,6 +80,8 @@
 if 'djangosaml2' in settings.INSTALLED_APPS:
     urlpatterns += (
         url(r'^accounts/', include('djangosaml2.urls')),
+        url(r'^samltest/', echo_attributes),
```
zqian: login_required for "samltest/"?

dashboard/settings.py
```diff
@@ -190,80 +191,84 @@
...
+    SAML_CONFIG = {
+        'xmlsec_binary': '/usr/bin/xmlsec1',
+        'entityid': '%smetadata/' % SAML2_URL_BASE,
+
+        # directory with attribute mapping
+        # 'attribute_map_dir': path.join(BASEDIR, 'attribute-maps'),
+        'name': 'Student Dashboard',
+        # this block states what services we provide
+        'service': {
+            # we are just a lonely SP
```
zqian: -:)

dashboard/urls.py
jonespm: I can double check, I don't think it will do anything if STUDENT_DASHBOARD_SAML == false because the settings for saml aren't setup in settings.py. This is one path I want to test. We really could probably remove this entirely after we see one time what values it provides on an actual instance.

dashboard/settings.py
jonespm: Ya, I just intended this, these comments were already there ;)

jonespm:
Ok I pushed all these changes. @pushyamig and you'll have to set 

STUDENT_DASHBOARD_SAML=false

Now in your env or you'll be blocked from using this locally.

PR #56, opened 2018-08-22
Updates d3-tip (and a few paths to other js libs) fixes #55

### LINKED ISSUE #55: Update d3-tip and obtain from npm
d3-tip is in NPM with a newer version, this should be used rather than using the local version

jonespm:


dashboard/templates/small_multiples_files_bar_chart.html
```diff
@@ -9,11 +9,10 @@
     <script src="{% static 'underscore/underscore-min.js' %}"></script>
```
pushyamig: This file can be removed from the repo as this is used for demo purposes

pushyamig:
Thanks for cleaning this up!

dashboard/templates/small_multiples_files_bar_chart.html
jonespm: Yeah that one and file_access_within_week.html used javascript resources from the web rather than locally, I switched that and this to use the local js files and seems like all the pages and tooltips worked fine.

jonespm: I wasn't able to test small_multiples as it didn't work anymore, it gives `GET http://localhost:5001/static/json/top_file_access_week_grade2.tsv 404 (Not Found)` I was going to ask you about that but am happy to just remove!

pushyamig:
The purpose of the library is for displying tool tip on the bars, please make sure one of the charts have this working after the version source update

jonespm:
Yup, and none of the pages (other than small_mutiples mentioned above) have any console errors.

screen shot 2018-08-22 at 4 09 07 pm

PR #62, opened 2018-09-04
issue_61 added the brush view of file access

zqian:


dashboard/static/css/file_access_within_week.css
```diff
@@ -1,5 +1,48 @@
...
+    font-size: 12px;
+    max-width: 620px;
+    margin: 0 auto;
+    padding-top: 10px;
+    color: #ababab;
+    font-weight: 300;
+}
+
+.y.axis line {
+    fill: none;
```
pushyamig: I think CSS naming convention is having dashes(-) `.y.axis` is confusing as one can think it is referring to css 2 class name, how about `.y-axis` and `.x-axis`?

dashboard/static/css/file_access_within_week.css
```diff
@@ -1,5 +1,48 @@
...
+    display: inline !important; /* show when empty */
+    fill: #7A7A7A;
+    fill-opacity: 1;
+    stroke: #7A7A7A;
+    stroke-width: 2px;
+}
+
+.bar {
+    /*shape-rendering: crispEdges;*/
 }
```
pushyamig: delete it not using

dashboard/views.py
```diff
@@ -150,6 +151,10 @@ def file_access_within_week(request):
     logger.info("start time=" + startTimeString + " end_time=" + endTimeString);
     df = pd.read_sql(sqlString, conn, params={"start_time": startTimeString,"end_time": endTimeString})
 
+    # return if there is no data during this interval
+    if (df.empty):
+        return HttpResponse("no data");
```
pushyamig: I don't know if this is being grabbed and displayed as part of the UI, if it is then having the response in the JSON format would make it consistent

PR #65, opened 2018-09-06
Lockdown versions and fix a few codacy deploy issues fixes #64

### LINKED ISSUE #64: Lockdown versions in requirements, fix a few codacy issues with deploy
An issue saml2 in the deploy today caused the server to not work. So this is going to lock down the requirements a little more and push back the saml2 version one version back so it will work without any local hacks.

jonespm:


Dockerfile
```diff
@@ -5,19 +5,20 @@ RUN pip install --upgrade pip
 COPY requirements.txt /requirements.txt
 RUN pip install -r /requirements.txt
 #FROM python:2-onbuild
-RUN apt-get install curl
+RUN apt-get install curl=7.52.1-5+deb9u6 --no-install-recommends
```
pushyamig: isn't this needs to be like `apt-get update && apt-get install curl `

pushyamig:
Looks Good!

Dockerfile
jonespm: Usually just need to do an update if the repositories change and to look for newer versions. 

By not running an update here the version really isn't needed at all. 

But later in the script I do update below to pickup the new yarn sources which will update the other packages there. 

This should probably be all re-organized, I'll update that all in the other PR when we switch from apt to apk.

PR #70, opened 2018-09-07
Update some settings for production rollout

jonespm:


jonespm:
Note, there is a default DJANGO_SECRET_KEY there only because collectstatic failed without it. This also means it will startup for you locally. You can set this in your .env if you'd like.

README.md
```diff
@@ -5,8 +5,16 @@ There is some environment configuration and addtional files needed for this app
```
pushyamig: I feel instead of having the `.env` the configuration in README create a file`sample.env` default to Local setting

dashboard/settings.py
```diff
@@ -31,13 +31,13 @@
...
+SECRET_KEY = os.environ.get('DJANGO_SECRET_KEY','GENERATE_NEW_KEY')
 
 # SECURITY WARNING: don't run with debug turned on in production!
-DEBUG = True
-TEMPLATE_DEBUG = DEBUG
+DEBUG = os.environ.get('DJANGO_DEBUG',True)
+TEMPLATE_DEBUG = os.environ.get('DJANGO_TEMPLATE_DEBUG',DEBUG)
 
-ALLOWED_HOSTS = ["127.0.0.1", "localhost","dashboard.tl.it.umich.edu"]
+ALLOWED_HOSTS = ["127.0.0.1", "localhost","dashboard.tl.it.umich.edu","dev.dashboard.tl.it.umich.edu"]
```
pushyamig: can this be set as environmental variable as well? so that we don't need to list them all here

jonespm: Cool, I added this as a csv to the config file.

README.md
jonespm: Great, I created .env.sample instead of this

jonespm:
Okay I removed the SECRET_KEY entirely but added it to the dockerfile so it builds.

It won't startup unless you generate and set a secret key locally

jonespm:
It might be safter to use this python decouple module to improve the config https://pypi.org/project/python-decouple/ maybe as an issue.

README.md
zqian: +1!

Dockerfile
```diff
@@ -39,7 +39,7 @@ COPY manage.py /manage.py
 
 COPY data/* /data/
 
-RUN echo yes | python manage.py collectstatic
+RUN echo yes | DJANGO_SECRET_KEY="collectstatic" python manage.py collectstatic
```
zqian: So user need to update the key value inside the docker file? 

Is this step documented inside the README file?

jonespm: This is in there so the python collectstatic runs, it needs something set in the environment.

When you start up the docker file it will read from the .env file.

That's in the README (or the new sample env)

PR #73, opened 2018-09-10
issue_72 changes according to feedbacks

zqian:


dashboard/templates/view_file_access_within_week.html
```diff
@@ -17,22 +17,25 @@
...
+
 
     <input type="text" id="slider" class="slider">
 
 
     <input type="hidden" name="slider_num" id="slider_num" />
     <div/>
-    Please choose Grade level:
+
+    Top fils access by
```
jonespm: fils is probably a misspelling

dashboard/views.py
```diff
@@ -39,13 +39,17 @@
 # Todo the framework needs to remember the course
 CANVAS_COURSE_ID =os.environ.get('CANVAS_COURSE_IDS', '')
 UDW_ID_PREFIX = "17700000000"
+
+# TODO: need to put this into env.sample file
```
jonespm: TODO probably should do this ;)

dashboard/templates/view_file_access_within_week.html
zqian: Indeed!

dashboard/views.py
zqian: Done. Added to .env.sample

Issue #71, opened 2018-09-07
Switch config over to decouple


jonespm:
This decouple library looks pretty good for having a better config parsing pattern. I'd consider using it.

https://pypi.org/project/python-decouple/

PR #75, opened 2018-09-11
Switching over all templates to use a base template. Fixes #74

### LINKED ISSUE #74: Add headers/footers, reorganize templates
Before we continue to add too many templates, we should reorganize them to be in a structure inheriting off a base. I have a WIP in for this which hopefully won't break too many existing templates.

I was thinking of just doing a few blocks

title (to define the page title)
css (to add additional css)
js (to add additional js) 
content (for the body of the page)

A lot of it is redundant and will just go into the template.

This will allow us to have headers/footers and start on some type of structure.

jonespm:
This is going to slightly break all of the templates. Feedback appreciated.

I'll wait to merge this as there's a few things that weren't working that I couldn't test as my local data was wrong.

I put ng-app="visApp" in the base.html, does this need to be configurable?
I left out the configrabably of ng-controller in the body

These could probably be pushed up with the {% with %}

Also this includes displaying the user name in the header which is untested locally but "should" work. ;)

### DIFF
--- a/dashboard/templates/base.html
+++ b/dashboard/templates/base.html
@@ -10,7 +10,7 @@
     <link rel="stylesheet" href="{% static 'bootstrap/dist/css/bootstrap.css' %}">
     {% block css %} {% endblock %}
 </head>
-<body class="container-fluid" ng-app="visApp">
+<body class="container-fluid">
     <header>Student Dashboard :: 
         {% if user.is_authenticated %}Welcome, {{ user.get_username }}. You can <a href="/accounts/logout">logout here</a>.
         {% else %}Welcome! Please <a href="/accounts/login">login</a>


dashboard/templates/assignments.html
```diff
@@ -94,13 +87,11 @@ <h4>Assignments Due By Date</h4>
 </div>
```
pushyamig: `<div ng-app="visApp" class="container" ng-controller="mainController">` we can add the `ng-app` directive to the `div` tag itself.

dashboard/templates/files.html
```diff
@@ -1,23 +1,19 @@
+{% extends 'base.html' %}
```
pushyamig: we can delete this file as not needed

jonespm: Okay great, I'll delete it in a separate cleanup. @zqian are there other files that can be deleted?

zqian: Pushyami is right that this file is OK to be removed. But please do not delete others!

dashboard/templates/base.html
```diff
@@ -0,0 +1,42 @@
...
+        {% else %}Welcome! Please <a href="/accounts/login">login</a>
+        {% endif %}
+        <hr>
+    </header>
+    {% block content %}
+    {% endblock %}
+    <footer><hr>Copyright 2018 University of Michigan ITS Teaching and Learning</footer>
+</body>
+<!-- Common Javascript -->
+<script async src="https://www.googletagmanager.com/gtag/js?id=UA-124181744-1"></script>
```
zqian: I tried to get the GA track id out to configuration, but was not successful. 

Putting it into the single file make a progress forward, at least.

jonespm: Yeah! It should be picked up by all the pages anyway. I think I can move this out now, or maybe after I fix #71 . That is up next!

PR #77, opened 2018-09-11
Switch build to alpine and 3.7. Pre-build base image. Fixes #76

### LINKED ISSUE #76: Update python from 3.5 to 3.6 research new bases
I'm planning on switching this over to alpine. I had a WIP PR on #28 but that needs to be re-worked now that I know more about this process. I'm going to create a Dockerfile where we can store our "Base image" then push that to both docker hub and openshift. Then we can use that in both places.

jonespm:


jonespm:
I pushed the base image here https://hub.docker.com/r/tlitsumichedu/student-dashboard-django-base/

I'll push it to the open shift repo then we should be all set with this. Looks good locally.

jonespm:
Okay this is all in openshift and seems ready to roll!

requirements-base.txt
```diff
@@ -0,0 +1,30 @@
+gunicorn>=19.9.0,<19.9.99
+
+# Not sure if Django 2 will work yet, hold back to LTS
+Django>=1.11,<1.11.99
+whitenoise>=3.3.1,<3.3.99
+
+django-nvd3>=0.9.7,<0.9.99
```
zqian: we might drop this too. I am no longer using nvd3, just d3 directly. @pushyamig how about you?

jonespm:
Ops wanted to talk to me about this one, I'll merge it this afternoon after that meeting.

jonespm:
Okay I'm closing this one again, third time will be a charm! ;)

Issue #78, opened 2018-09-12
Assignment progress view


pushyamig:
Last bit remaining is the having the markers for showing Current grade and Max Possible
Progress bar making it support mobile when resizing

pushyamig:
Close

PR #82, opened 2018-09-12
#78 assignment progress view

pushyamig:
This issue addresses the issue #78 , this issue still work in progress last bit remaining is adding marker denoting current and Max possible grade

### DIFF
--- a/dashboard/templates/assignments.html
+++ b/dashboard/templates/assignments.html
@@ -14,7 +14,7 @@
 
 <div class="container-fluid">
     <h5 style="padding-left: 8px">Progress Bar</h5>
-    <svg id="chart"></svg>
+    <div id="chart"></div>
 </div>
 
 <div class="container" ng-controller="mainController" ng-app="visApp">
@@ -93,7 +93,6 @@
 </div>
 
 </div>
-<div class ="error"></div>
 {% endblock %}
 
 {% block js %}
--- a/dashboard/templates/assignments.html
+++ b/dashboard/templates/assignments.html
@@ -12,13 +12,19 @@
     <link rel="stylesheet" href="/static/css/assign.css">
 </head>
 
-<body ng-app="visApp">
+<body>
 <div class="container-fluid">
     <div class="navbar-header mastHead">
         <h1 class="navbar-brand collapse-nav"><span>Student Dashboard</span></h1>
     </div>
 </div>
-<div class="container" ng-controller="mainController">
+
+<div class="container-fluid">
+    <h5 style="padding-left: 8px">Progress Bar</h5>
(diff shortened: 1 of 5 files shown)


dashboard/templates/assignments.html
```diff
@@ -87,11 +93,10 @@ <h4>Assignments Due By Date</h4>
...
 </div>
+<div class ="error"></div>
 {% endblock %}
 
 {% block js %}
-<script src="{% static 'underscore/underscore-min.js' %}"></script>
-<script src="/static/js/app.js"></script>
-<script src="/static/js/controllers.js"></script>
-<script src="/static/js/factories.js"></script>
+<script src="{% static 'd3-v4/build/d3.min.js' %}"></script>
```
jonespm: You said you wanted to use v3? Is v4 okay?

pushyamig: SOrry for the confusion, I need v4!

dashboard/views.py
```diff
@@ -259,10 +259,39 @@ def grade_distribution(request):
...
+def assignment_progress(request):
+    logger.info(assignment_view.__name__)
+    sql = "select assignment_id,local_graded_date as graded_date,score,name,assign_grp_name,local_date as due_date,points_possible,group_points,weight from (" \
+          "(select assignment_id,local_graded_date,score from" \
+          "(select id from USER where sis_name = %(current_user)s ) as u join" \
+          "(select user_id,assignment_id,local_graded_date,score from SUBMISSION) as sub on sub.user_id=u.id) as rock join" \
+          "(select assign_id,name,assign_grp_name,local_date,points_possible,group_points,weight from" \
+          "(select id as assign_id,assignment_group_id, local_date,name,points_possible from ASSIGNMENT where course_id = %(course_id)s) as a join" \
+          "(select id,name as assign_grp_name,group_points, weight from ASSIGNMENT_GROUPS) as ag on ag.id=a.assignment_group_id) as bottom on rock.assignment_id = bottom.assign_id)"
+    df = pd.read_sql(sql,conn,params={"current_user": current_user,'course_id': UDW_COURSE_ID},parse_dates={'due_date': '%Y-%m-%d','graded_date':'%m/%d/%Y'})
```
jonespm: We probably should start putting some more error checking in here incase there is no result returned. Maybe at least in a try/catch. If df is empty it's going to cause problems.

jonespm: Or maybe just something like what's already used

    # return if there is no data during this interval
    if (df.empty):
        return HttpResponse("no data")

jonespm:
Removed the no-semi colon rule from Codacy! https://standardjs.com/rules.html#semicolons

PR #87, opened 2018-09-12
Resolves #83 and adding tool tip to the vertical line

### LINKED ISSUE #83: fix grades distribution view not working problem
1. fix the not loading the grades  view
2. handle empty data case
3. adding tooltip to the vertical line

pushyamig:


dashboard/templates/base.html
```diff
@@ -34,9 +34,9 @@
 <script src="{% static 'jquery/dist/jquery.min.js' %}"></script>
 <script src="{% static 'angular/angular.js' %}"></script>
 <script src="{% static 'underscore/underscore-min.js' %}"></script>
-<script src="{% static 'js/app.js' %}"></script>
-<script src="{% static 'js/controllers.js' %}"></script>
-<script src="{% static 'js/factories.js' %}"></script>
+<!--<script src="{% static 'js/app.js' %}"></script>-->
```
jonespm: Were these not supposed to be in the global file? You can just delete them rather than uncomment them. I saw it in two different html files so I added it here. 

I know we're removing the files.html but it might need to be added back to the assignments.html if it's still used there? Otherwise just maybe clean all these up? :) Things changing fast!

pushyamig: Those files are specific to the assignment view, so no need for them in the base.html.

PR #93, opened 2018-09-13
Add link back to home, fixes some styles fixes #90

### LINKED ISSUE #90: Add a link back on header, fix styles
We should have a link back to the homepage on the header now that we have a header. Also improve the styling to keep it more consistent across pages.

jonespm:


jonespm:
@zqian @pushyamig Do we need these body styles still? I think we should move them to a a global level if you want to override bootstraps styles. (Now that we have a global template)  

I personally like the look better without (most of) these overrides.

jonespm:
The "default style" is bootstrap's reboot

http://v4-alpha.getbootstrap.com/content/reboot/

dashboard/static/css/grades.css
```diff
@@ -1,7 +1,3 @@
-body {
-    font: 10px sans-serif;
```
pushyamig: I would make sure removing would have little impact the view

jonespm: Thanks, some views might be impacted. If you want smaller fonts in some cases just target a class or something in the content rather than the body.

jonespm: You can even edit this PR if you like and add the changes here!

pushyamig: Too late I created the issue  and fixed it . once you merge this change I will merge mine. https://github.com/tl-its-umich-edu/student-dashboard-django/pull/98

Issue #29, opened 2018-08-14
Change table/column name in Mysql with lower case 


pushyamig:
Django MySQL notes advise against using mixed case or upper case table names due to a known bug in MySQL.
https://docs.djangoproject.com/en/2.1/ref/databases/#table-names

jonespm:
Are you working on this one @pushyamig ? This seems like it would be nice to have sooner rather than later. (Though we could run a conversion)

pushyamig:
I am not currently working, I opened as an issue for awareness/ should work pilot. If you like to work on it  go ahead

jonespm:
We really should be lowercase for everything, table names and column names. Otherwise it can cause problems depending on the OS and lots of confusion in the code. In our current database some columns are upper case some are lower case.

https://dev.mysql.com/doc/refman/8.0/en/identifier-case-sensitivity.html

jonespm:
Also, we're running
python manage.py migrate

But we're also putting a script in the docker startup location. We should just switch to django migrations similar to student explorer going forward? 

https://github.com/tl-its-umich-edu/student_explorer/tree/develop/management/migrations

Maybe a separate issue.

pushyamig:
Done

PR #94, opened 2018-09-13
#29 making all the table/column names lowercase

pushyamig:


jonespm:
I'll look at this and try it out this afternoon. This is a good improvement though!

jonespm:
Looks good, we'll have to drop the database to load this up. Maybe we'll have the dev database tomorrow and can start fresh on both of those?

PR #97, opened 2018-09-13
Clean up unusused urls, duplicates and slashes. Fixes #91

### LINKED ISSUE #91: Clean up urls.py, remove duplicate and slashes
I noticed in urls.py grade_distribution is duplicated and some values have a trailing slash and others don't.

APPEND_SLASH is set because of CommonMiddleware so we don't need the slashes anywhere.
https://docs.djangoproject.com/en/2.1/ref/settings/#std:setting-APPEND_SLASH

jonespm:


dashboard/urls.py
```diff
@@ -41,26 +41,26 @@ def login_required(func):
...
+    url(r'^file_access', login_required(views.file_access), name='file_access'),
+    url(r'^file_access_within_week', login_required(views.file_access_within_week), name='file_access_within_week'),
+    url(r'^view_file_access_within_week', login_required(TemplateView.as_view(template_name='view_file_access_within_week.html')), name="view_file_access_within_week"),
+    url(r'^assignment_view', login_required(views.assignment_view), name='assignment_view'),
+    url(r'^assignment_progress', login_required(views.assignment_progress), name='assignment_progress'),
 
+    # These methods are all for loading test data
+    # TODO: Move these to cron job
     # load data from UDW
     url(r'^update_with_udw_file', login_required(cron.update_with_udw_file), name='update_with_udw_file'),
```
zqian: I think we may hide the individual 'update*' end points, and keep only one endpoint like 'testloader' (we need to remove the test part from the name) to do all the data load, for the purpose of data consistency.

jonespm: Yeah, that would be fine. I didn't know if you ever called these individual end points directly anymore. The easiest way (when I first wrote testloader) was just to call them all. ;) 

But it would probably be easier now to consolidate them all together in the backend.

jonespm: I plan to add the idea of "admin" so we could protect this "testloader" (or data_loader if we called it that) to only us for now too.

dashboard/urls.py
```diff
@@ -41,26 +41,26 @@ def login_required(func):
...
+    url(r'^files', login_required(TemplateView.as_view(template_name='files.html')), name="files"),
+    url(r'^grades', login_required(TemplateView.as_view(template_name='grades.html')), name="grades"),
     url(r'^assignments', login_required(TemplateView.as_view(template_name='assignments.html')), name="assignments"),
 
-    # load file information
-    url(r'^load_data', login_required(views.load_data), name='load_data'),
+    # Thse URL's are data patterns
+    # get file access patterns
+    url(r'^grade_distribution', login_required(views.grade_distribution), name='grade_distribution'),
+    url(r'^file_access', login_required(views.file_access), name='file_access'),
```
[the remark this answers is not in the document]
jonespm: Ok I can wait for that to merge.

dashboard/urls.py
jonespm: I have a  fix to clean up all of these other urls and support multiple courses so maybe I'll just wait for that too? That depends on #101

PR #100, opened 2018-09-14
 Add watchman support, redo database configuration fixes #99

### LINKED ISSUE #99: Add watchman status checking
This app should expose the /status URL via the watchman module

https://github.com/mwarkentin/django-watchman

jonespm:


jonespm:
I switched all of this over to just use the django connections here. I retained UDW_ENDPOINT but also support UDW_HOST to be consistent with MYSQL (we could remove ENDPOINT). 

To access the UDW you just need to specify it as a connection like I have here with `django.db.connections` if we want it outside the cron.py, otherwise it will use the MySQL default. 

https://docs.djangoproject.com/en/1.11/topics/db/multi-db/

jonespm:
Thanks @pushyamig 

I'm thinking for UDW we'll still run the SQL. 

For MySQL we'll probably set up the Django models and can switch to using those if we'd want (but can still also run the SQL) 

Though the plan is that eventually we'll get rid of both of these anyway. Just worthwhile learning experience I guess.

dashboard/settings.py
```diff
@@ -41,6 +41,10 @@
 
 ALLOWED_HOSTS = os.environ.get("ALLOWED_HOSTS","127.0.0.1,localhost").split(',')
 
+WATCHMAN_TOKEN = getenv('DJANGO_WATCHMAN_TOKEN', None)
```
zqian: updates needed for the README file or .env.sample file?

jonespm: Oh oops! Forgot about this, I think it's can be any value that you'd just give to access this. By default this is unsecured. But should be in the sample. Thanks!

PR #101, opened 2018-09-14
Switch config to python decouple fixes #71

### LINKED ISSUE #71: Switch config over to decouple

jonespm:


jonespm:
Yeah, this is good because it will more easily allow us to set Csv values (like admin users for become user and course ids for multiple course ids) as well as the cast is safer for other types. I didn't change all files but will change more in future PR's

PR #105, opened 2018-09-17
Remove broken files link from home page

jonespm:


jonespm:
This is fixed in #108

PR #108, opened 2018-09-17
Issue 14 -- latest change on the file access html

zqian:


dashboard/templates/view_file_access_within_week.html
```diff
@@ -19,33 +19,31 @@
...
 
     <input type="text" id="slider" class="slider">
 
-
     <input type="hidden" name="slider_num" id="slider_num" />
     <div/>
-
-    Top files access by
+    <br/>
+    <h6 align="center">Files accessed by
```
jonespm: Minor but these align attribute should be using bootstrap class="text-center" as align is deprecated https://getbootstrap.com/docs/3.3/css/#type-alignment

dashboard/templates/view_file_access_within_week.html
```diff
@@ -19,33 +19,31 @@
...
+    </select> <span id="slider_label"> students from week 1 to: </span></h6>
+    <br/>
     <div id="chart"></div>
 {% endblock %}
 {% block js %}
     <script src="{% static 'd3/d3.js' %}"></script>
     <script src="{% static 'd3-tip/dist/index.js' %}"></script>
     <script src="{% static 'js/rSlider.min.js' %}"></script>
     <script src="{% static 'js/file_access_within_week.js'%}"></script>
+    <script src="http://labratrevenge.com/d3-tip/javascripts/d3.tip.v0.6.3.js"></script>
```
jonespm: Can you switch this to` {% static 'd3-tip/dist/index.js' %}` (0.9.1) already there. Or does it need this old version?

dashboard/views.py
```diff
@@ -64,46 +64,30 @@ def gpa_map(grade):
...
         return 'B'
-    else:
+    elif grade_float > 90:
         return 'A'
+    else:
+        return NO_GRADE_STRING
+
+def get_current_week_number(request):
+    # get term start date
+    # TODO: term id hardcoded now
```
jonespm: Should probably create these TODO's as new issues so we don't forget about them.

dashboard/views.py
```diff
@@ -208,7 +202,7 @@ def file_access_within_week(request):
     # time 100 to show the percentage
     output_df["total_count"] = output_df["total_count"] * 100
     # round all numbers to one decimal point
-    output_df = output_df.round(1)
+    output_df = output_df.round(2)
```
jonespm: This comment doesn't match the change here. Should the rounding precision be a constant?

zqian: fixed. Defined constant as DECIMAL_ROUND_DIGIT = 1

dashboard/views.py
zqian: now added a config variable "CURRENT_CANVAS_TERM_ID" into .env.sample, and the tool now read the setting from the .env file.

dashboard/templates/view_file_access_within_week.html
zqian: No, I can definitely use the existing version. Link removed.

dashboard/templates/view_file_access_within_week.html
zqian: updated!

PR #111, opened 2018-09-18
Fix TEMPLATE warning and fix SAML issue fixes #110

### LINKED ISSUE #110: Fix TEMPLATE warning and fix SAML issue
Saml was broken in the url change #97. Need to fix these login URL's again.

Also this warning has been poking around, should remove/clean up all of these

`?: (1_8.W001) The standalone TEMPLATE_* settings were deprecated in Django 1.8 and the TEMPLATES dictionary takes precedence. You must put the values of the following settings into your default TEMPLATES dict: TEMPLATE_DIRS, TEMPLATE_DEBUG, TEMPLATE_LOADERS.`

jonespm:


jonespm:
Looks like according to https://django.readthedocs.io/en/1.8.x/ref/templates/upgrading.html the TEMPLATE_LOADERS were no longer needed, seems to work fine without.

PR #114, opened 2018-09-19
Issue_113: read user name from request object

zqian:


dashboard/views.py
```diff
@@ -93,9 +90,20 @@ def get_current_week_number(request):
     data['currentWeekNumber'] = currentWeekNumber
     return HttpResponse(json.dumps(data))
 
+def get_current_user_from_request(request):
+    if request.user.is_authenticated:
```
jonespm: I think this is fine. We can assume in these views though that the user will always be authenticated. So probably don't even need this method and can just directly call `request.user.get_username()` in the protected functions.

There is no method in this view class that should be able be accessed without being authenticated. 

I guess locally since we don't require auth it's different but now that we actually have localhost setup to be able to work with a "test auth" we probably could also remove that "hack" that allows local to work unauthenticated.

jonespm: I think after this change we should require creating users on local to test anyway, so I can clean this up on a followup patch if you'd like to and you can just commit this "as-is".

zqian: Good I will commit as it is

PR #120, opened 2018-09-20
#i16,#i106-addressing the rules flow, bootstrap update

pushyamig:
1. Rules workflow
2. Bootstrap update ~4.1.3
3. CSS  fixes
4. Handles both courses for pilot
5. Current week workflow and view changes

What remains with this view will be Formatting issues and whatever comes up with the feedback

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,7 +12,7 @@ RUN curl -sL https://deb.nodesource.com/setup_10.x | bash - && \
 
 # apt-utils needs to be installed separately
 RUN apt-get update && \ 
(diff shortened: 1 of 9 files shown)


pushyamig:
@jonespm @zqian If you could review this in the couple of hours that would be great as I have some DB tables update I want to make this working in PROD/Dev before the demo. I tested my code by merging it with the latest from repo, tested it and I don't see any issues.

jonespm:
Try vim-tiny package instead of vim. Usually not recommended to install the
full vim in an image since it's big and not used much.

Looks good otherwise.

Also I think we should create a new ticket to review our MySQL datatypes.
All of them seem to be char(255) that might be too long for some and too
short for others

Issue #92, opened 2018-09-13
Move Google Analytics config into the env file


jonespm:
It was mentioned that we'd like to have the Google GA tags in the env files now rather than hardcoding them. 

Then we could also re-configure or disable locally and in dev.

PR #130, opened 2018-09-26
Fix SAML login to just go to one URL fixes #129

### LINKED ISSUE #129: Fix SAML login to just go to one URL
All we need to do is add a config parameter to the URL

jonespm:


jonespm:
This adds a django settings module, which will also allow us to easily fix #92

dashboard/settings.py
```diff
@@ -285,3 +291,5 @@
     AUTHENTICATION_BACKENDS += ('django.contrib.auth.backends.ModelBackend',)
     LOGIN_REDIRECT_URL = '/'
     
+
+SETTINGS_EXPORT = ['LOGIN_URL','LOGOUT_URL','DEBUG']
```
zqian: What's the purpose of this setting?

zqian: I see, it is used in base.html.

jonespm: Yeah I can add a comment here, it's for the settings module. Basically every setting we want to have available to in the templates needs to be in here. I added DEBUG as I figured we'd want that.

PR #131, opened 2018-09-26
Move Google Analytics to env file fixes #92

### LINKED ISSUE #92: Move Google Analytics config into the env file

jonespm:


jonespm:
I added this value to just the prod deployment

PR #134, opened 2018-09-26
Add course number to URL line fixes #127

### LINKED ISSUE #127: Add course number to the URL line
We need to have the course number in the URL line. This will take the same pattern as Canvas.

`https://dashboard.tl.it.umich.edu/<course id>/<page>` . . . (Any additional page parameters)

Where <course id> matches the value in Canvas and page is (currently) one of 

grades
view_file_access_within_week
assignment

This is currently hardcoded as CANVAS_COURSE_ID but will be passed to all methods via django.

jonespm:


jonespm:
I tested all of the existing views to work here. This needs a little more work.

- We need a method to look up the course name from the number and display a user friendlier course name. This is straightforward as it's in the course table already but figured I'd leave it separate.
- We need a something on the home page for a user so if they launch dashboard without a link or go back to the main page they'll have thing to do. I left this hardcoded at the moment. Maybe we want to give them access to all of their sites, and maybe users with no data should have any access. This is for another ticket too.

dashboard/settings.py
```diff
@@ -291,5 +291,12 @@
     AUTHENTICATION_BACKENDS += ('django.contrib.auth.backends.ModelBackend',)
     LOGIN_REDIRECT_URL = '/'
     
+# This is fixed from UDW
+UDW_ID_PREFIX = config("UDW_ID_PREFIX", default=17700000000, cast=int)
+
+# TODO: Select a default course for user from database, just here so things don't all break
+# This will probably need to move to views.py
+DEFAULT_COURSE_ID = config("CANVAS_COURSE_IDS", default=0, cast=int)
```
zqian: I would think the CANVAS_COURSE_IDS consists a course id array, unlike the DEFAULT_COURSE_ID is just one single course id.

jonespm: Yeah, at the moment this was only singular (even though the variable name says IDS)

I think we should remove ID's from this completely and store it in the database. I can work on this but the 'default' for a user is going to be different than what this array was. I just didn't want it to completely break on the dev/prod server(s)

PR #139, opened 2018-09-28
Change the URL pattern to include /courses and /api in the URL

jonespm:


jonespm:
FYI.

I added /api/v1 to everything that just has a data return. Eventually we may want to bump the version up on these as well.

I added /courses to anything that's a view.

Currently if the course id is NOT present it default it to 0 producing unexpected results, especially in the "View File Access by week" view. Will have to look into this.

PR #151, opened 2018-10-01
issue_104: fixes to file access view, grade drop choices, fixed the o…

zqian:
…verflow problem of brushing, etc

### DIFF
--- a/dashboard/templates/base.html
+++ b/dashboard/templates/base.html
@@ -32,7 +32,6 @@
     </header>
     {% block content %}
     {% endblock %}
-    <footer><hr>Copyright 2018 University of Michigan ITS Teaching and Learning</footer>
 </body>
 <!-- Common Javascript -->
 {% if settings.GA_ID %}
--- a/dashboard/templates/view_file_access_within_week.html
+++ b/dashboard/templates/view_file_access_within_week.html
@@ -27,16 +27,14 @@
     <input type="hidden" name="slider_num" id="slider_num" />
     <div/>
     <br/>
-    <h6 class="text-center">Files accessed by
+    <h6 class="text-center">Files accessed by students with grade
     <select class="grade" name="grade" id="grade">
         <option value="all" selected>All</option>
-        <option value="A" >A</option>
-        <option value="B" >B</option>
-        <option value="C" >C</option>
-        <option value="D" >D</option>
-        <option value="F" >F</option>
+        <option value=">=90" >90-100%</option>
+        <option value="80-89" >80-89%</option>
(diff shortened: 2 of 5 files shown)


dashboard/templates/base.html
```diff
@@ -32,7 +32,6 @@
     </header>
     {% block content %}
     {% endblock %}
-    <footer><hr>Copyright 2018 University of Michigan ITS Teaching and Learning</footer>
```
jonespm: Did you mean to remove the footer entirely in this PR?
I think it should match up with other sites and be

> © 2018 The Regents of the University of Michigan


Rather than what we have here. But we should have something.

dashboard/templates/view_file_access_within_week.html
```diff
@@ -27,16 +27,14 @@ <h4 class="text-center">What Students Are Reading</h4>
...
-    <h6 class="text-center">Files accessed by
+    <h6 class="text-center">Files accessed by students with grade
     <select class="grade" name="grade" id="grade">
         <option value="all" selected>All</option>
-        <option value="A" >A</option>
-        <option value="B" >B</option>
-        <option value="C" >C</option>
-        <option value="D" >D</option>
-        <option value="F" >F</option>
+        <option value=">=90" >90-100%</option>
```
jonespm: We probablay don't need > symbol in the text if you have the range here. Wouldn't be understood by all students anyway. Saying 90-100 seems the same thing.

dashboard/views.py
```diff
@@ -33,8 +33,14 @@
 CANVAS_FILE_PREFIX = config("CANVAS_FILE_PREFIX", default="")
 CANVAS_FILE_POSTFIX = config("CANVAS_FILE_POSTFIX", default="")
 CANVAS_FILE_ID_NAME_SEPARATOR = "|"
+# This is fixed from UDW
+UDW_ID_PREFIX = config("UDW_ID_PREFIX", default="17700000000")
```
jonespm: I'd just use settings.UDW_ID_PREFIX for this value when needed. Better to just define everything in settings. We need to fix cron.py reusing this value too.

jonespm: I don't even see this used in this change?

dashboard/templates/base.html
zqian: updated to "Copyright © 2018 The Regents of the University of Michigan"

dashboard/templates/view_file_access_within_week.html
zqian: it is not in the text. But I have it fixed anyway -- "90-100" is used

PR #155, opened 2018-10-01
Refactor some code in the grades view add additional error handling

### LINKED ISSUE #154: Refactor some code in the grades view add additional error handling
There was text in the grades view 
"You don\'t seems to have your grade yet"

That should be rewritten, also this is redundant and also there was a problem in the view when there was no grades calculated in the database.

jonespm:
fixes #154

### DIFF
--- a/dashboard/static/js/grade_distribution.js
+++ b/dashboard/static/js/grade_distribution.js
@@ -158,20 +158,16 @@ var makeGraph = function () {
                 .text("a simple tooltip");
         }
 
-        if(myscore==null){
-            var gradeInfo = d3.select(".grade-info").append("div")
-                .attr("class", "grade-details")
-                .html('<span>Number of students: <b>' + totalStudents + '</b></span><br>' +
-                    '<span>Average grade: <b>' + gradeAverage + '</b></span><br>' +
-                    '<span>Standard deviation of grades: <b>' + standardGradeDeviation + '</b></span><br>'+
-                     '<span>**You don\'t seems to have your grade yet</span><br>')
-        }else{
-            var gradeInfo = d3.select(".grade-info").append("div")
-                .attr("class", "grade-details")
-                .html('<span>Number of students: <b>' + totalStudents + '</b></span><br>' +
-                    '<span>Average grade: <b>' + gradeAverage + '</b></span><br>' +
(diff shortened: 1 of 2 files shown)


dashboard/views.py
```diff
@@ -213,7 +213,7 @@ def grade_distribution(request, course_id=0):
     grade_score_sql = "select current_grade,(select current_grade from user where sis_name=" \
                       "%(current_user)s and course_id=%(course_id)s) as current_user_grade from user where course_id=%(course_id)s;"
     df = pd.read_sql(grade_score_sql, conn, params={"current_user": current_user,'course_id': course_id})
-    if df.empty:
+    if df.empty or df['current_grade'].isnull().all():
```
pushyamig: Nice catch! all 3 courses have grades come in in my local DB, so I couldn't catch this bug

dashboard/static/js/grade_distribution.js
```diff
@@ -158,20 +158,16 @@ var makeGraph = function () {
...
-                .html('<span>Number of students: <b>' + totalStudents + '</b></span><br>' +
-                    '<span>Average grade: <b>' + gradeAverage + '</b></span><br>' +
-                    '<span>Standard deviation of grades: <b>' + standardGradeDeviation + '</b></span><br>')
+        var gradeDetailsHtml = 
+            '<span>Number of students: <b>' + totalStudents + '</b></span><br>' +
+            '<span>Average grade: <b>' + gradeAverage + '</b></span><br>' +
+            '<span>Standard deviation of grades: <b>' + standardGradeDeviation + '</b></span><br>';
+        if(myscore==null) {
+            gradeDetailsHtml += '<span><b>There are no grades yet for you in this course</b></span><br>';
         }
```
pushyamig: Nice

dashboard/views.py
jonespm: Yeah thanks I noticed a stack trace locally when there were no grades for this user at all. This caught that case.

pushyamig: I feel the stack trace would have occurred when there are no grades at all for the course vs no grade for user. The code is filtering out the null's and if user grade is null then it is shown as message of "no grades for you"

jonespm: Yeah, that might have been what happened too. Basically if the nulls are all filtered out and df was empty after that process was done. This just catches it a little bit earlier.

PR #158, opened 2018-10-01
Update the color of unread files fixes #150

### LINKED ISSUE #150: File view page: change orange to gray to match Assignments viz
change orange to gray to match Assignments viz

zqian:


dashboard/static/js/file_access_within_week.js
```diff
@@ -2,6 +2,10 @@
 ///////////////// Set-up SVG and wrappers ///////////////////
 /////////////////////////////////////////////////////////////
 
+// colors used for different file states
+var COLOR_ACCESSED_FILE ="steelblue",
```
jonespm: I wonder if we should define these in the top level dasbboard { } variable in the base? Then we can use them through other views and have them all in one place?

zqian: currently file view and assignment view uses different js and css. I will check in the change right now, and leave the code refactor to later issues.

PR #160, opened 2018-10-01
Fix django issue about is_anonymous django deprecation fixes #126

### LINKED ISSUE #126: Fix django issue about is_anonymous django deprecation
RemovedInDjango20Warning: Using user.is_authenticated() and user.is_anonymous() as a method is deprecated. Remove the parentheses to use it as an attribute.

jonespm:


manage.py
```diff
@@ -1,10 +1,15 @@
 #!/usr/bin/env python
 import os
 import sys
+import warnings
 
 if __name__ == "__main__":
     os.environ.setdefault("DJANGO_SETTINGS_MODULE", "dashboard.settings")
 
+    # Regex for which modules to ignore warnings from
```
zqian: does this mean any warning from djangosaml2 will all be ignored?

jonespm: Yeah, that should probably target the line specifically, let me look into that.

jonespm: I modified this so it only ignores DeprecationWarning(s). I could target line number but that might change if the code changes.

zqian: looks good!

PR #168, opened 2018-10-03
alignment changes for assignment fixes #141

### LINKED ISSUE #141: Assignment Planning: improve the visual design
1. Left align the top and bottom sections
2. Choose fewer font sizes.  One for text, one for header, one for sub-headers
3. Top level the largest - first thing you see ; sub-headings at next level
4. Add title to the page: Assignment Planning Tool
5. Put a little more space between the progress bar numbers and the header for the list below to make it clear that the sub-header goes with the section below
6. Move the text on the line markers on the progress bar above the bar (Current, Max Possible)

pushyamig:
Fixing the alignment issues with assignment view


dashboard/views.py
```diff
@@ -211,7 +211,7 @@ def grade_distribution(request, course_id=0):
     logger.info(grade_distribution.__name__)
 
     current_user = request.user.get_username()
-    grade_score_sql = "select current_grade,(select current_grade from user where sis_name=" \
+    grade_score_sql = "select distinct sis_name,current_grade,(select distinct current_grade from user where sis_name=" \
```
jonespm: This is fine but we should be adding these as unique values. I don't think we're doing this correctly. This is another ticket though.

pushyamig: I can open up another ticket. But In theory, I shouldn't be doing this as there should not be duplicate records.

pushyamig: I removed it from this commit, if the duplicate record issue is resolved then this can be used as a test case for testing it

Issue #186, opened 2018-10-04
Select the site a user is in if they don't specify a site id


jonespm:
Related to #177 

We need something that will select what site a user is in if they "leave out" or remove the siteid from the URL. If they are not in any sites known to the system they should be displayed some generic page saying that they don't have any sites available.

Issue #132, opened 2018-09-26
Add some security checks that user can access course


jonespm:
After we move to multiple courses, we should have a check that the user can access the data in the course. 

I think (for now) this is just checking the enrollment_dim table for user_id, course_id?

jonespm:
This check should be added for any methods called after the course_id is passed via URL in #127

jonespm:
This should be the check

`select course_id from user where sis_name = {} and course_id = {}`

This would also return the course(s) a user is in, if let without the course_id parameter.

jonespm:
This is duplicated in #314 so closing.

PR #187, opened 2018-10-04
This just gets the courses/<course_id> to work and removes the front

### LINKED ISSUE #177: Re-do URL's to have a new landing page
As discussed in chat

/courses/<course_id> should display the current landing page that the root shows with that specific course. 

/ and /courses without an id should show the /courses/<course_id> for whatever the first course found for that user is (currently there is just one course per user). If they are in no courses, just display a message of some kind.

In the future we may have another page where the user can pick their course.

All pages without a course id should show no data after the authorization fix is in.

jonespm:
page. Fixes #177

Still needs some additional fixes.

### DIFF
--- a/dashboard/urls.py
+++ b/dashboard/urls.py
@@ -37,8 +37,10 @@ urlpatterns = [
     # These URL's are for views, the accept an empty id
     url(r'^courses/(?P<course_id>[0-9]+|)/?grades', login_required(TemplateView.as_view(template_name='grades.html')), name="grades"),
     url(r'^courses/(?P<course_id>[0-9]+|)/?assignments', login_required(TemplateView.as_view(template_name='assignments.html')), name="assignments"),
-
     url(r'^courses/(?P<course_id>[0-9]+|)/?view_file_access_within_week', login_required(TemplateView.as_view(template_name='view_file_access_within_week.html')), name="view_file_access_within_week"),
+
+    # This is the courses catch-all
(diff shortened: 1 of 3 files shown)


jonespm:
@pushyamig I think this will fix the issue with the courses that you were working on providing links for. I plan to also work on the "default course" as part of #186 and #132

dashboard/urls.py
```diff
@@ -37,9 +37,11 @@
     # These URL's are for views, the accept an empty id
     url(r'^courses/(?P<course_id>[0-9]+|)/?grades', login_required(TemplateView.as_view(template_name='grades.html')), name="grades"),
     url(r'^courses/(?P<course_id>[0-9]+|)/?assignments', login_required(TemplateView.as_view(template_name='assignments.html')), name="assignments"),
-
     url(r'^courses/(?P<course_id>[0-9]+|)/?view_file_access_within_week', login_required(TemplateView.as_view(template_name='view_file_access_within_week.html')), name="view_file_access_within_week"),
 
+    # This is the courses catch-all
+    url(r'^courses/(?P<course_id>[0-9]+|)', login_required(TemplateView.as_view(template_name='courses.html')), name="courses"),
+
     # Thse URL's are data patterns
```
pushyamig: This all looks good.

zqian: Looks good!

jonespm:
Thanks, I'll merge this but it's going to break the main landing page from working, now only /courses is going to work. I'll fix up the main landing page in another PR.

Issue #182, opened 2018-10-04
Add a link to help documentation.


jennlove-um:
Add a link to the dashboard pages to the Google help documentation:

https://docs.google.com/document/d/1Av0SJQGsMf-EmChLpygUi5-a603w6D6zROT7p48mwkI/edit?usp=sharing

jennlove-um:
From the home page, link to the first page of the help doc: https://docs.google.com/document/d/1Av0SJQGsMf-EmChLpygUi5-a603w6D6zROT7p48mwkI/edit?usp=sharing

From the What Students Are Reading page, link to that section in the help: https://docs.google.com/document/d/1Av0SJQGsMf-EmChLpygUi5-a603w6D6zROT7p48mwkI/edit#heading=h.ei0djpnrbuw

From the Assignments Planning Tool, link to that section in the help: https://docs.google.com/document/d/1Av0SJQGsMf-EmChLpygUi5-a603w6D6zROT7p48mwkI/edit#bookmark=id.j326p1brruc1

From the Grade Distribution page, link to that section in the help: https://docs.google.com/document/d/1Av0SJQGsMf-EmChLpygUi5-a603w6D6zROT7p48mwkI/edit#bookmark=id.9qq8fcgci5m7

jonespm:
I'm not sure if we need context sensitive help currently. I also wouldn't want these help documents to be links to Google documents.

Minimally the documents we link should be published documents (File->Publish to the Web)
I think it would be better if all of these were on a Google Sites (sites.google.com) and organized together there rather than individual files here. 

If you created a new "Google Site" and had these pages with these titles and URL's on them it would be nicer.

jennlove-um:
Is this more like what you're thinking with a Google site?  https://sites.google.com/umich.edu/student-dashboard-help/home

jonespm:
1000% yes! ;)

pushyamig:
Done

PR #188, opened 2018-10-04
#182 adding help text to the header

pushyamig:


dashboard/templates/base.html
```diff
@@ -66,6 +66,22 @@
 <script src="{% static 'jquery/dist/jquery.min.js' %}"></script>
 <script src="{% static 'angular/angular.js' %}"></script>
 <script src="{% static 'underscore/underscore-min.js' %}"></script>
+<script>
+    $(document).ready(function () {
+        url = window.location.href;
+        var help_link='https://sites.google.com/umich.edu/student-dashboard-help/home';
```
zqian: seems to me that we can define the base url in a variable. otherwise, it all looks good!

jonespm: Yeah right, help_link is kind of the base, in all of the methods you can just go

help_link=help_link+"/grade-distribution" or whatever the link is for

I'd also make this into an else if just incase it matches multiple cases

PR #189, opened 2018-10-04
Add django-debug-toolbar fixes #185

### LINKED ISSUE #185: Add debug-toolbar to better assist debugging
Debug toolbar should be added to assist debugging. It should be restricted to the ports on the developer VPN. 

This might also need to use this pattern to get the IP https://stackoverflow.com/a/4581997/3708872

jonespm:


jonespm:
Yeah sure.

Basically if DEBUG is on and the user is a super user (in the database) it will show this toolbar. So just for local and non-prod. The toolbar displays a lot of useful information for debugging purposes. It won't even be loaded at all in production. I'm also looking at some additional extensions to help aid more in debugging.

PR #190, opened 2018-10-05
Issue 125 -- run cron job on separate pod

zqian:


.env.sample
```diff
@@ -50,7 +50,9 @@ UDW_PORT=
 UDW_ID_PREFIX=
 
 # Canvas Configuration
+# put course canvas ids here, separated by ","
 CANVAS_COURSE_IDS=
+CANVAS_COURSE_IDS=COURSE_ID_1,COURSE_ID_2
```
jonespm: I have a WIP PR to store these in a database table, this is fine for now though. I'll change it next week.

dashboard/cron.py
```diff
@@ -42,221 +41,60 @@
...
-    :return:
-    '''
-    logger.info("weight_consideration()")
-    is_weight_considered_url ="with course as (select course_id, sum(group_weight) as group_weight from assignment_group_fact " \
-                              "where course_id = '" + UDW_COURSE_ID + "' group by course_id having sum(group_weight)>1)"\
-                              "(select CASE WHEN EXISTS (SELECT * FROM course WHERE group_weight > 1) THEN CAST(1 AS BOOLEAN) ELSE CAST(0 AS BOOLEAN) END)"
-
-    return HttpResponse("loaded weight_consideration info: " + util_function(is_weight_considered_url,'assignment_weight_consideration', 'weight'))
+UDWCourseIdsString = config("CANVAS_COURSE_IDS", "")
+UDWCourseIds = UDWCourseIdsString.split(",")
```
jonespm: Rather than having this split, just change the line before to be 

config("CANVAS_COURSE_IDS", default="", cast=Csv()) 

This is already imported.

dashboard/cron.py
```diff
@@ -42,221 +41,60 @@
...
+logger.debug(UDWCourseIds)
+
+# minutes between cron jobs
+RUN_SCHEDULE = 24*60
+
+# get UDW course id array from configuration file
+def getUDWCourseIds():
 
+    # ### Get enrolled users and files within site
+    canvasCourseIdString =config('CANVAS_COURSE_IDS', default="")
```
jonespm: This is repeated from UDWCourseIdsString and UDWCourseIds? Are these going to be the same?

jonespm:
I fixed the things I commented about on https://github.com/tl-its-umich-edu/student-dashboard-django/pull/190/commits/7637936a5fb1fc4204760ae1049d360054ca51b5

I tested that the cron still runs as expected.

If you could look at what Codacy still suggests I'd think this PR is pretty good to go! Just have to set these course id's we're supporting in dev and prod.

dashboard/cron.py
```diff
@@ -42,221 +41,35 @@
...
-# ## Connect to Student Dashboard's MySQL database
-#
-# ### Get enrolled users and files within site
 db_name = settings.DATABASES['default']['NAME']
 db_user = settings.DATABASES['default']['USER']
 db_password = settings.DATABASES['default']['PASSWORD']
 db_host = settings.DATABASES['default']['HOST']
 db_port = settings.DATABASES['default']['PORT']
 
-engine = create_engine("mysql+pymysql://{user}:{password}@{host}:{port}/{db}"
```
jonespm: create_engine is duplicated a bunch of times in this code. Were you not able to re-use this engine?

dashboard/cron.py
```diff
@@ -268,4 +81,339 @@ def util_function(sql_string, mysql_table, table_identifier=None):
...
+# remove all records inside the specified table
+def deleteAllRecordInTable(tableName):
+    # delete all records in the table first
+    engine = create_engine("mysql+pymysql://{user}:{password}@{host}:{port}/{db}"
+                           .format(db = db_name,  # your mysql database name
+                                   user = db_user, # your mysql user for the database
+                                   password = db_password, # password for user
+                                   host = db_host,
+                                   port = db_port))
+    connection = engine.connect()
```
jonespm: [I saw that if you use the pattern
](https://docs.sqlalchemy.org/en/latest/core/connections.html#sqlalchemy.engine.Connection.detach)
```
with engine.connect() as con:
    conn.execute("SOME SQL") 
```

Then you don't have to close because connection is fully closed after that.

dashboard/cron.py
```diff
@@ -268,4 +81,339 @@ def util_function(sql_string, mysql_table, table_identifier=None):
...
+        logger.debug("in update with udw user")
+
+        # delete all records in the table first
+        status += deleteAllRecordInTable("user")
+
+        # loop through multiple course ids
+        for UDW_course_id in settings.DEFAULT_UDW_COURSE_IDS:
+
+            # select all student registered for the course
+            user_sql = "select u.name AS name, " \
```
jonespm: When I was writing these queries for validation

putting them in """ ... """ made the queries cleaner. https://github.com/tl-its-umich-edu/unizin-sis-validation/blob/master/dbqueries.py

You can use the regular string formatting with % or .format to insert the variables afterward.

dashboard/cron.py
```diff
@@ -268,4 +81,339 @@ def util_function(sql_string, mysql_table, table_identifier=None):
...
+def update_cron_status(status):
+    logger.debug("in update_cron_status")
+
+    engine = create_engine("mysql+pymysql://{user}:{password}@{host}:{port}/{db}"
+                           .format(db = db_name,  # your mysql database name
+                                   user = db_user, # your mysql user for the database
+                                   password = db_password, # password for user
+                                   host = db_host,
+                                   port = db_port))
+    connection = engine.connect()
```
jonespm: Same suggestion as in the delete

dashboard/cron.py
zqian: Done! Moved out from class and defined as global variable.

PR #206, opened 2018-10-08
Set MySQL to a fixed port

jonespm:


docker-compose.yml
```diff
@@ -20,7 +20,7 @@ services:
       - UDW_DATABASE=${UDW_DATABASE}
     entrypoint: ['docker-entrypoint.sh', '--default-authentication-plugin=mysql_native_password']
     ports:
-      - "3306"
+      - "5306:3306"
```
zqian: Why do we need to set MySQL to a fixed port? Why 5306?

jonespm: It could be any port, but if you don't it's randomized locally each time. This would allow us to use some client app like MySQL Workbench or DBeaver to connect to the instance and inspect the data on a fixed port.

It could be any port number, it probably shouldn't be 3306 because that would conflict if a local MySQL was installed as that's the default. I just chose 5306 because we have 5001 for the server right now.

jonespm: This only affects local host

PR #216, opened 2018-10-09
Sync up gitignore to githubs recommendations fixes #215

### LINKED ISSUE #215: Sync up gitignore to github's recommendations
We have some files that are getting through our gitignore. Should clean those up and use githubs recommended ignores.

https://github.com/github/gitignore/blob/master/Python.gitignore

jonespm:


jonespm:
Yeah, as it says we don't. These ones we have must have been added months ago before the .gitignore was setup. This commit is removing these files.

https://github.com/tl-its-umich-edu/student-dashboard-django/tree/master/dashboard/__pycache__

zqian:
I see. No more questions then!

PR #220, opened 2018-10-09
Remove pymysql fixes #219

### LINKED ISSUE #219: Remove pymysql
We have both mysqlclient and pymysql in the requirements.

I only see pymysql used for the cron and we can switch it over. For the standard cpython mysqlclient is much faster.

https://gist.github.com/methane/90ec97dda7fa9c7c4ef1
https://stackoverflow.com/a/46396881/3708872

jonespm:


jonespm:
Yeah the only "slight" negative about this is to build mysqlconnector on OSX takes a little work. This is only to avoid errors with a local IDE since the docker image has everything it needs! I had to do this though. Either seemed like the full MySQL had to be installed locally or a piece if it on homebrew.

https://zehengl.github.io/2017/12/17/pip-install-mysqlclient-on-macos/

Issue #222, opened 2018-10-09
Homepage link should keep the course info when a student clicks it from one of the visualizations


jennlove-um:
When I try to go from the Files Accessed page back to the home page, the dashboard forgets which course I'm in and gives the message "My Learning Analytics: This site must be launched from Canvas."

Here's the workflow: 
1. Go to the Files Accessed page for a course
2. Click on the My Learning Analytics link in the upper left 
3. Dashboard home page displays with no course selected
--> Expected result: User sees the home page for the course they are in

Is this something that is happening because I'm an admin and not officially enrolled in any course?  I also tested using the "login as" option, but saw the same results.

jonespm:
Yeah, that sounds like a good idea. I'm partially working on this as part of #186 but I think the ID should carry back to the URL link on the upper left.

PR #224, opened 2018-10-09
Select the site a user is in if they don't specify fixes #186 and #222

### LINKED ISSUE #186: Select the site a user is in if they don't specify a site id

jonespm:


dashboard/common/util/__init__.py
```diff
@@ -0,0 +1,42 @@
...
+        If there is no default return nothing
+    Arguments:
+        user_id {[int]} -- User Id
+    
+    Returns:
+        [str] -- Canvas Course Id
+    """
+    course_id = ""
+    with django.db.connection.cursor() as cursor:
+        cursor.execute("SELECT course_id FROM user WHERE sis_name= %s LIMIT 1", [user_id])
```
zqian: Should we add a "order by" selection here, to avoid random returns.

dashboard/settings.py
```diff
@@ -309,9 +310,6 @@
 # TODO: Select a default course for user from database, just here so things don't all break
 DEFAULT_COURSE_IDS = config("CANVAS_COURSE_IDS", default="0", cast=Csv())
```
zqian: There are two entries in the .env.sample for "CANVAS_COURSE_IDS". Can you also remove one there?

dashboard/templates/base.html
```diff
@@ -57,7 +57,7 @@
 <script>
 <!-- Setup a top level js object to hold global values -->
 var dashboard = {
-{% if course_id %}"course_id": '{{settings.UDW_ID_PREFIX}}'+'{{ course_id }}',
+{% if course_id %}"course_id": '{settings.UDW_ID_PREFIX}}'+'{{ course_id }}',
```
zqian: the curly braces does not seem to match

jonespm: Yeah, hmm I wonder how I broke that!

dashboard/common/util/__init__.py
jonespm: Yeah, phase 1 should only have 1 return, but we can add a predictable order.

dashboard/settings.py
jonespm: Sure! These "DEFAULT_COURSE_IDS" now are really just the default id's that are populated by the cron. This is no longer used otherwise after this change.

zqian:
Sphinx doc seems cool! We can do that update after the pilot release.

dashboard/context_processors.py
```diff
@@ -1,14 +1,15 @@
...
-        if (row != None):
-            course_name = row[0]
+        course_name = get_course_name_from_id(course_id)
 
     return {'course_name': course_name}
+
+def current_user_course_id(request):
+    course_id = str(request.resolver_match.kwargs.get('course_id'))
+    if not course_id:
+        course_id = get_default_user_course_id(request.user.username)
```
pushyamig: I would add a `log.info` statement capturing this case

dashboard/common/util/__init__.py
```diff
@@ -0,0 +1,40 @@
+# Some utility functions used by other classes in this project
+
+import django
+
+def get_course_name_from_id(course_id):
```
pushyamig: I feel we should log the function call as `logger.info(get_course_name_from_id.__name__)` as it is easy for debugging. I don't know it is a best practice.

jonespm: I commented on #227 that we should just probably look into using decorators instead. I can add this as a new issue.

dashboard/context_processors.py
jonespm: Yeah, this is going to be called many times in a single request so there'd be many duplicate records. I wonder if there's a way in python to buffer/compress log messages like for instance if I had.

logger.info("Course ID not in request for someuser")

And that was repeated like 6 times in the same requests it would just say

Course ID not in request for some user (Message repeated 6 times)

pushyamig: That many time it would be noise, But why is it calling 6 times in one request. I am assuming this is from the home page where course id is required for each view. But once you get it from the backend we should store it as JS variable and use it. I don't want to make it complicated if it is practical enough to do it. Since the `current_user_course_id ` is not making a DB call we should be fine for calling it multiple time

jonespm: Well this function is being used currently by a context_processor and on every page it's called 7 times. But you make a great point. These functions should for sure be cached or memoized on the backend and just called once by the view.

jonespm: I think caching could really improve this app, especially the views. This probably won't be a worry for the pilot but something we really should look at for a larger rollout.

https://docs.djangoproject.com/en/1.11/topics/cache/

jonespm: I added that log to the method, it looks like even though it's hit multiple times in the template it's only executed once.

dashboard/common/util/__init__.py
zqian: new issue is better. Let's get this checked in first.

PR #227, opened 2018-10-10
#226 minor log correction

pushyamig:


jonespm:
I feel like this fix is fine, but related to the comment on the other ticket I think we should do this with decorators if this is something we want to do. Less chance of error and less code.

Like this library had the @traced decorator.
https://pythonhosted.org/Autologging/index.html#

Maybe create a new Issue to switch the code over to logging decorators?

PR #234, opened 2018-10-10
Issue 233

zqian:


dashboard/settings.py
```diff
@@ -307,6 +307,9 @@
 # This is fixed from UDW
 UDW_ID_PREFIX = config("UDW_ID_PREFIX", default="17700000000")
 
+# This is fixed from UDW
+UDW_FILE_ID_PREFIX = config("UDW_FILE_ID_PREFIX", default="1770000000")
```
jonespm: I wonder if this is ever going to ever be different than UDW_ID_PREFIX?

jonespm: This value is also not in the .env.sample.

jonespm: I think for this we should just set the default value of UDW_ID_PREFIX and just add it to the sample, if we think it *might* be different.

zqian: The UDW_FILE_ID_PREFIX is less one zero compared to UDW_ID_PREFIX. Don't ask me how I found  out  -:(

zqian: I will add that into .env.sample

README.md
```diff
@@ -81,3 +81,16 @@ Then you can edit your files! (Probably in /dashboard/dashboard)
   - You have to be authenticated and a "super user" account. See step #1
   - The method that controls this access is in show_debug_toolbar(request):
   - Configuration of the panels is in DEBUG_TOOLBAR_PANELS as described on https://django-debug-toolbar.readthedocs.io/en/latest/configuration.html#debug-toolbar-panels
+
+## Steps for get localhost running
```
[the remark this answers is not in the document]
zqian: OK. I will consolidate with that section

dashboard/settings.py
zqian: .env.sample updated

PR #237, opened 2018-10-10
Show latest time updated fixes #165

### LINKED ISSUE #165: The page should indicate when data was last refreshed
Currently there's no indicate how current the data is. There should be some indication (maybe in the footer?) how fresh the data is.

jonespm:


Dockerfile
```diff
@@ -42,5 +42,9 @@ RUN yarn install && \
 # This sets up some initial MySQL tables
 COPY mysql/init.sql /docker-entrypoint-initdb.d
 
+# Sets the local timezone of the docker image
+ENV TZ=America/Detroit
+RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone
```
zqian: is "/usr/share/.." path available in OpenShift deploy?

jonespm: This was in the docker image which is running the same base as Openshift. I checked and it's there.

I hard-coded this all to America/Detroit. We currently don't support changing timezones. But I felt like displaying the time in the footer in Eastern time was more useful to 99% of the population than displaying it in UTC.

PR #238, opened 2018-10-11
Issue 229 - support opt out views

zqian:


jonespm:
All looks good to me. I was mentally debating whether these columns should just be added to the courses table,  a separate table with columns or in a separate key/value table. We can change it later if there are too many.

I think though once we link more to Canvas we can just store them either there via the API or with LTI eventually so might not matter.

mysql/init.sql
```diff
@@ -83,13 +83,26 @@ consider_weight BOOLEAN DEFAULT FALSE,
...
 INSERT INTO academic_terms (NAME, start_date, end_date) VALUES ('SUMMER 2018', '2018-06-27 04:00:00',  '2018-08-17 23:59:59' );
 INSERT INTO academic_terms (NAME, start_date, end_date) VALUES ('FALL 2018', '2018-09-03 04:00:00',  '2018-12-24 23:59:59' );
--- inert course
+-- insert course
 INSERT INTO course (id, name, term_id) VALUES ('17700000000230745', 'STATS 250 SU 2018', 1);
 INSERT INTO course (id, name, term_id) VALUES ('17700000000252307', 'SI 110 001 FA 2018', 2);
 INSERT INTO course (id, name, term_id) VALUES ('17700000000245664', 'SI 664 002 FA 2018', 2);
 INSERT INTO course (id, name, term_id) VALUES ('17700000000283292', 'HMP 654 001 FA 2018', 2);
+-- insert course view options
+INSERT INTO course_view_option (course_id, show_files_accessed, show_assignment_planning, show_grade_distribution) VALUES ('17700000000252307', FALSE, FALSE, FALSE);
```
pushyamig: The Option for this course should be True, True,False

zqian: Right.  That's the combination for production.

Those values are here for us to test no-show for every view in local deployment.

PR #241, opened 2018-10-12
Adding cronjob script to pod for cron fixes #240

### LINKED ISSUE #240: Fix up cronjob pod to run regularly
The cronjob doesn't look like it'd running regularly, need to fix this up!

jonespm:


start.sh
```diff
@@ -30,12 +30,16 @@ if [ -z "${IS_CRON_POD}" ]; then
...
+    # Make the log file available
+    touch /var/log/cron.log
 
-    python manage.py migrate django_cron
+    # Get the environment from docker saved
+    # https://ypereirareis.github.io/blog/2016/02/29/docker-crontab-environment-variables/
+    printenv | sed 's/^\([a-zA-Z0-9_]*\)=\(.*\)$/export \1="\2"/g' >> $HOME/.profile
 
-    python manage.py runcrons
+    echo "*/5 * * * * . $HOME/.profile; python /dashboard/manage.py runcrons >> /var/log/cron.log 2>&1" | crontab
```
zqian: every 5 minutes would be too much. can we change this to 2AM every morning?

jonespm: This is just how often the system cron will kick off the django cron. It probably could be every hour but I think might want it regularly instead of a specific time. The django cron runs on CRON_RUN_SCHEDULE so it could run at any time but will wait at *least* that many minutes. 

For instance with this setting, it will execute runcrons every 5 minutes but CRON_RUN_SCHEDULE being set to 24 hours will only allow it to run once a day. 

I think if we wanted it to be at 2AM precisely then it's still fine to have this cron fire off hourly, but we'll want to change the cron to be
    RUN_AT_TIMES = ['2:00']

Rather than having run_every_mins. 

What the cron library is getting us
- Log table with status of the cron runs in the database
- Django configuration of schedule
- Cluster wide protection (So if we trusted the cron and wanted to remove the cron pod we probably could, that would just mean any cron changes have to have a full app change but maybe will be a good idea)

However it still has to be executed regularly. We can review this more tomorrow and get the right settings.

jonespm: I'll fix this by tomorrow. I'll change the CRON_RUN_SCHEDULE parameter to be RUN_AT_TIMES and allow for a csv of times that could be run. 

Then we can have another parameter for the actual cron schedule in cron format. (Which should match the RUN_AT_TIMES but will need to be created separately) 

Like 2AM in cron format is "0 2 * * *"  "https://crontab.guru/#0_2_*_*_*". It's also fine if this cron format is more frequent than the RUN_AT_TIMES which I'll document, as it won't do anything but print messages to the logs that nothing was run.

jonespm:
I've updated this PR to support specific run times

jonespm:
Added more documentation on the cron job to readme

jonespm:
I merged this and set dev to be 
RUN_AT_TIMES=12:00,14:00,16:00,18:00,20:00
CRONTAB_SCHEDULE=0 12-20 * * *

This should enable us to test it a few hours running through the day before moving to prod pod with more limited runs.

PR #243, opened 2018-10-12
Fixed up a lot of Codacy issues Fixes #225

### LINKED ISSUE #225: Review and fix up some Codacy issues
Before launching pilot, Codacy should be reviewed for the easy issues and fixed up.

jonespm:
Cleaned up SQL in cron to use f-strings for readability

Fixes #225

### DIFF
--- a/dashboard/cron.py
+++ b/dashboard/cron.py
@@ -25,8 +25,8 @@ db_password = settings.DATABASES['default']['PASSWORD']
 db_host = settings.DATABASES['default']['HOST']
 db_port = settings.DATABASES['default']['PORT']
 
-logger.debug("db-name:" + db_name);
-logger.debug("db-user:" + db_user);
+logger.debug("db-name:" + db_name)
+logger.debug("db-user:" + db_user)
 
 engine = create_engine("mysql+mysqldb://{user}:{password}@{host}:{port}/{db}?charset=utf8mb4"
                        .format(db = db_name,  # your mysql database name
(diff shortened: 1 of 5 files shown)


jonespm:
There was some other SQL that probably could have been improved in the views.py but that can be left for another day. I'd really like to move most of that out to models anyway.

dashboard/cron.py
```diff
@@ -263,18 +267,16 @@ def update_assignment(self):
...
         # loop through multiple course ids
         for UDW_course_id in settings.DEFAULT_UDW_COURSE_IDS:
-            assignment_sql="with assignment_info as " \
-                           "(select ad.due_at AS due_date,ad.due_at at time zone 'utc' at time zone 'America/New_York' as local_date," \
-                           "ad.title AS name,af.course_id AS course_id,af.assignment_id AS id," \
-                           "af.points_possible AS points_possible,af.assignment_group_id AS assignment_group_id" \
-                           " from assignment_fact af inner join assignment_dim ad on af.assignment_id = ad.id where af.course_id='" + UDW_course_id + "'" \
-                                                                                                                                                      "and ad.visibility = 'everyone' and ad.workflow_state='published')" \
-                                                                                                                                                      "select * from assignment_info"
+            assignment_sql = f"""with assignment_info as
```
zqian: What is the syntax of f""" stands for? Formatted string value?

jonespm: Yeah, they're fstrings, new in Python 3.6. I kind of think it's pretty clean to have them like this.

https://realpython.com/python-f-strings/

We could change it to .format, but I think either .format or fstrings are easier to read than inline concatenation.

zqian: Nice to know!

PR #244, opened 2018-10-12
Use canvas data date for last updated fixes #242

### LINKED ISSUE #242: Use canvasdatadate from udw for last updated date
Currently we're using the date from the last time the cronjob ran as the updated date, however this isn't accurate. 

We should "really" be using the date from the UDW which is in unizin_metadata -> canvasdatadate. We can store this in some table during the cronjob run.

jonespm:


dashboard/context_processors.py
```diff
@@ -32,4 +32,4 @@ def course_view_option(request):
     return {"course_view_option": course_view_option}
 
 def last_updated(request):
-    return {'last_updated' : get_last_cron_run()}
\ No newline at end of file
+    return {'last_updated' : get_canvas_data_date()}
```
zqian: should we compare the last update time from both UDW and CRON, and show the earliest of the two?

jonespm: Hmm, I'd believe that this UDW time should always be the right one as the table is refreshed when we update? 

The only way I think it might not be correct is if it's not present at all which is why I didn't completely remove the other method yet.

zqian: Yes, the UDW table unizin_metadata is the one stores the latest update timestamp.

I will vote yes for the change and merge it in.

Issue #245, opened 2018-10-15
Change the Assignment Design that it is not closely tied to the student information


pushyamig:
This is mostly due to the way the DB query made closely w.r.t to Student data.  The view should show course data irrespective for admins and only course specific instructor

pushyamig:
Test Plan: 
Logging as admin/or any student that are not part of the course should be able to see the Non-personalized view. 

I won't see this is a ideal design for showing other who are not part of the course to see course related information but that what other 2 views are supporting now. That would be a separate issue to fix showing the content specific to the user

jennlove-um:
Testing passes.  The non-personalized view shows for the admin or for a student who is not in the course selected.  The personalized view shows for a student in the course (SI 110 tested).

PR #247, opened 2018-10-17
#245 non-personalized view for assignments

pushyamig:
#245 .
This PR has one DB column update and cron job query update. Although it was not greatly needed but I felt it will be cleaner that way. 

Currently anybody who has uniqname can access the tool and who ever figured out the course_number can access all the Views student dashboard has to offer. This is not just true for Assignment view and this PR is not handling that. That would be a different Issue.


jonespm:
Looks good to me, I like the cleanup here. I think we should change the format of the SQL in views to fstrings and triple quotes like cron.py file. But it probably should be removed from the views entirely. Good first step toward that cleanup.

Issue #248, opened 2018-10-17
Multiple students in different courses is not handled currently


pushyamig:
The pilot courses don't seems to have students in multiple courses but the `user` table that is created as such  you cannot add a same user with different course. `User` table PK needs to be based of `user_id and course_id`.

pushyamig:
Test Plan: This changes Primary Key(PK) for the user table. Making a new column as PK should not impact the views at all

1. Make sure Files/assignment/grades personalized and non-personalized view is loading correctly.
Testing this is more or less like a regression testing.

jennlove-um:
As far as I can tell this is working.  I see the correct view of the screens when I'm logged in as myself (non-personalized) and when I masquerade as a student in SI 110. 

The only view I can't test is the grade distribution view - we don't have a student from the SI 664 or HMP 654 to test with in dev.

PR #249, opened 2018-10-17
#248 supporting students in multiple courses

pushyamig:
#248 
I don't see a burning need to support this for pilot as no students seems to be part of multiple courses. But it was small enough fix to do


jonespm:
Yup, cool! Just going to have to run this as an alter on the dev/prod db's.

PR #250, opened 2018-10-17
Add custom checks for watchman to check cron status fixes #246

### LINKED ISSUE #246: Add retry on failure and notification to cron
Now that we have crontab and it's success/failure, we should be notified when there is an error on the pod.

jonespm:
Clean up a few imports of local modules


jonespm:
Hi Zhen,

Watchman is the python package we're using for the status checks. This adds a custom check for "cron" = "Status of the last cron job run success/failure" to the status URL. This would result in Nagios warning us when the cron failed.

The biggest negative of this I can think of is if it does fail and it's scheduled at 2AM we'll get a nagios alert.

We either should 
- Change the cron schedule to a more convenient time when we could more easily respond to it
- Implement the "retry after failure" to the cron to try to get it to retry and recover automatically. https://django-cron.readthedocs.io/en/latest/sample_cron_configurations.html#retry-after-failure-feature
- Not include this in watchman for nagios status check at all but just send an email when the cron fails. (Most of this code would still be relevant)

jonespm:
Yeah, right if cronjob failed it wouldn't be killing the app. Maybe I'll just change this to send a separate email and leave the "watchman" access point off for now.

PR #252, opened 2018-10-18
I251 muted grades

pushyamig:


jonespm:
The commit on 196acc4 looks good! Thanks for the quick fix. Need to rebase or resolve the conflict from the already merged 7d9a74f.

PR #253, opened 2018-10-18
Issue 20

zqian:


dashboard/views.py
```diff
@@ -79,6 +100,16 @@ def file_access_within_week(request, course_id=0):
...
     grade = request.GET.get('grade','all')
 
+    # json for eventlog
+    data = {
+        "week_num_start": week_num_start,
+        "week_num_end": week_num_end,
+        "grade": grade,
+        "course_id": course_id
+    }
+    eventlog(request.user, EVENT_VIEW_FILE_ACCESS, None, json.dumps(data))
```
pushyamig: where is this event is getting logged? in the DB or Inline( Doc says that can be done this way too)? if it is in DB then until the insert is done the Response is not sent to the view? Ideally this should be like a event log request be sent to the background thread and that process should be doing the insert with out blocking the original request.

start.sh
```diff
@@ -24,6 +24,9 @@ if [ -z "${IS_CRON_POD}" ]; then
     # Start Gunicorn processes
     echo Starting Gunicorn.
 
+    # prepare for event log
+    python manage.py migrate eventlog
```
jonespm: I don't think you need a separate migrate for eventlog, the general migrate above should pick this up.

dashboard/views.py
zqian: Events are stored in the database as eventlog_log table. It is also accessible from the admin interface.

dashboard/views.py
```diff
@@ -27,12 +29,31 @@
 GRADE_LOW="low_grade"
 NO_GRADE_STRING = "NO_GRADE"
 
+EVENT_VIEW_FILE_ACCESS="VIEW_FILE_ACCESS"
```
jonespm: These constants don't seem to add much value since they're only used once. I wonder if there's a better practice for this. 

Maybe a "little" better organization might be to put this in a class like 

```
class Events:
    VIEW_FILE_ACCESS = "VIEW_FILE_ACCESS"
    VIEW_ASSIGNMENT_PLANNING = "VIEW_ASSIGNMENT_PLANNING"
    VIEW_GRADE_DISTRIBUTION = "VIEW_GRADE_DISTRIBUTION"

```
Then you'd call Events.VIEW_GRADE_DISTRIBUTION

Just a thought, I don't know what's better. Probably both are about the same at the moment. ;)

zqian: I will keep that in mind. Maybe not changing this time. Will do that once we have more events.

start.sh
zqian: indeed! removed the line.

dashboard/views.py
jonespm: Yeah, I think once we get more events we could have the EVENT_ID as well as the data payload all defined inside the class. So all of these are better together. Would be nice as this event model expands out. A good idea for future refactor as we add more for sure.

PR #260, opened 2018-10-19
Fixes up some imports and cleans up some other code fixes #259

### LINKED ISSUE #259: Cleanup common util imports
We shouldn't be doing an import * in context_processors, this should be cleaned up a little

jonespm:


dashboard/settings.py
```diff
@@ -330,20 +330,7 @@
...
-    'debug_toolbar.panels.request.RequestPanel',
-    'debug_toolbar.panels.sql.SQLPanel',
-    'debug_toolbar.panels.staticfiles.StaticFilesPanel',
-    'debug_toolbar.panels.templates.TemplatesPanel',
-    'debug_toolbar.panels.cache.CachePanel',
-    'debug_toolbar.panels.signals.SignalsPanel',
-    'debug_toolbar.panels.logging.LoggingPanel',
-    'debug_toolbar.panels.redirects.RedirectsPanel',
-]
+DEBUG_TOOLBAR_PANELS = dt_settings.PANELS_DEFAULTS
```
zqian: Could not find PANELS_DEFAULTS defined in dt_settings.py

jonespm: Is this an error you see what you run this? It should be there https://github.com/jazzband/django-debug-toolbar/blob/master/debug_toolbar/settings.py and I'd thought it was working. I'll can look at this again.

PR #264, opened 2018-10-19
Add additional configuration to cron to allow retries and emailing

### LINKED ISSUE #246: Add retry on failure and notification to cron

jonespm:
admins. Fixes #246


dashboard/settings.py
```diff
@@ -320,11 +323,20 @@
...
+# Time to run cron, use the old RUN_AT_TIMES as a fallback
+CRON_RUN_AT_TIMES = config('CRON_RUN_AT_TIMES', default=config("RUN_AT_TIMES", default=""), cast=Csv())
+
+# Time to retry crons after failures
+CRON_RETRY_AFTER_FAILURE_MINS = config('CRON_RETRY_AFTER_FAILURE_MINS', default=None)
+
+# Minimum number of failures before an email is sent to ADMINS
+CRON_MIN_NUM_FAILURES = config('CRON_MIN_NUM_FAILURES', default=10, cast=int)
+
+FAILED_RUNS_CRONJOB_EMAIL_PREFIX = config('CRON_FAILED_RUNS_CRONJOB_EMAIL_PREFIX', default="[{0} :Cron Failure]".format(PROJECT_NAME))
```
zqian: 10 shall be a good retry number!

jonespm: Yeah, it's configurable so we can change it. 

Currently it wouldn't retry at all unless we set up the system cron to run more often. I think you'd have to have the system (Crontab) cron to run at least as often as the CRON_RETRY_AFTER_FAILURE_MINS setting for this to do anything.

Like if we have 
CRON_RETRY_AFTER_FAILURE_MINS=5 (5 minutes to retry after a failure)
and CRON_MIN_NUM_FAILURES=10, it would be 50 minutes till it signaled a failure. But we'd have to have crontab every 5 minutes. If we had the CRON_RETRY_AFTER_FAILURE_MINS = 1 we'd have to have crontab running every minute and it would send an email faster. It's fine to run it frequently, I just don't know what the best setting would be.

Maybe like 
CRON_RETRY_AFTER_FAILURE_MINS=5
CRON_MIN_NUM_FAILURES=3

And have crontab every 5 minutes, then we'd be notified in 15 minutes after 3 failures?

PR #269, opened 2018-10-22
Json data in event log is not properly formatted fixes #268

### LINKED ISSUE #268: Json Data in event log is not propertly formatted
Something @pushyamig noticed was that the json in the event log wasn't proper json format. It was incorrectly escaped. I think it was because it was double escaped by the code and the library. Will submit a PR for this to clean this up.

jonespm:


dashboard/views.py
```diff
@@ -107,7 +92,7 @@ def file_access_within_week(request, course_id=0):
         "grade": grade,
         "course_id": course_id
     }
-    eventlog(request.user, EVENT_VIEW_FILE_ACCESS, None, json.dumps(data))
+    eventlog(request.user, EVENT_VIEW_FILE_ACCESS, extra=data)
```
pushyamig: is there a way in the library we could mention this `extra` column data type so that we could make this column as JSON Obj instead of `longtext` now

jonespm: It doesn't look like we can support json column because mysql library we're using doesn't want to support it. https://github.com/PyMySQL/mysqlclient-python/issues/213

zqian: This is very elegant way of storing JSON data into extra field!

jonespm: Yeah, seemed super straightforward. I compared what's written out to the column and it's better. Looked like it was being double escaped before and not usable in tableau. Will just merge this.

Issue #261, opened 2018-10-19
Assignment Planning - sort order is sometimes wrong when using weight dropdown


jennlove-um:
I noticed in the HMP 654 class that when I select assignments weighted at 5% or higher, the assignments don't display in order by week.

![image](https://user-images.githubusercontent.com/39634416/47229683-51755800-d396-11e8-923f-4d9463342888.png)

jennlove-um:
This passes in testing on dev.  Thanks!

PR #270, opened 2018-10-25
#261 assignments are not sorted sometime

pushyamig:
#261

### DIFF
--- a/dashboard/views.py
+++ b/dashboard/views.py
@@ -312,7 +312,7 @@ def assignment_view(request, course_id=0):
     week_list = []
     for item in assignment_list:
         week_list.append(item['week'])
-    weeks = set(week_list)
+    weeks = sorted(set(week_list))
     full = []
     i = 1
     for week in weeks:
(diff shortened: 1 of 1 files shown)


dashboard/views.py
```diff
@@ -312,7 +312,7 @@ def assignment_view(request, course_id=0):
     week_list = []
     for item in assignment_list:
         week_list.append(item['week'])
-    weeks = set(week_list)
+    weeks = sorted(set(week_list))
```
jonespm: Why is there a new set variable here? Seems like it should have just been
week_list.sort()

And use week_list later in the for loop.

```suggestion
    week_list.sort()
```

And changing
`for week in week_list:`

jonespm: Is this just a set to remove duplicates?

pushyamig: Yes to remove the duplicated it is a `Set`

jonespm: Yeah, so then using set from the beginning probably would have made the code a little easier to read. https://github.com/tl-its-umich-edu/student-dashboard-django/compare/master...jonespm:issue_261-rewrite?expand=1

Yeah, minor cleanup but this file could probably a little more cleanups.

jonespm: Also another cool python function is [enumerate](http://book.pythontips.com/en/latest/enumerate.html) which I changed the for loop below to use
https://github.com/tl-its-umich-edu/student-dashboard-django/commit/54afbb42ce6edb451ecf5a3aa5c38bf9bfe922c5

PR #274, opened 2018-10-25
Build is failing due to some import problems

pushyamig:
Currently the OpenShift build is failing due to import issue in the `settings.py`. This fix should resolve that.

I tested this fix from my personal branch, So you might see latest as successful in Openshift


jonespm:
Oops! Thanks for the fix!

PR #278, opened 2018-11-01
Only watch default database, not udw

### LINKED ISSUE #276: Only watch the "default" database, not the UDW database.
We only should have the default database in the watchman check, not both.

jonespm:
This fixes #276


dashboard/settings.py
```diff
@@ -46,6 +46,9 @@
 
 WATCHMAN_TOKEN_NAME = config('DJANGO_WATCHMAN_TOKEN_NAME', default='token')
 
+# Only report on the default database
+WATCHMAN_DATABASES = ('default',)
```
pushyamig: is that extra comma needed?

PR #284, opened 2018-11-01
Add github version to footer

### LINKED ISSUE #282: Add github version number to the dashboard footer
Dashboard footer should show the github version number. I'm thinking this would be generated with

One idea is to have the DockerFile create this file 

`RUN git rev-parse --short HEAD > /dashboard/dashboard/version.py`

And then have the other file pull this in. However since we really just care about it in the template, it could just add it as an html template and maybe just include it in another template?

https://stackoverflow.com/a/10985987/3708872

jonespm:
This fixes #282

### DIFF
--- a/dashboard/templates/base.html
+++ b/dashboard/templates/base.html
@@ -46,8 +46,8 @@
     <footer>
         <hr>
         <div class="row">
-            <div class="col text-left">Copyright © 2018 The Regents of the University of Michigan</div>
-            <div class="col text-right">Last updated on {{last_updated|date:"m/d/Y P T"}}</div>
+            <div id="footer-left" class="col text-left"><span id="copyright">Copyright © 2018 The Regents of the University of Michigan</span></div>
+            <div id="footer-right" class="col text-right"><span id="updated">Data last updated on {{last_updated|date:"m/d/Y P T"}}</span><span id="version">:Version {% include "git_head_version.html" %}</span></div>
         </div>
     </footer>
 </body>
--- a/dashboard/templates/base.html
+++ b/dashboard/templates/base.html
@@ -47,7 +47,7 @@
         <hr>
         <div class="row">
             <div id="footer-left" class="col text-left"><span id="copyright">Copyright © 2018 The Regents of the University of Michigan</span></div>
-            <div id="footer-right" class="col text-right"><span id="updated">Data last updated on {{last_updated|date:"m/d/Y P T"}}</span><span id="version">:Version {% include "git_head_version.html" %}</span></div>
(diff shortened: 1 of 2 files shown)


dashboard/templates/base.html
```diff
@@ -46,8 +46,8 @@
     <footer>
         <hr>
         <div class="row">
-            <div class="col text-left">Copyright © 2018 The Regents of the University of Michigan</div>
-            <div class="col text-right">Last updated on {{last_updated|date:"m/d/Y P T"}}</div>
+            <div id="footer-left" class="col text-left"><span id="copyright">Copyright © 2018 The Regents of the University of Michigan</span></div>
+            <div id="footer-right" class="col text-right"><span id="updated">Data last updated on {{last_updated|date:"m/d/Y P T"}}</span><span id="version">:Version {% include "git_head_version.html" %}</span></div>
```
zqian: Should "Version" be "GitHub Version" to be more specific?

jonespm: Well maybe "Git" version. I didn't want to change it if we moved to GitLab

jonespm: Though when/if we do that we could worry about that then ;)

jonespm: image

pushyamig: This is good. But I feel that we need to have the Openshift Build info on the page some where invisible that would be good. That will be helpful in case of roll back of code to earlier version build/deployment. It would be a good gaze as we know that we have Build#123 or Deploy#23, git branch/tag. Much like what CCM/CPM doing it as an invisible button.

`OPENSHIFT_BUILD_NAMESPACE: ccm OPENSHIFT_BUILD_NAME: canvascoursemanager-12 GIT_SOURCE: https://github.com/tl-its-umich-edu/canvas_course_manager.git GIT_BRANCH: 4.0.1 GIT_COMMIT: fb68ce8d56741338f6c1fa414201be5c19b91023`

pushyamig: May be another ticket

jonespm: Yeah, while invisible button is convenient, it's not a good solution for accessibly reasons. Maybe for the python programs we can just add this information to the Django Admin UI. That seems a good solution all-around. Another ticket.

pushyamig: Yes, I think openshift variable is just for the dev to make sure the deploy/build it is what it is visually.

jonespm:
Yeah, I'd like to tag this release prior to deploying it. I'd also like to tag a few of the previous milestone releases. I hope to talk about this during the brief meeting this afternoon about what tag/versions to use. Probably the first "released" version as 1.0, then 1.1, etc.

Issue #279, opened 2018-11-01
grades distribution is not showing grades > 100%


pushyamig:
SI 664 course design seems to be student can score grades more than 100% and currently grades view is not accounting for it.

pushyamig:
Test Plan: 
1. the Number that appear on the bars should match up the grades that appears on the grade book, it not necessarily match up the number of students in the course as sometime student scores is empty so that data is removed from the distribution view.

Note: student 0 score will be reflected in the Grades view but not the empty value

pushyamig:
Done

PR #286, opened 2018-11-02
#279 show course with grades more than 100

pushyamig:
Relates to #279

### DIFF
--- a/dashboard/static/js/grade_distribution.js
+++ b/dashboard/static/js/grade_distribution.js
@@ -20,8 +20,9 @@ var makeGraph = function () {
 
         // get the current grade
         bin_grades = [];
+        data_more_than_100_percent = singleDataPoint.more_than_100_percent;
         data.forEach(function (e) {
-            bin_grades.push(e.current_grade_mod)
+            bin_grades.push(e.current_grade)
         })
 
         var width = 1000,
@@ -37,9 +38,10 @@ var makeGraph = function () {
         height = height - margin.top - margin.bottom;
         g = svg.append("g")
             .attr('transform', 'translate(' + margin.left + ',' + margin.top + ')');
+        var domain_values = (data_more_than_100_percent) ? [0, 110] : [0, 100]
 
         var x = d3.scaleLinear()
-            .domain([0, 100]).nice()
+            .domain(domain_values).nice()
             .rangeRound([0, width]);
 
         var bins = d3.histogram()
@@ -157,7 +159,7 @@ var makeGraph = function () {
                 .text("a simple tooltip");
         }
 
-        var gradeDetailsHtml = 
+        var gradeDetailsHtml =
             '<span>Number of students: <b>' + totalStudents + '</b></span><br>' +
             '<span>Average grade: <b>' + gradeAverage + '</b></span><br>';
         if(myscore==null) {
--- a/dashboard/static/js/grade_distribution.js
+++ b/dashboard/static/js/grade_distribution.js
@@ -20,7 +20,7 @@ var makeGraph = function () {
 
         // get the current grade
         bin_grades = [];
-        data_more_than_100_percent = singleDataPoint.more_than_100_percent;
+        get_domain_upper_limit=singleDataPoint.graph_upper_limit;
         data.forEach(function (e) {
             bin_grades.push(e.current_grade)
         })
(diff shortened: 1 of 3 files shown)


dashboard/views.py
```diff
@@ -216,8 +216,13 @@ def grade_distribution(request, course_id=0):
         return HttpResponse(json.dumps({}), content_type='application/json')
     number_of_students = df.shape[0]
     df = df[df['current_grade'].notnull()]
-    df['current_grade_mod'] = df['current_grade'].apply(lambda x: x.replace("100","99.99"))
-    average_grade = df['current_grade'].astype(float).mean().round(2)
+    df['current_grade'] = df['current_grade'].astype(float)
+    if df[df['current_grade'] > 100.0].shape[0] > 0:
+        df['more_than_100_percent'] = True
```
zqian: is the new ''more_than_100_percent' column with bool value used later in tool?

dashboard/templates/grades.html
```diff
@@ -14,12 +14,12 @@
...
         {% endblock %}
-        {% block js %}
-            <script src="{% static 'd3-v4/build/d3.min.js' %}"></script>
-            <script src="/static/js/grade_distribution.js"></script>
-        {% endblock %}
     {% else %}
         Grade Distribution view is hidden for this course.
     {% endif %}
 {% endblock %}
+{% block js %}
```
zqian: not sure why this change is needed for current issue: if grade distribution is hidden, no js is needed, right?

dashboard/static/js/grade_distribution.js
```diff
@@ -37,9 +38,10 @@ var makeGraph = function () {
         height = height - margin.top - margin.bottom;
         g = svg.append("g")
             .attr('transform', 'translate(' + margin.left + ',' + margin.top + ')');
+        var domain_values = (data_more_than_100_percent) ? [0, 110] : [0, 100]
```
zqian: Is 110 big enough to accommodate extra points?

dashboard/templates/grades.html
pushyamig: This is bug that resulted from JS error is the console  as this added the <script> tags twice in the html as resulted i guess from the  toggle view changes . So this is not related to current issue but I felt might as well fix it since I am here

dashboard/views.py
pushyamig: Yes

dashboard/static/js/grade_distribution.js
jonespm: Yeah, maybe it would be nice to see what the visualization looks like in this case. 

Extra credit in grading has somewhat confused me. Either there's still a hard limit of 100% and anything over it doesn't get you anything extra, or grades over 100% give you something different. It looks like in Canvas, the grading schema can't go over 100% but students can still score over 100%. https://community.canvaslms.com/thread/2127

I wonder if our visualization needs to account for this? Maybe have a little "piece" on the end that's "up-to" 10% in width with a tooltip that says how far over 100% the students total grade is? Is this is what this code is attempting to do? Maybe an idea for phase 2.

pushyamig: Should be good or have to do the math based on the percent appropriate. I could also tell D3 that fix the domain limit based on values in the list. The problem with that is if the class is in 90-100 the distribution range it looks odd.  I will lie if I say i did not think is 110 enough? But being practical will the % grade ever grow higher than  110%, I felt unlikely so I made the hardcoded number.

pushyamig: screen shot 2018-11-02 at 9 30 30 am
screen shot 2018-11-02 at 9 30 46 am

jonespm: Yeah this looks good. This 110% only shows if there are grades over 100% right? I think it's nice to show. If there are any grades over 110% it could probably change the label to "110%+" and just show them on the "110%+ line". We might need to add this logic to the files page too. 

I think we *could* have just even left it at 100% and put all the grades >= 100% on a "100%+" line but it's probably nice to have this extra bit of visualization.

jonespm: Just wait till we have the gamification class that goes up to 1000% or something someday. ;)

pushyamig: OK, I have made the upper limit adaptive based on the data. so if we have course with grade distribution like 150% the view will serve it than abandoning scores as before
@jonespm I like your idea of making it something like 110%+, I think code needs to change the logic around `MyGrade` line if a student has a score 115% (say) that means you i have put the orange line on 110%+ marker as that is the last marker on the line. So I feel the docs needs to be updated with this feature. We will keep it simple for now for everybody. We can think about this idea for next phase.

DanielDugan: I agree with the above comments. I think a 110%+ column is the best visualization option. We push the distribution of the graph to far to the left in scenarios where a handful of students score in 1.4, 1.5, and 10.1 percentages.

PR #302, opened 2018-11-19
single query to display both progress and plan view

pushyamig:
This PR is addressing the assignment view backed call processing. instead of 2 separate calls to present assignment progress and plan view single DB query is doing it all. Now the Controller is handling both the data for the views. The Progress JS is loaded after assignment data call so that progress data is made available  

The `assign_progress.js` has a lot of white space changes in order to ignore the white space changes use the GitHub UI as show here
screen shot 2018-11-19 at 8 09 22 am

### DIFF
--- a/dashboard/static/js/assign-progress.js
+++ b/dashboard/static/js/assign-progress.js
@@ -1,219 +1,200 @@
-var makeGraph = function () {
+jQuery(document).ready(function ($) {
     d3.selectAll("#chart > *").remove();
-    $.getJSON("/api/v1/courses/"+dashboard.course_id+"/assignment_progress", function (initResult) {
-        data = initResult;
-        if(_.isEmpty(data)){
-            var gradeInfo = d3.select("#chart").append("div")
-                .attr("class", "alert alert-info")
-                .attr("role","alert")
-                .style("font-size", "20px")
-                .style("font-weight", "bold")
-                .html('No data for the view')
-            return
-        }
+    var ctrlName = 'mainController'
+    var sel = 'div[ng-controller="' + ctrlName + '"]';
+    var data = angular.element(sel).scope().progress
+    var tooltip = d3.select("body").append("div").attr("class", "tooltip").style("opacity", 0);
 
-        var tooltip = d3.select("body").append("div").attr("class", "tooltip").style("opacity", 0);
+    var width = 1500,
+        height = 100;
 
-        var width = 1500,
-            height = 100;
+    var margin = {
+        top: 20,
+        right: 30,
+        bottom: 30,
+        left: 30
+    };
 
-        var margin = {
-            top: 20,
-            right: 30,
-            bottom: 30,
-            left: 30
-        };
+    var svg = d3.select("#chart").append("svg")
+        .attr('width', '100%')
+        .attr('height', '100%')
+        .attr('viewBox', '0 0 ' + width + ' ' + height)
+        .attr("preserveAspectRatio", "xMinYMin")
+        .append('g');
 
-        var svg = d3.select("#chart").append("svg")
-            .attr('width', '100%')
-            .attr('height', '100%')
-            .attr('viewBox', '0 0 ' + width + ' ' + height)
-            .attr("preserveAspectRatio", "xMinYMin")
-            .append('g');
-
-        width = width - margin.left - margin.right,
+    width = width - margin.left - margin.right,
         height = height - margin.top - margin.bottom;
 
-        svg.attr('transform', 'translate(' + margin.left + ',' + margin.top + ')');
+    svg.attr('transform', 'translate(' + margin.left + ',' + margin.top + ')');
 
-        // Create x and y scale.
+    // Create x and y scale.
 
-        var xScale = d3.scaleLinear()
-            .range([0, width]);
+    var xScale = d3.scaleLinear()
+        .range([0, width]);
 
-        var yScale = d3.scaleBand()
-            .range([0, height]);
+    var yScale = d3.scaleBand()
+        .range([0, height]);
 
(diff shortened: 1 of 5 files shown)


dashboard/views.py
```diff
@@ -297,6 +268,20 @@ def assignment_view(request, course_id=0):
     df.drop(columns=['assignment_id', 'due_date'], inplace=True)
     df.drop_duplicates(keep='first', inplace=True)
 
+    df3 = df[df['towards_final_grade']>0.0]
```
jonespm:
```suggestion
    df3 = df[df['towards_final_grade'] > 0.0]
```

dashboard/views.py
```diff
@@ -297,6 +268,20 @@ def assignment_view(request, course_id=0):
...
+    df3[['score']] = df3[['score']].astype(float)
+    df3['percent_gotten']=df3.apply(user_percent,axis=1)
```
jonespm:
```suggestion
    df3['percent_gotten'] = df3.apply(user_percent, axis = 1)
```

jonespm: I'm not going to reformat all of these, just a few here.

dashboard/urls.py
```diff
@@ -48,8 +48,7 @@
     # get file access patterns
     url(r'^api/v1/courses/(?P<course_id>[0-9]+)/grade_distribution', login_required(views.grade_distribution), name='grade_distribution'),
     url(r'^api/v1/courses/(?P<course_id>[0-9]+)/file_access_within_week', login_required(views.file_access_within_week), name='file_access_within_week'),
-    url(r'^api/v1/courses/(?P<course_id>[0-9]+)/assignment_view', login_required(views.assignment_view), name='assignment_view'),
-    url(r'^api/v1/courses/(?P<course_id>[0-9]+)/assignment_progress', login_required(views.assignment_progress), name='assignment_progress'),
+    url(r'^api/v1/courses/(?P<course_id>[0-9]+)/assignments', login_required(views.assignments), name='assignments'),
```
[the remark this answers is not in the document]
DanielDugan: I like the idea of consolidating the calls. I sat in for a bit while Pushyami worked on this.

dashboard/static/js/controllers.js
```diff
@@ -1,14 +1,24 @@
...
-                    return scrollId=e.id;
+            if (_.isEmpty(data.data)) {
+                $scope.plan = []
+                $scope.progress = []
+                return;
+            }
+            $scope.plan = data.data.plan
+            $scope.progress = data.data.progress
+            // call the js file onload only
+            if (call == 1) {
```
jonespm: This would probably be clearer if you used an object lookup. Maybe like
```

var dataScripts = {
    progress: '/static/js/assign-progress.js',
}
```

Then you could call it as dataCall("progress") instead of dataCall(1) and look it up in this object? The way it is now it's not clear what calling this with anything does. I think it would even be better if you just called it with the name of the script like dataCall ("assign-progress.js") rather than an integer. Then it can concat to /static/js ?

dashboard/views.py
jonespm: Looks like equals are supposed to be spaced out between an assignment or binary operators but not  in arguments.

dashboard/static/js/assign-progress.js
```diff
@@ -1,219 +1,200 @@
-var makeGraph = function () {
+jQuery(document).ready(function ($) {
```
zqian: Should the javascript for displaying file access also follow the same pattern?

dashboard/templates/assignments.html
```diff
@@ -7,14 +7,18 @@
 
 {% block content %}
 {% if course_view_option.show_assignment_planning %}
+<div ng-controller="mainController" ng-app="visApp">
 <div class="container-fluid">
     <h5 style="padding-left: 10px">Progress toward Final Grade</h5>
     <div id="chart"></div>
+    <div ng-show="progress.length == 0" class="alert alert-info">
+        There's no data
```
zqian: since we changed "data" into "plan", should the message in UI also say "There is no assignment plan", or "There is no assignment data for planning", or something similar?

dashboard/views.py
zqian: Is there an auto-way to do the reformatting job from IDE?

jonespm:
@zqian It looks like in VSCode you can select an area and do "Format Selection". It prompts you to install a PEP8 filter then it will format.

By default it also enforces the line length restriction, but I can commit a settings.json to override this to only do the formatting. Looks like it works really nice!

dashboard/templates/assignments.html
pushyamig: I changed the working for clarity

pushyamig:
Yes, IntelliJ IDE has the option to do that. I did not pay much attention to the formatting part and caused so may comments around it

dashboard/static/js/controllers.js
pushyamig: Ok, I changed it as per the second suggestion

dashboard/views.py
pushyamig: Yes, I fixed it with IDE help.

dashboard/views.py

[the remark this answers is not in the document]
pushyamig: Done on Intellj the option for formatting is `alt+cmd L`

dashboard/static/js/assign-progress.js
pushyamig: I think the purpose of `$(document).ready()` to wait to do our stuff until the DOM is fully loaded. So maybe the code that is in there wait and do it supposed to be doing. It doesn't hurt to wrap the code in the `$(document).ready)`

Issue #287, opened 2018-11-02
Add more version/build information to django admin


jonespm:
It was suggested on #284 to add more information. I believe we could add this to the Django Admin as a module.

`OPENSHIFT_BUILD_NAMESPACE: ccm OPENSHIFT_BUILD_NAME: canvascoursemanager-12 GIT_SOURCE: https://github.com/tl-its-umich-edu/canvas_course_manager.git GIT_BRANCH: 4.0.1 GIT_COMMIT: fb68ce8d56741338f6c1fa414201be5c19b91023`

pushyamig:
Done

PR #303, opened 2018-11-19
#287 adding build info to UI

pushyamig:
The build info is displayed in the admin. Let me know your comments on this
screen shot 2018-11-19 at 3 10 43 pm
view only

### DIFF
--- a/dashboard/common/utils.py
+++ b/dashboard/common/utils.py
@@ -1,0 +1,13 @@
+import logging
+from decouple import config
+logger = logging.getLogger(__name__)
+
+
+def get_build_info():
+    logger.info(get_build_info.__name__)
+    git_commit=config("OPENSHIFT_BUILD_COMMIT",default="")
+    build_namespace=config("OPENSHIFT_BUILD_NAMESPACE",default="")
+    git_branch=config("OPENSHIFT_BUILD_REFERENCE",default="")
+    build_source=config("OPENSHIFT_BUILD_SOURCE",default="")
+    build_name=config("OPENSHIFT_BUILD_NAME",default="")
+    return 'Build_Namespace:'+build_namespace+' Build_Name:'+build_name+' Git_Source: '+build_source+' Git_Branch:'+git_branch+' Git_Commit:'+git_commit
--- a/dashboard/context_processors.py
+++ b/dashboard/context_processors.py
@@ -1,8 +1,10 @@
 from dashboard.common import db_util
+from dashboard.common import utils
 
 import logging
 
 logger = logging.getLogger(__name__)
+
 
 def course_name(request):
     course_id = str(request.resolver_match.kwargs.get('course_id'))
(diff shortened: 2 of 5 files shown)


jonespm:
If don't think we need to hide it behind a button if this is already on as a super user check. I was thinking of just adding it to the django admin template page, but this makes sense too to just check for super user and display as the footer.

I wonder if we should display anything to a regular user? Maybe the git commit? Or just hide that entirely?

dashboard/templates/home.html
```diff
@@ -1,5 +1,9 @@
 {% extends 'base.html' %}
 {% load static %}
 {% block content %}
+{% if user.is_superuser %}
+<button style="position:fixed;left:0;top:60px;width:20px;height:20px;background:transparent;border: none !important;"
```
jonespm: As mentioned in the comments I'd probably remove this button and just display this informational div all the time if the user is_superuser.

dashboard/context_processors.py
```diff
@@ -31,5 +35,11 @@ def course_view_option(request):
...
 
+
 def last_updated(request):
-    return {'last_updated' : db_util.get_canvas_data_date()}
\ No newline at end of file
+    return {'last_updated' : db_util.get_canvas_data_date()}
+
+
+def get_build_info(request):
+    logger.info("are you atleat here")
```
jonespm: This logger can be removed

dashboard/common/utils.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+def get_build_info():
+    logger.info(get_build_info.__name__)
+    git_commit=config("OPENSHIFT_BUILD_COMMIT",default="")
+    build_namespace=config("OPENSHIFT_BUILD_NAMESPACE",default="")
+    git_branch=config("OPENSHIFT_BUILD_REFERENCE",default="")
+    build_source=config("OPENSHIFT_BUILD_SOURCE",default="")
+    build_name=config("OPENSHIFT_BUILD_NAME",default="")
+    return 'Build_Namespace:'+build_namespace+' Build_Name:'+build_name+' Git_Source: '+build_source+' Git_Branch:'+git_branch+' Git_Commit:'+git_commit
```
jonespm: This can be written cleaner with fstrings

```suggestion
    return f'Build_Namespace:{build_namespace} Build_Name: {build_name} Git_Source: {build_source} Git_Branch: {git_branch} Git_Commit: {git_commit}'
```

pushyamig: Done

dashboard/context_processors.py
pushyamig: Done

dashboard/templates/home.html
pushyamig: This is just for us for checking build info. I think this will work for now.

pushyamig:
I don't think we need to present to anything to the regular user.
 We just want to check after the new deploy what is the version it is showing. For that, this was the simplest I felt to do. I did not find it an easy way to add a single line to the admin template. I felt like I need to create python file to add this to the template so I did not see the need for doing it.

PR #305, opened 2018-11-19
Add database table to store the courses we currently support

### LINKED ISSUE #138: Add a database table to store the courses that we currently support
I was thinking of just using a database table that could be maintained through the django admin kind of like
https://djangobook.com/adding-models-to-django-admin/

Where we could have like supported_courses.

In that table we could put all of the course ids we want to allow onto the dashboard. Then the cron process would load them all in. Currently this is hardcoded as a CSV CANVAS_COURSE_IDS(but it really only contains one value)

jonespm:
This fixes #138 

This is currently in progress as it doesn't use this table yet, it still uses the environment variable. 

Not all of the models are 100% yet either as it was created from a dump.

### DIFF
--- a/dashboard/models.py
+++ b/dashboard/models.py
@@ -1,0 +1,166 @@
+# This is an auto-generated Django model module.
+# You'll have to do the following manually to clean this up:
+#   * Rearrange models' order
+#   * Make sure each model has one field with primary_key=True
+#   * Make sure each ForeignKey has `on_delete` set to the desired behavior.
+#   * Remove `managed = False` lines if you wish to allow Django to create, modify, and delete the table
+# Feel free to rename the models, but don't rename db_table values or field names.
+from __future__ import unicode_literals
+
+from django.db import models
+import logging
+logger = logging.getLogger(__name__)
+
+
+class AcademicTerms(models.Model):
+    term_id = models.AutoField(primary_key=True, verbose_name="Term Id")
+    name = models.CharField(max_length=255, verbose_name="Name")
+    start_date = models.DateField(verbose_name="Start Date")
+    end_date = models.DateField(verbose_name="End Date")
+
+    def __str__(self):
+        return self.name
+
+    class Meta:
+        managed = False
+        db_table = 'academic_terms'
+        verbose_name = "Academic Terms"
+        verbose_name_plural = "Academic Terms"
+
+class Assignment(models.Model):
+    id = models.CharField(primary_key=True, max_length=255, verbose_name="Assignment Id")
+    name = models.CharField(max_length=255, verbose_name="Name")
+    due_date = models.DateTimeField(blank=True, null=True, verbose_name="Due DateTime")
+    local_date = models.DateTimeField(blank=True, null=True, verbose_name="Local DateTime")
+    points_possible = models.CharField(max_length=255, blank=True, null=True, verbose_name="Points Possible")
+    course_id = models.CharField(max_length=255, verbose_name="Course Id")
+    assignment_group_id = models.CharField(max_length=255, verbose_name="Assignment Group Id")
+
+    def __str__(self):
+        return self.name
+
+    class Meta:
+        managed = False
+        db_table = 'assignment'
+
+
+class AssignmentGroups(models.Model):
+    id = models.CharField(primary_key=True, max_length=255, verbose_name="Assignment Group Id")
+    name = models.CharField(max_length=255, verbose_name="Name")
+    weight = models.CharField(max_length=255, blank=True, null=True, verbose_name="Weight")
+    group_points = models.CharField(max_length=255, blank=True, null=True, verbose_name="Group Points")
+    course_id = models.CharField(max_length=255, verbose_name="Course Id")
+    drop_lowest = models.CharField(max_length=255, blank=True, null=True, verbose_name="Drop Lowest")
+    drop_highest = models.CharField(max_length=255, blank=True, null=True, verbose_name="Drop Highest")
+
+    def __str__(self):
+        return self.name
+
+    class Meta:
+        managed = False
+        db_table = 'assignment_groups'
+        verbose_name = "Assignment Groups"
+        verbose_name_plural = "Assignment Groups"
+
+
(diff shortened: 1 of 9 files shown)


DanielDugan:
I think this is valuable to have once we start getting more courses.
Codacy doesn't like the lowercase starting characters for class attribute names, but I am ok with those for certain common variables. (i.e id and logger)
"Class attribute name "id" doesn't conform to '([A-Za-z_][A-Za-z0-9_]{2,30}|(__.*__))$' pattern"

jonespm:
Yeah, I'm going to relax this pattern to
'([A-Za-z_][A-Za-z0-9_]{1,30}|(.*))$'

This is the name in the database. For the logger, it should match the constant name patterns I'll create a new issue for that.

These Codacy issues are recommendations that we can configure or disable.

dashboard/models.py
```diff
@@ -0,0 +1,166 @@
...
+class Course(models.Model):
+    id = models.CharField(primary_key=True, max_length=255, verbose_name="Canvas Course Id")
+    term_id = models.ForeignKey(AcademicTerms, verbose_name="Term Id", on_delete=models.SET_NULL, db_column='term_id')
+    name = models.CharField(max_length=255, verbose_name="Name")
+
+    def __str__(self):
+        return self.name
+
+    class Meta:
+        managed = False
```
pushyamig: what does the `managed=False` mean i.e what is the purpose/use?

pushyamig:
This is great as we wanted to do this for a log time. We don't want to do too much work as we will be relying on DPDB for data for visuals

pushyamig:
Since we have Django models now, how will it alter the behaviour of the queries that we are making in the `views.py`?

dashboard/models.py
zqian: From the above comment, line 6: 
"Remove `managed = False` lines if you wish to allow Django to create, modify, and delete the table"

dashboard/models.py
```diff
@@ -0,0 +1,166 @@
...
+class Submission(models.Model):
+    id = models.CharField(primary_key=True, max_length=255, verbose_name="Submission Id")
+    assignment_id = models.CharField(max_length=255, verbose_name="Assignment Id")
+    course_id = models.CharField(max_length=255, verbose_name="Course Id")
+    user_id = models.CharField(max_length=255, verbose_name="User Id")
+    score = models.CharField(max_length=255, blank=True, null=True, verbose_name="Score")
+    graded_date = models.DateTimeField(blank=True, null=True, verbose_name="Graded DateTime")
+
+    def __str__(self):
+        return "TODO"
```
zqian: update the "TODO"

dashboard/models.py
```diff
@@ -0,0 +1,166 @@
...
+class UnizinMetadata(models.Model):
+    pkey = models.CharField(primary_key=True, max_length=20, verbose_name="Key")
+    pvalue = models.CharField(max_length=100, blank=True, null=True, verbose_name="Value")
+
+    class Meta:
+        managed = False
+        db_table = 'unizin_metadata'
+
+
+class User(models.Model):
```
zqian: at some point of time, we can separate the user information from the registration information, by splitting this into two tables.

dashboard/models.py
```diff
@@ -0,0 +1,166 @@
...
+
+    class Meta:
+        managed = False
+        db_table = 'user'
+        unique_together = (('id', 'course_id'),)
+
+class FileAccess(models.Model):
+    file = models.OneToOneField(File, on_delete=models.CASCADE, primary_key=True, verbose_name="File")
+    user = models.OneToOneField(User, on_delete=models.CASCADE, primary_key=True, verbose_name="User")
+    access_time = models.DateTimeField(verbose_name="Access Time")
```
zqian: this should also be in the primary key

dashboard/models.py
```diff
@@ -0,0 +1,166 @@
...
+        managed = False
+        db_table = 'unizin_metadata'
+
+
+class User(models.Model):
+    id = models.CharField(primary_key=True, max_length=255, verbose_name="User Id")
+    name = models.CharField(max_length=255, verbose_name="Name")
+    sis_id = models.CharField(max_length=255, blank=True, null=True, verbose_name="SIS Id")
+    sis_name = models.CharField(max_length=255, blank=True, null=True, verbose_name="SIS Name")
+    course_id = models.CharField(max_length=255, verbose_name="Course Id")
```
pushyamig: course_id should also be the primary key for the cases the student is part of the 2 courses, it won't throw the id already error

jonespm: Thanks I'll double check those are correct in the model.

I only verified a few of the tables are actually correct as I'm not using everything here yet.

dashboard/models.py
jonespm: Right, currently these are being created by the init.sql script. If we want the model to create these tables we'll have to modify the script. This could be changed in a future commit.

dashboard/models.py
jonespm: I think leaving managed=False in there is "probably" safer for now. I can create a new issue #317 for this. I updated the TODO

dashboard/models.py
jonespm: In the django model this is represented by the Meta 
        unique_together = (('id', 'course_id'),)

This was already set.

dashboard/models.py
jonespm: I don't see any primary key in the original table. It probably should be a unique_together on file_id and user_id right?

jonespm:
I've almost got this done, I just need to update the README for what the current process is and test it a little more. The process will change though after #315

jonespm:
This is complete, this is going to require a new column in courses (canvas_id) and populated with the canvas id. This is just a UI assistance since everything in the database has the UDW_ID prefix.

dashboard/models.py
jonespm: Yeah, it sounds like this is going to be a future feature

zqian:
PR #319 contains this PR content. Unfortunately, I have put my comments there already. So I will mark this PR as done.

PR #308, opened 2018-11-19
Improve the LOGGING block 

### LINKED ISSUE #293: Cron logs should be timestamped
There is no timestamp in the current cron logs, these should be timestamped for easier parsing. The existing times that are inserted should be removed. This might be related to some setting in LOGGING. 

Also I noticed we still have duplicates for some things, like the SQL debug events.

jonespm:
Fixes #293

### DIFF
--- a/dashboard/settings.py
+++ b/dashboard/settings.py
@@ -181,14 +181,24 @@ STATICFILES_FINDERS = (
 LOGGING = {
     'version': 1,
     'disable_existing_loggers': False,
+    # Gunicorns logging format https://github.com/benoitc/gunicorn/blob/19.x/gunicorn/glogging.py
+    'formatters': {
+        "generic": {
+            "format": "%(asctime)s [%(process)d] [%(levelname)s] %(message)s",
+            "datefmt": "[%Y-%m-%d %H:%M:%S %z]",
+            "class": "logging.Formatter",
+        }
+    },
     'handlers': {
         'console': {
             'class': 'logging.StreamHandler',
+            'formatter': 'generic',
         },
     },
     'loggers': {
         'django': {
             'handlers': ['console'],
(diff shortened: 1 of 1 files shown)


dashboard/settings.py
```diff
@@ -181,14 +181,24 @@
 LOGGING = {
     'version': 1,
     'disable_existing_loggers': False,
+    # Gunicorns logging format https://github.com/benoitc/gunicorn/blob/19.x/gunicorn/glogging.py
+    'formatters': {
+        "generic": {
+            "format": "%(asctime)s [%(process)d] [%(levelname)s] %(message)s",
+            "datefmt": "[%Y-%m-%d %H:%M:%S %z]",
```
pushyamig: I feel we should be log time in the EST than UTC

dashboard/settings.py
```diff
@@ -181,14 +181,24 @@
...
         'console': {
             'class': 'logging.StreamHandler',
+            'formatter': 'generic',
         },
     },
     'loggers': {
         'django': {
             'handlers': ['console'],
+            'propagate': False,
             'level': config('DJANGO_LOG_LEVEL', default='INFO'),
```
pushyamig: Currently, the prod logging level is Warning, I would make it INFO at least so that we could watch for errors

dashboard/settings.py
jonespm: I think it was logging in the timezone of the server, which is set to Eastern time in the Dockerfile, I can double check this though. The %z just confirms what time zone it's in.

dashboard/settings.py
jonespm: This is configurable via env variable. I don't think anything needs to be done here. This is just the default.

PR #319, opened 2018-12-04
Support more than one term. Connect terms better with courses.

### LINKED ISSUE #315: Support more than one term connect terms better with courses
Currently in this project term_id is hardcoded to "2" which equals "FALL 2018"

To support Winter 2018 we'll have to manually change this term everywhere in the code. This isn't an ideal solution so this should be fixed. 

Here's the idea I've got
- Remove the current academic_terms table since it doesn't match the UDW data
- Remove everything that references this table

- Retrieve the terms from the UDW enrollment_dim, populate the local database
  - This should only retrieve "recent" terms since really old ones are supported either
    - Filter based on date_start and maybe only pull the terms > 2018-06-27 (when this project started)
    - Filter based on the earliest event time in enriched_events (currently) 2016-11-15
  - We can just have this date as a variable

Then the software can determine what term we're in as well as link this term up to the courses correctly. 

Questions: 
- When it gets to the end of the semester and the site is left on what should happen? 

  - The file access attempts to get the current week number but this will fail on the next semester. 
    - Maybe just leave it with the entire bar selected?

jonespm:
Fixes #315 

Needs to be rebased when #305 is merged.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,9 +36,29 @@ The bq_cred.json is service account for Big Query, it needs to be supplied and p
 
 ## Add a default course to test with
 
-TODO
+There are some default courses pre-populated by the mysql/init.sql script. If you'd like to add additional courses you need to add them as an admin user. (See next section on adding an admin user)
 
-## Load user, file, file access data into database (Local)
+## Create a super user to test login. 
+
+On the both local dev and remote the users are stored in the local database. However on local the users have to be created via the command line, on Openshift they are created either manually in the database or when logged in via Shibboleth.
+
+# Localhost (Dev) process
+
+The password will be displayed to the scree unless you specify it with --password. You can run this multiple times to change a password but you need to delete/modify super users via the Admin login (appears when logged in as admin). You can also add new users in there.
+
+`docker exec -it student_dashboard python manage.py createuser --superuser --username=root --email=root@example.edu`
+
+You can create regular users to test with without the superuser flag via the command line without using the --superuser.
+
+`docker exec -it student_dashboard python manage.py createuser --username=student --email=student@example.edu`
+
+Note: You can also make a user a "super user" by connecting to the database, editing the record in auth_user and setting is_staff=1 and is_superuser=1.
+
+# Openshift process
+
+You should login via Shibboleth into the application. Once you do that for the first admin you'll have to go into the database auth_user table and change is_staff and is_superuser to both be true. After doing this you can change future users with any admin via the GUI.
+
+## Load user, file, file access data into database
 Users and files are loaded now with the cron job. This is run on a separate pod in Openshift when the environment variable `IS_CRON_POD=true`.
 
 Crons are configured in this project with django-cron. Django-cron is executed whenever `python manage.py runcrons` is run but it is limited via a few environment variables.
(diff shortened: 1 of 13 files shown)


dashboard/cron.py
```diff
@@ -351,15 +329,31 @@ def weight_consideration(self):
...
+
     def do(self):
         logger.info("************ dashboard cron tab")
 
         status = ""
 
         status += "Start cron at: " +  str(datetime.datetime.now()) + ";"
 
-        logger.info("************ course")
-        status += self.update_with_udw_course()
```
zqian: Not sure why the course update function was deleted...

dashboard/cron.py
```diff
@@ -110,7 +88,7 @@ def update_with_udw_user(self):
         status += deleteAllRecordInTable("user")
 
         # loop through multiple course ids
-        for UDW_course_id in settings.DEFAULT_UDW_COURSE_IDS:
+        for UDW_course_id in db_util.get_supported_courses():
```
zqian: I do not see the get_supported_courses() function defined in db_util.

dashboard/cron.py
```diff
@@ -170,7 +148,7 @@ def update_with_udw_file(self):
         status += deleteAllRecordInTable("file")
 
         #select file record from UDW
-        for UDW_course_id in settings.DEFAULT_UDW_COURSE_IDS:
+        for UDW_course_id in db_util.get_supported_courses():
```
zqian: I do not see the get_supported_courses() function defined in db_util.

migrations/r_12_05.sql
```diff
@@ -0,0 +1,13 @@
+ALTER TABLE academic_terms CHANGE term_id id BIGINT UNSIGNED;
+ALTER TABLE academic_terms CHANGE start_date date_start TIMESTAMP DEFAULT CURRENT_TIMESTAMP;
+ALTER TABLE academic_terms CHANGE end_date date_end TIMESTAMP DEFAULT CURRENT_TIMESTAMP;
+
+ALTER TABLE academic_terms ADD COLUMN canvas_id CHAR(255) AFTER id;
+
+ALTER TABLE course MODIFY term_id BIGINT UNSIGNED NOT NULL DEFAULT 1;
+
+UPDATE academic_terms SET id = 17700000000000108, canvas_id = "108" WHERE id = 1;
```
zqian: Are those migrations/r_12_xx.sql script just for localhost deploy? I see the hardcoded canvas_id here.

README.md
```diff
@@ -34,7 +34,32 @@ The bq_cred.json is service account for Big Query, it needs to be supplied and p
 `docker cp ~/secrets student_dashboard:/secrets`
 5. Initialize the MySQL database by loading the users and files on the next step. You'll need to be on VPN for this to work.
 
-## Load user, file, file access data into database (Local)
+## Add a default course to test with
+
+<<<<<<< HEAD
```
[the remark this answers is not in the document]
jonespm: That looks like a mistake in a merge, will fix thanks!

dashboard/settings.py
```diff
@@ -320,19 +320,13 @@
...
 
-# TODO: Select a default course for user from database, just here so things don't all break
-DEFAULT_COURSE_IDS = config("CANVAS_COURSE_IDS", default="0", cast=Csv())
-
-# Course IDS with the UDW ID appended to them
-DEFAULT_UDW_COURSE_IDS = [UDW_ID_PREFIX + course_id for course_id in DEFAULT_COURSE_IDS]  
-
-# set the current term id from config
-CURRENT_CANVAS_TERM_ID =config('CURRENT_CANVAS_TERM_ID', default="2")
+# set the earliest start date for terms
```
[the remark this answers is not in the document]
jonespm: That's right. I didn't want older (2015, 2016) terms before we actually could possibly have data to show up in the terms dropdown. Eventually we'll probably want something better like maybe a default filter to only show the terms for the current year? 

It looks like Canvas has a variety of dates on it's terms table (like dates instructors can access) but I don't see those dates exposed in the enrollment_term_dim.

dashboard/cron.py
jonespm: This was deleted because courses are intended to be populated manually now via the admin UI as part of #305. We didn't need to have a cron to update this table anymore. (Some of this came over from that PR)

dashboard/cron.py
jonespm: I don't either, I think it was missing from some other change on #305, I'll need to merge that first, this was based on that commit.

jonespm: I'll merge that and rebase this and it should be there.

migrations/r_12_05.sql
jonespm: This conversion script is actually for production. These are the hard coded ids in our prod instance for academic terms. This conversion script *should* work fine with production. With dev we'll probably have to wipe out the data and start fresh. This academic_terms table will populate from canvas during the cron runs. I'll backup the database table before we roll these changes out though.

PR #322, opened 2018-12-07
 Support more than one term. Connect terms better with courses.

jonespm:
Fixes issue #315

### DIFF
--- a/dashboard/views.py
+++ b/dashboard/views.py
@@ -53,15 +53,15 @@ def gpa_map(grade):
 
 def get_current_week_number(request, course_id=0):
     # get current term start date
-    term_start_date = AcademicTerms.objects.filter(course__id=course_id).start_date
+    term_date_start = AcademicTerms.objects.filter(course__id=course_id).date_start
 
     today = datetime.now().date()
 
-    logger.info(term_start_date.date())
+    logger.info(term_date_start.date())
     logger.info(today)
 
     ## calculate the week number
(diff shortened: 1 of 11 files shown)


jonespm:
Thanks, I'm going to review the Codacy issues and merge. It also looks like the migrations as written were only able to be applied manually. I think it would be nice if migrations could be automated in the future, and the database structure could also be managed by Django. This would take a little work. Maybe we can just wait and see what UDP brings.

dashboard/views.py
```diff
@@ -411,9 +407,9 @@ def is_weight_considered(course_id):
 
 def get_term_dates_for_course(course_id):
     logger.info(get_term_dates_for_course.__name__)
-    sql = "select a.start_date from course c, academic_terms a where c.id = %(course_id)s and c.term_id=a.term_id;"
-    df = pd.read_sql(sql, conn, params={"course_id": course_id}, parse_dates={'start_date': '%Y-%m-%d'})
-    return df['start_date'].iloc[0]
+    sql = "select a.date_start from course c, academic_terms a where c.id = %(course_id)s and c.term_id=a.id;"
+    df = pd.read_sql(sql, conn, params={"course_id": course_id}, parse_dates={'date_start': '%Y-%m-%d'})
+    return df['date_start'].iloc[0]
```
pushyamig: can't we use models to get this piece of information instead of querying the DB?

PR #325, opened 2018-12-10
Ignore null user ids in cron run

### LINKED ISSUE #324: Cron run giving integrity error when null user id returned
It seems like for some reason duplicate user(s) with null id's are being returned. These users should be ignored. These are currently failing the cron for one course.

jonespm:
Fixes #324

### DIFF
--- a/dashboard/cron.py
+++ b/dashboard/cron.py
@@ -48,9 +48,6 @@ def util_function(UDW_course_id, sql_string, mysql_table, table_identifier=None)
 
     # drop duplicates
     df.drop_duplicates(keep='first', inplace=True)
-    
-    logger.debug("Duplicates dropped")
-    logger.debug(df)
 
     logger.debug(" table: " + mysql_table + " insert size: " + str(df.shape[0]))
 
@@ -58,7 +55,7 @@ def util_function(UDW_course_id, sql_string, mysql_table, table_identifier=None)
     try:
         df.to_sql(con=engine, name=mysql_table, if_exists='append', index=False)
     except Exception as e:
-        logger.debug(e)
+        logger.warn(e)
         raise
 
     # returns the row size of dataframe
--- a/dashboard/cron.py
+++ b/dashboard/cron.py
@@ -48,11 +48,18 @@ def util_function(UDW_course_id, sql_string, mysql_table, table_identifier=None)
 
     # drop duplicates
     df.drop_duplicates(keep='first', inplace=True)
+    
+    logger.debug("Duplicates dropped")
+    logger.debug(df)
 
     logger.debug(" table: " + mysql_table + " insert size: " + str(df.shape[0]))
 
     # write to MySQL
-    df.to_sql(con=engine, name=mysql_table, if_exists='append', index=False)
+    try:
+        df.to_sql(con=engine, name=mysql_table, if_exists='append', index=False)
+    except Exception as e:
+        logger.debug(e)
+        raise
 
     # returns the row size of dataframe
     return f"inserted + {str(df.shape[0])} rows in table {mysql_table} for course {UDW_course_id};"
@@ -107,7 +114,8 @@ class DashboardCronJob(CronJobBase):
                         and e.workflow_state='active' ) as e 
                         where p.user_id=u.id 
                         and u.id = e.user_id 
-                        and c.enrollment_id =  e.enrollment_id
+                        and c.enrollment_id =  e.enrollment_id 
+                        and p.sis_user_id is not null
                         """
             logger.debug(user_sql)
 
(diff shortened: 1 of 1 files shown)


jonespm:
I could not figure out any easy way in Pandas to_sql to ignore this IntegrityError. I saw this open issue here
https://github.com/pandas-dev/pandas/issues/15988

And I saw a suggestion on Stackoverflow about doing each insert individually, but this makes the process a LOT slower. It could also do manipulation on the dataframe rather than in the query but this isn't technically a duplicate.

I added the catch for the Exception here as that will not print out which primary keys cause the exception but to_sql loads them all at once so not sure what more we can do for that.

dashboard/cron.py
```diff
@@ -48,11 +48,18 @@ def util_function(UDW_course_id, sql_string, mysql_table, table_identifier=None)
 
     # drop duplicates
     df.drop_duplicates(keep='first', inplace=True)
+    
+    logger.debug("Duplicates dropped")
```
zqian: Not sure about the usage of the debug message here.

dashboard/cron.py
```diff
@@ -48,11 +48,18 @@ def util_function(UDW_course_id, sql_string, mysql_table, table_identifier=None)
...
+    logger.debug(df)
 
     logger.debug(" table: " + mysql_table + " insert size: " + str(df.shape[0]))
 
     # write to MySQL
-    df.to_sql(con=engine, name=mysql_table, if_exists='append', index=False)
+    try:
+        df.to_sql(con=engine, name=mysql_table, if_exists='append', index=False)
+    except Exception as e:
+        logger.debug(e)
```
zqian: use warn level instead of debug, and also log the mysql_table value. 

Not sure whether we can differentiate different Exception types in Python, and handle them differently.

dashboard/cron.py
jonespm: I thought the issue was in the duplicate removal. @pushyamig commented that the SQL could possibly be improved. I could remove this message but it still was useful to see the dataframe before and after duplicates to rule that out as a problem.

dashboard/cron.py
jonespm: Yeah, that makes sense. The cron pod "probably" has debug on but warn is better.

The problem here is there isn't a lot to log. This is all done in the pandas function to_sql. Normally you can differentiate Exception types but the only Exception this looks like it throws is
```
ValueError
When the table already exists and if_exists is ‘fail’ (the default).
```

Everything else is an error from SQL Alchemy and there really is no recovery from this that I could see. It just tries to insert the entire dataframe at once here and if one value fails it all fails. This was failing because we were getting a null value back for one of the primary keys.

We could work around this by inserting one row at a time as suggested like here
https://stackoverflow.com/questions/30337394/pandas-to-sql-fails-on-duplicate-primary-key/35881631

But that would really slow things down. In any case, we cannot have a null as the primary.

dashboard/cron.py
```diff
@@ -48,17 +48,14 @@ def util_function(UDW_course_id, sql_string, mysql_table, table_identifier=None)
...
-    logger.debug(df)
 
     logger.debug(" table: " + mysql_table + " insert size: " + str(df.shape[0]))
 
     # write to MySQL
     try:
         df.to_sql(con=engine, name=mysql_table, if_exists='append', index=False)
     except Exception as e:
-        logger.debug(e)
+        logger.warn(e)
```
zqian: How about log the mysql_table name in the log file?

Otherwise, it is OK to merge

dashboard/cron.py
```diff
@@ -52,7 +52,11 @@ def util_function(UDW_course_id, sql_string, mysql_table, table_identifier=None)
     logger.debug(" table: " + mysql_table + " insert size: " + str(df.shape[0]))
 
     # write to MySQL
-    df.to_sql(con=engine, name=mysql_table, if_exists='append', index=False)
+    try:
+        df.to_sql(con=engine, name=mysql_table, if_exists='append', index=False)
+    except Exception as e:
+        logger.warn(e)
```
pushyamig: I feel this should be logged as an error and not warn.

dashboard/cron.py
jonespm:
```suggestion
        logger.exception(f"Error running to_sql on table {mysql_table}")
```

dashboard/cron.py
jonespm: Yup! I changed this to logger.exception which logs as an error and prints a message.

jonespm: It prints out something like 

```
[2018-12-11 13:16:23 -0500] [34] [ERROR] Error running to_sql on table user
  File "/usr/local/lib/python3.6/site-packages/sqlalchemy/engine/base.py", line 1170, in _execute_context
    context)
  File "/usr/local/lib/python3.6/site-packages/sqlalchemy/dialects/mysql/mysqldb.py", line 105, in do_executemany
    rowcount = cursor.executemany(statement, parameters)
```

Issue #320, opened 2018-12-05
Assignment Planning: Add the average grade on an assignment to the mouse-over text


jennlove-um:
In the assignment details, add the average grade in the mouse-over.  This should show in the progress bar mouse-over.

Business Rules: 
Canvas allows instructors to hide the grade distribution by assignment in a course.  If they have hidden this data in the course, we should not show it in the visualization.

Canvas only shows the grade distribution if 5 other students have submitted the assignment.  Our visualization should follow the same logic.

jennlove-um:
[https://community.canvaslms.com/docs/DOC-9540-4212724430](url)

pushyamig:
A sample set of queries to validate this issue
1. select s.*,a.name from submission s join assignment a on s.assignment_id=a.id and s.course_id='17700000000245664';

pushyamig:
@jennlove-um fixed the minor formatting issue

jennlove-um:
Update the online help to include the average assignment grade information.  

I'll add this before closing this issue.

jennlove-um:
Online help has been updated.

PR #326, opened 2018-12-13
#320 getting the avg grade of assignment

pushyamig:
This pull request is for #320

### DIFF
--- a/dashboard/cron.py
+++ b/dashboard/cron.py
@@ -304,8 +304,9 @@ class DashboardCronJob(CronJobBase):
                              sub_with_enroll as (select sf.* from sub_fact sf join enrollment e on e.user_id = sf.user_id),
                              submission_time as (select sd.id, sd.graded_at from submission_dim sd join sub_fact suf on sd.id=suf.submission_id),
                              assign_fact as (select s.*,a.title from assignment_dim a join sub_with_enroll s on s.assignment_id=a.id where a.course_id='{UDW_course_id}' and a.workflow_state='published' and muted = false),
-                             assign_sub_time as (select a.*, t.graded_at from assign_fact a join submission_time t on a.submission_id = t.id)
-                             select submission_id AS id, assignment_id AS assignment_id, course_id, global_canvas_id AS user_id, published_score AS score, graded_at AS graded_date from assign_sub_time
+                             assign_sub_time as (select a.*, t.graded_at from assign_fact a join submission_time t on a.submission_id = t.id),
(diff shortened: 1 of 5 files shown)


dashboard/views.py
```diff
@@ -363,10 +373,10 @@ def get_user_assignment_submission(current_user,assignments_in_course_df, course
     return assignment_submissions
 
 
-def hide_score_on_mute(row):
-    if row['grade_muted'] and row['score'] is not None:
-        return None
-    else:return row['score']
```
pushyamig: `hide_Score_on_mute` is a redundant probably did not clean up when pulling only graded that are unmuted that is done as part of cron sql query earlier.

dashboard/cron.py
```diff
@@ -304,8 +304,9 @@ def submission(self):
...
                              assign_fact as (select s.*,a.title from assignment_dim a join sub_with_enroll s on s.assignment_id=a.id where a.course_id='{UDW_course_id}' and a.workflow_state='published' and muted = false),
-                             assign_sub_time as (select a.*, t.graded_at from assign_fact a join submission_time t on a.submission_id = t.id)
-                             select submission_id AS id, assignment_id AS assignment_id, course_id, global_canvas_id AS user_id, published_score AS score, graded_at AS graded_date from assign_sub_time
+                             assign_sub_time as (select a.*, t.graded_at from assign_fact a join submission_time t on a.submission_id = t.id),
+                             all_assign_sub as (select submission_id AS id, assignment_id AS assignment_id, course_id, global_canvas_id AS user_id, published_score AS score, graded_at AS graded_date from assign_sub_time order by assignment_id)
+                             select f.*, f1.avg_score from all_assign_sub f join (select assignment_id, round(avg(score),1) as avg_score from all_assign_sub group by assignment_id) as f1 on f.assignment_id = f1.assignment_id
```
jonespm: Okay I had some concerns about the f.* incase the table changed it might break our query, but I see that it's pulling from the WITH query table so it's safe.

dashboard/views.py
```diff
@@ -363,10 +373,10 @@ def get_user_assignment_submission(current_user,assignments_in_course_df, course
...
+def no_show_avg_score_for_ungraded_assignments(row):
```
jonespm: Could you add some comments to this method? Why are you returning avg_score if the score is None?

pushyamig: I have added comments.  ` don't show the avg scores for a student when an individual assignment is not graded as canvas currently don't show it`
I think I return N/A for avg_score when the individual score is not available.

dashboard/cron.py
pushyamig: line 308 return all the columns the DB has and I did not want to make the line longer than required so I used ` f.*`.

PR #333, opened 2018-12-24
Provide a JSON feed for the courses supported

### LINKED ISSUE #332: Provide a JSON feed that would give the courses supported and the features supported
If we had a JSON feed then we could have Canvas load this to avoid having to configure Canvas every term.

This is an example of how we could add this to the Canvas theme script
https://github.com/robert-carroll/ccsd-canvas/blob/master/admintray-subaccmenu/admintray-subaccmenu.inc.js

I'm thinking JSON like this should work?

```
{
  "241011": {
    "files": "true",
    "assigment": "true",
    "grade": "true"
  },
  "029921": {
    "files": "true",
    "assigment": "true",
    "grade": "false"
  }
}
```

jonespm:
Also fixes a few other pylint issues that I'd found and an issue with
the default SQL mapping

Fixes #332

### DIFF
--- a/dashboard/models.py
+++ b/dashboard/models.py
@@ -17,7 +17,7 @@ class AcademicTermsQuerySet(models.QuerySet):
     def course_date_start(self, course_id):
         try:
             return self.get(course__id=str(course_id)).date_start
-        except AcademicTerms.DoesNotExist:
+        except self.model.DoesNotExist:
             logger.debug(f"Could not find term for course {course_id}")
             return datetime.min
 
@@ -99,7 +99,7 @@ class CourseQuerySet(models.QuerySet):
         """
         try:
             return self.values_list('id', flat=True)
-        except Course.DoesNotExist:
+        except self.model.DoesNotExist:
             logger.info("Courses did not exist", exc_info = True)
         return []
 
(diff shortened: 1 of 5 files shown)


jonespm:
Directly it would yes. @zqian and I talked before the break about this. I believe it would be easy to cache the URL on Amazon Cloudfront and never go to it directly. This is where Canvas gets it's cached data from anyway. If that for some reason didn't work we could get some process added to the daily cron [upload directly to Canvas](https://canvas.instructure.com/doc/api/file.file_uploads.html). Possibly with the UM API. 

For immediate testing (after this is merged) we could just copy/paste the JSON directly into the Canvas scripts and then switch over to using Cloudfront later.

pushyamig:
uploading directly to canvas seems to be a viable solution than using Amazon services. AWS services come at a price as we will be paying the price of canvas load. 
If we can load the feed/file to Canvas as part our daily cron set up then why do we need AWS service. It would be good to exercise this with cloud solution for short-term goal but not the long-term.

dashboard/models.py
```diff
@@ -140,6 +140,23 @@ def __str__(self):
...
+        :return: JSON formatted CourseViewOption
+        :rtype: Dict
+        """
+
+        return {
+            self.course.canvas_id: {
+                'cn': self.course.name, 
+                'fa': int(self.show_files_accessed),
+                'ap': int(self.show_assignment_planning),
+                'gd': int(self.show_grade_distribution),
```
pushyamig: will the json feed will have the abbreviated properties like `cn`,`fa`? I feel we should have clear names for readability purposes

jonespm:
Yeah, the Cloudfront solution *should* be really inexpensive but don't know 100%. Each course supported adds ~50 bytes so 20 courses would be 1KB. We'd only invalidate the cache at most daily so even if we had 50,000 unique users a day (unlikely especially on weekends) it would still be less than a 2 million requests month. This is less than $1. ;)

dashboard/models.py
jonespm: I have it abbreviated for bandwidth cost savings. We're charged by KB and also it would be a quicker download for the user. This will be 100% machine read. It lowers the per course size from about 100 bytes to like 50.

jonespm: This isn't a *huge* savings right now (maybe a few dollars) but if we supported a lot of courses it would be could be a lot. Though we might have to think of a better solution at that time.

jonespm: Look at #332 it's an example of the JSON generated by this.

pushyamig:
ok, convinced! if this is not going to cost us a lot then we should try CloudFront option for the winter term and see benefit and loss for using it

Issue #329, opened 2018-12-18
Files visualization needs to account for full number of weeks in term


jennlove-um:
The Files Accessed default is not working correctly in finals week.  Now that we're in week 16 (finals week), the default is showing up as Wk2 to Wk 2 since there is no week 16 in the visualization.

The default needs to work all the way through the end of finals week for each term.

jonespm:
I think it'll also have to calculate the actual number of weeks in a course based on the start/end date of the term the course is in. There are spring/summer half term courses that don't even run the full 16 weeks. These *might* be ones we look to trying to support soon and at least eventually with the wider rollout we will be getting to these eventually.

jonespm:
I worked on this a little, here's a graph for week 5 of a 12 week course
![image](https://user-images.githubusercontent.com/27447/50614012-7ebda680-0ead-11e9-82ac-f8ca97207100.png)

This is week 1 of a 8 week course
![image](https://user-images.githubusercontent.com/27447/50614508-0f48b680-0eaf-11e9-8515-1d6f954b63b3.png)

This is when the 16 week course is past the date
![image](https://user-images.githubusercontent.com/27447/50614673-a7df3680-0eaf-11e9-9180-cc3851a82fb4.png)

jennlove-um:
16 weeks display correctly for courses in the Files Accessed slider.

PR #334, opened 2019-01-03
Files visualization needs to account for full number of weeks in term

jonespm:
Fixes issue #329

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -116,3 +116,6 @@ RUN_AT_TIMES =
 # This should run as frequently or more frequently than RUN_AT_TIMES
 # Use https://crontab.guru to validate your schedule
 CRONTAB_SCHEDULE = 
+
+# Change this to set the max default weeks to allow. Default is currently 16. The issue is the end dates in Canvas currently are set 10 years out so it can't calculate the range.
+MAX_DEFAULT_WEEKS
--- a/dashboard/views.py
+++ b/dashboard/views.py
@@ -2,11 +2,10 @@ from django.shortcuts import redirect
 from django.http import HttpResponse
 from django.contrib import auth
 from django.db import connection as conn
-from django.core import serializers
+from dashboard.models import AcademicTerms, Course, CourseViewOption
+from django.core.exceptions import ObjectDoesNotExist
 
-from dashboard.models import AcademicTerms, CourseViewOption
-
-
+from django.forms.models import model_to_dict
 
 import random, math, json, logging
 from datetime import datetime, timedelta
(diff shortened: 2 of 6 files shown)


.env.sample
```diff
@@ -115,4 +115,7 @@ RUN_AT_TIMES =
 # An example to match the above would be 0 2,4 * * *
 # This should run as frequently or more frequently than RUN_AT_TIMES
 # Use https://crontab.guru to validate your schedule
-CRONTAB_SCHEDULE = 
\ No newline at end of file
+CRONTAB_SCHEDULE = 
+
+# Change this to set the max default weeks to allow. Default is currently 16. The issue is the end dates in Canvas currently are set 10 years out so it can't calculate the range.
```
pushyamig: What is the need for setting the max default week value? can we have calculate this based on the start and end date of the term and add it to the json feed as a property like
```
{
....
"max_week":8
...
}
```

jonespm: Yeah I am calculating it.

However some weeks in Canvas are set to many years in the future per KB0023342. I do have code in the view that will set the end_year to be the same as the start_year if the year difference is > 1. This should fix this canvas issue, and this really probably isn't ever needed.  It does this calculation correctly if the input is correct.

We talked about using a different source of data (like UM API) to pull the "actual" term dates (not the ones coded into Canvas) but this would be extra effort to do this as well as match them up to the canvas terms. Maybe if it's necessary in the future.

pushyamig: Yes, I do feel the need to get the start and end term date from Some UM API. We can't trust canvas course start and end date as that is extended for some business reasons. 
 As for I know, we are still hard-coding the start and end term date in DB so I think we can safely assume that we get the start/end date no matter how it is fed to the DB table( via API or manual placing)

jonespm: After some recent changes it's actually pulling the academic term dates down from Canvas, we don't have local control over that anymore in our database. 
https://github.com/tl-its-umich-edu/my-learning-analytics/blob/b8f2659ca869b59d7b45d4d7d885e46fe4137ceb/dashboard/cron.py#L338

I thought we could at least trust the "dates" if not the years but it looks like the dates for winter are even incorrect

Winter 2019 = Term Runs from | Jan 9 at 12am | to | Dec 31, 2029 at 12am

I'll have to see what's the easiest fix for this. I'll check with Dan, maybe we can at least guarantee the "Date" is correct if not the year. This is the first time it looks like that wasn't the case.

jonespm: This was resolved, we set the dates correctly.

pushyamig: For the assignments view there is no need for the end date but only start date. So based on the start date I am calculating the week and if you have a question how do I know how many weeks then I don't need to know when the term end, as assignment due, is what I care or student care. So in Assignment view there is not burning need. 

But it will useful if we have the term end date. 
I don't know pulling that start/end date from canvas is a good idea, but with DPDB we might get it from there

jonespm:
@zqian Assigning to you since you originally worked on this. See screenshots on #329

dashboard/views.py
```diff
@@ -55,22 +54,48 @@ def gpa_map(grade):
...
+        logger.debug(f'{term.date_end.year} - {term.date_start.year} greater than 1 so setting end year to match start year.')
+        term.date_end = term.date_end.replace(year=term.date_start.year)
+
+    current_week_number = math.ceil((today - term.date_start).days/7)
+    total_weeks = math.ceil((term.date_end - term.date_start).days/7)
+    
+    resp['term'] = model_to_dict(term)
+
+    # Have a fixed maximum number of weeks
+    if total_weeks > settings.MAX_DEFAULT_WEEKS:
```
zqian: So MAX_DEFAULT_WEEKS setting still matters. Is value of 16 good enough?

zqian: So it looks like the total week number calculation is done whenever the file access view is generated. 

I wonder whether it is possible to cache the value, and use the same output for all courses within the same term?

jonespm: Yeah this is possible. Previously the code was calling get_current_week_number which wasn't much more efficient than this process is. (Both required a database query)

I don't believe caching is an issue specific to this, we should be utilizing [Django's built in caching. ](https://docs.djangoproject.com/en/1.11/topics/cache/)

We could easily cache to memory or filesystem. Caching to memcache would require a little more setup.

jonespm: 16 is the longer number of actual weeks I've seen so far. I think we could increase this if we can ensure the start/end dates are correct, maybe a little longer to at least 20? I didn't test out the slider with more than 16 weeks though. :)

jonespm: The slider gets kind of bad out in week 20 . . . I think this might need to be redesigned to remove the "Week" name if we wanted more than 16 weeks. We might want to remove "Week" anyway as it would work better on Mobile? Maybe that's a different problem.

image

jonespm: I created new issue about the visual here as #336

zqian: Sounds good. I think we are done with this issue.

Issue #121, opened 2018-09-20
Improve dev docker experience


jonespm:
As noted on another ticket by @pushyamig it would be convenient if we could install dev utilities easier. I don't think it's necessary to include them in every image though. But having an easy command/alias that can be run to install them would be nice. 

The ones that come to mind are

vim, mysql-client, silversearcher-ag

There may be some others . . .

jonespm:
This has changed a bit, I think we'd like a better experience for development in docker. This includes
- Mapping the local file system
- Reloading changes dynamically
- Allowing for debugging in development (maybe with Werkzeug or runserver_plus or ptvsd for those using VSCode)

jonespm:
This is going to require some research. Assigning to @DanielDugan to look at this sometime.

DanielDugan:
Will review and provide a solution. I agree that automatic rebuilds will be valuable in dev environments.

jonespm:
Ok, I liked your suggestion but I got pushback from ops when I implemented it.
https://github.com/tl-its-umich-edu/my-learning-analytics/pull/77

Ops said we can only go from a "Base Image" that's already published in Docker Hub, we cannot curate our own partial images (even if it's going to save us some time). Docker caches steps and doesn't run the steps that haven't changed. Openshift has a different (smaller) subset of images than Docker Hub has.

That said we could "potentially" have a different Dockerfile for dev verses production, but I was trying to avoid this. In Student Explorer the initial pass of converting to Docker I had two separate ones but I combined them because they were 99% the same there. 

docker-compose is only used for the local build so we can make any changes there.

DanielDugan:
Okie dokie. Hmmmmmm. Maybe a selective rebuild on that volume. Investigating.

DanielDugan:
Holding on this to prioritize other work [#293](https://github.com/tl-its-umich-edu/my-learning-analytics/issues/293)

PR #346, opened 2019-01-09
Improve docker dev environment

andrew-gardener:
- add gunicorn reload option for dev
- load local code changes into image (compose-file used `/code` but dockerfile was using `/dashbaord`. switch dockerfile to use `/code`)
- persist mysql container data

Related #121

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -22,6 +22,7 @@ services:
     ports:
       - "5306:3306"
     volumes:
+      - ./.data/mysql:/var/lib/mysql
       - ./mysql:/docker-entrypoint-initdb.d/:ro
     container_name: student_dashboard_mysql
   web:
@@ -29,6 +30,8 @@ services:
     command: bash -c "./start.sh"
     volumes:
       - .:/code
+      # use the container's node_modules folder (don't override)
+      - /code/node_modules/
     ports:
       - "5001:5000"
       - "3000:3000"
@@ -37,4 +40,6 @@ services:
       - mysql
     env_file:
       - .env
+    environment:
+      - GUNICORN_RELOAD=True
     container_name: student_dashboard
--- a/.gitignore
+++ b/.gitignore
@@ -117,3 +117,5 @@ venv.bak/
 .mypy_cache/
 .dmypy.json
 dmypy.json
+
+.data/
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,8 +20,8 @@ COPY start.sh /start.sh
 
 # EXPOSE port 5000 to allow communication to/from server
 EXPOSE 5000
-WORKDIR /dashboard/
-COPY . /dashboard/
+WORKDIR /code/
+COPY . /code/
 
 COPY manage.py /manage.py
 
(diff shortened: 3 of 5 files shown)


jonespm:
One of the issues I felt like I'd run into with this is the user problem.

The code in Docker currently runs as root (We'd like to change this I just haven't created an issue for it)

This results in any files written back to the local file system being owned by root! 

You can pass a username and group to Docker and also via docker-compose. The idea on localhost is to pass the currently running user, then it will create files appropriately so the local user can manipulate them. If it's not, I guess it doesn't matter if it matches up as long as it's not root.

https://medium.com/redbubble/running-a-docker-container-as-a-non-root-user-7d2e00f8ee15

Anyway, this was on my "TODO" list but I haven't created an issue. I did solve it on another project and it was a annoying dealing with all the local files owned as root. ;)

andrew-gardener:
I've honestly not run into that issue before but it turns out it might only be an issue for linux users (using a mac). 

Anyways, adding something like this https://jtreminio.com/blog/running-docker-containers-as-current-host-user/#make-it-dynamic could solve it (would just need to set uid and gui in the .env file after its in place).

jonespm:
Yeah, I have only run into this on Linux as well. I mostly use a Mac but I don't know if it would be an issue and Linux VM is the probably the best option for those on Windows too. (Only real option for Windows 10 Home) Was aggravating like that article said tho. ;)

docker-compose.yml
```diff
@@ -22,17 +22,22 @@ services:
     ports:
       - "5306:3306"
     volumes:
+      - ./.data/mysql:/var/lib/mysql
```
zqian: Will the MySQL db be removed after "docker-compose down"?

Currently, if I keep running "docker-compose stop", without using "docker-compose down", the MySQL database is persistent

docker-compose.yml
```diff
@@ -22,17 +22,22 @@ services:
...
       - ./mysql:/docker-entrypoint-initdb.d/:ro
     container_name: student_dashboard_mysql
   web:
     build: .
     command: bash -c "./start.sh"
     volumes:
       - .:/code
+      # use the container's node_modules folder (don't override)
```
zqian: still keep line 32?

docker-compose.yml
andrew-gardener: > Will the MySQL db be removed after "docker-compose down"?
No by mounting the folder outside the docker container, it will persists through a `docker-compose down` (the contents of `/var/lib/mysql` within the container will be in the folder `.data/mysql` within the project folder so it will not be lost when the container is destroyed)

docker-compose.yml
andrew-gardener: This is a little trick for development environments. By mounting the all the host project files within the container (in this case `.:/code`) you can have your local changes be available to the container without rebuilding. 

This however can be alloying with certain subfolders containing dependencies (in this case `/code/node_modules`). What you can then do to get around having to build this folder locally is add a volume for the subfolder without any host (in this case `/code/node_modules/`) to basically tell docker to use the container's version of the folder instead of the host's

docker-compose.yml
jonespm: This is a good idea, the only thing that ever got me was that you have to remember to wipe out the database you'll have to delete. We should add this to the README.

jonespm:
We're sorry about holding all these PR's up. We were waiting to create a release branch for this semester (which we did earlier this week). We also wanted to  get a LICENSE here so we don't have to revert your commits afterward if it was one you didn't agree with. I believe we planning on resolving the license issue by the end of this week.

jonespm:
I merged this and tested it locally but it's not starting up when connecting to the port. I don't have time to debug this today and may it needs a PR after this one? It does start up on our dev server in Openshift.

Are you able to get it started up locally? It looks like in the local container the /code/static directory doesn't exist.

```
    raise ValueError("Missing staticfiles manifest entry for '%s'" % clean_name)
ValueError: Missing staticfiles manifest entry for 'bootstrap/dist/css/bootstrap.css'
```

If this other exception is added this works however I guess that would prevent static files from being "reloaded". Though that seems okay.

```
diff --git a/docker-compose.yml b/docker-compose.yml
index 8d255fa..55dcd81 100644
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -35,6 +35,7 @@ services:
       - .:/code
       - /code/node_modules/
+      - /code/static/
     ports:
       - "5001:5000"
       - "3000:3000"
```

Issue #266, opened 2018-10-22
Making the Dashboard tool as LTI tool


pushyamig:
For the Pilot we adopted the approach as standalone tool, but this needs to be a LTI tool. For the Next Phase of the student dashboard will this be in the scope?

PR #347, opened 2019-01-09
Add basic LTI 1.1.1 support

andrew-gardener:
- Auto creates user auth accounts if needed
- Doesn't create course records or modify course view options

Related #266

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -61,6 +61,14 @@ CANVAS_FILE_POSTFIX=/download?download_frd=1
 # Paths to special login credentials (uses SAML)
 GOOGLE_APPLICATION_CREDENTIALS=/secrets/bq_cred.json
 
+# LTI Configuration
+# LTI is disabled by default
+STUDENT_DASHBOARD_LTI=False
+
+# list LTI consumers as per https://github.com/rohitjose/django-lti-auth#usage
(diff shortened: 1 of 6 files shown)


jonespm:
Cool, I'll try this out tomorrow. Do you know if it coexists with the authentication backend currently enabled (which is either ModelBackend or Saml2Backend? We probably should have an option to disable the ModelBackend if only launches via LTI are desired.

andrew-gardener:
It can co-exist with both. It basically just uses `django_lti_auth` to validate the lti launch and the in https://github.com/tl-its-umich-edu/my-learning-analytics/pull/347/files#diff-50e78b1017eaf6cd684310c9d3b8ae7dR15 it will find/create the user and log them in.

Wasn't sure how you wanted to handle access control issues so it doesn't look at roles (but that could be added if your interested in tackling that with LTI)

jonespm:
Saying "tomorrow" may have been a little optimistic. We're currently working on getting this out next week for our Spring courses and I have to work on some other things but we will get these all reviewed and merged before hopefully any conflicts. 

Up until now it's only been our team locally working on this. I prefer [Trunk Based Development](https://trunkbaseddevelopment.com/) and will probably setup a release branch (before each term) before merging these all in to master next week so they don't cause any issues with our running production.

PR #348, opened 2019-01-10
Allows control over view options globally

andrew-gardener:
- Each of the view options (files accessed, assignment planning, and grade distribution) can now be disabled globally via the VIEWS_DISABLED environment variable.
- If files accessed is disabled, will no longer attempt to collect data file date from Canvas Data or big query
- Admin course view options will hide disabled view options
- added default value of true for course view options

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,14 @@ The bq_cred.json is service account for Big Query, it needs to be supplied and p
 	student-dashboard-saml.key
 	student-dashboard-saml.pem
 
+# Control course view options
+
+View options can be controlled at the global and course level. If a view is disabled globally, it will be disabled for each course, even if previously enabled at the course level. If a view is not globally disabled, it can still be disabled at the course level.
+
+`VIEWS_DISABLED` comma delimited list of views to disable (default empty). The expected name of the view is the same as the view's column name in the `course_view_option` table. Example value of `show_files_accessed,show_grade_distribution` will disable both the Files Accessed and Grade Distribution views.
+
+Note that by default all views are enabled when a course is added.
+
 # LTI v1.1.1 Configuration
 
 Only basic LTI launches are supported at the moment (automatic account creation and redirection to the correct course). New courses are not added nor are course view options modified.
--- a/dashboard/models.py
+++ b/dashboard/models.py
@@ -8,8 +8,7 @@
 from __future__ import unicode_literals
 
 from django.db import models
-from django.core.exceptions import ObjectDoesNotExist
-
+from django.conf import settings
 import logging
 logger = logging.getLogger(__name__)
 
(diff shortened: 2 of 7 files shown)


dashboard/models.py
```diff
@@ -130,9 +131,9 @@ class Meta:
 
 class CourseViewOption(models.Model):
     course = models.OneToOneField(Course, on_delete=models.CASCADE, primary_key=True, verbose_name="Course View Option Id")
-    show_files_accessed = models.BooleanField(blank=False, null=False, verbose_name="Show Files Accessed View")
-    show_assignment_planning = models.BooleanField(blank=False, null=False, verbose_name="Show Assignment Planning View")
-    show_grade_distribution = models.BooleanField(blank=False, null=False, verbose_name="Show Grade Distribution View")
+    show_files_accessed = models.BooleanField(blank=False, null=False, default=False, verbose_name="Show Files Accessed View")
```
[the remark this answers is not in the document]
andrew-gardener: Fixed (switched default of all three to `True`)

README.md
```diff
@@ -23,7 +23,17 @@ The bq_cred.json is service account for Big Query, it needs to be supplied and p
 	student-dashboard-saml.key
 	student-dashboard-saml.pem
 
-## Docker commands for deploying the app
+# Control course view options
+
+View options can be controlled at the global and course level.
```
jonespm: I had a little bit of confusion around these settings, these actually enable/disable the ability to turn on the view at all. You still have to go in and enable/disable the view in the database. 

Maybe just a comment to this point could be added. Like

For each of these view settings set to True you need to configure at the course level which views to enable (by default all views are off)
For each of these view settings set to False that view will be disabled for that view even if enabled via the UI at the course level.

dashboard/settings.py
```diff
@@ -325,6 +325,11 @@
 # This is fixed from UDW
 UDW_FILE_ID_PREFIX = config("UDW_FILE_ID_PREFIX", default="1770000000")
 
+# Allow enabling/disabling the View options globally
+ENABLE_FILES_ACCESSED = config('ENABLE_FILES_ACCESSED', default=True, cast=bool)
```
jonespm: We're anticipating adding many more views in the future. 

I wonder if it would be better to rewrite this property as 

`VIEWS_DISABLED = config('VIEWS_DISABLED', default = '', cast= Csv())`

By default we'd have
`VIEWS_DISABLED=`

But you could have
`VIEWS_DISABLED=show_files_accessed,show_assignment_planning,show_grade_distribution`

Right? Then in the code just do an "in" check for this value to see if a specific view is disabled. And this is already the CSV excluded list. Then we don't have to add new settings for new views.

Maybe call it VIEW_EXCLUDED to be consistent with the Django admin.

andrew-gardener: Sure

andrew-gardener: I do prefer `VIEWS_DISABLED` over `VIEW_EXCLUDED` since it helps infer the idea that the view is fully disabled and not just globally hidden. Ex: data that is exclusively used for a disabled view should not be fetched in the cron (since it saves time and will help when certain views aren't yet supportable)

andrew-gardener:
@jonespm modified readme and switched to `VIEWS_DISABLED` as discussed (also rebased to resolve conflicts)

jonespm:
Yeah, this looks like it'll work well, thanks!

PR #349, opened 2019-01-10
Support non-unizin Canvas Data sources

andrew-gardener:
- Renames UDW to DATA_WAREHOUSE to be more data source generic
- adds env variable to disable unizin only features/data. Currently is this only the `unizin_metadata` table and the Canvas Data last update datetime (falling back unto last cron run for now)
- removes UDW environment variables from mysql container in docker compose file

### DIFF
--- a/dashboard/common/db_util.py
+++ b/dashboard/common/db_util.py
@@ -14,14 +14,14 @@ logger = logging.getLogger(__name__)
 
 def get_course_name_from_id(course_id):
     """[Get the long course name from the id]
-    
-    :param course_id: [Canvas course ID without UDW PREFIX]
+
+    :param course_id: [Canvas course ID without DATA_WAREHOUSE PREFIX]
     :type course_id: [str]
     :return: [Course Name of course or blank not found]
     :rtype: [str]
     """
     logger.info(get_course_name_from_id.__name__)
-    course_id = settings.UDW_ID_PREFIX + str(course_id)
+    course_id = settings.DATA_WAREHOUSE_ID_PREFIX + str(course_id)
     course_name = ""
     if (course_id):
         with django.db.connection.cursor() as cursor:
@@ -34,7 +34,7 @@ def get_course_name_from_id(course_id):
 def get_course_view_options (course_id):
 
     logger.info(get_course_view_options.__name__)
-    course_id = settings.UDW_ID_PREFIX + str(course_id)
+    course_id = settings.DATA_WAREHOUSE_ID_PREFIX + str(course_id)
     logger.debug("course_id=" + str(course_id))
     course_view_option = ""
     if (course_id):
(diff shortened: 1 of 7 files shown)


jonespm:
So reviewing these commits and working on another project since this was done, I've noticed that we actually were using this as a "CANVAS_ACCOUNT_NUMBER" rather than anything related to UDW or Unizin. I'm thinking now that this is less about a data warehouse and more specific to Canvas. . .

andrew-gardener:
Rename to `CANVAS_DATA` or `CANVAS_DATA_SOURCE` then?

jonespm:
Something like that! ;) 

It's like a CANVAS_DATA_ID_ADDITION, but that's a little wordy. I suppose CANVAS_DATA_SOURCE is fine. Thanks!

jonespm:
We just made a few incorrect assumptions about this, and got lucky ALL of our course id's were 6 digits so far. :)

andrew-gardener:
oh, the comment was about the prefix name, I thought you meant the rename from `UDW` to `DATA_WAREHOUSE` (since it was in the PR)

I'm fine with `UDW_ID_PREFIX`+`UDW_FILE_ID_PREFIX` to `CANVAS_DATA_ID_INCREMENT` but I'll do it in PR #350 (since I tried to fix it there).

I can also go from `UDW` to `CANVAS_DATA_SOURCE` in this PR if you prefer it over `DATA_WAREHOUSE`

Does that sound good?

jonespm:
Yeah I think it was more about these 3 values. On double checking there *really* isn't anything specific to Unizin (besides the cron process) and these ID's and "PREFIX's" (which it really isn't a prefix but a integer addition) are coming from the Canvas Data.

I guess if someone used it outside of Canvas maybe it would be different. Maybe it's still fine to keep these "as-is". The other changes look okay that go to the actual data warehouse. (As that may or may not be UDW) 

```
DATA_WAREHOUSE_IS_UNIZIN = config("DATA_WAREHOUSE_IS_UNIZIN", default=True, cast=bool)

DATA_WAREHOUSE_ID_PREFIX = config("DATA_WAREHOUSE_ID_PREFIX", default="17700000000", cast=str)

# This is fixed from DATA_WAREHOUSE
DATA_WAREHOUSE_FILE_ID_PREFIX = config("DATA_WAREHOUSE_FILE_ID_PREFIX", default="1770000000")
```

dashboard/common/db_util.py
```diff
@@ -75,6 +75,9 @@ def get_last_cron_run():
     return datetime.min
 
 def get_canvas_data_date():
+    if not settings.DATA_WAREHOUSE_IS_UNIZIN:
```
jonespm: I feel like this "last cron run" isn't specific to Unizin and this is the only thing this value is used for. This is if the cron is run at all. If you're loading using a separate process and aren't populating the django_cron_cronjoblog then that's when it probably wouldn't be useful.

If you wanted to hide the date or check if the return value was datetime.min and not display it that seems okay. But I think if it didn't return a date it would mess up the base.html temple where last_updated is used.

andrew-gardener: the `DATA_WAREHOUSE_IS_UNIZIN` is currently only used here to avoid the fetch from the`unizin_metadata` table below (currently this is the only unizin specific data that is used from what I can tell). I was looking for a somewhat quick solution so I fell back to what was previously used before the `unizin_metadata` table call was used (which is the call to `get_last_cron_run`).

andrew-gardener:
fixed the merge conflict

PR #350, opened 2019-01-10
Refactor canvas id prefixing

andrew-gardener:
- Switched from prefixing canvas ids to incrementing the id by a set integer from the `CANVAS_DATA_ID_INCREMENT` environment variable.
- This handles how Canvas Data creates their ids better
- DATA_WAREHOUSE_FILE_ID_PREFIX and DATA_WAREHOUSE_ID_PREFIX are removed (instead CANVAS_DATA_ID_INCREMENT is used)

Note: this is a breaking change to how `DATA_WAREHOUSE_ID_PREFIX` worked. Existing installation will need to update the value with the correct canvas id increment.

Depends on PR #349

### DIFF
--- a/dashboard/common/db_util.py
+++ b/dashboard/common/db_util.py
@@ -12,16 +12,32 @@ from django.conf import settings
 
 logger = logging.getLogger(__name__)
 
+def canvas_id_to_incremented_id(canvas_id):
+    try:
+        int(canvas_id)
+    except ValueError:
+        return None
+
+    return str(int(canvas_id) + settings.CANVAS_DATA_ID_INCREMENT)
+
+def incremented_id_to_canvas_id(incremented_id):
+    try:
+        int(incremented_id)
+    except ValueError:
+        return None
+
+    return str(int(incremented_id) - settings.CANVAS_DATA_ID_INCREMENT)
+
 def get_course_name_from_id(course_id):
     """[Get the long course name from the id]
 
-    :param course_id: [Canvas course ID without DATA_WAREHOUSE PREFIX]
+    :param course_id: [Canvas course ID without the Canvas Data increment]
     :type course_id: [str]
     :return: [Course Name of course or blank not found]
     :rtype: [str]
     """
     logger.info(get_course_name_from_id.__name__)
-    course_id = settings.DATA_WAREHOUSE_ID_PREFIX + str(course_id)
+    course_id = canvas_id_to_incremented_id(course_id)
     course_name = ""
     if (course_id):
         with django.db.connection.cursor() as cursor:
@@ -34,7 +50,7 @@ def get_course_name_from_id(course_id):
 def get_course_view_options (course_id):
 
     logger.info(get_course_view_options.__name__)
-    course_id = settings.DATA_WAREHOUSE_ID_PREFIX + str(course_id)
+    course_id = canvas_id_to_incremented_id(course_id)
     logger.debug("course_id=" + str(course_id))
     course_view_option = ""
     if (course_id):
@@ -61,8 +77,7 @@ def get_default_user_course_id(user_id):
         cursor.execute("SELECT course_id FROM user WHERE sis_name= %s ORDER BY course_id DESC LIMIT 1", [user_id])
         row = cursor.fetchone()
         if (row != None):
-            #Remove the DATA_WAREHOUSE_ID_PREFIX and just return the course_id
-            course_id = str(row[0]).replace(settings.DATA_WAREHOUSE_ID_PREFIX, "")
+            course_id = canvas_id_to_incremented_id(row[0])
     return course_id
 
 def get_last_cron_run():
--- a/dashboard/context_processors.py
+++ b/dashboard/context_processors.py
@@ -35,6 +35,13 @@ def course_view_option(request):
 
     return {"course_view_option": course_view_option}
 
+def current_user_incremented_course_id(request):
+    course_id = str(request.resolver_match.kwargs.get('course_id'))
+    if not course_id:
+        logger.info(f"Course ID could not be determined from request, attempting to look up for user {request.user.username}")
+        course_id = db_util.get_default_user_course_id(request.user.username)
+    incremented_course_id = db_util.canvas_id_to_incremented_id(course_id)
+    return {'current_user_incremented_course_id': incremented_course_id}
 
 def last_updated(request):
     return {'last_updated': db_util.get_canvas_data_date()}
--- a/dashboard/settings.py
+++ b/dashboard/settings.py
@@ -105,6 +105,7 @@ TEMPLATES = [
                 'django_settings_export.settings_export',
                 'dashboard.context_processors.course_name',
                 'dashboard.context_processors.current_user_course_id',
+                'dashboard.context_processors.current_user_incremented_course_id',
                 'dashboard.context_processors.course_view_option',
                 'dashboard.context_processors.last_updated',
                 'dashboard.context_processors.get_build_info',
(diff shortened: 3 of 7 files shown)


dashboard/common/db_util.py
```diff
@@ -12,16 +13,29 @@
...
 logger = logging.getLogger(__name__)
 
+def canvas_id_to_prefixed_id(canvas_id):
+    # Add the DATA_WAREHOUSE_ID_PREFIX and to the id
+    canvas_id = str(canvas_id)
+    prefixed_id = str(settings.DATA_WAREHOUSE_ID_PREFIX)[0:-len(canvas_id)] + canvas_id
+    return prefixed_id
+
+def prefixed_id_to_canvas_id(prefixed_id):
+    # Remove the DATA_WAREHOUSE_ID_PREFIX and just return the id
```
zqian: However, settings. DATA_WAREHOUSE_ID_PREFIX was not used directly in line 24-26?

dashboard/context_processors.py
```diff
@@ -35,6 +35,14 @@ def course_view_option(request):
 
     return {"course_view_option": course_view_option}
 
+def current_user_prefixed_course_id(request):
+    course_id = str(request.resolver_match.kwargs.get('course_id'))
+    if not course_id:
+        logger.info("Course ID could not be determined from request, attempting to look up for user {}".format(
```
zqian: use of Python3's f-string to do logging?

https://realpython.com/python-f-strings/

dashboard/cron.py
```diff
@@ -375,9 +375,10 @@ def do(self):
 
         status += self.submission()
         status += self.weight_consideration()
-        
-        logger.info("************ informational")
-        status += self.update_unizin_metadata()
+
+        if settings.DATA_WAREHOUSE_IS_UNIZIN:
```
zqian: if the DATA_WAREHOUSE_IS_UNIZIN is false, then there is no metadata shown at all?

dashboard/settings.py
```diff
@@ -318,12 +319,12 @@
 else: 
     AUTHENTICATION_BACKENDS += ('django.contrib.auth.backends.ModelBackend',)
     LOGIN_REDIRECT_URL = '/'
-    
-# This is fixed from UDW
-UDW_ID_PREFIX = config("UDW_ID_PREFIX", default="17700000000", cast=str)
 
-# This is fixed from UDW
-UDW_FILE_ID_PREFIX = config("UDW_FILE_ID_PREFIX", default="1770000000")
+# controls whether unizen specific features/data is available from the Canvas Data source
```
zqian: "Unizin" instead of "Unizen"

dashboard/cron.py
andrew-gardener: Yes as per #349 the metadata table seems to be a Unizin table (not part of Canvas Data). from what I understand (unless I missed something) the metadata is only being used for displaying the last updated time to the user currently. When `DATA_WAREHOUSE_IS_UNIZIN` is false, I switched it to show the last time the cron job ran (which was what was used before switching to Unizin metadata).

dashboard/common/db_util.py
```diff
@@ -12,16 +13,29 @@
 
 logger = logging.getLogger(__name__)
 
+def canvas_id_to_prefixed_id(canvas_id):
+    # Add the DATA_WAREHOUSE_ID_PREFIX and to the id
+    canvas_id = str(canvas_id)
+    prefixed_id = str(settings.DATA_WAREHOUSE_ID_PREFIX)[0:-len(canvas_id)] + canvas_id
```
pushyamig: This particular code end up creating a wrong course id. This particular function is called from from admin view for adding courses from django admin view and that statement will end up in wrong long canvas id.  this `[0:-len(canvas_id)]` needs to removed in order to have right id format. what was the reason behind substring in the above statement?
```
canvas_id = '235999'
DATA_WAREHOUSE_ID_PREFIX=17700000000
prefixed_id = str(DATA_WAREHOUSE_ID_PREFIX)[0:-len(canvas_id)] + canvas_id
print(prefixed_id) 

o/p: 17700235999 instead of 17700000000235999
```

andrew-gardener: > Note: this is a breaking change to DATA_WAREHOUSE_ID_PREFIX. Existing installation will need to update the value with the correct number of additional 0s added to the end

Try changing your `DATA_WAREHOUSE_ID_PREFIX` from `17700000000` to `17700000000000000` (six extra zeros).

I switched to using a safer approach to adding the prefix instead of simply putting a static prefix in front of the canvas id since doing that won't work for canvas id of different orders (ex: canvas id 100 and 1000 won't be prefixed properly if you simply put the `17700000000` in front of both of them). instead it will insert the canvas id into the prefix string as you saw (ex: when using `17700000000000000` it will do `17700000000000100` or `17700000000001000` for ids 100 and 1000)

andrew-gardener: I'll also rename the prefix variable to template since it isn't really a straight prefix anymore

pushyamig: @andrew-gardener yes, the name(prefix) is what caused the confusion. Changing it as `DATA_WAREHOUSE_ID_TEMPLATE` or `DATA_WAREHOUSE_EMPTY_ID` will make sense.

dashboard/common/db_util.py
andrew-gardener: I've rewritten the comments explain what `canvas_id_to_prefixed_id` and `prefixed_id_to_canvas_id` do (hopefully better!)

dashboard/context_processors.py
andrew-gardener: Switched to f-string. 

Note that the other logging in the file is done in the same format I used. Is this a change you'd want to apply to `.format()` uses in the project?

dashboard/settings.py
andrew-gardener: fixed typo

dashboard/common/db_util.py
andrew-gardener: Renamed to `DATA_WAREHOUSE_ID_TEMPLATE` and improved the comments explaining how `canvas_id_to_prefixed_id` and `prefixed_id_to_canvas_id` work

jonespm:
So in looking at this PR and this doc I noticed we were using this ID incorrectly. This is actually from Canvas and _"Is made up by adding a large number to the ID of the source table."_ It "seems" like the value can be calculated from the data as.

`select id-canvas_id AS account_num from account_dim where parent_account IS NULL; `

So this is also not a string prefix that could be appended, but an integer that should be added (or possibility subtracted) 

If the (original value - account_num) < 0 then it didn't have this change to begin with.

dashboard/common/db_util.py
```diff
@@ -12,16 +13,38 @@
...
+    Insert the Canvas ID into the beginging of the DATA_WAREHOUSE_ID_TEMPLATE string
+    Ex: if DATA_WAREHOUSE_ID_TEMPLATE=11000000 and the Canvas ID = 999 then the prefixed_id will be 11000999
+    """
+    canvas_id = str(canvas_id)
+    prefixed_id = str(settings.DATA_WAREHOUSE_ID_TEMPLATE)[0:-len(canvas_id)] + canvas_id
+    return prefixed_id
+
+def prefixed_id_to_canvas_id(prefixed_id):
+    """
+    Remove the prefix from prefix_id to get the original Canvas ID
```
jonespm: I think we're actually going to have to change this up to treat this value as an integer rather than a string. This should simplify a lot of the code here.

andrew-gardener: very true, I'll fix it up

andrew-gardener:
@jonespm I've changed the prefixing/templating into a simpler increment `CANVAS_DATA_ID_INCREMENT` as discussed

pushyamig:
This change is approved, @andrew-gardener can you fix the conflict for merging this

andrew-gardener:
fixed conflict!

Issue #321, opened 2018-12-06
Allow students to set default for Files and Assignment views


jennlove-um:
Provide a way for students to set their defaults for the Files and Assignment views: 

File page: Allow them to set the default for files accessed by students in grade group All grades, 90-100%, 80-89%, or 70-79%.

Assignment page: Allow them to set the default for the assignment weight.

The default should persist until they change their preference.

jennlove-um:
Outstanding questions:
Should changing preferences be done in a separate Preferences (or Settings) section of the Dashboard?
Should changing preferences be done by simply persisting the student's most recent choice in the visualization?

pushyamig:
So far I have created a DB table that looks like this and I have made the default value selection in json so for user in a course would have one record.
<img width="674" alt="screen shot 2019-01-03 at 9 56 21 am" src="https://user-images.githubusercontent.com/8579775/50644153-0226d880-0f3e-11e9-9fdf-89ef43f02010.png">
<img width="607" alt="screen shot 2019-01-03 at 9 56 32 am" src="https://user-images.githubusercontent.com/8579775/50644159-06eb8c80-0f3e-11e9-830c-d80f919de462.png">
This is the query that does both insert/update the table as approperiate 
[insert into user_views_default_selection values (17700000000245664, 'user5','{"assignment":50}')
on duplicate key update default_types_value = json_set(default_types_value,'$.assignment',50);](url)

Note: please ignore the `grades` json property it will be actually be files as it is associated with that view.

pushyamig:
MySQL version 5.7.22 supports JSON data types and this will also for using the json structure. 
Let me know if any questions

jonespm:
I suggested we could just have 2 columns key/value instead of a single one holding both in a json array.

Also do you need to have course_id? I feel like the user preference system wide is enough. Would someone really want/need a different setting per course? (I don't think this was a requirement?)

pushyamig:
I don't think there was a mention if the user needs defaults per course. The user might want a different selection per course that's what I felt should happen. But @jennlove-um what do you think should the behavior in case a user is part of 2 courses in MyLA? how are his/her selections should be chosen?

pushyamig:
@jonespm why do you think we should have 2 columns holding different default selection instead if a JSON array?

jennlove-um:
My inclination is to allow students to set the default on a per course basis for now.

[...]

PR #354, opened 2019-01-11
#321 adding default settings option to views

pushyamig:
1. Handles setting the option for both assignment and files view
2. can be operated using Django models, I have migration file committed as well. I have added the table in the `init.sql` file. we can decide how to create the `user_default_seletion` table
3. st defaults are added Eventlog table as ` {'course_id': '17700000000245664', 'default_type': 'files', 'default_value': '90-100'}` as type 'VIEW_SET_DEFAULT`
4. refactored code associated with eventLog refactor


mysql/init.sql
```diff
@@ -98,6 +98,16 @@ CREATE TABLE IF NOT EXISTS unizin_metadata (
...
 );
 
+-- CREATE TABLE IF NOT EXISTS user_default_selection (
+--      id int NOT NULL AUTO_INCREMENT,
+--      course_id CHAR(255),
+--      user_id CHAR(255),
+--      default_view_type CHAR(225),
+--      default_view_value CHAR(225),
+--      PRIMARY KEY (course_id, user_id, default_view_type)
+-- );
```
pushyamig: I commented out the create table so that if we want to use Django model to create this table.

dashboard/event_logs_types/event_logs_types.py
```diff
@@ -0,0 +1,13 @@
+from enum import Enum, unique
+
+
+@unique
+class EventLogTypes(Enum):
+    EVENT_VIEW_FILE_ACCESS = "VIEW_FILE_ACCESS"
+    EVENT_VIEW_ASSIGNMENT_PLANNING = "VIEW_ASSIGNMENT_PLANNING"
+    EVENT_VIEW_GRADE_DISTRIBUTION = "VIEW_GRADE_DISTRIBUTION"
+    EVENT_VIEW_SET_DEFAULT = "VIEW_SET_DEFAULT"
+
```
pushyamig: I felt the event log should all be coming from one place instead of the `views.py` so separated as an Enum class

dashboard/migrations/0001_initial.py
```diff
@@ -0,0 +1,178 @@
+# -*- coding: utf-8 -*-
```
pushyamig: generated the migration file with `managed=True` for user_selection_table` in case if we want to create the tables using django. So I wasn't sure if this is the directory that should be placed in but that what Django doing it so i followed the approach

dashboard/views.py
```diff
@@ -1,23 +1,24 @@
...
+from django.contrib import auth
+from django.db import connection as conn
+from django.http import HttpResponse
+from django.shortcuts import redirect
 from pinax.eventlog.models import log as eventlog
+from dashboard.event_logs_types.event_logs_types import EventLogTypes
+
+from django.core.exceptions import ObjectDoesNotExist
+
+from dashboard.models import AcademicTerms, UserDefaultSelection
```
pushyamig: Sorry for the mess here just shuffled the imports

dashboard/views.py
```diff
@@ -119,7 +117,7 @@ def file_access_within_week(request, course_id=0):
         "grade": grade,
         "course_id": course_id
     }
-    eventlog(request.user, EVENT_VIEW_FILE_ACCESS, extra=data)
+    eventlog(request.user, EventLogTypes.EVENT_VIEW_FILE_ACCESS.value, extra=data)
```
zqian: I think EventLogTypes.EVENT_VIEW_FILE_ACCESS should be good enough, without use ".value"

dashboard/views.py
```diff
@@ -255,11 +253,61 @@ def grade_distribution(request, course_id=0):
     data = {
         "course_id": course_id
     }
-    eventlog(request.user, EVENT_VIEW_GRADE_DISTRIBUTION, extra=data)
+    eventlog(request.user, EventLogTypes.EVENT_VIEW_GRADE_DISTRIBUTION.value, extra=data)
```
zqian: Same as above: EventLogTypes.EVENT_VIEW_GRADE_DISTRIBUTION, without ".value"

dashboard/event_logs_types/event_logs_types.py
zqian: some of your PR comments could be embedded inside the code as Python comments.

dashboard/static/js/controllers.js
```diff
@@ -1,8 +1,10 @@
 visApp.controller('mainController', ['$scope', 'Get','$location', '$anchorScroll', function($scope, Get,$location,$anchorScroll) {
-
-   function dataCall(call){
-        var dataFile = '/api/v1/courses/'+dashboard.course_id+'/assignments?percent='+$scope.percent_filter;
-        Get.getData(dataFile).then(function(data) {
+    function dataCall(call, default_value){
```
zqian: rename to something like "resetToDefault"?

dashboard/static/js/controllers.js
```diff
@@ -26,18 +28,59 @@ visApp.controller('mainController', ['$scope', 'Get','$location', '$anchorScroll
             $anchorScroll();
         });
     }
+    // get the user default selection from DB and if not available selects the first item from the dropdown list
+    get_default_selection = function () {
+        let get_defaults_url = '/api/v1/courses/' + dashboard.course_id + '/get_user_default_selection?default_type=assignment';
```
zqian: is "let" necessary here?

dashboard/static/js/controllers.js
```diff
@@ -26,18 +28,59 @@ visApp.controller('mainController', ['$scope', 'Get','$location', '$anchorScroll
...
+        let re = Get.getData(get_defaults_url)
```
zqian: same thing: is "let" necessary here?

dashboard/views.py
pushyamig: as per the suggestion, it will load the event as `EventLogTypes.EVENT_VIEW_FILE_ACCESS` instead of `VIEW_SET_DEFAULT` or `VIEW_FILE_ACCESS`

dashboard/static/js/controllers.js
pushyamig: This function is actually pulling the data needed for visualization and I am passing the default value as one of the parameters if available. So the name suggested wouldn't be appropriate

dashboard/static/js/controllers.js
pushyamig: `let` has a local scope so it makes sense to use that here within the function scope

dashboard/event_logs_types/event_logs_types.py
pushyamig: Done

dashboard/static/js/controllers.js
jonespm: I had an email hanging around in draft about this for awhile, let is the more appropriate thing to use here.

README.md
```diff
@@ -34,6 +34,10 @@ The bq_cred.json is service account for Big Query, it needs to be supplied and p
 `docker cp ~/secrets student_dashboard:/secrets`
 5. Initialize the MySQL database by loading the users and files on the next step. You'll need to be on VPN for this to work.
 
+## Making migration using Django models
+1. creating the file `./manage.py makemigrations dashboard`
+2. applying the changes to DB `./manage.py migrate dashboard`
```
jonespm: To do this locally you actually have to install the requirements, probably with a venv? 

This can also be done by connecting to the docker instance and running it in there, but then you have to copy the file out (also with docker). Maybe we should mention a little more about this since the default path doesn't involve venv.

dashboard/migrations/0001_initial.py
jonespm: Yeah I think this is the way we'd like to do it rather than use the script

dashboard/models.py
```diff
@@ -46,6 +46,54 @@ class Meta:
...
+    id = models.AutoField(primary_key=True, verbose_name="Table Id")
+    user_id = models.CharField(max_length=255, verbose_name="SIS Id")
+    course_id = models.CharField(max_length=255, verbose_name="Course Id")
+    default_view_type = models.CharField(max_length=255, verbose_name="Default Type")
+    default_view_value = models.CharField(max_length=255, verbose_name="Default Value")
+
+    objects = UserDefaultManager()
+
+    class Meta:
+        managed = False
```
jonespm: Shouldn't this be managed = True if we're using the migrations?

dashboard/templates/assignments.html
```diff
@@ -23,9 +23,14 @@ <h5 style="padding-left: 10px">Progress toward Final Grade</h5>
         <div class="column">
             <h5>Assignments Due By Date</h5>
             <span>Show assignments that weigh at least</span>
+            {% verbatim %}
             <p ng-show="percentFilterEnabled">
                 <select ng-change="updatePercentFilter()" ng-model="percent_filter" ng-options="obj.value as obj.text for obj in percent"></select>
+                &nbsp;
+                <input ng-show="isResetMyDefaultCheckboxEnabled" type="checkbox" ng-model="default_selection_model" ng-change="sendDefaults()">
+                <label>{{default_setting.msg}}</label>
```
jonespm: Does this label print correct within the verbatim tags? You might need to move the endverbatim up a little. Not 100% sure why it's needed here? Maybe you can add some comments?

pushyamig: Django itself uses tags like this `{% endblock %}` which are placed in the curly braces. Django for some reason get confused if don't have angular tags `{{}}`in verbatim tags.

jonespm: Interesting, maybe I never saw it used. 

Perhaps we should look at the [django-angular](https://github.com/jrief/django-angular) package which has an [{% angularjs %} tag.
](https://django-angular.readthedocs.io/en/latest/template-sharing.html)

README.md
pushyamig: I have added more details as suggested

dashboard/models.py
pushyamig: Ya. I wanted to know if we want to create tables using Django models or not for this release. I changed to True. We might want to enable all the table `managed = True` or make it a configurable option so that it can be turned on via .env

dashboard/templates/assignments.html
pushyamig: I did not look into this so deeply at this during winter pilot. I am reluctant to use any more libraries supporting angularJS as it is old, entered into [long term support](https://blog.angular.io/stable-angularjs-and-long-term-support-7e077635ee9c) and we should move away from it for `assignment` view and some time in future this version won't be supported and it will become ancient.

dashboard/views.py
```diff
@@ -255,11 +253,61 @@ def grade_distribution(request, course_id=0):
...
+    return HttpResponse(json.dumps({key: value}))
+
+
+def get_user_default_selection(request, course_id=0):
+    logger.info(get_user_default_selection.__name__)
+    user_id = request.user.get_username()
+    default_view_type = request.GET.get('default_type')
+    key = 'default'
+    no_user_default_response = json.dumps({key: ''})
+    logger.info(f"the default option request from user {user_id} in course {course_id} of type: {default_view_type}")
```
jonespm: I think we'll need to reduce the log level of the messages in both of these methods.

PR #356, opened 2019-01-14
Support for debugging with ptvsd

jonespm:


jonespm:
@pushyamig here's the work in progress patch I have to get this to work with VSCode debugging. It opens up the port 3001. I still need to

- Add to the README more information about this
- Open up additional port(s) and document a process for being able to debug the cron (and possibly other jobs)

jonespm:
I added support for debugging the cron job to this. We'll wait to merge as it's going to need a change after #346

.vscode/launch.json
```diff
@@ -0,0 +1,31 @@
+{
```
pushyamig: will this file used? or this is for documentation purposes?

pushyamig:
you can merge this. I use IntelliJ so this setup is focused on Visual studio code. So I might need to figure out how to do this in IntelliJ

start.sh
```diff
@@ -21,13 +21,21 @@ echo Running python startups
 python manage.py migrate
 
 if [ -z "${IS_CRON_POD}" ]; then
-    # Start Gunicorn processes
-    echo Starting Gunicorn.
+    if [ -z "{PTVSD_DEBUG}" ]; then
```
pushyamig: do I still need to set the `PTVSD` to True in order to work locally even though I won't be using the remote debugging?

jonespm: I think if you want to debug at all (with VSCode) you'd have to set this to true.

.vscode/launch.json
jonespm: This file is used by VSCode to define the default debug configurations.
image

start.sh
jonespm: If you want to switch to VSCode I can show you how it works. :) I feel like PyCharm is what Jetbrains expects you to use with python. IntelliJ is more for Java but works with other languages.

pushyamig: yes, if you enable the python plugin with IntelliJ I have the full version of PyCham.  when I launch the IDE it always says "IntelliJ" So that's where I got stuck in my head

jonespm: I see, it sounds like PyCharm might support ptvsd. If it doesn't it's probably something really similar.

jonespm:
Now that we have the branch I'm going to merge this.

PR #358, opened 2019-01-15
292 Cron should fail with an error if course data is missing from UDW

DanielDugan:
Cron fails when course data is missing

### DIFF
--- a/dashboard/cron.py
+++ b/dashboard/cron.py
@@ -61,8 +61,7 @@ def util_function(UDW_course_id, sql_string, mysql_table, table_identifier=None)
     # returns the row size of dataframe
     return f"inserted + {str(df.shape[0])} rows in table {mysql_table} for course {UDW_course_id}\n"
 
-# Exit with log if UDW dataset is missing courses
-def data_validation (incoming_courses,existing_courses):
+    # exit cron with log if UDW dataset is missing courses
     incoming_courses = df['course_id']
     existing_courses = get_queryset().get_supported_courses()
 
@@ -73,7 +72,6 @@ def data_validation (incoming_courses,existing_courses):
                     continue
                 else:
                     logger.exception(f"missing {course} from incoming dataset from UDW.")
-
                     return
 
     except Exception:
@@ -168,8 +166,6 @@ class DashboardCronJob(CronJobBase):
         status = ""
 
         logger.debug("in update with udw file")
-
-        #Log an error if the course is invalid
 
         # delete all records in the table first
         status += deleteAllRecordInTable("file")
--- a/dashboard/cron.py
+++ b/dashboard/cron.py
@@ -54,18 +54,38 @@ def util_function(UDW_course_id, sql_string, mysql_table, table_identifier=None)
     # write to MySQL
     try:
         df.to_sql(con=engine, name=mysql_table, if_exists='append', index=False)
-    except Exception as e:
+    except Exception:
         logger.exception(f"Error running to_sql on table {mysql_table}")
         raise
 
     # returns the row size of dataframe
     return f"inserted + {str(df.shape[0])} rows in table {mysql_table} for course {UDW_course_id}\n"
 
+# Exit with log if UDW dataset is missing courses
+def data_validation (incoming_courses,existing_courses):
+    incoming_courses = df['course_id']
+    existing_courses = get_queryset().get_supported_courses()
+
+    try:
+        for course in existing_courses:
+                if course in incoming_courses:
+                    print(f"{course} found")
+                    continue
+                else:
+                    logger.exception(f"missing {course} from incoming dataset from UDW.")
+
+                    return
+
+    except Exception:
+        logger.exception(f"Missing course data for {mysql_table}")
+        raise
+
 # execute database query
 def executeDbQuery(query):
     with engine.connect() as connection:
         connection.detach()
         connection.execute(query)
+
 
 # remove all records inside the specified table
 def deleteAllRecordInTable(tableName):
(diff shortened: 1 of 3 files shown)


zqian:
I dropped the "Deadline met!" label. I have not review the code yet, and we cannot make the claim by only creating pull request.

dashboard/cron.py
```diff
@@ -149,13 +169,17 @@ def update_with_udw_file(self):
 
         logger.debug("in update with udw file")
 
+        #Log an error if the course is invalid
+
```
zqian: where is the log message for errors?

dashboard/cron.py
```diff
@@ -54,19 +54,39 @@ def util_function(UDW_course_id, sql_string, mysql_table, table_identifier=None)
...
         logger.exception(f"Error running to_sql on table {mysql_table}")
         raise
 
     # returns the row size of dataframe
     return f"inserted + {str(df.shape[0])} rows in table {mysql_table} for course {UDW_course_id}\n"
 
+# Exit with log if UDW dataset is missing courses
+def data_validation (incoming_courses,existing_courses):
+    incoming_courses = df['course_id']
+    existing_courses = get_queryset().get_supported_courses()
```
zqian: we have discussed yesterday that the get_supported_courses() function merely returns what's inside the Admin configuration. 

It does not verify against the UDW to check whether the given course id is a valid Canvas course id.

The whole purpose of this request is to check the course id against UDW.

DanielDugan:
I understand.
I will work to re-earn that label.

dashboard/cron.py
DanielDugan: That is accurate.
The existing courses are pulled from the Admin config.

The feed from UDW is "incoming_courses"

zqian: This is new function "data_validation" added in this pull request. 

Where is it used?

How do you know the parameter "incoming_courses" is the feed from UDW?

DanielDugan: This is new function "data_validation" added in this pull request.

Where is it used?

I created it as a function so it can be called in the future.
I want to reuse this code so that we can load individual courses when the data feed is incomplete.
It is currently taking in incoming and existing courses as arguments and comparing them to see if data is missing.
If data is missing, the cron fails.

DanielDugan: How do you know the parameter "incoming_courses" is the feed from UDW?

There is a pandas dataframe generated earlier in the code.
Would you like me to review that with you?

DanielDugan: I will go ahead and take initiative here.

The code itself is very well documented, so it will be easy to review on how I know it is the feed.
```

# the util function
def util_function(UDW_course_id, sql_string, mysql_table, table_identifier=None):
    df = pd.read_sql(sql_string, conns['UDW'])
    logger.debug(df)

    # Sql returns boolean value so grouping course info along with it so that this could be stored in the DB table.
    if table_identifier == 'weight' and UDW_course_id:
        df['course_id']=UDW_course_id
        df.columns=['consider_weight','course_id']

    # drop duplicates

    logger.debug(" table: " + mysql_table + " insert size: " + str(df.shape[0]))

    # write to MySQL
    try:
        df.to_sql(con=engine, name=mysql_table, if_exists='append', index=False)
    except Exception:
        logger.exception(f"Error running to_sql on table {mysql_table}")
        raise

    # returns the row size of dataframe
    return f"inserted + {str(df.shape[0])} rows in table {mysql_table} for course {UDW_course_id}\n"
```

Lines 39-62 of cron.py

DanielDugan: I then extract the course id's from the UDW dataframe with 
`incoming_courses = df['course_id']`

Does that clarify your inquiry?

zqian: >I then extract the course id's from the UDW dataframe with
>incoming_courses = df['course_id']
> Does that clarify your inquiry?

No, it does not. 

The df variable is not a global variable, nor it is defined within data_validation().

This line will throw error when the data_validation() is called.

DanielDugan: Thanks for the insight.
That hasn't been my experience, but I will investigate it.
Have you tested it locally to confirm an error?

dashboard/cron.py
DanielDugan: I will submit updated code today.

dashboard/cron.py
DanielDugan: Updated for 1:1

dashboard/cron.py
```diff
@@ -61,8 +61,7 @@ def util_function(UDW_course_id, sql_string, mysql_table, table_identifier=None)
     # returns the row size of dataframe
     return f"inserted + {str(df.shape[0])} rows in table {mysql_table} for course {UDW_course_id}\n"
 
-# Exit with log if UDW dataset is missing courses
-def data_validation (incoming_courses,existing_courses):
+    # exit cron with log if UDW dataset is missing courses
```
zqian: This is a fundamental error in coding:

No line of code will be executed after the return statement on line 62

DanielDugan: Nice catch!
That is not code I edited, but I should have seen that.
Thank you for informing me.

DanielDugan: In this line of code, I removed the unused cast of an Exception "as e", as that wasn't used.
It may be wise for future releases to include casting an Exception as e to widen the "net" for for caught exeptions.

https://stackoverflow.com/questions/18982610/difference-between-except-and-except-exception-as-e-in-python

zqian: What do you mean by "That is not code I edited"? 

The removal of line 64 and line 65 is inside the 474e717d5acf3ae411ee1d16347d1f7b27c4476f commit.

zqian: There is also a warning from Codacy for "Unreachable code".

zqian:
Zhen has reworked issue_292 and merged in the changes. Close this PR.

Issue #359, opened 2019-01-18
Text changes to assignment and files view 


pushyamig:
Change “Reset as my default” to “Remember my setting”
Change “Current Default” to “My current setting”
Change text to “Files accessed from week 1 to 16 by students with these grades:”  
Change “All Grades” to “All”

pushyamig:
This is ready for testing

jennlove-um:
Testing passes.  Text has been updated on Files Accessed and Assignment Planning pages.

PR #360, opened 2019-01-18
#359 text changes for assignment and files view

pushyamig:


dashboard/static/js/file_access_within_week.js
```diff
@@ -18,6 +18,10 @@ var mini_margin = {top: 50, right: 10, bottom: 50, left: 10},
     mini_width = 100 - mini_margin.left - mini_margin.right,
     mini_height = 400 - mini_margin.top - mini_margin.bottom;
 
+const REMEMBER_MY_SETTING = 'Remember my setting';
```
jonespm: Since this text is duplicated in 2 files I think we should define these in a separate file and include that file before both of these 2 javascript files are included. That would help i18n future consideration.

pushyamig: ya. I have been thinking about that. We need to have a solution to have a JS static content loaded from the separate file. I have the issue opened [here](https://github.com/tl-its-umich-edu/my-learning-analytics/issues/361)

jonespm: Thanks, something that would be nice to have for spring/summer.

PR #366, opened 2019-01-20
issue_292 check for invalid course ids

zqian:


dashboard/cron.py
```diff
@@ -80,6 +80,37 @@ class DashboardCronJob(CronJobBase):
...
+    def invalid_course_ids(self):
+
+        # whether all course ids are valid ids
+        invalid_course_id_list = []
+
+        logger.debug("in checking course")
+
+        # loop through multiple course ids
+        for course_id in Course.objects.get_supported_courses():
+            if (not course_id.isdigit()):
```
pushyamig:
```suggestion
            if not course_id.isdigit():
```

dashboard/cron.py
```diff
@@ -358,26 +389,34 @@ def do(self):
 
         status += "Start cron at: " +  str(datetime.datetime.now()) + "\n"
 
-        logger.info("************ term")
-        status += self.update_with_udw_term()
+        invalid_course_id_list = self.invalid_course_ids()
+        logger.debug(f"invalid id {invalid_course_id_list}")
+        if (len(invalid_course_id_list) > 0):
```
pushyamig:
```suggestion
        if len(invalid_course_id_list) > 0:
```

dashboard/cron.py
```diff
@@ -80,6 +80,37 @@ class DashboardCronJob(CronJobBase):
     schedule = Schedule(run_at_times=settings.RUN_AT_TIMES)
     code = 'dashboard.DashboardCronJob'    # a unique code
 
+    # return a list of invalid course ids
+    def invalid_course_ids(self):
+
+        # whether all course ids are valid ids
+        invalid_course_id_list = []
```
pushyamig: Optional: more [info](https://docs.python.org/3/library/typing.html#type-aliases)
```
from typing import List
invalid_course_id_list = List[str] 
```

dashboard/cron.py
```diff
@@ -80,6 +80,37 @@ class DashboardCronJob(CronJobBase):
...
+                course_sql = f"""select * 
+                            from course_dim c
+                            where c.id = '{course_id}'
+                            """
+                logger.debug(course_sql)
+                course_df = pd.read_sql(course_sql, conns['UDW'])
+
+                # error out when course id is invalid
+                if course_df.empty:
+                    logger.error(f"""Course {course_id} is invalid. """)
```
pushyamig: I don't think it is an invalid course_id but there is no info in the UDW for that course. So that msg should is appropriate to that `f"Course {course_id} don't have the entry in UDW yet"`

dashboard/cron.py
jonespm: There's not much in the style guide about if statements. I believe either form is _fine_ but best practice is for single line if statement to **not** have parenthesis but multiple lines should, so looks good.

https://www.python.org/dev/peps/pep-0008/#multiline-if-statements

dashboard/cron.py
```diff
@@ -80,6 +80,37 @@ class DashboardCronJob(CronJobBase):
     schedule = Schedule(run_at_times=settings.RUN_AT_TIMES)
     code = 'dashboard.DashboardCronJob'    # a unique code
 
+    # return a list of invalid course ids
+    def invalid_course_ids(self):
```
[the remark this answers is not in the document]
jonespm: It sounds like it's more like **verify_course_ids**

dashboard/cron.py
jonespm: I'd probably make this an attribute in the class so it could be used by the other methods later. We could alter this to have the other methods just skip the invalid courses rather than stop the job entirely.

jonespm: This is probably fine that it is for now, that's an easy change later. I'm thinking this is the "good" solution. The better solution would be if this could continue to process the good courses, but I wouldn't worry about that for now.

dashboard/cron.py
```diff
@@ -358,26 +389,34 @@ def do(self):
 
         status += "Start cron at: " +  str(datetime.datetime.now()) + "\n"
 
-        logger.info("************ term")
-        status += self.update_with_udw_term()
+        invalid_course_id_list = self.invalid_course_ids()
+        logger.debug(f"invalid id {invalid_course_id_list}")
+        if (len(invalid_course_id_list) > 0):
+            # error out and stop cron job
+            status += f"ERROR: Those course ids are invalid: {invalid_course_id_list}\n"
```
jonespm: Rather than indenting all of this in an else you could have just 

`return status`

here.

dashboard/cron.py
zqian: parentheses removed.

PR #375, opened 2019-01-23
All API urls should have client side caching enabled

### LINKED ISSUE #374: All API urls should have client side caching enabled
We've noticed that many of the URL's take awhile to load and only really change once every 24 hours. We should implement client side caching to both give the user a better experience as well as the server an easier time. This would also greatly benefit the new courses enabled api.

Some screenshots

Before (Almost a second to generate this last URL)
image

(Since it hasn't changed and browser is caching it it's instant)
image

The best solution might be to calculate when the next run of the cron will be and use that value, but a set value should be good for now.

jonespm:
Fixes #374

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -120,3 +120,6 @@ CRONTAB_SCHEDULE =
 
 # Change this to set the max default weeks to allow. Default is currently 16. The issue is the end dates in Canvas currently are set 10 years out so it can't calculate the range.
 MAX_DEFAULT_WEEKS = 
+
+# How long to cache some URL's on the client (Defualt 3600 seconds)
+CLIENT_CACHE_TIME =
--- a/dashboard/urls.py
+++ b/dashboard/urls.py
@@ -48,23 +48,24 @@ urlpatterns = [
     url(r'^courses/(?P<course_id>[0-9]+|)', login_required(TemplateView.as_view(template_name='courses.html')), name="courses"),
 
     # Thse URL's are data patterns
-    # get file access patterns
+    # GET access patterns
     url(r'^api/v1/courses/(?P<course_id>[0-9]+)/grade_distribution', 
         cache_page(settings.CLIENT_CACHE_TIME)(login_required(views.grade_distribution)), name='grade_distribution'),
     url(r'^api/v1/courses/(?P<course_id>[0-9]+)/file_access_within_week', 
         cache_page(settings.CLIENT_CACHE_TIME)(login_required(views.file_access_within_week)), name='file_access_within_week'),
     url(r'^api/v1/courses/(?P<course_id>[0-9]+)/assignments', 
         cache_page(settings.CLIENT_CACHE_TIME)(login_required(views.assignments)), name='assignments'),
-    url(r'^api/v1/courses/(?P<course_id>[0-9]+)/set_user_default_selection', 
-        cache_page(settings.CLIENT_CACHE_TIME)(login_required(views.update_user_default_selection_for_views)), name='update_user_default_selection_for_views'),
     url(r'^api/v1/courses/(?P<course_id>[0-9]+)/get_user_default_selection', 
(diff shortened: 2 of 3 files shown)


dashboard/urls.py
```diff
@@ -46,15 +49,22 @@
...
-    url(r'^api/v1/courses/(?P<course_id>[0-9]+)/set_user_default_selection', login_required(views.update_user_default_selection_for_views), name='update_user_default_selection_for_views'),
-    url(r'^api/v1/courses/(?P<course_id>[0-9]+)/get_user_default_selection', login_required(views.get_user_default_selection), name='get_user_default_selection'),
-    url(r'^api/v1/courses/(?P<course_id>[0-9]+)/info', login_required(views.get_course_info), name='get_course_info'),
+    url(r'^api/v1/courses/(?P<course_id>[0-9]+)/grade_distribution', 
+        cache_page(settings.CLIENT_CACHE_TIME)(login_required(views.grade_distribution)), name='grade_distribution'),
+    url(r'^api/v1/courses/(?P<course_id>[0-9]+)/file_access_within_week', 
+        cache_page(settings.CLIENT_CACHE_TIME)(login_required(views.file_access_within_week)), name='file_access_within_week'),
+    url(r'^api/v1/courses/(?P<course_id>[0-9]+)/assignments', 
+        cache_page(settings.CLIENT_CACHE_TIME)(login_required(views.assignments)), name='assignments'),
+    url(r'^api/v1/courses/(?P<course_id>[0-9]+)/set_user_default_selection',
```
pushyamig: This is actually a PUT call based on user default selection why this needs to be cached?

jonespm: Which URL are you referring to? I thought these were all GET's.

pushyamig: `/set_user_default_selection`

jonespm: Oh you mean the set_user_default_selection? Yeah that probably should be excluded thanks for checking. I wasn't knew there might have been 1-2 that weren't GET's

pushyamig:
I was just reading about this [cache_page](url), is there any reason why you did not use `@cache_page` decorator in the `view.py` across each request. I get it as having [cache_page](url) in url.py is clean.

.env.sample
```diff
@@ -119,4 +119,7 @@ CRONTAB_SCHEDULE =
 
 
 # Change this to set the max default weeks to allow. Default is currently 16. The issue is the end dates in Canvas currently are set 10 years out so it can't calculate the range.
-MAX_DEFAULT_WEEKS = 
\ No newline at end of file
+MAX_DEFAULT_WEEKS = 
+
+# How long to cache some URL's on the client (Defualt 3600 seconds)
```
pushyamig: what do think our configuration option will be? based on the cron update schedule?

jonespm:
Yeah, I feel like I had the same thought back when implementing the login_required. If it's in the urls then it's obvious which URL's are protected and not and easier to see at a glance if one is missing (like for instance the case you pointed out with the setter). That could have also been used as a decorator in the views. 

It's more of a "which one is easier to spot any potential issues and have everything in one place". That's really the reason I chose urls over views. 

I feel like if we wanted to refactor our views file to have an authenticated views and unauthenticated views in separate files that could work. Just like if we wanted to split up cached and uncached apis in separate files. There's probably other ways to refactor this as the project grows? I don't know best practice.

.env.sample
jonespm: Ideally yes, it should calculate when the cron run will next occur and use that. 

However we don't always set the RUN_AT_TIMES value on this server (though we could) and that value would be something like 2:00AM. So it would have to be like

`
# If RUN_AT_TIMES is defined calculate how long we should cache for
if RUN_AT_TIMES:
  # Do some calculation to subtract the next time cron will run from the current time and set that as the cache time. Maybe add a 15 minutes for padding since cron takes awhile.
  # TODO Implement code that does this  ;)
`

jonespm: I don't think this would be very many lines of code but not feels like it would require more testing than just setting a fixed value, so could create a new issue.

Issue #376, opened 2019-01-23
Calls to set_user_default_selection should be POST or PUT


jonespm:
The calls to set_user_default_selection are via GET. 

A GET request should be able to be cached and have a predictable result.
https://softwareengineering.stackexchange.com/a/188861/169485

In HTTP the correct very is PUT to update, or POST to create (or often also update).

jennlove-um:
I am able to set and reset the defaults on both the Files Accessed and Assignment Planning pages.  The correct default displays for the student.

PR #377, opened 2019-01-24
#376 making the set default calls as PUT

pushyamig:
#376, Making the call that save the default as PUT instead of GET. Followed the guidelines prescribed by Django doc
https://docs.djangoproject.com/en/2.1/ref/csrf/#acquiring-the-token-if-csrf-use-sessions-or-csrf-cookie-httponly-is-true

### DIFF
--- a/dashboard/static/js/controllers.js
+++ b/dashboard/static/js/controllers.js
@@ -1,10 +1,10 @@
-visApp.controller('mainController', ['$scope', 'Get','$location', '$anchorScroll', function($scope, Get,$location,$anchorScroll) {
+visApp.controller('mainController', ['$scope', 'Service','$location', '$anchorScroll', function($scope, Service,$location,$anchorScroll) {
     function dataCall(call, default_value){
         if (typeof default_value !== 'undefined') {
             $scope.percent_filter = default_value;
         }
         var dataFile = '/api/v1/courses/' + dashboard.course_id + '/assignments?percent=' + $scope.percent_filter;
-        Get.getData(dataFile).then(function (data) {
+        Service.getData(dataFile).then(function (data) {
             if (_.isEmpty(data.data)) {
                 $scope.plan = [];
                 $scope.progress = [];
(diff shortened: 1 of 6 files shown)


dashboard/static/js/controllers.js
```diff
@@ -71,8 +71,9 @@ visApp.controller('mainController', ['$scope', 'Get','$location', '$anchorScroll
     $scope.sendDefaults = function () {
         $scope.isResetMyDefaultCheckboxEnabled = false;
         if ($scope.default_selection_model) {
-            var defaultsURL = '/api/v1/courses/' + dashboard.course_id + '/set_user_default_selection?assignment=' + $scope.percent_filter;
-            Get.getData(defaultsURL).then(function (data) {
+            var url = '/api/v1/courses/' + dashboard.course_id + '/set_user_default_selection';
+            let putData = {"assignment":$scope.percent_filter};
```
jonespm: Nice, this actually looks better than the old code too.

dashboard/static/js/file_access_within_week.js
```diff
@@ -584,18 +584,30 @@ default_selection_logic_on_grade_selection = function(){
...
-        if (initResult.default === 'fail') {
-            $("#label_for_default_selection").html(SETTING_NOT_UPDATED_MSG);
-            $('#default_selection').prop('checked', false);
-            return;
+    let data = {"file":selection};
+    // https://docs.djangoproject.com/en/2.1/ref/csrf/#acquiring-the-token-if-csrf-use-sessions-or-csrf-cookie-httponly-is-true
+    $.ajax({
+        url: "/api/v1/courses/" + dashboard.course_id + "/set_user_default_selection",
+        method: 'PUT',
+        data: JSON.stringify(data),
```
jonespm: I think you want to set the contentType here to application/json

`    contentType: "application/json; charset=utf-8",
`

jonespm: Do you know why this has to be json encoded at all Rather than just plain ol x-www-form-urlencoded? 

I'll send a diff suggestion in a sec.

jonespm: Nevermind, it *sounds* like the problem is it's hard to get the angular form to send regular application/x-www-form-urlencoded data. I'd just add this contentType here and leave it as-is. 

Othewise this jquery is sending this as 
"Content-Type: application/x-www-form-urlencoded; charset=UTF-8"

pushyamig: Done updated to json content type

pushyamig: I think if the format is json when it reaches backend the parsing json is better. I looked few code sample they were using json format with PUT call and felt of doing the same. i did not think much on this

jonespm: I the code having it urlencoded I just had to pass to QueryDict and it came out almost the same as this code. But I couldn't get Angular to do anything different than json easy so just gave up. Seemed about the same though.

PR #378, opened 2019-01-24
Added eslint config, fixed undeclared var and removed unused var

justin0022:
Noticed that there were undeclared/unused variables in the JavaScript files, so I fixed them. 

I intentionally didn't touch `app.js`, `assign-progress.js` and `controllers.js` - I'll fix it after https://github.com/tl-its-umich-edu/my-learning-analytics/pull/377 lands.


jonespm:
It looks like all of these var should be let instead.

justin0022:
I've changed all `var` to let`.

PR #381, opened 2019-01-25
 Enable managed database 

### LINKED ISSUE #317: Removed managed=false, create database from models
After #305 we will have models but the models currently aren't being used to create or manage the database. We should remove the "manged=false" and create migrations so that django can manage the schema of the tables.

andrew-gardener:
- Tried make models match as closely as possible to current `mysql/init.sql`
- Added management functions to add terms and courses
- added instructions to add default terms + courses in README.md
- Cleaned up docker-compose for mysql (default mysql env variables similarly named to app env variables so just use the .env file for both)
- docker-compose mysql uses utf8mb4 by default now (no need to set in init script though prod installs will still need to manually set this)
- Cleaned up existing migrations and db init scripts

Depends on PR #346 (Also mentions `DATA_WAREHOUSE_ID_TEMPLATE` in README.md from PR #350)
Closes #317


jonespm:
I feel like this change will be at the end because of dependencies, so I'll wait to review it until the other changes are merged. I was also planning on testing this one out.

jonespm:
Okay, I think master is pretty stable now, I plan to try this one out soon.

docker-compose.yml
```diff
@@ -5,25 +5,15 @@ services:
...
-      - MYSQL_PASSWORD=${MYSQL_PASSWORD}
-      - MYSQL_PORT=${MYSQL_PORT}
-
-      - UDW_ENDPOINT=${UDW_ENDPOINT}
-      - UDW_USER=${UDW_USER}
-      - UDW_PASSWORD=${UDW_PASSWORD}
-      - UDW_PORT=${UDW_PORT}
-      - UDW_DATABASE=${UDW_DATABASE}
-    entrypoint: ['docker-entrypoint.sh', '--default-authentication-plugin=mysql_native_password']
+      - MYSQL_ROOT_PASSWORD=student_dashboard_root_pw
```
jonespm: Do you know why this particular value was left here?

andrew-gardener: I left it there since `MYSQL_ROOT_PASSWORD` isn't in `.env.sample` (so wouldn't be included automatically in a normal setup) and I wasn't sure if you guys would still need it or not

jonespm: [I see it in there!](https://github.com/tl-its-umich-edu/my-learning-analytics/blob/f814032134c51ad5bf6a0b99eb55c9321330a7cc/.env.sample#L36) ;) I don't see anything recent with the blame for that.

This file has a conflict too, though I didn't see what the conflict was on.

andrew-gardener: oh right, sorry about that. Its been a while since I last looked at this

jonespm: I know, me too, sorry about that. This was going to be a pretty big change but will be great to have!

andrew-gardener: I'll remove this env variable, rebase to fix the conflict, and take a look over to make sure there haven't been any db changes since I last looked

andrew-gardener: alright, remove `MYSQL_ROOT_PASSWORD`, fixed conflict, and double check the models so there shouldn't be any problems atm

jonespm:
I finally got around to testing this and this looks great. Two thoughts.

We're going to need a conversion script to catch our existing databases up to this schema (since they already have this structure). I'm thinking that this would do it.

```
-- Add in the new migrations
INSERT INTO django_migrations (app, name, applied) VALUES ('dashboard','0001_initial', now());
```

We also want to convert the existing database and tables over to mb4 so we'll have to generate a script for that.

Also I think it would be nicer if we just had a shell script for creating demo courses rather than having to run a number of commands. I can submit a commit to move all those commands you had into their own script and update the docs, unless you have time to do that. Otherwise, it looks great and fixes some errors I had in the models. (I didn't test out all of the models)

Thanks!

README.md
```diff
@@ -51,29 +51,50 @@ Environment variables:
...
 
-## Add a default course to test with
+Before adding adding initial terms and courses, ensure that the `DATA_WAREHOUSE_ID_TEMPLATE` environment variable is set correctly
 
-There are some default courses pre-populated by the mysql/init.sql script. If you'd like to add additional courses you need to add them as an admin user. (See next section on adding an admin user)
+### Add terms
 
-## Create a super user to test login. 
+    docker exec -it student_dashboard bash
+    ./manage.py term --term_id=108 --name='SUMMER 2018' --date_start='2018-06-27 04:00:00' --date_end='2018-08-17 23:59:59'
```
jonespm: One thing to change, move all of these demo commands into their own shell script so we can just run one command. It's fine to have these documented here (For example how to manually add a term or course) but we should have some added as a demo via a single script. We *may* want to have and environment variable also set that would just automatically run this demo script, but don't have to do that now.

dashboard/models.py
```diff
@@ -82,22 +81,21 @@ def set_user_defaults(self, course_id, user_id, default_view_type, default_view_
 
 class UserDefaultSelection(models.Model):
     id = models.AutoField(primary_key=True, verbose_name="Table Id")
-    user_id = models.CharField(max_length=255, verbose_name="SIS Id")
-    course_id = models.CharField(max_length=255, verbose_name="Course Id")
-    default_view_type = models.CharField(max_length=255, verbose_name="Default Type")
-    default_view_value = models.CharField(max_length=255, verbose_name="Default Value")
+    course_id = models.CharField(max_length=255, blank=True, null=True, verbose_name="Course Id")
+    user_id = models.CharField(max_length=255, blank=True, null=True, verbose_name="SIS Id")
```
jonespm: I think I mixed this up initially but it's still mixed up
```suggestion
    user_id = models.CharField(max_length=255, blank=True, null=True, verbose_name="User Id")
```

README.md
andrew-gardener: created `demo_init.sh` with the term/course create commands and updated the readme

dashboard/models.py
andrew-gardener: fixed

Issue #384, opened 2019-01-29
browser caching is not helping with default setting option when changed by user


pushyamig:
All the GET calls are cached now that also means the GET call to get the defaults call too. When a user changes there a default setting in file/assignment view and refreshes or pop out from the views momentarily in the site and come back again since the get_default call is cache it serves the content from the cached value. The easy solution would remove the get_default call from the cached configuration. But looking for a solution that let to wipe the cache on demand during the setting of the default.

pushyamig:
Test Plan: from the chrome dev tool Network tab
1. `get_user_default_selection?default_type=` is removed from cached option so this URL will make a DB call to get the data.
2. All the other call that gets the data like `grade_distribution`, `file_access_within_week`, `assignments`, `info', `courses_enabled` will be coming from the cache if available during the 24 hour period.

jennlove-um:
The selected default displays correctly when changed within the 24 hour period.  It is not coming from the cached data.

PR #385, opened 2019-01-29
#384 removing the get defaults from caching

pushyamig:


jonespm:
Good catch, this was probably a little too aggressive on this caching. Rather than having this as an API call we might just want to embed the javascript for the defaults right on the page. Maybe we can revisit this later, but thanks for the fix!

PR #386, opened 2019-01-31
Changing the BigQuery Data set to new one

pushyamig:


pushyamig:
Merging it but still can review it later

jonespm:
Was there an issue describing this change?

jonespm:
It looks like it reduces the cost of running these queries by almost 99%. If this provides the same data this is great.

PR #388, opened 2019-02-04
Improve the courses admin view to display more details in the columns

### LINKED ISSUE #387: Improve the courses admin view to display more details in the columns
The courses admin view currently only displays the course name. It would be more helpful at a glance to display more details.

A mockup is below

image

jonespm:
Fixes #387

### DIFF
--- a/dashboard/admin.py
+++ b/dashboard/admin.py
@@ -1,5 +1,7 @@
 from django.contrib import admin
 from django.conf import settings
+from django.utils.safestring import mark_safe
+from django.template.defaultfilters import linebreaksbr
 
 from .models import CourseViewOption, Course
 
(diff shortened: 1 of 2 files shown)


dashboard/admin.py
```diff
@@ -11,6 +13,13 @@ class CourseViewOptionInline(admin.StackedInline):
 
 class CourseAdmin(admin.ModelAdmin):
     inlines = [CourseViewOptionInline,]
+    list_display = ('name', 'term_id','_courseviewoption')
```
zqian: For human readabilities:

name -> course name or user name
_courseviewoption -> course view option

zqian: do not see list_display or list_select_realted used anywhere....

jonespm: These are attributes of ModelAdmin (which this is extending)
https://docs.djangoproject.com/en/1.11/ref/contrib/admin/#django.contrib.admin.ModelAdmin.list_display

They are used to define which columns appear in the table. It's pretty neat. 

These fields have to either map against the columns in the database or a method.

The verbose_name from the model is used in the display, except for this special new method I added to display the courseviewoptions table. I had to add this special method here so it would display the newlines as html <br> on separate lines for better readability in the table.

zqian: Nice!

PR #393, opened 2019-02-05
AttributeError in CourseViewOption json if course does not exist

jonespm:


dashboard/models.py
```diff
@@ -197,13 +199,18 @@ def json(self):
...
+        try:
+            return {
+                self.course.canvas_id: {
+                    'fa': int(self.show_files_accessed),
+                    'ap': int(self.show_assignment_planning),
+                    'gd': int(self.show_grade_distribution),
+                }
             }
-        }
+        except ObjectDoesNotExist:
```
pushyamig:
```suggestion
        except (ObjectDoesNotExist, Exception) as e:
```

pushyamig: Sometimes we get exceptions that are unknown and  I would like to also capture the generic case

dashboard/models.py
```diff
@@ -197,13 +199,18 @@ def json(self):
...
+            logger.warn(f"CourseViewOption does not exist in Course table, skipping")
```
pushyamig: I feel the reason for exception should be printed from the Exception object as it would have additional info. isn't this should be an error than a warning?

dashboard/models.py
jonespm: Yeah, I could catch more here, though it's looks like the best patterns are to except the specific exceptions separate from the less specific ones.

https://docs.quantifiedcode.com/python-anti-patterns/correctness/bad_except_clauses_order.html

Catching the general Exception case usually isn't good either as it can potentially hide real issues that aren't noticed. I think for this feed we never want an error, maybe the general Exception should be moved up into the view instead.

dashboard/models.py
jonespm: I tried to get additional information (like the id that was causing an issue) but I couldn't figure out how to get it after about an hour of research. If you try to access "self" it causes an Exception. It might be easier to figure this out if that debugging patch was applied. I feel like the exception might have to be re-thrown and caught at a higher level for more details.

PR #396, opened 2019-02-06
Missing $ in startup script breaks ptvsd

### LINKED ISSUE #339: Add support for debugging with ptvsd
It would be very useful if we could debug in VSCode via ptvsd. This would save me some time testing some of the issues. It looks like it's not currently compatible with gunicorn so we'll have to use the runserver and have a separate launch for that.

https://gist.github.com/veuncent/1e7fcfe891883dfc52516443a008cfcb

jonespm:
Also remove the nothreading option as it works fine without.

Fixes #339

### DIFF
--- a/start.sh
+++ b/start.sh
@@ -21,7 +21,7 @@ echo Running python startups
 python manage.py migrate
 
 if [ -z "${IS_CRON_POD}" ]; then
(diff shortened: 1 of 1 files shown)


jonespm:
@pushyamig This is why nagios is giving an error in dev. ptvsd has to run as runserver. 

We "probably" should make nagios respect either

PR #398, opened 2019-02-08
Issue - 389 - Add total bytes billed for BQ call to cron logs

### LINKED ISSUE #389: Add total bytes billed for BQ call to the cron logs
It would be useful to have details on the BQ billed status to be inserted into the cron logs.

jonespm:
Fixes #389


jonespm:
Because of this issue https://github.com/Tivix/django-cron/issues/149 messages in the django-cron are limited to 1000 characters. So I shortened up some of the messages so they all print.

jonespm:
This log now displays like below, which gives the user a better sense (if they look for it) how much the query cost to run.

image

PR #400, opened 2019-02-15
Add Apache 2 License

jonespm:


zqian:
According to [the Applying Apache License site ](http://www.apache.org/dev/apply-license.html#new),  we also need a NOTICE file at the top level:

"In addition, a correct NOTICE file MUST be included in the same directory as the LICENSE file."

jonespm:
Ok, I'll add a notice tomorrow (Tuesday)

I also wanted to scan our dependencies with maybe something like
https://github.com/dhatim/python-license-check

Just to ensure the license is compatible with everything we're currently using. ;)

jonespm:
I have added the requested file(s), please review.

Issue #402, opened 2019-02-21
Updates to the "files" data to support other caliper events


jonespm:
Currently update_with_bq_access is only pulling down views for the files. Here's some changes that I'd like to implement to support more tools

- Also support ViewEvent/Viewed
  - This is the more appropriate event for what we're visualizing. However it looks like Canvas is currently only sending NavigationEvent/NavigatedTo for the files at least, both of these should be supported pretty similarly with the same query
- Change the "Files" view around to be more generic

We'll need to find a course that has some of these "Viewed" events from the textbook like VitalSource or Runestone first to implement this.

pushyamig:
For **Leccap data** the query to use is. Leccap data provides only short canvas course/user id
```
select 
CAST(SUBSTR(JSON_EXTRACT_SCALAR(event, '$.object.id'), 48) AS STRING) AS resource_id,
JSON_EXTRACT_SCALAR(event, '$.federatedSession.messageParameters.custom_canvas_user_id') AS user_id,
JSON_EXTRACT_SCALAR(event, "$.federatedSession.messageParameters.custom_canvas_course_id") As course_id,
datetime(EVENT_TIME) as access_time
FROM event_store.events
where JSON_EXTRACT_SCALAR(event, '$.edApp') = 'https://leccap.engin.umich.edu/#applicationName=Lecture+Capture'
and type = 'MediaEvent'
and JSON_EXTRACT_SCALAR(event, '$.action') = 'Started'
and JSON_EXTRACT_SCALAR(event, '$.federatedSession.messageParameters.custom_canvas_course_id') is not null
and JSON_EXTRACT_SCALAR(event, "$.federatedSession.messageParameters.custom_canvas_course_id") like '%298053%';
```
Leccap Additional insights: ed_app is used for Session#loggedIn events

```select type, action  From event_store.events where ed_app ='https://leccap-beta.engin.umich.edu/#applicationName=Lecture+Capture';```
looking at different leccap ed_app
`select distinct ed_app from event_store.events where ed_app like '%leccap%';`

For **Files canvas** data, only those 2 ed_app generates file looked event as a form of navigation events. You may also find a bunch of cluster machines generated event which can be ignores as these are duplicates of user action. `http://a1770-9792514.cluster32.canvas-user-content.com/ `


```
select event, CAST(SUBSTR(JSON_EXTRACT_SCALAR(event, '$.object.id'), 35) AS STRING) AS file_id,
[...]

PR #404, opened 2019-02-21
Improve BQ call to return multiple courses

### LINKED ISSUE #403: Improve BQ call to return multiple courses in one query
We're currently iterating and calling BQ individually for each course. This requires the large data set to be processed through each time. If we pulled down all (or more) courses in the same query it would be a lot more efficient and save some money/time!

jonespm:
Fixes #403

### DIFF
--- a/dashboard/cron.py
+++ b/dashboard/cron.py
@@ -36,6 +36,10 @@ engine = create_engine("mysql+mysqldb://{user}:{password}@{host}:{port}/{db}?cha
                                host = db_host,
                                port = db_port))
 
+# Split a list into *size* shorter pieces
+def split_list(l: list, size: int = 10):
+    return [l[i:i + size] for i in range(0, len(l), size)]
+
 # the util function
 def util_function(UDW_course_id, sql_string, mysql_table, table_identifier=None):
     df = pd.read_sql(sql_string, conns['UDW'])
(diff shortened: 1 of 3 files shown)


jonespm:
I compared the results and the table before and after this change had the same data in it. 

The data only processed through since it processes in groups of 10 (I think I'll increase it to 20 and make it configurable) at the moment. This could *probably* be increased to larger size groups, but this is just a small optimization.

dashboard/cron.py
```diff
@@ -226,24 +230,25 @@ def update_with_bq_access(self):
                     for table in tables:
                         if ("enriched_events" == table.table_id):
                             logger.debug('\t{}'.format("found table"))
-                            # loop through multiple course ids
-                            for UDW_course_id in Course.objects.get_supported_courses():
+                            # loop through multiple course ids, 10 at a time
```
zqian: The comment mentioned "10 at a time", but I did not see the limit specified elsewhere.

dashboard/cron.py
```diff
@@ -226,24 +230,25 @@ def update_with_bq_access(self):
...
+                                        where JSON_EXTRACT_SCALAR(event, '$.edApp.id') = 'http://umich.instructure.com/'
+                                        and type = 'NavigationEvent'
+                                        and JSON_EXTRACT_SCALAR(event, '$.object.name') = 'attachment'
+                                        and JSON_EXTRACT_SCALAR(event, '$.action') = 'NavigatedTo'
+                                        and JSON_EXTRACT_SCALAR(event, '$.membership.member.id') is not null
+                                        and SUBSTR(JSON_EXTRACT_SCALAR(event, "$.group.id"),31) IN UNNEST(@course_ids)
+                                        """
                                 logger.debug(query)
+                                logger.debug(udw_course_ids)
                                 query_params =[
```
zqian: Maybe it would be better to move the param (@course_ids) definition in front of the query line (line 237)

dashboard/cron.py
```diff
@@ -226,24 +230,25 @@ def update_with_bq_access(self):
                     for table in tables:
```
pushyamig: in the function, `update_with_bq_access` from line 220 - 272 need to be chunked to the small set of pieces. It's almost unreadable unless you pay close attention.

dashboard/cron.py
jonespm: This actually increased to 20 by default, it's in the split_list method. I can be more explicit about this comment.

dashboard/cron.py
jonespm: This is just to debug print both statements, it doesn't use it until line 251. This debug statement could probably be improved.

dashboard/cron.py
jonespm: Yeah, it is complex, but I didn't write it the first time! ;) Some of these if's are probably unnecessary. 

I think this will have to be rewritten completely to be able to handle other BQ calls like for textbooks, just wanted to make this change now (to retrieve records in bulk) as it's a cost/time savings. I think we could address the method complexity in #402

jonespm: I also feel like we might be "close" to having unittests for some of these methods too, when the app has more mock data. That would make large refactors a little safer.

pushyamig: I agree, unit test make it easier to test and we don't need to rely on functional testing each time

jonespm: I rewrote this whole method eliminating most of the if statements as they looked to be unnecessary and were actually outdated based on the recent changes. I'm not sure why there were in there in the first place. Maybe it looks better for you now?

Issue #414, opened 2019-03-01
Bootstrap alert


zqian:
CVE-2019-8331 More information
moderate severity
Vulnerable versions: >= 4.0.0, < 4.3.1
Patched version: 4.3.1
In Bootstrap 4 before 4.3.1 and Bootstrap 3 before 3.4.1, XSS is possible in the tooltip or popover data-template attribute. For more information, see: https://blog.getbootstrap.com/2019/02/13/bootstrap-4-3-1-and-3-4-1/

zqian:
Bootstrap version upgraded, and GitHub no longer shows alert.

PR #430, opened 2019-03-12
#414 changing the boostrap version

pushyamig:


package.json
```diff
@@ -9,7 +9,7 @@
     "d3-tip":"~0.9.1",
     "jquery": "~3.3.1",
     "underscore":"",
-    "bootstrap": "~4.1.3",
+    "bootstrap": "~4.3.1",
```
jonespm: It looks like popper.js also has a new version of 1.14.7. Though I guess the ^ syntax will pick this up.

https://stackoverflow.com/a/22345808/3708872

So that's probably ok.

jonespm: Though we might want to avoid using the ^ syntax for slightly more controlled versions, then it will only pick up the latest minors.

package.json
```diff
@@ -9,7 +9,7 @@
...
 	"popper.js":"^1.14.3",
```
jonespm:
```suggestion
	"popper.js":"~1.14.7",
```

Issue #428, opened 2019-03-11
clean up the React code that has lot of visuals not related to MyLA project


pushyamig:
Currently, I see a lot of visuals in the code like `donut chart` `Sankey chart ` not related to the project. These things needs to be removed to keep things simple and clear

justin0022:
Those visuals are components that can be used for future visualizations. I'll remove them, since the components exist [here](https://github.com/justin0022/dashboard).

PR #431, opened 2019-03-12
Remove unused components

justin0022:
Addresses https://github.com/tl-its-umich-edu/my-learning-analytics/issues/428

Left in some unused components that I thought would be used in the near future, like `Scatterplot`, `LineChart`, and `EmojiFeedback`. 

All removed components can be found in [this repo](https://github.com/justin0022/dashboard), in case we need them in the far future.


assets/src/components/GradeDistribution.js
```diff
@@ -1,9 +0,0 @@
-import createHistogram from './d3/createHistogram'
-import withResponsiveness from './withResponsiveness'
-import createChartComponent from './createChartComponent'
-import compose from '../util/compose'
-
-export default compose(
-  withResponsiveness,
-  createChartComponent
-)(createHistogram)
```
pushyamig: I thought this file is being used for Grades view?

pushyamig:
You can leave the `FileAccessed.js` I know it will be used. But other files are not relevant so it would be a distraction.

assets/src/components/GradeDistribution.js
justin0022: The file name is misleading, it isn't being used. 

The dependencies for Grade Distribution view are:
```javascript
import React from 'react'
import { renderToString } from 'react-dom/server'
import useFetch from '../hooks/useFetch'
import { withStyles } from '@material-ui/core/styles'
import Paper from '@material-ui/core/Paper'
import Grid from '@material-ui/core/Grid'
import Histogram from '../components/Histogram'
import Spinner from '../components/Spinner'
import createToolTip from '../util/createToolTip'
import Table from '../components/Table'
import { average } from '../util/math'
```

justin0022:
> You can leave the `FileAccessed.js` I know it will be used. But other files are not relevant so it would be a distraction.

Done!

Issue #432, opened 2019-03-12
useFetch hook does not properly handle subsequent API calls 


justin0022:
Currently, the `useFetch` hook implementation relies on the data value being `null` until the GET request completes. The problem is, on subsequent calls, such as when the `assignmentFilter` value is changed via the dropdown selector, the data value is set no longer `null`, as it was set on the previous useFetch call.
https://github.com/tl-its-umich-edu/my-learning-analytics/blob/7913f803d7c0a3a4669d8d783e9d63805f74b073/assets/src/containers/AssignmentPlanning.js#L30

I'm going to reimplement `useFetch` so that it returns a tuple of variables as follows: `[loading, data]` 

This will allow us to handle when new data is being fetched - for instance, showing a spinner.

PR #433, opened 2019-03-12
Refactored useFetch to return tuple

justin0022:
Fixes https://github.com/tl-its-umich-edu/my-learning-analytics/issues/432

Especially useful for handing multiple calls to `useFetch` on one page - with this implementation, user can be informed while new data is being fetched.


justin0022:
Closing so I can make this PR from my fork.

PR #434, opened 2019-03-13
Added Jest snapshot test

justin0022:
`createBarChart.test.js.snap` file contains the automatically generated Jest snapshot.

It gets automatically generated on the first test run, and future test runs will compare the snapshot to this one, and fail if the snapshot does not match. 

The command to run the tests is: `docker exec -it webpack_watcher npm test`
The command to update the snapshots (if the new snapshots are correct, and the old snapshot is outdated): `docker exec -it webpack_watcher npm run update-snapshot`


justin0022:
If snapshot testing for D3 looks good, I'll add snapshots for all of the D3 components.

Adonais0:
The D3 snapshot testing looks good to me. We might want to split the testing into three parts: 

- Snapshot test for D3
- Unit test for utility functions
- Enzyme test for React components

assets/src/__tests__/util/math.test.js
```diff
@@ -0,0 +1,7 @@
+import { average } from '../../util/math'
+
+describe('average', () => {
+  it('takes an array of numbers and calculates average', () => {
+    expect(average([1,2,3,4,5])).toEqual(3)
+  })
+})
```
Adonais0: It's better to create another pull request for the unit test for utility functions I think?

pushyamig: I think it is fine to have all test in one PR, it doesn't hurt that way.

Adonais0: gotcha

justin0022:
> The D3 snapshot testing looks good to me. We might want to split the testing into three parts:
> 
> * Snapshot test for D3
> * Unit test for utility functions
> * Enzyme test for React components

That sounds good to me.

pushyamig:
@justin0022 you can merge this PR once the conflicts are resolved. The conflicts are due to Andrews changes to the docker build got merged before this. Sorry!

PR #435, opened 2019-03-13
Refactored useFetch to return tuple

justin0022:
Fixes https://github.com/tl-its-umich-edu/my-learning-analytics/issues/432

Especially useful for handing multiple calls to `useFetch` on one page - with this implementation, user can be informed while new data is being fetched.

### DIFF
--- a/assets/src/containers/AssignmentPlanning.js
+++ b/assets/src/containers/AssignmentPlanning.js
@@ -27,7 +27,7 @@ function AssignmentPlanning (props) {
   const { classes, match } = props
   const currentCourseId = match.params.courseId
   const [assignmentFilter, setAssignmentFilter] = useState(0)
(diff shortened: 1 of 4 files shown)


assets/src/containers/AssignmentPlanning.js
```diff
@@ -72,27 +72,24 @@ function AssignmentPlanning (props) {
...
-                    <MenuItem value={20}>20%</MenuItem>
-                    <MenuItem value={50}>50%</MenuItem>
-                    <MenuItem value={75}>75%</MenuItem>
-                  </Select>
-                </FormControl>
-                {AssignmentTable(assignmentData.plan)}
-              </>
-              : <Spinner />
-            }
+            <>
```
Adonais0: Agree with this change. But just wonder why in the previous code the `<FormControl>` was also included in the chunk after the data is fetched.

justin0022: It's to make sure the dropdown selector doesn't disappear while the new data is being fetched.

Issue #420, opened 2019-03-11
add initial React+Webpack+Django plumbing


pushyamig:
This issue handles how React and Django communicate with each other. The `Web pack` is doing the heavy lifting and building the react components and `django-webpack-loader` is helping in finding the react static assets to django.

pushyamig:
@andrew-gardener I see that in the `settings.py`  `WEBPACK_LOADER` variable is referenced twice, I supposed it is by mistake?

pushyamig:
The docker build is much slower now is there a way to make this faster and I also see a lot of warning during the docker build some are related to babel.

justin0022:
I'll look into this tomorrow - I suspect swapping out Yarn for NPM and running the command via `npm ci` might speed up the dependency installation.

andrew-gardener:
I'll consider this done for now (will reopen if another issue comes up)

PR #436, opened 2019-03-13
Improve multi-stage build 

andrew-gardener:
- removed package-lock.json
- switched from node-slim to node-alpine for node containers in multistage build (runs much faster than node-slim)
- ensured that only production node_modules dependencies are added to the image (`--production` flag wasn't working for some reason, instead install all dependencies in webpack container and in prod container, prune to only production)
- improved `.dockerignore` file (basically the same as `.gitignore`)
- switch from yarn to npm (including django-yarn to django-npm)
- moved most libraries in package.json to dev dependencies. Technically only front-end only dependencies that aren't used by react should go in in non devDependencies since and these will mostly be removed/only used by react soon
- removed second declaration of `WEBPACK_LOADER` in settings.py
- updated a few node dependencies (there are still some warnings when building though)

Related to #420, https://github.com/tl-its-umich-edu/my-learning-analytics/issues/420#issuecomment-471711115, and https://github.com/tl-its-umich-edu/my-learning-analytics/commit/8c885b3b4056ef34dd233b6eb3500e4f402b10bd#commitcomment-32688738


pushyamig:
@andrew-gardener can you open the PR from you forked branch instead of creating the branch in the Myla Main repo. Like done [here](https://github.com/tl-its-umich-edu/my-learning-analytics/pull/434)

PR #437, opened 2019-03-13
Improve multi-stage build 

andrew-gardener:
- removed package-lock.json
- switched from node-slim to node-alpine for node containers in multistage build (runs much faster than node-slim)
- ensured that only production node_modules dependencies are added to the image (`--production` flag wasn't working for some reason, instead install all dependencies in webpack container and in prod container, prune to only production)
- improved `.dockerignore` file (basically the same as `.gitignore`)
- switch from yarn to npm (including django-yarn to django-npm)
- moved most libraries in package.json to dev dependencies. Technically only front-end only dependencies that aren't used by react should go in in non devDependencies since and these will mostly be removed/only used by react soon
- removed second declaration of `WEBPACK_LOADER` in settings.py
- updated a few node dependencies (there are still some warnings when building though)

Related to #420 and https://github.com/tl-its-umich-edu/my-learning-analytics/commit/8c885b3b4056ef34dd233b6eb3500e4f402b10bd#commitcomment-32688738


dashboard/templates/grades.html
```diff
@@ -19,7 +19,7 @@
     {% endif %}
 {% endblock %}
 {% block js %}
-    <script src="{% static 'd3-v4/build/d3.min.js' %}"></script>
+    <script src="{% static 'd3/dist/d3.min.js' %}"></script>
```
pushyamig: The reason we used `yarn` is we are using 2 different versions of the `d3 library` and `yarn` supported to install 2 different version of same library. I don't think it would be a problem as the React seems to be using the latest3` version. I am thinking the existing viz might not load properly

pushyamig:
I haven't tried running it locally but this looks good!

dashboard/templates/grades.html
andrew-gardener: yes, I actually removed the d3@4 version since we're probably only need one version once the react rewrite is done. I don't think we're currently using vis in the rewrite yet, but if its a problem we can try importing a second version again

jonespm: Ya that's right, yarn [seemed the best way to have two versions of the same library available.](https://stackoverflow.com/questions/26414587/how-to-install-multiple-versions-of-package-using-npm)

Back when this was created we were using nvd3 for some charts and that only supported d3 v3. It looks like that still doesn't support anything newer. However we migrated away from that and some pages used d3 v4 but some just didn't get checked for upgrading.

Issue #407, opened 2019-02-25
User table display user info irrespective of Course score info


pushyamig:
User table is a combination of user info and course grade info. Ideally, we want these 2 in separate tables. The issues is when there is no course grade info at least the user info should be populated in the table leaving the current and final score to null. This fix is to perform a left join then inner join.

pushyamig:
This is resolved with this PR https://github.com/tl-its-umich-edu/my-learning-analytics/commit/8405cd2beedc70e168484683909c9846996b0f99. I am closing this

jennlove-um:
Testing appears to pass.  Teachers are an example of users who will always have no course grades.  Logging in as a teacher, I can still see all the course visualizations correctly, so this appears to work as intended.

PR #438, opened 2019-03-14
Add basic access control

### LINKED ISSUE #314: Restrict the MyLA tool access to the people belonging to the  course only
Currently, the tool is accessible to anyone who has uniqname, once anybody figured out the course id can access the course analytics views. 
This could be controlled by restricting access only to the people in the course. The DB tables only have student enrollment data. We need to bring in the Teacher enrollment data to restrict instructors viewing only his course content. A question to answer is which enrollment type need access to tool? like TA/Observer.

andrew-gardener:
- added `enrollment_type` to user table (currently only stores `StudentEnrollment`, `TaEnrollment`, and `TeacherEnrollment` enrolments)
- modified user cron query to fetch student, ta, and instructor enrollments (instead of just students) as well as passing along `enrollment_type` info
- added rules (django-rules) for basic access control
	- Note missing 403 error handling to api calls from front end (didn't add due to react frontend rewrite) as well as a 403.html template

Related to #314


dashboard/cron.py
```diff
@@ -128,25 +128,29 @@ def update_user(self):
...
+                               u.global_canvas_id as id,
+                               e.current_score as current_grade,
+                               e.final_score as final_grade,
+                               e.type as enrollment_type,
+                               '{data_warehouse_course_id}' as course_id
+                        from user_dim u
+                        join pseudonym_dim p ON p.user_id = u.id
+                        join enrollment e ON u.id = e.user_id
+                        where p.workflow_state = 'active'
+                        and p.authentication_provider_id = '{settings.CANVAS_PRIMARY_AUTH_PROVIDER_ID}'
```
jonespm: This value authentication_provider_id looks to be "null" for all of our values in the table. Joining this this against pseudonym_dim would have to be changed to allow for "is null" or be an optional condition.

jonespm:
This needs to be updated to support the models change, init.sql has been removed.

dashboard/cron.py
```diff
@@ -128,25 +128,29 @@ def update_user(self):
...
-                        where e.course_id = '{data_warehouse_course_id}'
-                        and e.type='StudentEnrollment'
-                        and e.workflow_state='active' ) as e
-                        where p.user_id=u.id
-                        and u.id = e.user_id
-                        and c.enrollment_id =  e.enrollment_id
-                        and p.sis_user_id is not null
+            # select all student, tas, and instructors registered for the course
+            user_sql = f"""
+                        with enrollment as (
```
pushyamig: There was [commit](https://github.com/tl-its-umich-edu/my-learning-analytics/commit/8405cd2beedc70e168484683909c9846996b0f99) that made, I don't see that this PR is picking up that changes. The commit was based of this [issue](https://github.com/tl-its-umich-edu/my-learning-analytics/issues/407),those changes need to be included as well

dashboard/rules.py
```diff
@@ -0,0 +1,66 @@
...
+@rules.predicate
+def is_enrolled_in_course(user, course):
+    try:
+        User.objects.get(
+            sis_name=user.get_username(),
+            course_id=course.id,
+        )
+        return True
+    except User.DoesNotExist:
+        return False
```
pushyamig: log an error message referencing user x, course 1233 info

dashboard/cron.py
```diff
@@ -128,25 +128,29 @@ def update_user(self):
...
-                        (select e.user_id as user_id, e.id as enrollment_id from enrollment_dim e
-                        where e.course_id = '{data_warehouse_course_id}'
-                        and e.type='StudentEnrollment'
-                        and e.workflow_state='active' ) as e
-                        where p.user_id=u.id
-                        and u.id = e.user_id
-                        and c.enrollment_id =  e.enrollment_id
-                        and p.sis_user_id is not null
+            # select all student, tas, and instructors registered for the course
+            user_sql = f"""
```
pushyamig: There is something wrong with this query, I see that if you just have single course the loading the data through cron process works fine, but if you have multiple courses i see this error.
```sqlalchemy.exc.IntegrityError: (_mysql_exceptions.IntegrityError) (1062, "Duplicate entry '17700000000123456-17700000000098765' for key 'PRIMARY'") (Background on this error at: http://sqlalche.me/e/gkpj)```

File: dashboard/rules.py [hunk 5]
```diff
@@ -0,0 +1,66 @@
...
+
+is_admin_or_enrolled_in_course = is_admin | is_enrolled_in_course
+
+# api
+rules.add_perm('dashboard.get_course_info', is_admin_or_enrolled_in_course)
+rules.add_perm('dashboard.file_access_within_week', is_admin_or_enrolled_in_course)
+rules.add_perm('dashboard.grade_distribution', is_admin_or_enrolled_in_course)
+rules.add_perm('dashboard.update_user_default_selection_for_views', is_admin_or_enrolled_in_course)
+rules.add_perm('dashboard.get_user_default_selection', is_admin_or_enrolled_in_course)
+rules.add_perm('dashboard.assignments', is_admin_or_enrolled_in_course)
```
pushyamig: The rules feature looks fine to me but I couldn't actually test this locally due to load problem error. One concern for me is these rules provide a way to preventing users from accessing other course content. In that case, this PR is not handling it, If not this PR at least another issue must be created stating to handle it from UI, like showing "Permission denied" . I know with rewrite of UI in react it would be a duplicate effort for handling it.

andrew-gardener:
@jonespm @pushyamig I've rebased, made some of the fixes you requested, and simplified a few aspects of the way I implemented it.

quick summary:
- simplified the user cron query (https://github.com/tl-its-umich-edu/my-learning-analytics/commit/8405cd2beedc70e168484683909c9846996b0f99 is a much more elegant of doing it than what I did. I just needed to pass through the enrollment type)
- removed the `CANVAS_PRIMARY_AUTH_PROVIDER_ID` setting (no longer needed)
- added the permission checking to the view templates (needed to create new view functions since the rules library doesn't wrap around templates directly very well)
- there is still no permissions within the view, no 403 status checking in ajax requests, or a 403 template at the moment (I think some of this should be addressed in a separate issue after the react rewrite is in place)

PR #439, opened 2019-03-14
Progress on grades distribution view

justin0022:
* improved styling of text on graph
* set a minimum width of any graph component to be 400 pixels (reduces risk of negative value errors in d3)
* x axis label now appears in the middle of x axis instead of at the end
* added "There are no grades yet for you in this course" message if `current_user_grade` is null

I think this view is more or less complete. The text of the 'My Grade' currently appears on the right side of the line, instead of the left, but I believe that's the only major difference.


justin0022:
Going to make a new PR

Issue #421, opened 2019-03-11
switch over grade distribution dashboard (partial)


pushyamig:
We show 2 different types of view based on the logged in the user.
1. Aggregation view where it is generic to anyone who has access to the visual
2. Student view where we overlay what is the student current grade in the grade distribution.

The view needs to know logged in user role and show respective content in the view.

PR #440, opened 2019-03-14
Updated grades view

justin0022:
Addresses https://github.com/tl-its-umich-edu/my-learning-analytics/issues/421

* improved styling of text on graph
* set a minimum width of any graph component to be 400 pixels (reduces risk of negative value errors in d3)
* x axis label now appears in the middle of x axis instead of at the end
* added "There are no grades yet for you in this course" message if `current_user_grade` is null

I think this view is more or less complete. The text of the 'My Grade' currently appears on the right side of the line, instead of the left, but I believe that's the only major difference. 

Todo:
- [X] Add grade to line hover tooltip or add grade to label itself (awaiting direction from UI/UX, added to label itself for now) 
- [X] Add number of students to tip of histogram bar
- [X] Add `There are no grades yet for you in this course` message if no grades exist
- [X] Handle grades greater than 100%
- [ ] ~Remove tooltip from bars? @pushyamig could you clarify if the tooltip for bars should stay or be removed? If it stays, what info should be displayed?~


pushyamig:
@justin0022 Great work! This is my observation with grade distribution

1. When the user clicks on the  MyGrade line his current grade in the class should appear as he mouse over the line, is there any reason why you did not implement that? here is the screenshot from that
Screen Shot 2019-03-17 at 8 35 29 AM
2. I think there is no need to have the MyGrade info in the short description table to the side and instead as a tooltip on the MyGrade like should work as per the existing implemetation
2. Currently I don't see that grade distribution handle grades if grades have more than 100%. 
Screen Shot 2019-03-17 at 8 44 59 AM
3. I think rather than adding tool tip on hover over about number of student info it should be added to the tip of the bars. The above error occurred in the <Link> component:
    in Link (at DashboardAppBar.js:60)
Screen Shot 2019-03-17 at 8 52 09 AM
4. I feel No need to add Avg.grade in the tool tip on hover as it is provided in the short description table. 
5. "There are no grades yet for you in this course", that msg should popup when required. 
React App observation:
1. I see lot of warning when load the React app
Screen Shot 2019-03-17 at 8 59 09 AM
2. I can't load the `/grades` or` /assignment `curl it throws me an error 
`The context `router` is marked as required in `Link`, but its value is `undefined``.
`browser.js:34 Uncaught Invariant Violation: You should not use <Link> outside a <Router>`
`
In order to fix this error I had to comment Line 60 with `<Link>` componenet in the `DashboardAppBar.js`
```
              {/*<Link to='/' className={classes.homeButton}>My Learning Analytics</Link>*/}
  </Typography>
```

justin0022:
> @justin0022 Great work!
> 
> This is my observation with grade distribution
> 
> 1. When the user clicks on the  MyGrade line his current grade in the class should appear as he mouse over the line, is there any reason why you did not implement that? here is the screenshot from that

I can implement this, but I found the line quite difficult to hover on, and generally line charts have tooltips implemented [like this](https://bl.ocks.org/alandunning/cfb7dcd7951826b9eacd54f0647f48d3), where hovering over anywhere on the visualization shows the tooltip. 

Could we just add the grade % to the label itself, so beside the line it'll say something like `My Grade: 79.4%`? I feel like most users won't think to hover over the line itself.

pushyamig:
> I can implement this, but I found the line quite difficult to hover on, and generally line charts have tooltips implemented like this, where hovering over anywhere on the visualization shows the tooltip.

I agree


Could we just add the grade % to the label itself, so besides the line it'll say something like `My Grade: 79.4%`?
@jennlove-um  what do think?

jennlove-um:
I think it would be okay to show the grade, but I'd like to run it by the UI/UX experts on the MyLA team.  I couldn't find any definitive reason in the initial design discussions that said it had to be a mouse-over rather than always available.  But it would be worth asking them in this week's meeting.

pushyamig:
> React App observation:
> 
> I see lot of warning when load the React app
> 
> I can't load the /grades or/assignmenturl it throws me an error
> The context routeris marked as required inLink, but its value is undefined.
> browser.js:34 Uncaught Invariant Violation: You should not use <Link> outside a <Router>
> In order to fix this error I had to comment Line 60 with <Link> componenet in the DashboardAppBar.js
> <Typography variant='h6' color='inherit' className={classes.grow}>
>               {/*<Link to='/' className={classes.homeButton}>My Learning Analytics</Link>*/}
>   </Typography>

Magically all the errors are vanished after a new build today. I go back and forth between `master` and `react` branches.  Earlier I made sure that i wiped out docker images/containers/volumes/networks before loading any branch. but still I was seeing the problem earlier and not today. Strange!

justin0022:
> Magically all the errors are vanished after a new build today. I go back and forth between `master` and `react` branches. Earlier I made sure that i wiped out docker images/containers/volumes/networks before loading any branch. but still I was seeing the problem earlier and not today. Strange!

I've also noticed some strangeness with random errors that disappear after builds. Sometimes I restart my computer too, and that seems to help. Hard to determine the source of the error - is it my code? react? webpack? docker?

justin0022:
> 6\. `There are no grades yet for you in this course`, that msg should popup when required.

In the table, the `There are no grades yet for you in this course` message appears for the `My Grade` row. If we remove the `My Grade` row from the table, should this message appear elsewhere?

pushyamig:
> In the table, the There are no grades yet for you in this course message appears for the My Grade row. If we remove the My Grade row from the table, should this message appear elsewhere?

Just leave it the way it is. based on your design it makes sense to have MyGrade row in the table to show % grade or mgs in case you don't have it.

pushyamig:
> Remove tooltip from bars? @pushyamig could you clarify if the tooltip for bars should stay or be removed? If it stays, what info should be displayed?
Screen Shot 2019-03-19 at 9 01 06 AM
can the count info be added to the tip of the bar? instead of pop up on hover.

justin0022:
Could either @pushyamig or @Adonais0 pull the latest grades view and check to see if there's any further changes required before we can merge this in?

pushyamig:
@justin0022 I will take a look at this today or tomorrow

justin0022:
Perfect, thanks!

assets/src/containers/GradeDistribution.js
```diff
@@ -48,14 +48,6 @@ function GradeDistribution (props) {
                   <Grid item xs={12} lg={10}>
                     <Histogram
                       data={gradeData.map(x => x.current_grade)}
-                      tip={createToolTip(d => renderToString(
```
pushyamig: @justin0022 I have removed the tool-tip from the bars as it is redundant. Sorry if I caused any confusion on this. All the info that D3 tool-tip is displaying is already present either in the side table or count of students on the bar itself

justin0022: Sounds good, not a problem!

assets/src/components/d3/createHistogram.js
```diff
@@ -34,7 +34,7 @@ function createHistogram ({ data, width, height, el, tip, xAxisLabel, yAxisLabel
 
   bar.append('text')
     .attr('x', d => x((d.x1 + d.x0) / 2))
-    .attr('y', d => y(d.length) + 14)
+    .attr('y', d => y(d.length) + margin.top-5)
```
pushyamig: added the MyGrade text top of the line as to make it stand out for student to see where his grade in the distribution

justin0022:
I noticed a small bug with the `My Grade` label overlapping with line, so I just rounded the `My Grade` to be one decimal place (e.g. 80.88% => 80.9%).

assets/src/components/d3/createHistogram.js
```diff
@@ -92,22 +92,15 @@ function createHistogram ({ data, width, height, el, tip, xAxisLabel, yAxisLabel
       .attr('stroke-width', '2')
     svg.append('text')
       .attr('x', x(myGrade) - 110)
-      .attr('d', '1em')
```
pushyamig: > I noticed a small bug with the My Grade label overlapping with line, so I just rounded the My Grade to be one decimal place (e.g. 80.88% => 80.9%).

Probably that is happening because i removed `d` attribute . I just wanted the Mygrade text to be close line

pushyamig:
I see that Snapshot test is failing for histogram. I also see that the test doesn't consider the MyGrade line into consideration probably because test data for in the `d3TestData.js` don't have it. 
So test is only written to the histogram and not to the side table grid gives a summary of the grade distribution. I am not telling it must be done or not. Trying to understand the efficient way of testing the UI. 
Screen Shot 2019-04-01 at 9 12 59 AM

justin0022:
Okay, I think the snapshot has to be updated since the histogram was changed. I'll look into it tomorrow (I'm off today).

assets/src/__tests__/d3/createHistogram.test.js
```diff
@@ -1,11 +1,21 @@
...
+  test('should build a histogram bar chart with MyGrade line', () => {
+    const div = document.createElement('div')
+    createHistogram({ data: histogramData, width: 1000, height: 500,  el: div, myGrade:79.8})
+    expect(div).toMatchSnapshot()
+  })
+  test('should build a histogram bar chart with MyGrade line and distribution more than 100', () => {
+    const div = document.createElement('div')
+    createHistogram({ data: histogramDataMoreThan100Percent, width: 1000, height: 500,  el: div, myGrade:56.8, maxGrade: 110})
+    expect(div).toMatchSnapshot()
+  })
```
pushyamig: Just added a few tests with various cases for grades distribution. please check it. 
A question do we need to add tests so that both the side grid table and distribution is loaded as part of the view. Seems to be a reasonable test. But the table is not a d3 element and the `Table` component is at Container level. so how this will be handled?

pushyamig:
one thing I noticed with Jest test is while running the command 'docker exec -it webpack_watcher npm test' this will run all tests and inherently it will do `jest --watchall` . But when you change a test I would suppose it should rerun the test again but I end up seeing errors continually and I have to stop the docker run and start it back again to run the test.

assets/src/__tests__/d3/createHistogram.test.js
justin0022: Yes, I think for the React tests we'll be using [Enzyme](https://airbnb.io/enzyme/) and Jest. Soo and I are still exploring the best way to test containers. She refactored the API calls made from container components into a separate folder as part of this PR: https://github.com/tl-its-umich-edu/my-learning-analytics/pull/461

We'll be using Jest to mock API calls and then asserting that the React components have the correct output.

justin0022:
What errors are you seeing? It watches for my changes and reruns properly for me. 

All tests are passing on my end. Should we merge this PR into `React` branch now?

pushyamig:
Test are passing! I noticed that when writing test and if you change something while `--watchall` option is going on then it generates errors. I will open a different issue as that is not related to the grades view test. I think there is no info in the readme added steps for testing snapshot test, so you might see no errors but some others do

This PR can be merged!

PR #446, opened 2019-03-15
Fix error when data is an empty object

Adonais0:
The `GradeDistribution` page and `AssignmentPlanning` page shows error when `gradeData` or `assignmentData` is an empty object.

### DIFF
--- a/assets/src/containers/AssignmentPlanning.js
+++ b/assets/src/containers/AssignmentPlanning.js
@@ -66,13 +66,11 @@ function AssignmentPlanning (props) {
     />
   )
 
-  const tableBuilder = (loaded, assignmentData, table) => {
+  const tableBuilder = (assignmentData) => {
     if (!assignmentData || Object.keys(assignmentData).length === 0) {
       return (<p>No data provided</p>)
-    } else if (loaded) {
-      return AssignmentTable(assignmentData.plan)
-    } 
-    return <Spinner />
+    }
+    return AssignmentTable(assignmentData)
   }
 
   return (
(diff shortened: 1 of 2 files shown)


assets/src/containers/AssignmentPlanning.js
```diff
@@ -66,6 +66,15 @@ function AssignmentPlanning (props) {
     />
   )
 
+  const tableBuilder = (loaded, assignmentData, table) => {
```
justin0022: `table` param isn't used - could you please remove it?

assets/src/containers/AssignmentPlanning.js
```diff
@@ -88,7 +97,7 @@ function AssignmentPlanning (props) {
                   <MenuItem value={75}>75%</MenuItem>
                 </Select>
               </FormControl>
-              {loaded ? AssignmentTable(assignmentData.plan) : <Spinner />}
+              { tableBuilder(loaded, assignmentData) }
```
justin0022: Just curious, why provided `tableBuilder` with the `loaded` variable? 

If we make it `{loaded ? tableBuilder(assignmentData.plan) : <Spinner />}`? That way, `tableBuilder` is only called when the data is loaded. As written, `tableBuilder` is called while data is loading. Not a huge deal though, just a thought!

Adonais0: Sure, done.

assets/src/containers/AssignmentPlanning.js
Adonais0: Yep, removed

assets/src/containers/AssignmentPlanning.js
justin0022: Perfect, looks good to me!

Issue #422, opened 2019-03-11
switch over assignment planning dashboard


pushyamig:
This has 2 views to it and both need to be served up from the same page 

1. Assignment planning tabular view 
2. Assignment progress view

We show 2 different types of view based on the logged in the user.

Aggregation view where it is generic to anyone who has access to the visual
Student view where we overlay what is the student current grade in the grade distribution.
The view needs to know logged in user role and show respective content in the view.

pushyamig:
@justin0022 I am also assigning this to me in case if you want to split the work as this is a big chunk

pushyamig:
In assignment view and files view both has a 1. dropdown box 2. checkbox to select the default option. This could be separated as a generic component that populates the dropdown and based on the view and would have an action event that does the PUT call for storing the user selected default.

justin0022:
Merged: https://github.com/tl-its-umich-edu/my-learning-analytics/pull/447

PR #447, opened 2019-03-15
Assignment view (in progress, help appreciated)

justin0022:
https://github.com/tl-its-umich-edu/my-learning-analytics/issues/422

* Progress bar is implemented (no text inside the rectangles yet)
* Tooltip implemented (`drop_lowest` and `drop_highest` isn't yet added. Tooltip is not bounded to viewport, so sometimes part of it is not inside browser window - would appreciate some help in thinking of best way to implement this)
* Might be missing some empty data/null checks

It's getting close. Please feel free to contribute to this PR if you're able to add any fixes. 

Todo:
- [X] Add `x%` beside horizontal bar inside table
- [X] Fix formatting of table to be nested table for assignments in same week
- [X] adding a default bar of small width for 0%  visually will be informational to the user
- [X] Add "Assignment Status" legend
- [X] Add text of assignment name to progress bar
- [X] add percentage info to the bars for a tabular view
- [x] Progress bar current/max possible line logic fixing
- [x] drop_lowest and drop_highest logic
- [x] Add hover to horizontal bar
- [x] Make the nested table assignment view scrollable vertically and don't take up rest of the space
- [x] making the `Percent of final grade` column bars and text closer, add a vertical line between `Title` and 'Percent of final grade' and have the bar touch the line
- [x] scroll to the particular row in the assignment table view based on the current week in the term. current week info is available in the Json feed
- [x] Make progress bar tooltip take up less real estate
- [x] Add "Remember my settings" feature


pushyamig:
@justin0022  I will contribute. stay tuned!

pushyamig:
> Tooltip is not bounded to viewport, so sometimes part of it is not inside browser window - would appreciate some help in thinking of best way to implement this)
1. I feel tool tip need to know the position the user hovered(like end/start of the viz). Based on the current  implementation the tool tip is a react component and some position info need to be passed on from the D3 (`d3/createProgressBar.js`) to react componenet(`ProgressBar`) to position the tool tip. I don't know if that is practical to do that.
2. I feel the the progress bar Tool tip is a table component and it is taking alot of real estate as it is falling into Assignment table view. The Table needs to be compact for viewable reason. Can it be something different than Table Componenet?
3. Another way is implement the tool-tip in the D3 itself(not as Table component) and when when we know it is end or start of progress bar we can change the position to be top/bottom/right/left (what ever make sense. Current implementation Progress handles this case in the D3 itself. 
```
if ((width - coordinates[0]) < 50) {
                tooltip.html(tooltipVal)
                    .style("left", (d3.event.pageX) - (d3.event.pageY * 2) + "px")
                    .style("top", (d3.event.pageY) - 150 + "px");
            } else {
                tooltip.html(tooltipVal)
                    .style("left", (d3.event.pageX) + "px")
                    .style("top", (d3.event.pageY) + "px");
            }
```

pushyamig:
Assignment Table View: Still work in progress but current implementation doesn't seems to fit the table structure of current implementation. It uses nested tables to structure based the our design requirment. It needs to be scrollable currently it takes up rest of the page. The size needs to be decreased and based on the how far along course is it should auto scroll to that week. 

I can work on this if you are not too far into this.

justin0022:
> Assignment Table View: Still work in progress but current implementation doesn't seems to fit the table structure of current implementation. It uses nested tables to structure based the our design requirment. It needs to be scrollable currently it takes up rest of the page. The size needs to be decreased and based on the how far along course is it should auto scroll to that week.
> 
> I can work on this if you are not too far into this.

Sure please feel free to work on the table. I'll be working on addressing https://github.com/tl-its-umich-edu/my-learning-analytics/pull/440#issuecomment-473663848 today.

justin0022:
> > Tooltip is not bounded to viewport, so sometimes part of it is not inside browser window - would appreciate some help in thinking of best way to implement this)
> 
> 1. I feel tool tip need to know the position the user hovered(like end/start of the viz). Based on the current  implementation the tool tip is a react component and some position info need to be passed on from the D3 (`d3/createProgressBar.js`) to react componenet(`ProgressBar.js`) to position the tool tip. is it doable to pass info like that? Thoughts???

It is possible, via [lifting state up](https://reactjs.org/docs/lifting-state-up.html). That was one of the ways I was thinking of implementing this. 

> 2. I feel the the progress bar Tool tip is a table component and it is taking alot of real estate as it is falling into Assignment table view. The Table needs to be compact for viewable reason. Can we make it be something different than Table Componenet?

Yup, we can change it to be something else. Could just be paragraph tags, for example.

> 3. Another way is implement handle `tool-tip` in the D3 itself(not as Table component) and when we know user hovered is end or start of progress bar we can change the position to be top/bottom/right/left (what ever make sense. Current implementation Progress handles this case in the D3 itself.

This is another option for sure. I think getting the styling to align with Material-UI will be a bit more challenging, but will simplify the calculation of bounding box.

> 
> ```
> if ((width - coordinates[0]) < 50) {
>                 tooltip.html(tooltipVal)
>                     .style("left", (d3.event.pageX) - (d3.event.pageY * 2) + "px")
>                     .style("top", (d3.event.pageY) - 150 + "px");
>             } else {
>                 tooltip.html(tooltipVal)
>                     .style("left", (d3.event.pageX) + "px")
>                     .style("top", (d3.event.pageY) + "px");
>             }
> ```
> 1. Another requirment often asked up can we track the hover over info of the user. So when a user hover pass that info to backend and log it for leaning analytics purpose. Current implementation doesn't do this but often talked upon in our requirment session for MyLA. Just sharing the info

Good to know, this will definitely be done by lifting state up to the container level from the d3 component and making the appropriate `PUT` or `POST` request.

pushyamig:
I am getting familiar with the design flow with React in the App.
With the original implementation of the assignment table, it uses nested tables I have used custom styling to make some of the table lines disappear. Also, I see that you have used D3 for bars implementation
Next step for me to see with the existing design and JSON data if the original idea is still valid. I.e having nested tables with custom styling along with D3 components.

assets/src/components/d3/createProgressBar.js
```diff
@@ -104,6 +110,8 @@ function createProgressBar ({ data, width, height, el, tip }) {
 
   if (tip) {
     tip.direction('s')
+      .offset([0, 60])
+      .style('max-width', '300px')
```
justin0022: I added an offset and a max-width of 300px to the tooltip to prevent the overflow issue for now

assets/src/components/d3/createProgressBar.js
```diff
@@ -47,8 +47,14 @@ function createProgressBar ({ data, width, height, el, tip }) {
     .style('outline-style', 'solid')
     .style('outline-width', '1px')
 
+  // bar.append('text')
```
justin0022: Commented out the text because it was messing up the alignment of the tooltip, as the `<text>` is not bounded to the `<rect>`. Is it worth keeping the text here? In the previous implementation, I believe the first 3 or 4 characters of the text were sliced and displayed. I'm wondering if there's a more UI friendly way to implement this.

One idea that was floated around in our office was highlighting the progress bar rectangle when user hovers over corresponding table row, and vice versa. Thoughts?

pushyamig: > Commented out the text because it was messing up the alignment of the tooltip, as the <text> is not bounded to the <rect>. Is it worth keeping the text here? In the previous implementation, I believe the first 3 or 4 characters of the text were sliced and displayed.
It was one of the requirements that was asked, we could we CSS overflow:ellipsis. So we still need that. But we don't want to add test with bar % is too small.

pushyamig: > One idea that was floated around in our office was highlighting the progress bar rectangle when user hovers over corresponding table row, and vice versa. Thoughts?

It is interesting idea, but for assignments with 0% will end up in table view but not in the progress view. So we might need to consider that case when implementing.

justin0022: > It was one of the requirements that was asked, we could we CSS overflow:ellipsis. So we still need that. But we don't want to add text when bar % is too small. I think having the text on progress bar in there will let users hover over it for more info.

Okay I've implemented this here: https://github.com/tl-its-umich-edu/my-learning-analytics/pull/447/commits/36d7b2a24553000097c819992240a94dbae3a958

assets/src/components/TableAssignment.js
```diff
@@ -0,0 +1,112 @@
...
+              {tableHead.map((prop, key) => {
+                return (<TableCell className={classes.tableCell + ' ' + classes.tableHeadCell}
+                                   key={key}>{prop}</TableCell>)
+              })}
+            </TableRow>
+          </TableHead>
+        ) : null}</Table>
+      <Table><TableBody>
+        {
+          tableData.map((week, key) => {
```
pushyamig: Here is the logic that goes with nested tables. It still have alignment issues to be fixed

assets/src/components/TableAssignment.js
```diff
@@ -0,0 +1,112 @@
...
+            return (
+              <TableRow key={key}>
+                <TableCell className={classes.tableCell}
+                           key={key + 'week'}>Week {week.week}</TableCell>
+                <TableCell className={classes.tableCell} key={key + 'weekRest'}>
+                  <Table><TableBody>{
```
pushyamig: The nested `<Table>` element is leading to the dis-alignment. But if you don't have the `<Table>` element and instead just `<TableBody>` alignment is preserved between the 2nd column and 3rd column but I see warning regarding. 

> Warning: validateDOMNesting(...): <tbody> cannot appear as a child of <td>.
>     in tbody (created by TableBody)
>     in TableBody (created by WithStyles(TableBody))
>     in WithStyles(TableBody) (at TableAssignment.js:62)
>     in td (created by Context.Consumer)
>     in TableCell (created by WithStyles(TableCell))
>     in WithStyles(TableCell) (at TableAssignment.js:61)
>     in tr (created by Context.Consumer)
>     in TableRow (created by WithStyles(TableRow))
>     in WithStyles(TableRow) (at TableAssignment.js:58)
>     in tbody (created by TableBody)
>     in TableBody (created by WithStyles(TableBody))
>     in WithStyles(TableBody) (at TableAssignment.js:54)
>     in table (created by Table)

But the [issue](https://github.com/facebook/react/issues/5652) here suggest to wrap `<TableBody>` inside `<Table>` that make sense.

assets/src/containers/AssignmentPlanning.js
```diff
@@ -66,19 +66,9 @@ function AssignmentPlanning(props) {
   }
 
   const AssignmentTable = plan => (
-    <Table
+    <TableAssignment
       tableHead={['Week', 'Due', 'Title', 'Percent of final grade']}
-      tableData={generateAssignmentTable(plan)
-        .map(row => {
-          const { percentOfFinalGrade, graded } = row.pop()
-          row.push(<HorizontalBar
```
pushyamig: Took the `HorizontalBar` component from the container and added to the `TableAssignment` component. Felt passing the Data is simple(line in 71 here) without nesting the component inside the data. Thoughts??

justin0022:
@pushyamig could you pull my latest change? I fixed the styling issues by removing `border-bottom` and `props` of certain cells if the previous or next row had the same week. It works on my test course, please check with yours!

assets/src/components/TableAssignment.js
```diff
@@ -89,13 +89,21 @@ function CustomAssignmentTable (props) {
                   let displayProp = true
                   let displayBorder = true
                   if (data[i - 1]) {
-                    if (data[i][0] === data[i - 1][0] && j < 2) {
+                    // first item in the row the logic handles display of week property
+                    if (j == 0 && data[i][0] === data[i - 1][0]) {
+                      displayProp = false
+                    }
+                    // second item in the row the logic handles display of Due date property
+                    if(j == 1 && data[i][1] === data[i-1][1]){
```
pushyamig: @pushyamig could you pull my latest change? I fixed the styling issues by removing border-bottom and props of certain cells if the previous or next row had the same week. It works on my test course, please check with yours!
I changed few things in logic to take into account due date 

 some week may have more than one due date Adding logic to consider due date as well

assets/src/components/d3/createProgressBar.js
```diff
@@ -72,46 +72,60 @@ function createProgressBar ({ data, width, height, el, tip }) {
...
+  const showMaxLine = (data) => {
+    // filtering the graded and ungraded values, list would be either 2 or 1
+    const gradedList = [...new Set(data.map(item => item.graded))]
+    if(gradedList.length >1){
+      return false
+    }
+    return gradedList[0]
+  }
+
+  if(!showMaxLine(data)) {
```
justin0022: Could you explain what `showMaxLine` does?

I'm a bit confused because it looks like `maxLine` is appended only if `showMaxLine(data)` _isn't_ true. Shouldn't it be appended if `showMaxLine(data)` _is_ true?

pushyamig: MaxLine should be shown
1. When you have a mix of graded and ungraded assignment, this is usually in the mids of the term
2. When No Graded assignments at all, this is the case for admin/teacher whose assignment won't be graded

Similarly, CurrentLine should be shown
1. when data has a mix of graded and ungraded assignment
2. When all assignments are graded.

pushyamig: OK, I changed the logic to make it less confusing

justin0022: Perfect, makes sense!

assets/src/components/TableAssignment.js
```diff
@@ -85,40 +105,54 @@ function CustomAssignmentTable (props) {
...
             return (
               <TableRow key={i}>
-                {row.map((prop, j) => {
-                  let displayProp = true
-                  let displayBorder = true
-                  if (data[i - 1]) {
-                    // first item in the row the logic handles display of week property
-                    if (j == 0 && data[i][0] === data[i - 1][0]) {
-                      displayProp = false
+                {
```
pushyamig: This commit is 
1. reducing the size of the table, making it scrollable
2. current week indicator with the color change

pushyamig:
I was also trying to make the table scroll to current week (in the term) when student loads the assignment view, I looked it up online and a lot of suggestion was around using `React.createRef` but they tend to be for a class-based component. But our components are stateless, one idea is to use [React hooks useRef](https://reactjs.org/docs/hooks-reference.html#useref). I tried little bit. Do you have any suggestion on this?

assets/src/components/TableAssignment.js
justin0022: I like the current week indicator!

A question about making the table smaller - why set a fixed height for the table at `400px`? Especially on mobile devices, nearly half of the bottom of the screen is blank, and users are forced to scroll.

pushyamig: The idea with assignment table design is when the user loads the page the table will be auto-scrolled to the `current_week` and that way student is guided what needs to be done for that week in the course. So that's why we had to make the table size smaller. So if they want to see a previous assignment they can scroll and look around.

The Week number is aligned with the start of the course term. So this makes them easier thinking for assignment/quiz preparation how far along they are in a term.

I know that leave the rest of the screen is blank for mobile. It has a better visual from the landscape mode.  

I was playing around with various devices from the Chrome Dev tool. I noticed for tablets the screen is blank from the bottom and left side as well.

is there a better way we can design this?

justin0022: > I noticed for tablets the screen is blank from the bottom and left side as well.

If you refresh after you open Chrome dev tools in the tablet/phone emulator, this blank spacing should go away.

I'm not sure what the best way to design this is. Perhaps having the table take up more space and highlighting the current week is sufficient.

pushyamig: auto-scroll feature to the current week is a feature that asked for, I don't know if we let the table take the rest of the space. I will work auto-scroll tomorrow and see the feasibility of it. I see scroll to row is supported as part of [react-virtualization](https://material-ui.com/demos/tables/#virtualized-table) table. It would be a change to the existing setup which is using simple material-UI table.

justin0022: Okay sounds good!

assets/src/containers/AssignmentPlanning.js
```diff
@@ -37,22 +37,35 @@ const styles = theme => ({
   }
 })
 
+const getCurrentWeek = assignmentData => {
+  // default week
+  let currentWeek = 0
```
justin0022: @pushyamig if you want to test, change `currentWeek` here to be, say , `5`, and you'll see the table will scroll to the 6th element (since index starts at 0).

assets/src/components/TableAssignment.js
```diff
@@ -32,7 +32,7 @@ const tableStyle = theme => ({
     verticalAlign: 'middle'
   },
   tableResponsive: {
-    // width: '70%',
+    width: '70%',
```
pushyamig: Changing the width not take up the same, for now. Will decide to condense or expand one formatting and hover on the bars implemented

assets/src/components/TableAssignment.js
```diff
@@ -125,14 +125,22 @@ function CustomAssignmentTable (props) {
...
+      const tableHeaderOffset = 50
       tableRef.current.parentNode.scrollTo({
-        top: currentWeekRow.current.offsetTop - tableHeaderOffset,
+        top: currentWeekRow.current.offsetTop-tableHeaderOffset,
         behavior: 'smooth'
       })
     }
   })
 
+  const isItFirstCurrentWeekIndicator= i => {
```
pushyamig: This logic only adds the `rootRef` for the first row when the week property is actually visible. With this, I see that the scroll is rolling to a currect position

assets/src/containers/AssignmentPlanning.js
```diff
@@ -39,7 +39,7 @@ const styles = theme => ({
 
 const getCurrentWeek = assignmentData => {
   // default week
-  let currentWeek = 0
+  let currentWeek = null
```
pushyamig: if 0 is returned there is a case that if assignments don't have a due date and no current week the table will have `week 0` row. In that case `week 0` is color coded, which we don't want as it doesn't make sense.

justin0022: Got it, thanks, this makes more sense.

assets/src/components/TableAssignment.js
justin0022: Sounds good

assets/src/components/TableAssignment.js
```diff
@@ -140,10 +151,33 @@ function CustomAssignmentTable (props) {
...
             borderAndCurrentWeekStyle['color'] = 'orange'
           }
+          // styling for assignment title column
+          if (j === 2) {
+            borderAndCurrentWeekStyle['width'] = '30%'
+            borderAndCurrentWeekStyle['maxWidth'] = 0
+            borderAndCurrentWeekStyle['overflow'] = 'hidden'
+            borderAndCurrentWeekStyle['textOverflow'] = 'ellipsis'
+            borderAndCurrentWeekStyle['whiteSpace'] = 'nowrap'
+            borderAndCurrentWeekStyle['borderRight'] = '0.1em solid rgba(224, 224, 224, 1)'
```
pushyamig: Assignment titles can get long and we don't want table column to expand as that will extend the Percent toward grade column to further away from the view. So using ellipsis for truncating the tile.

assets/src/components/TableAssignment.js
```diff
@@ -140,10 +151,33 @@ function CustomAssignmentTable (props) {
...
+          }
```
pushyamig: Also adding the border between title and percent columns for visual identification of title association with percent column

assets/src/components/TableAssignment.js
```diff
@@ -140,10 +151,33 @@ function CustomAssignmentTable (props) {
...
+          if (j === 3) {
+            borderAndCurrentWeekStyle['paddingLeft'] = 0
+          }
           return (
-            <TableCell className={classes.tableCell} key={j} style={borderAndCurrentWeekStyle}>
-              {displayProp ? prop : null}
-            </TableCell>
+            j === 2 ?
+              <TableCell title={displayProp ? prop : null} className={classes.tableCell} key={j}
+                         style={borderAndCurrentWeekStyle}>
```
pushyamig: since long assignment get truncated user is given the option on hover to review the full assignment time. Well, this feature in mobile is not able to see the full title as the title seems to work only with hover and not with on click on mobile. Another option is adding the CSS tooltip, thoughts???

assets/src/containers/AssignmentPlanning.js
```diff
@@ -20,83 +22,103 @@ const styles = theme => ({
...
+  },
+  ungraded: {
+    width: '10px',
+    height: '10px',
+    background: 'gray',
+    display: 'inline-block'
   }
 })
 
+const getCurrentWeek = assignmentData => {
```
Adonais0: I thought the `currentWeek` data should be fetched from `/info`?

pushyamig: Assignments view is different the current week info is baked in the assignment data itself when calculating the which assignment belongs to which week.
 I don't know why files view made a decision to get it from a separate backend call. good question!

justin0022:
Just had a chance to pull the latest, and it's looking good!

pushyamig:
That's pretty much a wrap in terms of the features assignment views

justin0022:
Awesome!

pushyamig:
@justin0022 I am merging the Assignment view changes to react branch.

PR #451, opened 2019-03-16
#450 handling weighted assignments groups with no assignments yet

### LINKED ISSUE #450: empty assignment groups with weights resulting in incomplete progress bar
When assignments groups are created with weights with no assignments the percent toward final grade calculation is screwed and the progress bar is not 100 %. The Assignment views design need to include this case as well.
Screen Shot 2019-03-16 at 10 28 12 AM

The proposed solution to handle this case is to fill in that gap with Place holder assignments with respective assignment groups name prepended.

pushyamig:
fixes #450

### DIFF
--- a/dashboard/views.py
+++ b/dashboard/views.py
@@ -346,7 +346,7 @@ def assignments(request, course_id=0):
         return HttpResponse(json.dumps([]), content_type='application/json')
 
     df.sort_values(by='due_date', inplace=True)
-    df.drop(columns=['assignment_id', 'due_date'], inplace=True)
+    df.drop(columns=['assignment_id', 'due_date','grp_id'], inplace=True)
     df.drop_duplicates(keep='first', inplace=True)
 
     # instructor might not ever see the avg score as he don't have grade in assignment. we don't have role described in the flow to open the gates for him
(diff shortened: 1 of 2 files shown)


dashboard/views.py
```diff
@@ -499,6 +518,36 @@ def get_term_dates_for_course(course_id):
...
 
+def are_weighted_assignments_hidden(course_id):
+    """
+    if assignments are weighted then assignment groups weight totals =100% . The code is checking if assignment groups
+    has group points corresponding to group weight and if not assignments are hidden
+    :param course_id:
+    :return:
+    """
+    logger.info(are_weighted_assignments_hidden.__name__)
+    sql = "select weight, group_points from assignment_groups where course_id=%(course_id)s"
```
jonespm: As I'd commented (in person) this would be better if this took a dataframe as a parameter rather than running the sql on the course id. This would make it easier to unit test in the future. Some of these methods may want to be refactored to make them easier to test and reduce the complexity. 

Doesn't the assignments_in_course already have this data? Does this need to be double queried?

jonespm: Yeah, it looks like if you just passed in assignments_in_course DataFrame it should have this data already in there? Saving an SQL call here and making it easier to test in the future. 

If you're worried about manipulating it you could copy it.

pushyamig: Ok, I will give it a try. I think the problem is `assignments_in_course` has a mix for assignments and groups and weight together and I well I can extract just assignment group info from Data frame. I have to give it a try.

jonespm: Ok, if it's not easy to split them from this DataFrame, I'd prefer that the method get_course_assignments run the 2 SQL calls and pass the DataFrame in, rather than this utility method run the SQL.

pushyamig: OK, I removed the extra SQL query instead just sliced the weight and group_points info from the df

PR #456, opened 2019-03-18
Issue 455 Put up a warning if the term is incorrect (partial fix)

### LINKED ISSUE #455: If course is on wrong term, there is a ValueError in assignment view
If you have the course on the wrong term (for instance if you have a course that was given Winter 2019 in the Fall 2019 term) MyLA throws this exception.

```
  File "/code/dashboard/views.py", line 430, in <lambda>
    assignments_in_course['week']=assignments_in_course['calender_week'].apply(lambda x: 0 if x == 0 else week_list.index(x)+1)
ValueError: 5 is not in list
```

It looks like the code here is getting the minimum week from the term dates associated with the course and returning the "week". So for Fall it returns something like 36
`min_week=find_min_week(course_id)`

Then this code runs getting the max value of calendar_week from the DataFrame. which in this case is 13.
`max_week=assignments_in_course['calender_week'].max()`

There is an error because min_week (36) is greater than max_week (13)

I feel like either these should be consistent, either both using the DataFrame or both using the term. I'm not sure why it is mixed up here.

On a course where the term is correct this min_week "seems" to be correct. I'm not sure which way works best yet.

jonespm:
Fixes #455

### DIFF
--- a/dashboard/views.py
+++ b/dashboard/views.py
@@ -419,14 +419,16 @@ def get_course_assignments(course_id):
     consider_weight=is_weight_considered(course_id)
     total_points=assignments_in_course['points_possible'].sum()
     assignments_in_course['towards_final_grade']=assignments_in_course.apply(lambda x: percent_calculation(consider_weight, total_points,x), axis=1)
-    assignments_in_course['calender_week']=assignments_in_course['due_date'].dt.week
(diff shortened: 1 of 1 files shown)


jonespm:
Also, lost a lot of debugging time because calendar was spelled wrong. ;)

PR #460, opened 2019-03-19
When adding a course in admin non numeric id results in 500 error

### LINKED ISSUE #459: When adding a course in the course admin adding a non-numeric id results in 500 error
If you're in the admin interface and add a course using a non-numeric id (like using text, "test") it says

**Server Error (500)**

This error should be handled better rather than throwing this error

jonespm:
Fixes #459

### DIFF
--- a/dashboard/admin.py
+++ b/dashboard/admin.py
@@ -1,4 +1,5 @@
 from django.contrib import admin
+from django import forms
 from django.conf import settings
 from django.utils.safestring import mark_safe
 from django.template.defaultfilters import linebreaksbr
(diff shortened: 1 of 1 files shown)


dashboard/admin.py
```diff
@@ -1,26 +1,35 @@
...
             exclude += (view,)
 
+class CourseForm(forms.ModelForm):
+    class Meta:
+        model = Course
+        exclude = ()
+
+    def clean(self):
+        canvas_id = self.cleaned_data.get('canvas_id')
+        if not canvas_id.isdigit():
```
andrew-gardener: `if not canvas_id or not canvas_id.isdigit():` (leaving blank causes error)

dashboard/admin.py
```diff
@@ -1,26 +1,35 @@
...
+            raise forms.ValidationError("Canvas ID must be an integer value")
```
pushyamig: I think we should log the course_id so the message should say 
```suggestion
            raise forms.ValidationError(f"Course ID {canvas_id} must be an integer value")
```

dashboard/admin.py
jonespm: Thanks, updated!

dashboard/admin.py
jonespm: Thanks! Updated.

PR #461, opened 2019-03-21
Abstract api calls into api folder to allow for testing with mock calls

soohyunkim:
In order to use mock objects for testing purposes, I had to abstract current API calls into separate functions. This will lead to a separate PR, which will use mock API calls (that will be stored with fake data in a `__mock__` folder) to test for react containers.

### DIFF
--- a/assets/src/service/api.js
+++ b/assets/src/service/api.js
@@ -1,8 +1,5 @@
-/* global fetch */
+import useFetch from '../hooks/useFetch'
 
-const get = url => {
-  return fetch(url)
-    .then(res => res.json())
-}
-
-export default get
+export const useGradeData = useFetch(`http://localhost:5001/api/v1/courses/${currentCourseId}/grade_distribution`)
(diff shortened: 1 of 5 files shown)


assets/src/service/api.js
```diff
@@ -1,8 +1,5 @@
-/* global fetch */
+import useFetch from '../hooks/useFetch'
 
-const get = url => {
-  return fetch(url)
-    .then(res => res.json())
-}
-
-export default get
+export const useGradeData = useFetch(`http://localhost:5001/api/v1/courses/${currentCourseId}/grade_distribution`)
```
Adonais0: The `currentCourseId` is not passed in here, the application shows `currentCourseId is not defined` error

soohyunkim: thanks for pointing that out, i don't have access to any of the data so I am unable to test the app before I make a PR :/

Adonais0:
Cool, works fine for me now.

PR #467, opened 2019-03-28
Merging master changes into react

andrew-gardener:
I had some issues when checking out react branch due to environment variables changes (mostly UDW -> DATA_WAREHOUSE and canvas id prefixing to increment changes). This is just to pull master changes in so there are less issues switching dev branches


pushyamig:
This looks good, most of the changes are in the backend and cron. I was to do the cron run successfully and was able to load the React UI Components. All the commit since the reach branched out seem to be added too. Looks Good!

PR #471, opened 2019-04-01
Fixes #463 Support hash tags in tracking info

### LINKED ISSUE #463: Track initial launch as a separate event
It was suggested that we should track the initial user launch (from Canvas) as a separate event.

I believe the easiest way would be to use an anchor on the URL like /#launch 

However GA doesn't track that by default, it only tracks the URL's and query parameters (which might also work out) though [this describes](https://stackoverflow.com/a/48267830/3708872) how we'd change the page_path to add the hash. Then we'd just have to update the Canvas javascript to also have these for the various links.

jonespm:


dashboard/templates/base.html
```diff
@@ -62,7 +62,9 @@
     window.dataLayer = window.dataLayer || [];
     function gtag(){dataLayer.push(arguments);}
     gtag('js', new Date());
-    gtag('config', '{{ settings.GA_ID }}');
+    gtag('config', '{{ settings.GA_ID }}', {
+        'page_path': location.pathname + location.search + location.hash
```
zqian: Need to have path dividers between location.pathname and location.search? Also between location.search and location.hash?

How do we track the invoke from the left-side tool bar entry? By LTI invoke history?

jonespm: location.search is the query part of the string. I put a screenshot on #463 but we don't have any  location.search there just the location.pathname and location.hash.

This pattern was the one I found with [over 100 upvotes](https://stackoverflow.com/a/4813223/3708872). 

The links from the left navigation come from a referrer of "edu-apps.org". This seems like what all LTI apps come from. I don't know of any way to get more information about this unless we launched our tool directly via LTI rather than relying on the [LTI redirect tool](https://www.eduappcenter.com/apps/63?filter_text=#.XKJY0-tKiL4). We can probably do this with the 2019.02 major release after that feature is included.

PR #473, opened 2019-04-01
Fixes #472 - Return course settings (view enabled) from course info feed

### LINKED ISSUE #472: Return course settings (views enabled) from course info feed
The course info feed doesn't provide information about the views that are enabled disabled, this could be added to this feed for this specific course similar to the courses_enabled feed for all courses.

jonespm:


jonespm:
This is what the feed looks like after this. I did a little refactoring and pep8 cleanup. 

`{"canvas_id": "421436", "term_id": 17700000000000111, "name": "GEOG 964 007 FA 2063", "term": {"id": 17700000000000111, "canvas_id": "111", "name": "Fall 2018", "date_start": "2018-09-04 00:00:00", "date_end": "2018-12-31 00:00:00"}, "current_week_number": 30, "total_weeks": 17, "course_view_options": {"fa": 1, "ap": 1, "gd": 1}}`

jonespm:
@pushyamig said this feed may change so it takes a canvas id instead of an incremented ID and do the conversion in the views.py. If we do this we *probably* should have a v2 URL but I think we can just update it since we're the only consumers.

pushyamig:
`/info` = > should give all the info that is related to the course/terms. This is b'coz since the templates are not controlled by Django (with React re-write) which view should be enabled/disabled info should be called with every view launch. After user login with `http://myla.dashboard.edu` he could directly hit `http://myla.dashboard.edu/12233`  or `http://myla.dashboard.edu/12233/grades` so in that case `/info` URL is the first thing that needs to be called checking to see if view is viewable in the course or not.  

 I feel `resp['course_view_options'] = CourseViewOption.objects.get(course=course).json(include_id=False)` `include_id` is redundant for future as we always was course view option info

If this change merged to master it will be pulled into React branch for further development

jonespm:
@pushyamig I have include_id there because sometimes we don't want the course_id included (that part is redundant. This call on the model is used by two views. On the public courses_enabled view (that is available via jsonp from Canvas) it has the course_id's, but if you call this function on the model and don't need the course_id returned and just want the view options we'd pass false to it. 

Passing False returns
`{"fa": 1, "ap": 1, "gd": 1}
`
Passing True (or nothing) returns
`'421436 ': {"fa": 1, "ap": 1, "gd": 1}`

dashboard/models.py
```diff
@@ -197,22 +199,26 @@ class Meta:
...
+                       'ap': int(self.show_assignment_planning and 'show_assignment_planning'
+                                 not in settings.VIEWS_DISABLED),
+                       'gd': int(self.show_grade_distribution and 'show_grade_distribution'
+                                 not in settings.VIEWS_DISABLED),}
+            if include_id:
+                return {self.course.canvas_id: options}
+            else:
+                return options
         except ObjectDoesNotExist:
             logger.warn(f"CourseViewOption does not exist in Course table, skipping")
```
pushyamig: `warn` seems to be deprecated. can you change it to `warning`?

jonespm:
```suggestion
            logger.warning(f"CourseViewOption does not exist in Course table, skipping")
```

jonespm:
```suggestion
            logger.warning(f"CourseViewOption does not exist in Course table, skipping")
```

pushyamig:
@justin0022 this endpoint should suffice for getting the course_info that is needed

justin0022:
Thanks I'll try it out

dashboard/models.py
```diff
@@ -197,24 +199,28 @@ class Meta:
...
     
-    def json(self):
+    def json(self, include_id=True):
         """Format the json output that we want for this record
         
+        :param include_id: Whether or not to include the id in the return 
         This should be of the format canvas_id : {options}
         :return: JSON formatted CourseViewOption
         :rtype: Dict
         """
```
justin0022: One suggestion I have is to add this line in here: `course_id = canvas_id_to_incremented_id(course_id)`.

For the React rewrite, we send requests using the short course_id.

PR #474, opened 2019-04-02
Fix index display bug

### LINKED ISSUE #465: Adding new items to index page causes formatting issues
Screen Shot 2019-03-27 at 3 47 31 PM

I think it may be because here:
https://github.com/tl-its-umich-edu/my-learning-analytics/blob/517e1baae1f569ce4cf4f40f0437dd0bedf696f3/assets/src/containers/IndexPage.js#L30-L35
The `SelectCard` component should be wrapped in a `Grid` component with an explicitly set width.

Adonais0:
Closes #465


justin0022:
Screen Shot 2019-04-02 at 11 44 00 AM
Almost there! I think the 4th item should be left-aligned though

Adonais0:
Yeah, I was matching with the older MyLA layout. But I guess a left align would look better in this case. Will figure it out!

Screen Shot 2019-04-02 at 3 38 33 PM

justin0022:
Looks good, merging now.

Issue #475, opened 2019-04-02
Support PUT request in React front-end


Adonais0:
Currently the `useFetch` hook only supports `get` requests, we'll also need support for `put` and probably `post` requests in the future

PR #476, opened 2019-04-02
Expanded useFetch to support 'put', 'post', etc.

justin0022:
Addresses https://github.com/tl-its-umich-edu/my-learning-analytics/issues/475.

`GET` is set by default, and the second parameter to `useFetch` can be used to pass `body` and `method` options to the [fetch api](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch).

### DIFF
--- a/assets/src/hooks/useFetch.js
+++ b/assets/src/hooks/useFetch.js
@@ -2,33 +2,46 @@
 import { useState, useEffect } from 'react'
 
 const cache = new Map()
-// https://sudo.isl.co/fetch-me-that-json-from-django/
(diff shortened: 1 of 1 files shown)


justin0022:
@pushyamig if you have a chance, could you pull the latest `React` branch with the modified `useFetch` code to make sure the views run on Firefox/Safari correctly? I suspect it should but thought we ought to double check.

pushyamig:
Yes, I will give it a try later and update here

justin0022:
Thanks!

PR #477, opened 2019-04-08
Disable views for course in React if they are disabled on back end

justin0022:
Fixes https://github.com/tl-its-umich-edu/my-learning-analytics/issues/469.

I included updates from master branch into this PR.

### DIFF
--- a/assets/src/routes/routes.js
+++ b/assets/src/routes/routes.js
@@ -1,27 +1,33 @@
 import { Grade } from '@material-ui/icons'
 
-const routes = courseId => [
-  {
-    path: `/${courseId}/grades`,
-    title: 'Grade Distribution',
-    icon: Grade,
-    description: "See where your grade sits within the course grade distribution.",
-    image: "/static/images/grade_distribution_icon.png",
-  },
-  {
-    path: `/${courseId}/assignment`,
-    title: 'Assignment Planning',
-    icon: Grade,
-    description: "See what assignments have the greatest impact on your grade.",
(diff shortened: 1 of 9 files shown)


assets/src/routes/routes.js
```diff
@@ -1,27 +1,33 @@
...
-]
+const routes = (courseId, activeViews) => {
+  const allViews = [
+    {
+      path: `/${courseId}/grades`,
+      title: 'Grade Distribution',
+      icon: Grade,
+      description: 'See where your grade sits within the course grade distribution.',
+      image: '/static/images/grade_distribution_icon.png',
+      viewCode: 'gd'
```
justin0022: This is slightly _hacky_, based on the JSON sent back from the `/info` endpoint. Could be improved/modified in the future if we change the APIs.

justin0022:
Just a note, both the `IndexPage` and the side drawer should show only the views that are activated for the course. `IndexPage` url is `/test/courses/{courseid}`.

pushyamig:
Previous commit history should not be added for this PR, I feel there is something wrong the way this branch is created

justin0022:
The previous commits are from the `master` branch. Should I update the `react` branch from `master` first, then make this PR?

pushyamig:
I think Andrew already pulled those commits into upstream react branch. You might need to pull the latest changes from the upstream react branch to your local forked branch and them make a new branch from there to not see the commit history.

justin0022:
Okay sure I'll try again then

PR #478, opened 2019-04-09
Merging the master changes to react

pushyamig:
#472 
#471 
#460

### DIFF
--- a/dashboard/admin.py
+++ b/dashboard/admin.py
@@ -1,4 +1,5 @@
 from django.contrib import admin
+from django import forms
 from django.conf import settings
 from django.utils.safestring import mark_safe
 from django.template.defaultfilters import linebreaksbr
@@ -16,8 +17,6 @@ class AlwaysChangedModelForm(ModelForm):
             return True
         return super(AlwaysChangedModelForm, self).has_changed()
 
-class CourseInline(admin.TabularInline):
-    model = Course
 
 class CourseViewOptionInline(admin.StackedInline):
     model = CourseViewOption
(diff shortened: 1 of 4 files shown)


pushyamig:
Trying to keep master changes in sync with react branch. I don't know why it has all the commit history other than 472, 471,460 commits

pushyamig:
All I did was `git merge origin/master' the commit history that is shown here is since react is branched out. when the first merge andrew did from master to react for syncing changes did not added the master commit history. All the merge happened was like a since merge commit. So now this is the secod merge for syncing master and react, I see all the commit history is reappearing here. The file changes reflect the correct difference but not the commit history.

pushyamig:
Ideally this is fine when we merge `react` branch changes to `master` as we don't want all the commit history appear again( i don't know if it will) all the commit would say is`merge commit from master to react'

jonespm:
Yeah, if you squash and merge this again it should be fine. I think it's because you've been git merging verses rebasing. Something to read here
https://www.atlassian.com/git/tutorials/merging-vs-rebasing

If we rebased this branch it might have looked better. We may want to rebase this a bit before it ends up back in master.

jonespm:
Because the previous #467 was merged down to one commit, git can't tell that these 'commits' haven't been included yet. That's what rebase would do since the commits would match up. 

When this is prepared as a PR against master again nothing the only thing that would happen with these merge commits is any changes that were made to resolve conflicts as part of those commits. We could probably rebase the history here to squash a few commits and clean up the history, but it shouldn't be too bad. (As long as this gets merged pretty soon)

jonespm:
You also could have probably git cherry-pick these specific commits ids in rather than merge and that should have avoided this entire history being included in this PR.

pushyamig:
I tried to create [PR](https://github.com/tl-its-umich-edu/my-learning-analytics/pull/479) rebased from master instead of merge it listed 100 files changed while merge only listed 4 files changed.  I will just squash and merge this now and worry about this rebase,history saving later

PR #480, opened 2019-04-09
React disable view

justin0022:


justin0022:
Actually I'll rebase first

PR #481, opened 2019-04-09
Linked 'MY LEARNING ANALYTICS' title and disabled views

justin0022:
Resolves https://github.com/tl-its-umich-edu/my-learning-analytics/issues/469

Only views that are active on back end should appear, both in the side drawer and on the index page.


assets/src/containers/SideDrawer.js
```diff
@@ -30,11 +32,19 @@ function SideDrawer (props) {
 
   const courseId = location.pathname.split('/')[1]
   const [selectedIndex, setSelectedIndex] = useState(false)
+  const [loaded, courseInfo] = useCourseInfo(courseId)
+  const [activeCourseViews, setActiveCourseViews] = useState({})
+
+  useEffect(() => {
+    if (loaded) {
+      setActiveCourseViews(courseInfo.course_view_options)
```
Adonais0: The code looks good to me. 

One thing to be noticed is that the current cards displayed on the index page are always left aligned, with a fix-width center-aligned flex wrapper. It looks a little bit odd when the number of cards is less than 3 since all the cards are left-aligned. I'm thinking of using the number of cards to dynamically change the maxWidth of the wrapper element. Something like:

```
const cardWidth = 340;
const [numOfCard, setNumOfCard] = useState(0)

useEffect(() => {
    if (loaded) {
      setActiveCourseViews(courseInfo.course_view_options)
      let cnt = 0;
      for (let option of Object.keys(courseInfo.course_view_options)) {
        if (courseInfo.course_view_options[option] === 1) {
          cnt ++;
        }
      }
      setNumOfCard(Math.min(3, cnt));
    }
  }, [loaded])

...
<Grid
   container
   className={classes.wrapper}
   style = {{maxWidth: numOfCard * cardWidth + "px"}}
   spacing={8}>
```

It's a little bit hacky so feel free to suggest any other possible solutions for this styling problem. I can open another PR for this if needed.

justin0022: I think it's okay for it to be left aligned. The code above makes the cards 'stretch' out if there's fewer than 3?

Adonais0: No, the above code will make sure the cards are center-aligned even when there're fewer than 3 cards. The problem is that if there are only one or two cards, they'll not be centered on the page by current styling. Because they are wrapped in a left-aligned flex box.

assets/src/containers/DashboardAppBar.js
```diff
@@ -57,7 +66,7 @@ function DashboardAppBar (props) {
           </IconButton>
           <Button>
             <Typography variant='h6' color='inherit' className={classes.grow}>
-              <Link to='/' className={classes.homeButton}>My Learning Analytics</Link>
+              <Link to={path} className={classes.homeButton}>My Learning Analytics</Link>
```
pushyamig: I don't see the Course name is not displaying that is a real piece of information that is needed.

assets/src/containers/IndexPage.js
```diff
@@ -21,24 +23,34 @@ const styles = theme => ({
   }
 })
 
-function IndexPage (props) {
+function IndexPage(props) {
   const { classes, match } = props
-  const currentCourseId = match.params.courseId
+  const courseId = match.params.courseId
+  const [loaded, courseInfo] = useCourseInfo(courseId)
```
pushyamig: My observation: 
The `useCourseInfo` is calling the `/info` endpoint and I see when the user first type or loads the Index page `/courses/12233`  `/info` is being called from `sideDrawer.js` and `IndexPage.js`. The chrome dev tools it is firing two backend call
Screen Shot 2019-04-11 at 2 50 31 PM

Instead, the user just directly goes `/course/1233/grades` bypassing the index page `/info` is fired one time and thereafter he moves to index page the data is pulled from the cache( as part of the `useFetch.js` logic).

Can this `/info` data be stored as a global piece of info somewhere so that subsequent flows would fetch is from that place if not present then make a backend call?

assets/src/containers/DashboardAppBar.js
pushyamig: The course name info is already available in the `/info` endpoint

assets/src/routes/routes.js
```diff
@@ -1,27 +1,33 @@
...
+      viewCode: 'ap'
+    },
+    {
+      path: `/${courseId}/files`,
+      title: 'Files Accessed',
+      icon: Grade,
+      description: 'See what files you and your peers are reading.',
+      image: '/static/images/file_access_trends_icon.png',
+      viewCode: 'fa'
+    }
```
pushyamig: Here is another thing I noticed if I disable say grades view in a course and I see that it is taken out from `index page` and `sidebar`. But if I go directly to grades page of the course like `/course/12233/grades` I see that display of the grades view. instead, it should display a message that says 'Grade Distribution view is hidden for this course.`

pushyamig:
I have been reading about this global state management in the [artical](https://medium.com/javascript-in-plain-english/state-management-with-react-hooks-no-redux-or-context-api-8b3035ceecf8)

justin0022:
> I have been reading about this global state management in the [artical](https://medium.com/javascript-in-plain-english/state-management-with-react-hooks-no-redux-or-context-api-8b3035ceecf8)

I think we can use either the [Context API](https://reactjs.org/docs/context.html) or just pass down the global state via `props` that is held in `App.js`. We probably will only need to pass props down one level, from `App` to the various other containers (e.g. `GradeDistribution`,  `AssignmentPlanning`, etc.)

pushyamig:
These are my thoughts...
The purpose for us seems to be simple, one piece of information that needed (by all the views to be passed as `props` around )is a `view toggle option`, and file view probably uses the `current week` info for showing it in the slider.  So I am think passing props seems to solve the problem since we don't have so many nested components. 
But I just want to explore this option of having global info for next time if it surfaces again. The problem I see is going forward since the global state of info pattern is not set up in project people might tend to pass props around as convenient option

justin0022:
That makes sense. 

If state becomes large and difficult to manage over time, we can also look at using [redux](https://redux.js.org/). Hoping it'll stay relatively manageable though!

assets/src/containers/SideDrawer.js
justin0022: Ah okay. I personally think left-aligned is okay.

assets/src/containers/DashboardAppBar.js
justin0022: Just getting to this - how should the course name be displayed?

assets/src/routes/routes.js
justin0022: Got it, that makes sense.

assets/src/containers/DashboardAppBar.js
justin0022: Okay I've added the course name so that it shows up as "MY LEARNING ANALYTICS: Course Name"

assets/src/containers/IndexPage.js
justin0022: Good point. `useFetch` caches responses, but it looks like in this case the requests happen all at the same time, before any responses return. 

The better way would be to call `useCourseInfo` in `App.js` just once, and then pass down the `courseInfo` as props to both `sideDrawer.js` and `IndexPage.js`. I'll make this change.

justin0022: I've implemented this change here: https://github.com/tl-its-umich-edu/my-learning-analytics/pull/481/commits/03f8e0c7cda0508e1ac09bd45015c0b39d591752

assets/src/containers/DashboardAppBar.js
```diff
@@ -28,7 +27,8 @@ const styles = theme => ({
   },
   homeButton: {
     textDecoration: 'none',
-    color: 'white'
+    color: 'white',
+    textTransform:'capitalize'
```
pushyamig: This change will make the title only to capitalize the First letter in the work

assets/src/containers/DashboardAppBar.js
```diff
@@ -64,7 +63,7 @@ function DashboardAppBar (props) {
           </IconButton>
           <Button>
             <Typography variant='h6' color='inherit' className={classes.grow}>
-              <Link to={courseId} className={classes.homeButton}>My Learning Analytics: {courseName} </Link>
+              <Link to={{pathname:`/${courseId}`}} className={classes.homeButton}>My Learning Analytics: {courseName} </Link>
```
pushyamig: This is will also lead the user to navigate to `/course/234555/` instead of `/course/234555/234555`

pushyamig:
There are a few things that need to work on this issue before we merge it. I will update the list of todo's.

assets/src/containers/IndexPage.js
```diff
@@ -35,19 +35,17 @@ function IndexPage (props) {
...
-            spacing={8}
-          >
-            {routes(courseId, activeCourseViews).map((props, key) =>
-              <Link style={{ textDecoration: 'none' }} to={props.path} key={key}>
-                <SelectCard cardData={props} />
-              </Link>
-            )}
-          </Grid>
-          : <Spinner />}
+        <Grid
```
pushyamig: This is what you assume when a prop is passed down to further down we assume the value is fetched. But earlier behavior is not like that just rendering all the component when things are getting warmed up

justin0022: It should be okay since we're only rendering this component once `courseInfo` is loaded in App? Or does it need to be reverted?

pushyamig: I don't know what you mean by reverting. The index page will show what all the views the course has so it relies on `courseInfo` data. if that courseInfo is not fetched index page won't load and spinner keeps on spinning.  

Another thing I realized is that in the `useFetch` we are assuming that data we are getting is correct and no error handling is done. 
 ```
fetch(dataURL, fetchOptions)
          .then(res => res.json())
          .then(data => {
            cache.set(dataURL, data)
            setData(data)
            setLoaded(true)
          })
```
If something went wrong in fetching the data the subsequest receipent of the `courseInfo` data just grab it as with out considering bad data as an example 
```
loaded
          ? <>
<DashboardAppBar
              onMenuBarClick={setSideDrawerState}
              sideDrawerState={sideDrawerState}
              user={user}
              **courseName={courseInfo.name}**
              courseId={courseId} />
```

Comment (justin0022) [reply assets/src/containers/IndexPage.js]: If `courseInfo` is not fetched, should we display an error message?

I'll add error handling to `useFetch`, maybe by having it return `[loaded, error, data]`. Every component that loads data will then have to check if the data loaded correctly, and if there's an error, display some error message to the user.

Comment (pushyamig) [reply assets/src/containers/IndexPage.js]: When `App.js` don't get the `courseInfo` yes It needs to display error message app is not loaded for reasons.   `grades/assignments/files` views when making their content call error handling is done. None of the views `useFetch.js` (files might be using it now but needs to change to accommodate view disabled msg). But once good `courseInfo` data is fetched the following components like `index`, `DashboardAppBar`,`SideDrawer` etc can simply grab it. 

If the subsequents component can check but what is the need for it if we are catching it at higher up in the component tress

Comment (justin0022) [reply assets/src/containers/IndexPage.js]: Good idea, so if `courseInfo` isn't fetched correctly in `App.js`, then it'll just display an error message.

I'll take this one.

Comment (pushyamig) [reply assets/src/containers/IndexPage.js]: if some error occurs the `/info` call is responding with `{}` if a course doesn't exist. This could happen if a course is not there in DB or DB connectivity problems. But it won't respond with a status code. The `courseInfo` json object has various parts to it that pulling info from different DB tables, I need to make sure all the error has the same error message for consistency sake. I think if it responds with a status code that it might be easier. Let me analyze what the backend call is doing so that Frontend can predict the format. Which do you think you prefer status code or `200 Ok with {}`

Comment (pushyamig) [reply assets/src/containers/IndexPage.js]: OK, I have reviewed this use case. I think if `courseInfo` don't return valid data then the BE is returning with `{}`. This is also the case when user types in bad course number that is not part of the Myla participation then the message should say "Tool is not enabled for this course"

Comment (justin0022) [reply assets/src/containers/IndexPage.js]: That makes sense. If there's a network error or the server is down, then should the same message ("Tool is not enabled for this course") be shown, or should there be a different message shown?

Comment (pushyamig) [reply assets/src/containers/IndexPage.js]: I think if you see the response as {} that means course data couldn't be fetched. So that message should be fine. But the `useFetch`  FetchAPI should handle when the promise fails, currently, the code only handles success case only.

Comment (justin0022) [reply assets/src/containers/IndexPage.js]: This should be resolved now.

File: assets/src/containers/App.js [hunk 8]
```diff
@@ -49,7 +49,7 @@ function App (props) {
               render={props => <GradeDistribution {...props} disabled={!courseInfo.course_view_options.gd}
                 courseId={courseId} />} />
             <Route path='/:courseId/assignment'
```
pushyamig: This commit handles disabling the view and other error cases. On to the files view now

justin0022: Nice this looks good

assets/src/containers/FilesAccessed.js
```diff
@@ -9,9 +9,11 @@ import RangeSlider from '../components/RangeSlider'
 import FormControl from '@material-ui/core/FormControl'
```
pushyamig: This commit takes care of the Files enable and disable view. I have simplified code a bit

pushyamig:
I see one issue I am not sure if it an issue or not. the call to `/info` Endpoint is made with user types in `/courses/12344/` and after that, the user navigates between view I don't see the `/info` called. With all the activity done before(as described above)  if he manually types in the URL in Browser like say `/courses/1233/file` then I see another extra `/info` call is made. I am thinking something to do with the Router logic? Thoughts???

justin0022:
I think when navigating between views using the router, the `info` call is cached, but if the URL is put in directly, then the page reloads, and since `info` isn't cached anymore, the network call is made.

justin0022:
Is this PR good to be merged?

pushyamig:
Yes I feel this is ready to merge

PR #498, opened 2019-04-25
Issues with file_access and other tables after initial migration

### LINKED ISSUE #494: file_access table is created wrong using Django Migration
It looks like with recent [commit](https://github.com/tl-its-umich-edu/my-learning-analytics/commit/e8288a8f094202cf5bf1f4da7be817f099d369ad) the file access table column name look like this `[file_id_id, user_id_id]` instead of `[file_id, user_id]`. 
 The simple fix seems to be adding this attribute to the `FileAccess` model for `file_access` `db_column=file_id` and `db_column=user_id`

This is leading to cron data is not able to load properly and cron is not event erroring saying the table column name mismatch.

Probably the [existing](https://github.com/tl-its-umich-edu/my-learning-analytics/blob/master/dashboard/migrations/0001_initial.py) migration files need to be dropped and new one should be created.

Unless this issue is fixed the migration changes should not be pulled into the react branch

jonespm:
Fixes #494

### DIFF
--- a/dashboard/migrations/0001_initial.py
+++ b/dashboard/migrations/0001_initial.py
@@ -95,17 +95,6 @@ class Migration(migrations.Migration):
             },
         ),
         migrations.CreateModel(
-            name='FileAccess',
-            fields=[
-                ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
-                ('access_time', models.DateTimeField(verbose_name='Access Time')),
-                ('file_id', models.ForeignKey(on_delete=django.db.models.deletion.CASCADE, to='dashboard.File', verbose_name='File')),
-            ],
-            options={
-                'db_table': 'file_access',
-            },
-        ),
-        migrations.CreateModel(
             name='Submission',
             fields=[
                 ('id', models.CharField(max_length=255, primary_key=True, serialize=False, verbose_name='Submission Id')),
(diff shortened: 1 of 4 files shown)


jonespm:
I tested all of the features of the (existing) tool. It loads the cron and runs on the views.

The changes I had to make here
* Because Django doesn't support having a table without a primary key (via migrations) I had to add an id field to file_access.
* Because file_access may have entries that aren't in the file table (as files may have been deleted but still show access) I had to remove the foreign key constraints
* In the user table we had a multiple column primary key on user_id and course_id. Django migrations [still don't support this ](https://code.djangoproject.com/ticket/373) so I just added a regular AutoField here and removed the dual primary key. The design of this table *should* change so grades aren't stored the in this table but that would be a future change.
* For Pandas, I added a some exception handling to the to_sql as some errors were hard to detect here.

This is going to require a few changes to existing (production)databases in a conversion I'll create. Local databases should just be dropped and re-created.

jonespm:
I could have created this as a 0002 migration file to fix these issues but seems like since it's just affecting dev (and prod requires a separate fix) it's okay just to fix it in 0001 and require local devs to rm the **.data** directory

dashboard/migrations/0001_initial.py
```diff
@@ -133,7 +122,8 @@ class Migration(migrations.Migration):
         migrations.CreateModel(
             name='User',
             fields=[
-                ('id', models.CharField(max_length=255, primary_key=True, serialize=False, verbose_name='User Id')),
+                ('id', models.AutoField(primary_key=True, serialize=False, verbose_name='Table Id')),
+                ('user_id', models.CharField(blank=False, max_length=255, null=False, verbose_name='User Id')),
                 ('name', models.CharField(max_length=255, verbose_name='Name')),
```
pushyamig: In `views.py` the call to `get_user_assignment_submission` function need to change to grab the `user_id` instead of `id`
```
sql = "select assignment_id, score, graded_date from submission where " \
          "user_id=(select **user_id** from user where sis_name = %(current_user)s and course_id = %(course_id)s ) and course_id = %(course_id)s"
```

jonespm: Good catch, #506 should refactor this table and use this id directly, but until that happens I think it will have to link up as you describe still.

jonespm:
I plan to merge this this afternoon and update dev to match this structure. Going forward after this we'll want to use the migration system.

jonespm:
```
-- Alter user
ALTER TABLE user RENAME COLUMN id TO user_id;
ALTER TABLE user ADD id INT PRIMARY KEY AUTO_INCREMENT;

```

I think this is the only conversion we need.

PR #499, opened 2019-04-25
Added equivalent hooks to HOCs

justin0022:
Here's an example of converting higher-order components (HOCs) to hooks. I think it's easier to understand, and it also doesn't cause unnecessary component nesting (since each HOC component adds a new layer of nesting). 

Let me know what you think! 

Happy to rewrite the rest of the React D3 components so they use hooks instead of HOCs.

### DIFF
--- a/assets/src/components/Histogram.js
+++ b/assets/src/components/Histogram.js
@@ -1,11 +1,17 @@
+import React, { useState } from 'react'
 import createHistogram from './d3/createHistogram'
-import withResponsiveness from './withResponsiveness'
-import createChartComponent from './createChartComponent'
-import compose from '../util/compose'
+import useChart from '../hooks/useChart'
+import useResponsiveness from '../hooks/useResponsiveness'
 
-const Histogram = compose(
-  withResponsiveness,
-  createChartComponent
-)(createHistogram)
+function Histogram (props) {
+  const [el, setEl] = useState(null)
+
+  const [width, height] = useResponsiveness({ ...props, el })
+  useChart({ ...props, el, width, height }, createHistogram)
+
+  return (
+    <div ref={el => setEl(el)} />
+  )
+}
 
 export default Histogram
--- a/assets/src/components/Histogram.js
+++ b/assets/src/components/Histogram.js
@@ -7,7 +7,7 @@ function Histogram (props) {
   const [el, setEl] = useState(null)
 
   const [width, height] = useResponsiveness({ ...props, el })
-  useChart({ ...props, el, width, height }, createHistogram)
+  useCreateChart({ ...props, el, width, height }, createHistogram)
 
   return (
     <div ref={el => setEl(el)} />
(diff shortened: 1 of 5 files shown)


assets/src/components/Histogram.js
```diff
@@ -1,11 +1,17 @@
...
 import createHistogram from './d3/createHistogram'
-import withResponsiveness from './withResponsiveness'
-import createChartComponent from './createChartComponent'
-import compose from '../util/compose'
-
-const Histogram = compose(
-  withResponsiveness,
-  createChartComponent
-)(createHistogram)
+import useChart from '../hooks/useChart'
```
pushyamig:
```suggestion
import useCreateChart from '../hooks/useCreateChart'
```

assets/src/components/Histogram.js
```diff
@@ -1,11 +1,17 @@
...
+import useResponsiveness from '../hooks/useResponsiveness'
+
+function Histogram (props) {
+  const [el, setEl] = useState(null)
+
+  const [width, height] = useResponsiveness({ ...props, el })
+  useChart({ ...props, el, width, height }, createHistogram)
```
pushyamig:
```suggestion
  useCreateChart({ ...props, el, width, height }, createHistogram)
```

assets/src/components/Histogram.js
```diff
@@ -1,11 +1,17 @@
...
-
-const Histogram = compose(
-  withResponsiveness,
-  createChartComponent
-)(createHistogram)
+import useChart from '../hooks/useChart'
+import useResponsiveness from '../hooks/useResponsiveness'
+
+function Histogram (props) {
+  const [el, setEl] = useState(null)
```
pushyamig: can this have better names for understanding purposes?

pushyamig:
This looks ok, logically make sense. I have few comments based on PR draft.
I know this would be a separate issue, I see the container being called multiple time this probably related to re-render issue. This is true across all the visuals. We need to fix that, I have opened up PR here to keep [track](https://github.com/tl-its-umich-edu/my-learning-analytics/issues/504)

assets/src/components/Histogram.js
justin0022: The `[el, setEL]` bit? Yes, how about `[element, setElement]`. Or maybe `[domElement, setDomElement]`?

pushyamig: [domElement, setDomElement]

justin0022: Resolved here: https://github.com/tl-its-umich-edu/my-learning-analytics/pull/499/commits/795a1c5b70445bff14a2aa344c47419084e9605b

pushyamig:
This PR is opened as Draft I have changed to review mode. This can be merged if all the work is done. I will open up PR with Assignment/Files D3 componenets

pushyamig:
can this be merged?

justin0022:
> can this be merged?

I think so. I'll merge now.

PR #509, opened 2019-05-01
Hoc with hooks for assignment and files view

pushyamig:


assets/src/containers/FilesAccessed.js
```diff
@@ -42,6 +42,7 @@ const defaultFetchOptions = {
 }
 
 function FilesAccessed (props) {
+  console.log("Files Accessed")
```
justin0022: The `console.log` can be removed.

pushyamig:
I Will open a new PR after  I fix my commit history

PR #511, opened 2019-05-01
Removed HOCs and replaced completely with hooks

justin0022:
* The higher order components `withResponsiveness` and `createChartComponent` were replaced with hooks `useResponsiveness` and `useCreateChart`
* `el` was renamed to `domElement` to improve readability.


pushyamig:
Here are my views on this PR, Some of the code that is refactored here are the JS file that we don't use it in the project like `LineChart.js` `HorizontalBarChart.js`, `ScatterPlot.js` etc. Since we are not using it we need to remove from the project. Just as we removed `withResponsiveness.js` and `createChartComponenet.js`. It creates confusion for somebody new coming into the project these files there if not using. If the worry the work being lost then you can store in a local forked branch and pull it up when needed.  If the ideas about the code maintenance are different among others we can have a discussion on this in our next meeting

justin0022:
I'm happy to remove unused D3 components. I have them stored in a separate repo anyway since I use them for other projects.

assets/src/components/BarChart.js
```diff
@@ -1,11 +1,17 @@
...
-import withResponsiveness from './withResponsiveness'
-import createChartComponent from './createChartComponent'
-import compose from '../util/compose'
-
-const BarChart = compose(
-  withResponsiveness,
-  createChartComponent
-)(createBarChart)
+import useResponsiveness from '../hooks/useResponsiveness'
+import useCreateChart from '../hooks/useCreateChart'
```
pushyamig: This can be deleted as well as we are not using it either for Assignment view or files view

assets/src/components/d3/createLineChart.js
```diff
@@ -23,7 +23,7 @@ function createLineChart ({ data, width, height, el }) {
     .attr('transform', `translate(${margin.left},0)`)
     .call(d3.axisLeft(y))
 
-  const svg = createSVG({ el, width, height })
+  const svg = createSVG({ domElement, width, height })
```
pushyamig: This can be deleted as well

assets/src/components/BarChart.js
justin0022: Okay I've removed this one as well

assets/src/components/d3/createLineChart.js
justin0022: This one has been removed

assets/src/components/BarChart.js
justin0022: I've deleted these now

assets/src/components/d3/createLineChart.js
justin0022: I've deleted these now

assets/src/containers/App.js
```diff
@@ -18,14 +22,24 @@ function App () {
...
+          sideDrawerState={sideDrawerState}
+          courseId={courseId}
+          courseInfo={loaded ? courseInfo : null} />
+        <Route path='/:courseId/' exact
+          render={props => <IndexPage {...props} courseInfo={loaded ? courseInfo : null} courseId={courseId} />} />
+        <Route path='/:courseId/grades' render={props => <GradeDistribution {...props} courseId={courseId} />} />
+        <Route path='/:courseId/assignment' render={props => <AssignmentPlanning {...props} courseId={courseId} />} />
+        <Route path='/:courseId/files' render={props => <FilesAccessed {...props} courseId={courseId} />} />
+      </>
     </Router>
```
pushyamig: How is this commit relevant here? this PR is addressing removing HOC with hooks whereas this looks like calling hooks for getting courseInfo and some other changes based on that call.

isn't this supposed to be [PR#481](https://github.com/tl-its-umich-edu/my-learning-analytics/pull/481)

justin0022: Sorry, I think this was a mistake, I'll revert the `App.js` back to how it was.

PR #515, opened 2019-05-10
WIP: React login page

justin0022:
* Login route added, ULR is: `http://localhost:5001/test/courses/account/login`. The route path will be changed in the future. 
* Redirects to `IndexPage` if the username/password is correct.
* Relies on `redirected` response from server to determine if username/password is correct. Probably a better way to verify this.
* Currently course is hardcoded to `/24693/`. There needs to be either a way for user to select which course, or the login page should be hit with a course already selected (i.e `http://localhost:5001/test/courses/account/24693/login`)

### DIFF
--- a/assets/src/containers/App.js
+++ b/assets/src/containers/App.js
@@ -6,6 +6,7 @@ import GradeDistribution from './GradeDistribution'
 import AssignmentPlanning from './AssignmentPlanning'
 import FilesAccessed from './FilesAccessed'
 import IndexPage from './IndexPage'
+import Login from './Login'
 
 function App () {
   const [sideDrawerState, setSideDrawerState] = useState(false)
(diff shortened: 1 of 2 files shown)


justin0022:
Screenshot 2019-05-10 12 50 36
Screenshot 2019-05-10 12 51 34

pushyamig:
A custom Login page is not needed if a form is posted to `/account/login` Django default login should come up as it is done now. In dev/prod env it will be a single sign-on that get handles this for you.

Something in these lines should work 
```
<form action="/account/login" method="post">
                        <CSRFToken />
                        <button type="submit">Send</button>
                 </form>
```

https://www.techiediaries.com/django-react-forms-csrf-axios/

jonespm:
Yeah, I'm not sure if we really need a custom login page either. 

As Pushyami said the login really only exists for local dev testing. It will either be launched in production via SAML or LTI and these auth's are only in the backend. I believe that we'll have to add some additional data to the end-points or provide that data via a javascript variable in #524 but I think it's fine if the login isn't part of the SPA.

justin0022:
Ah got it, that makes sense. I'll close this PR.

PR #525, opened 2019-05-23
React django mix

jonespm:


jonespm:
This is an idea of how we could easily merge the some of the previous django context processors and have it still work with the great work going on in the react side. It looks like the way you implemented it in the with the webpack_loader this was really straightforward. I believe most of the functionality on the title should be replicated in react rather than using HTML in Django because I like the hamburger and the styling there. I think we can provide these as javascript values from the backend. We can trust this because all of the checks are being done against the backend so even if a user changes it they wouldn't get any different access. Here's the current version. The old Django header is just there as a quick reference to what needs to be recreated in react. 

image

jonespm:
FYI: I communicated to @pushyamig that we had sample data used last week at the Unizin Hackathon but there is still some issue with the assignment data that I need to resolve by next week. You can have that link but we'd also need to merge #381 #498 and #522 (and possibly others) into this branch so the model is synced.

justin0022:
This is great! React just requires an entry point (in this project `<div id="root"></div>`), so I don't see any problems with this.

justin0022:
I think once the Django header is removed, this can be merged.

jonespm:
Thanks, I'll clean up the base.html file and commit this on Tuesday (Monday's a holiday for us).

I'll also create 2 new issues.
1) Remove code and templates for unused Django old views
2) Implement features in previous header in new react header 
  (These included Login link, access to Admin, logout of logged in and some indication of which view was displayed)

justin0022:
Sounds good, I'll be happy to take the second issue on.

Issue #528, opened 2019-05-28
Implement features in previous header in new react header


jonespm:
In the previous header there were these features

If the user is not logged in:
- Provide a login link to `settings.LOGIN_URL`

If the user is logged in
- Provide a logout link to settings.LOGOUT_URL
- Indicate the currently logged in username
- Indicate the current course id (if you're in a course

If the user is an admin
- Show the admin link to /admin

The values of whether a use is logged in and if they're an admin or not can just be provided via javascript. There's no security risk for exposing the /admin URL as it's protected from someone not admin going to it.

jonespm:
We'll also need to add a link to "Help"

Previously this was hardcoded so this will also need a new setting, probably passed to the frontend.

The old logic added a view specific URL to the help link. This looks like it's being cleaned up in #529

justin0022:
> * Indicate the current course id (if you're in a course

Should the course ID be displayed, or the course name? In the `React` branch the course name is currently being displayed.

pushyamig:
Displaying Course name and not ID as it is done currently

justin0022:
Implemented here: https://github.com/tl-its-umich-edu/my-learning-analytics/pull/529

Issue #523, opened 2019-05-22
Adding the Copyright info to the footer of each page


pushyamig:
Since the project will be used by other than umich, we need to come up with a way to add copyright info. currently, the code has Copyright © 2018 The Regents of the University of Michigan

justin0022:
Added in this PR: https://github.com/tl-its-umich-edu/my-learning-analytics/pull/529

justin0022:
Currently the copyright notice is hardcoded, but it is added to the footer of every page.

pushyamig:
We can work on this after the react branch merged to master. The effort is small so later could be worked on

PR #529, opened 2019-05-28
Cleanup unused files, simplify base.html

justin0022:
Made this a draft PR since there's likely additional things that can be removed.

Fixes formatting issues caused by imported CSS and made the footer not rely on Bootstrap anymore.

Remaining tasks:
- [ ] Redirect to login page if user is not logged in
- [X] Add user information to Avatar component
- [X] Logout button in Avatar should correctly route
- [ ] ~~Add user information to `myla_globals`~~
- [ ] Copyright info should be configurable, not hard coded
- [X] If user is admin, show link (probably in avatar) to `/admin` route
- [ ] ~~Display current course ID~~
- [X] Show context-sensitive help page (e.g. if on Grades view, link to grades-distribution help page)


dashboard/templates/base.html
```diff
@@ -64,24 +56,4 @@
     });
 </script>
 {% endif %}
-<script src="{% static 'moment/min/moment.min.js' %}"></script>
-<script src="{% static 'jquery/dist/jquery.min.js' %}"></script>
-<script src="{% static 'angular/angular.js' %}"></script>
-<script src="{% static 'underscore/underscore-min.js' %}"></script>
-<script>
-    $(document).ready(function () {
```
jonespm: This Help Link code will need to be re-added to the frontend too in #528 but good to remove here. 

The link should really also be provided via a setting.

dashboard/templates/base.html
```diff
@@ -36,16 +31,13 @@
...
     {% include "su/is_su.html" %}
     {% block content %}
     {% endblock %}
     <footer>
-        <hr>
-        <div class="row">
-            <div class="col text-left">Copyright © 2018 The Regents of the University of Michigan</div>
-            <div class="col text-right">Data last updated on {{last_updated|date:"m/d/Y P T"}}</div>
-        </div>
+        <div style="float: left; width: 50%">Copyright © 2018 The Regents of the University of Michigan</div>
```
jonespm: Fixing this was part of #523 but it's fine to leave it as it is for now.

dashboard/urls.py
```diff
@@ -45,28 +45,28 @@
     url(r'^courses/(?P<course_id>[0-9]+|)/?view_file_access_within_week', login_required(TemplateView.as_view(template_name='view_file_access_within_week.html')), name="view_file_access_within_week"),
```
jonespm: All of these url's that reference the removed *.html files can also be removed

dashboard/templates/base.html
justin0022: Right, this should come from the env variables.

dashboard/templates/base.html
justin0022: Screenshot 2019-05-30 10 01 37
The help is actually already linked. But ya the link should be configurable, also in env variable?

jonespm: Yeah right in AvatarModel it looks like it's hardcoded to the umich.edu help. This will need to be an env for like BASE_HELP_URL or something. 

Also the help in base.html appended the specific view this was on to give it "context sensitive" help. I'm not sure how easy that would be in React to have that happen.

dashboard/urls.py
justin0022: Could all the files under `dashboard/static` be removed as well?

jonespm: Yeah, if you have it all somewhere else. The 
dashboard/static/json is all old and unused.
dashboard/static/images were the images that we might want to retain if they fit into the design
dashboard/static/js had the code for creating the d3 for the views that you've probably re-created. 
dashboard/static/css I'm guessing isn't needed.

justin0022: The images are in `assets/src/static`, so I think the `dashboard/static` folder can be removed then.

justin0022: I've removed these files now.

pushyamig:
I see a problem with the coding changes in this PR. The URL `localhost:5001` doesn't work and we need to hook up log in and logout buttons to the react. This is the exception it throws now`django.template.exceptions.TemplateDoesNotExist: home.html`
The only way we can log in now by going to the URL `localhost:5001/accounts/login` logout: `localhost:5001/accounts/logout`

`base.html` `myla_globals` variable will be populated and based on the that we need to control the flow.
```
var myla_globals = {
        {% if course_id %}
            "course_id": '{{ current_user_incremented_course_id }}',
        {% else %}
            "course_id": 0,
        {% endif %}
            "is_superuser": {{user.is_superuser|lower}},
        {% if user.is_authenticated %}
            "username": '{{ user.get_username }}',
        {% else %}
            "username": ""
        {% endif %}

        };
```

We need to have a page home page in react with login flow. once the login is successful somehow the current `APP.js` code should be handled. To the code in the `App.js` need to change based on the `myla_globals`.

This is what the flow should be when user hits `localhost:5001` a landing page should appear with login button and once he click it for local development django admin flow takes care of that. But for the dev SAML flow. once that is done the `myla_globals` get populated and then user can type the course to get the visuals. I see `App.js` needs to change from the way it is implemented.

This is the final piece that glues all the project together for a finished product with react.

justin0022:
In the `myla_globals`, how can we tell if user is not currently logged in? I think we can use React-Router to determine if a user is not logged in from `myla_globals`, and redirect to the login page.

dashboard/templates/base.html
justin0022: This should be pretty straight forward

justin0022:
I thought I had wired the logout button correctly; in line 85 of `AvatarModal`
```js
<ListItemText inset primary='Logout' onClick={() => (window.location.href = `${window.location.hostname}:${window.location.port}/accounts/logout`)} />
```
But for some reason when the logout button is pressed, I see the status show `(canceled)`. Will need to investigate further.

pushyamig:
@jonespm 
1. For leveraging the Django Login flow for local development, It is needed that we still hold `registration/login.html` file for loading the login Django template. The URL for login/logout are exposed via js variables in the `base.html`. 
2. Reacts routing controlling navigations to the view once the login is successful based on the JS variable which is populated from Django. ``` {% if user.is_authenticated %}
            "username": '{{ user.get_username }}',
        {% else %}
            "username": "",
        {% endif %}```
the only thing that is coming from the user is `course_id` once they type in from the browser. I see that BE Django calculates it 'current_user_course_id` and this variable is also exposed to FE as well but won't be of any use if the user admin or user is multiple courses
```{% if course_id %}
            "course_id": '{{ current_user_incremented_course_id }}',
        {% else %}
            "course_id": 0,
        {% endif %}```
So the FE flow is relying on the user to input the course id.
One thing with this flow we don't need a login page anymore once we see if the user is not authenticated via JS variable then we can present with the flow like this 
```
 const user = {
    admin: myla_globals.is_superuser
  }

  if (!user.username) return (
    window.location.href = 'http://localhost:5001/accounts/login'
  )

  return (
    <Router basename='/courses/'>
      <div>
        <DashboardAppBar onMenuBarClick={setSideDrawerState} sideDrawerState={sideDrawerState} user={user} />
        <SideDrawer toggleDrawer={setSideDrawerState} sideDrawerState={sideDrawerState} />
        <Route path='/:courseId' exact component={IndexPage} />
        <Route path='/:courseId/grades' component={GradeDistribution} />
        <Route path='/:courseId/assignment' component={AssignmentPlanning} />
        <Route path='/:courseId/files' component={FilesAccessed} />
      </div>
    </Router>
  )```
This will eliminate the necessity for the login page. Thoughts???

justin0022:
Here's a link of what this login flow looks like: https://www.dropbox.com/s/9wipxrz2bve5vec/login.mp4?dl=0

dashboard/templates/base.html
```diff
@@ -42,6 +42,17 @@
         {% else %}
         "logout": "",
         {% endif %}
+        {% if current_user_courses_info%}
+            "user_courses_info":'{{current_user_courses_info|safe}}',
```
pushyamig: after the user login to the course this variable will be populated like below `"user_courses_info":'[{"course_id":"556114","course_name":"LATIN 842 009 SU 2062"},{"course_id":"717406","course_name":"DENT 877 005 SU 2067"}]'`

Issue #532, opened 2019-06-03
add more "Resource Access" data into File Access view


zqian:
The File Access view currently only show Canvas File Access pattern. 

There are several Caliper data types inside UDP, which include video view data from lecture capture and Kaltura, eTextbook access data from VitalSource and Runestone Interactive. 

It would be beneficial to include those caliper data into the File Access view, too, and provide filter to let student view all or one specific source stream.

**This issue will focus on adding the Lecture Capture view data.** 

Here is the task list to start with, please feel free to modify:

- [x] create sample/de-identified video access events. 
- [x] make sure the above events can be loaded successfully into MyLA db. Both "file" and "file_access" table will be updated
- [x] make sure the data shows correctly in the File Access view
- [x] add filter so that student can see all events combined, or "Access Canvas Files" or "View Lecture Capture videos", respectively.

jennlove-um:
Testing passes: Resources view is correctly showing Lecture Capture and Files resources for courses that have multiple types of resources.

PR #534, opened 2019-06-07
Issue 532 should be done

vjcao:
views.py, .env, view_file_access_within_week.html, file_access_within_week.js, and settings.py were modified to function with lecture capture events and other file types in the future. Added .env and sql files as references of what was used to test the file view access portion of myLA.


.env
```diff
@@ -0,0 +1,165 @@
+# Required to set this with some value
```
pushyamig: `.env` should never is committed

pushyamig:
The work done as part of the FE needs to be changed based on the React flow. The BE work is still valid. This PR need to adapt to React branch to get merged

jonespm:
I have a incomplete PR that's changing the backend. I'll talk to @vjcao about it and he can incorporate that as well as rework this for React.

vjcao:
Redoing #532 in react

PR #535, opened 2019-06-10
Allows user to select course after login

justin0022:
After login, uses `myla_globals.user_course_info` to create cards that show what courses the user has access to. 

I'm not sure if this is the best UI for this - I used the `SelectCard` component, and there's no image or description I can use. Also, the App Bar doesn't appear, since this component loads before `App.js` loads. 

I initially tried to add this component into `App.js`, but it didn't really make sense to me. 

- [x] file missing warning with respect to the font file. This is possible because it is looking in the `/static directory but the fonts are actually in `assets/dist/static`. This might be Django configuration
- [x] Fix the become user message is on the top of the page instead of the bottom. or where ever that looks appealing
Screen Shot 2019-06-13 at 12 39 00 PM
- [x] Admin should have the AppBar after logging in. Currently only displays a simple message "Select a course you choose to look in". Admin what to become a user and so having a pop-up avatar model will be helpful
Screen Shot 2019-06-13 at 12 44 03 PM


assets/src/containers/CourseList.js
```diff
@@ -0,0 +1,44 @@
...
+    margin: theme.spacing.unit * 2 + 'px auto',
+    flexDirection: 'row',
+    justifyContent: 'flex-start'
+  }
+})
+
+function CourseList (props) {
+  const { classes } = props
+  // currently JSON.parse is required because Django is returning string instead of object
+  const userCourseInfo = JSON.parse(myla_globals.user_courses_info)
```
pushyamig: I think this can't be returned as an object. since the context processor returning a dictionary so it is technically a string. Context processor won't return a response object as a typical API call does. So we need to do a JSON.parse to convert it into an object.

justin0022:
Even after hitting `accounts/logout`, if the browser back button is pressed, it appears the user is still logged in.

pushyamig:
In local logging out in not logging out the user. You need to clear the active logins. But if your dev/prod instance has shib integration and user hits logout it actually clears the user session in BE django

pushyamig:
A couple of things
1. I feel the `CourseList` page need a `<AppBar>` Separate from What `/courses` route uses. That Bar will have just the `AvatorModel` info and no sidebar it would look nice.

```
 <AppBar className={classes.root} position='static'>
      <Toolbar>
        <Button>
            My Learning Analytics: {courseName}
          </Typography>
        </Button>
        <div className={classes.grow} />
        <IconButton
          aria-owns={avatarOpen ? 'simple-popper' : undefined}
          onClick={event => setAvatarEl(event.currentTarget)}
          color='inherit'
          aria-haspopup='true'
          variant='contained'>
          <Avatar>{user.username.slice(0,1)}</Avatar>
        </IconButton>
        <Popover
          open={avatarOpen}
          anchorEl={avatarEl}
          onClose={() => setAvatarEl(null)}
          anchorOrigin={{
            vertical: 'bottom',
            horizontal: 'center'
          }}
          transformOrigin={{
            vertical: 'top',
            horizontal: 'center'
          }}
        >
          <AvatarModal user={user} />
        </Popover>
      </Toolbar>
    </AppBar>```
2. I see this warnining ` You are attempting to use a basename on a page whose URL path does not begin with the basename. Expected path "/" to begin with "/courses".`

**Comment by justin0022** [comment]
Okay I'll add just the app bar in, with no sidebar. And I'll look into that error.

File: assets/src/containers/CourseList.js [hunk 2]
```diff
@@ -60,7 +60,7 @@ function CourseList (props) {
       <AppBar className={classes.root} position='static'>
```
pushyamig: This commit resolves the warning message
`tiny-warning.esm.js:13 Warning: You are attempting to use a basename on a page whose URL path does not begin with the basename. Expected path "/" to begin with "/courses".`

pushyamig:
I have a few remaining items on this issue added a description. I will look into the missing font files piece

assets/src/containers/CourseList.js
```diff
@@ -88,15 +88,18 @@ function CourseList (props) {
...
         </Toolbar>
       </AppBar>
-      <Grid container spacing={16} className={classes.root}>
-        <Grid item xs={12} className={classes.container}>
-          {userCourseInfo.map((course, key) =>
-            <Link style={{ textDecoration: 'none' }} to={`/courses/${course.course_id}`} key={key}>
-              <SelectCard cardData={{ title: course.course_name }} />
-            </Link>
-          )}
+      {isSuperuser ? <Paper className={classes.paper}><Typography>Select a course you choose to look
```
pushyamig: the `isSuperuser` variable control the content to show.

Dockerfile
```diff
@@ -45,6 +45,7 @@ COPY . /code/
 # copy built react and node libraries for production mode
 COPY --from=node-prod-deps /usr/src/app/package-lock.json /code/package-lock.json
 COPY --from=node-prod-deps /usr/src/app/webpack-stats.json /code/webpack-stats.json
+COPY --from=node-prod-deps /usr/src/app/assets/images /code/static
```
pushyamig: Images can be contained in the `assets` directory instead of `dashboard/static` director

assets/src/index.js
```diff
@@ -3,12 +3,17 @@ import ReactDOM from 'react-dom'
...
 import './index.css'
 import App from './containers/App'
+import CourseList from './containers/CourseList'
 import 'typeface-roboto'
 // import * as serviceWorker from './serviceWorker'
 
 ReactDOM.render(
-  <Router basename='/courses/'>
-    <Route path='/:courseId' component={App} />
+  <Router basename='/'>
```
pushyamig: This commit seems to resolves the warning message
`tiny-warning.esm.js:13 Warning: You are attempting to use a basename on a page whose URL path does not begin with the basename. Expected path "/" to begin with "/courses".`

dashboard/templates/base.html
```diff
@@ -13,11 +13,26 @@
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
     <meta name="viewport" content="width=device-width, initial-scale=1">
     <title>{% use_macro title %}</title>
+    <style>
```
pushyamig: With this styling the logged in as another user message is place at the bottom

dashboard/urls.py
```diff
@@ -81,7 +81,7 @@
     # Login patterns for testing, SAML should be installed in prod
     urlpatterns += (
         url(r'^accounts/login', auth_views.LoginView.as_view(), name='login'),
-        url(r'^accounts/logout', auth_views.LoginView.as_view(), name='logout'),
+        url(r'^accounts/logout', auth_views.logout, name='logout'),
```
pushyamig: This is a nice Find!

webpack.config.js
```diff
@@ -18,7 +18,7 @@ module.exports = {
         use: ['style-loader', 'css-loader']
       },
       {
-        test: /\.(bmp|gif|jpe?g|png|woff|woff2|eot|ttf|svg)$/,
+        test: /\.(bmp|gif|jpeg|png|woff|woff2|eot|ttf|svg)$/,
```
pushyamig: I suppose this is a mistake so removed the `?`

Dockerfile
```diff
@@ -45,7 +45,7 @@ COPY . /code/
 # copy built react and node libraries for production mode
 COPY --from=node-prod-deps /usr/src/app/package-lock.json /code/package-lock.json
 COPY --from=node-prod-deps /usr/src/app/webpack-stats.json /code/webpack-stats.json
-COPY --from=node-prod-deps /usr/src/app/assets/images /code/static
+COPY --from=node-prod-deps /usr/src/app/assets/dist/static /code/static
 COPY --from=node-prod-deps /usr/src/app/assets /code/assets
```
pushyamig: This fixes this error 
Screen Shot 2019-06-14 at 10 44 42 AM

pushyamig:
I see this warning regarding with fonts as you start navigating to the site. The number of warnings increases. I am not sure how to fix this.  This probably seems to be due to the routes as it cannot find the font there. `http://localhost:5001/courses/static/roboto-latin-400.5d4aeb4e.woff2`. Obviously, the fonts are not placed there but in the `/static`. 
Screen Shot 2019-06-14 at 8 06 01 AM

I think other than this warning the flow from start to finish seems to works fine.

jonespm:
This is a great feature, looking forward to it. Just need to look at the fonts issue and resolve the conflicts!

pushyamig:
I see the warning w.r.t to fonts is happening only in chrome and not in FF and Safari. I tried a few things like setting mimeType, removed the [hash], etc did not seems to work
```
{
        test:/\.(woff2|woff)$/,
        loader: 'url-loader',
        options: {
          mimetype: 'application/font-woff',
          name: 'static/[name].[ext]',
          limit: 10000
        },
      }
```
The version of Chrome browser I am using for test is `Version 75.0.3770.80 (Official Build) (64-bit)`. I don't know if this is true for other chrome version browsers. 
The warning happens when the user using the browser to navigate to a view like `/courses/1233/grades` and not with UI flow. Like from the index page (which has grades/assignments/files) click on the grades page the warning doesn't happen. 

To me this is a minor issue can be fixed later thoughts??

justin0022:
I think it is a minor issue - I'll take a look at this next week.

pushyamig:
I created a separate [issue](https://github.com/tl-its-umich-edu/my-learning-analytics/issues/547) for fixing the warning in the browser. You can merge this request once the conflicts are resolved

justin0022:
I think it's ready to merge now.

pushyamig:
I thought font Roboto library is needed since Material-UI default is that and it won't ship with it so we need to include that in the package.json

pushyamig:
more info here under Typography https://blog.logrocket.com/the-definitive-guide-to-react-material-d730c8a3e8ba/

justin0022:
Okay I've added the typeface back to package.json. I'm certain that the import in `index.js` was the cause of the warnings.

pushyamig:
I think we can merge this PR to react and reopen issue #547 to work on later.

Dockerfile
```diff
@@ -45,7 +45,6 @@ COPY . /code/
 # copy built react and node libraries for production mode
 COPY --from=node-prod-deps /usr/src/app/package-lock.json /code/package-lock.json
 COPY --from=node-prod-deps /usr/src/app/webpack-stats.json /code/webpack-stats.json
-COPY --from=node-prod-deps /usr/src/app/assets/dist/static /code/static
```
pushyamig: I had to remove the line since (`import 'typeface-roboto'`)  the roboto font now is not imported anymore with this [change](https://github.com/tl-its-umich-edu/my-learning-analytics/pull/535/commits/f5318bb7588a45cf06aa5f1d05d414506b8b5e82)

Dockerfile
pushyamig: This error is never happening now with roboto not importing

pushyamig:
I resolved the conflicts

pushyamig:
@justin0022 If the roboto is added in `package.json` and never imported in the code, what difference it makes it having it in `package.json`. The app seems to be functional without the import and `dist/static` directory is never created

PR #539, opened 2019-06-13
Add google analytics to react

### LINKED ISSUE #470: Support Google analytics with React
The current implementation uses Google analytics for tracking purpose, we still want to support that as apart of that effort

https://medium.com/alturasoluciones/how-to-set-up-and-use-google-analytics-in-react-apps-fb057d195d13

andrew-gardener:
- GoogleAnalyticsTracking component will track page views when initialized and every time the route is changed
- refactored how routers/routers are setup so there is only one router in the application (app is now the main routing entry point which can have components added like CourseList and the new Course)
- Note: static standalone pages are not currently tracked (ex: login page which will eventually be converted to a react page)

closes #470

### DIFF
--- a/assets/src/components/GoogleAnalyticsTracking.js
+++ b/assets/src/components/GoogleAnalyticsTracking.js
@@ -1,0 +1,29 @@
+import React, { useState, useEffect } from "react"
+import GoogleAnalytics from "react-ga"
+import { withRouter } from 'react-router-dom'
+
+function GoogleAnalyticsTracking (props) {
+  const {
+    gaId
+  } = props
+
+  const [initialized, setInitialized] = useState(false)
+  const [previousPage, setPreviousPage] = useState(null)
+
+  if (gaId && !initialized) {
+    setInitialized(true)
+    GoogleAnalytics.initialize(gaId)
+  }
+
+  useEffect(() => {
+    const page = window.location.pathname + window.location.search + window.location.hash
+    if (gaId && page != previousPage) {
+      setPreviousPage(page)
+      GoogleAnalytics.pageview(page)
+    }
+  })
+
+  return null
+}
+
+export default withRouter(GoogleAnalyticsTracking)
--- a/assets/src/containers/App.js
+++ b/assets/src/containers/App.js
@@ -1,63 +1,34 @@
 import React, { useState } from 'react'
-import { BrowserRouter as Router, Route } from 'react-router-dom'
-import DashboardAppBar from './DashboardAppBar'
-import SideDrawer from './SideDrawer'
-import GradeDistribution from './GradeDistribution'
-import AssignmentPlanning from './AssignmentPlanning'
-import FilesAccessed from './FilesAccessed'
-import IndexPage from './IndexPage'
-import Spinner from '../components/Spinner'
-import Error from './Error'
-import { isObjectEmpty } from '../util/object'
-import { useCourseInfo } from '../service/api'
+import { Route, withRouter } from 'react-router-dom'
+import { matchPath } from 'react-router';
+import GoogleAnalyticsTracking from '../components/GoogleAnalyticsTracking'
+import CourseList from './CourseList'
+import Course from './Course'
 
 function App (props) {
-  const { match } = props
-  const courseId = match.params.courseId
-  const [loaded, error, courseInfo] = useCourseInfo(courseId)
-  const [sideDrawerState, setSideDrawerState] = useState(false)
+  const {
+    location
+  } = props
+  const isLoggedIn = !!myla_globals.username
+  // replace with redirect to react login page later
+  if (!isLoggedIn) {
+    return (window.location.href = myla_globals.login)
+  }
 
+  const coursePageMatch = matchPath(location.pathname, '/courses/:courseId/')
+  const courseId = coursePageMatch ? coursePageMatch.params.courseId : null
   const user = {
     username: myla_globals.username,
     admin: myla_globals.is_superuser
   }
-
-  if (!user.username) return (window.location.href = myla_globals.login)
-  if (error) return (<Error>Something went wrong, please try again later.</Error>)
-  if (loaded && isObjectEmpty(courseInfo)) return (<Error>Tool is not enabled for this course.</Error>)
-
   return (
-    <Router basename='/courses/'>
-      {
-        loaded
-          ? <>
-            <DashboardAppBar
-              onMenuBarClick={setSideDrawerState}
-              sideDrawerState={sideDrawerState}
-              user={user}
-              courseName={courseInfo.name}
-              courseId={courseId} />
-            <SideDrawer
-              toggleDrawer={setSideDrawerState}
-              sideDrawerState={sideDrawerState}
-              courseId={courseId}
-              courseInfo={courseInfo} />
-            <Route path='/:courseId/' exact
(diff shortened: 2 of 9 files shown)


assets/src/components/GoogleAnalyticsTracking.js
```diff
@@ -0,0 +1,42 @@
...
+    ga_initialized = true
+    GoogleAnalytics.initialize(google_analytics_id)
+  }
+  if (google_analytics_id) {
+    GoogleAnalytics.set({ page, ...options, })
+    GoogleAnalytics.pageview(page)
+  }
+}
+
+const GoogleAnalyticsTracking = class extends Component {
```
pushyamig: is there any reason why you used the class-based component instead of function based components. The whole react design is based on using functional components using Hooks for handling the side effects

dashboard/templates/registration/login.html
```diff
@@ -9,4 +9,5 @@ <h2>Login</h2>
     {{ form.as_p }}
     <button type="submit">Login</button>
   </form>
+  {% include "ga/include_ga.html" %}
```
pushyamig: Here are my thoughts on this, the `login.html` is just for local development purposes and in the dev/prod. The trigger for Google analytics starts from here and I don't know if the tracking will start taking effect in higher env. I need to test this.

assets/src/containers/App.js
```diff
@@ -43,6 +44,7 @@ function App (props) {
               sideDrawerState={sideDrawerState}
               courseId={courseId}
               courseInfo={courseInfo} />
+            <GoogleAnalyticsTracking />
```
pushyamig: This may not be the only tracking point to start with. With this PR[https://github.com/tl-its-umich-edu/my-learning-analytics/pull/535](url) we are adding the courseList page. I wonder if that get tracked at all with the change of [Router](https://github.com/tl-its-umich-edu/my-learning-analytics/pull/535/files#diff-e70b04386946903cb7c87b89a6de7cc0R3) setup.

pushyamig:
I might need to test this change in one of over dev environment.

assets/src/components/GoogleAnalyticsTracking.js
justin0022: I think Andrew found this code from here: https://github.com/react-ga/react-ga/wiki/React-Router-v4-withTracker

The hooks version didn't work for some reason.

pushyamig: I'm fine with it at this point. I was surprised that it was not done using functional component since your design is based on using it since beggining

justin0022: Me too. We can change in the future once the higher-priority items are complete.

dashboard/templates/registration/login.html
andrew-gardener: I've removed the tracking from the login page for now since it will also be rewritten in react (just mentioned he had a branch for it)

assets/src/containers/App.js
andrew-gardener: I've rebased the changes on #535 and refactored how the routers are setup. There should now only be one root level router (which GoogleAnalyticsTracking can be included in)

assets/src/components/GoogleAnalyticsTracking.js
andrew-gardener: I was previously using a slightly modified version of the link Justin provided but have now re-written it to be hook based

assets/src/containers/App.js
```diff
@@ -1,64 +1,33 @@
...
-                courseId={courseId} />} />
-            <Route path='/:courseId/files'
-              render={props => <FilesAccessed {...props} courseInfo={courseInfo}
-                courseId={courseId} />} />
-          </>
-          : <Spinner />
-      }
-    </Router>
+    <>
+      <GoogleAnalyticsTracking gaId={myla_globals.google_analytics_id} {...props} />
```
justin0022: You don't need to send the rest of the props over to `GoogleAnalyticsTracking`.

assets/src/containers/App.js
```diff
@@ -1,64 +1,33 @@
...
+  } = props
+  const isLoggedIn = !!myla_globals.username
+  // replace with redirect to react login page later
+  if (!isLoggedIn) {
+    return (window.location.href = myla_globals.login)
+  }
 
-  // this is temporary
+  const coursePageMatch = matchPath(location.pathname, '/courses/:courseId/')
+  const courseId = coursePageMatch ? coursePageMatch.params.courseId : null
```
pushyamig: This is a nice refactoring.

pushyamig:
I tested this PR running on `https://dev-myla.tl.it.umich.edu/`.you can also use the chrome extension `Google Analytics Debugger` for tracking the events if the GA account is not setup

assets/src/routes/routes.js
```diff
@@ -3,23 +3,23 @@ import { Grade } from '@material-ui/icons'
...
+      path: `/courses/${courseId}/grades`,
       title: 'Grade Distribution',
       icon: Grade,
       description: 'See where your grade sits within the course grade distribution.',
       image: '/static/images/grade_distribution_icon.png',
       viewCode: 'gd'
     },
     {
-      path: `/${courseId}/assignment`,
+      path: `/courses/${courseId}/assignment`,
```
pushyamig: This is changed to `/courses/${courseId}/assignments`

assets/src/containers/App.js
pushyamig: I feel there should be separate URLs for courses list page `http://dev.edu/courses` the base URL which is `http://dev.edu`. Currently, both are tracked under `/`

assets/src/containers/App.js
andrew-gardener: Fixed

assets/src/containers/App.js
andrew-gardener: for now I've added another route (`/courses`) that also shows the CourseList screen

assets/src/routes/routes.js
andrew-gardener: fixed here as well as in the Course commponent

assets/src/containers/App.js
pushyamig: What I mean is `CourseList` page should only be used for `/courses`

andrew-gardener: done

pushyamig: We still need `<Route path='/' exact render={props => <CourseList {...props} user={user} />} />` since after login it is redirecting to no component. This is fine for now. We need to change later

pushyamig:
```suggestion
 <Route path='/' exact render={props => <CourseList {...props} user={user} />} />
```

pushyamig:
After the [suggested](https://github.com/tl-its-umich-edu/my-learning-analytics/pull/539#discussion_r295829836) changes are fixed, and  the #535  is merged and conflicts are resolved, merge this issue

assets/src/containers/App.js
andrew-gardener: I've added `<Route path='/' exact render={props => <CourseList {...props} user={user} />} />` back in

pushyamig:
@andrew-gardener I have merged the #535 change, you can merge your changes after resolving the confilcts

PR #540, opened 2019-06-13
Update model schema

### LINKED ISSUE #318: Change char(255) types over to BIGINT
It would better match the source data if our char(255) data types that are numeric (like id, canvas_id) were switched from char(255) over to BIGINT. This is what they are in postgres. This would probably also result in a slight performance increase. This isn't the highest of priority since we'd like to remove all this code but might still stick around awhile.

andrew-gardener:
- Change all ids from CharField to BigIntegerField
- change `assignment.points_possible`, `assignment_groups.weight`, `assignment_groups.group_points`, `submission.score`, `user.current_grade`, and `user.final_grade` from CharField to FloatField
- Note there are three migrations instead of one in order to get around the foreign key constraint in the `course_view_option` table

Closes #318


andrew-gardener:
Note: I didn't test running files_access view or file fetching in the cron but I looked them over by eye so there (hopefully) aren't any issue with them

dashboard/models.py
```diff
@@ -283,8 +283,8 @@ class Meta:
 
 class FileAccess(models.Model):
     id = models.AutoField(primary_key=True, verbose_name="Table Id")
-    file_id = models.CharField(blank=True, max_length=255, null=False, verbose_name='File Id')
-    user_id = models.CharField(blank=True, max_length=255, null=False, verbose_name='User Id')
+    file_id = models.BigIntegerField(blank=True, null=False, verbose_name='File Id')
+    user_id = models.BigIntegerField(blank=True, null=False, verbose_name='User Id')
```
pushyamig: Due to this change, the file view is broken. It a minor change something like this needs to be added in the `views.py` 
```
df['file_id_name'] = df['file_id'].astype(str).str.cat(df['file_name'], sep=';')
```

dashboard/models.py
```diff
@@ -84,8 +84,8 @@ def set_user_defaults(self, course_id, user_id, default_view_type, default_view_
 
 class UserDefaultSelection(models.Model):
     id = models.AutoField(primary_key=True, verbose_name="Table Id")
-    course_id = models.CharField(max_length=255, blank=True, null=True, verbose_name="Course Id")
-    user_id = models.CharField(max_length=255, blank=True, null=True, verbose_name="User Id")
+    course_id = models.BigIntegerField(blank=True, null=True, verbose_name="Course Id")
+    user_id = models.BigIntegerField(blank=True, null=True, verbose_name="User Id")
```
pushyamig: The user default piece is broken due to this change. the `user_id` is actually the sis_user_name not the user_id. So this still needs to be `CharFiled`. due to the confusion in the naming convention, I felt the column name needs to be changed. Then it will be a new migration file will be generated something like this 
```
class Migration(migrations.Migration):

    dependencies = [
        ('dashboard', '0005_auto_20190613_1502'),
    ]

    operations = [
        migrations.AddField(
            model_name='userdefaultselection',
            name='user_sis_name',
            field=models.CharField(blank=True, max_length=255, null=True, verbose_name='User SIS name'),
        ),
        migrations.RemoveField(
            model_name='userdefaultselection',
            name='user_id',
        ),
        migrations.AlterUniqueTogether(
            name='userdefaultselection',
            unique_together=set([('user_sis_name', 'course_id', 'default_view_type')]),
        ),
    ]
```
 I can commit both the files change and the new migration change if you want me to.

pushyamig:
I am getting permission denied error when committing to this branch, SO I am going to merge this can create a new PR

PR #544, opened 2019-06-18
Lecture Capture Compatibility

### LINKED ISSUE #532: add more "Resource Access" data into File Access view

vjcao:
Adds backend functionality to be compatible with lecture capture data. This PR redid #534 to be compatible with the React Branch. Closes #532.


pushyamig:
you should always keep 2 different issues (Lecture capture data and Footer changes) in separate PR.

zqian:
@vjcao see if you can address the Codacy review issue first.

dashboard/settings.py
```diff
@@ -32,6 +32,10 @@
 # Google Analytics ID
 GA_ID = config('GA_ID', default='')
 
+# File values from .env
+CANVAS_FILE = config('CANVAS_FILE', default='0')
+LECCAP_FILE = config('LECCAP_FILE', default='1')
```
pushyamig: What is the purpose of having these as environmental variables?

assets/src/containers/FilesAccessed.js
```diff
@@ -189,9 +200,23 @@ function FilesAccessed (props) {
               onWeekChange={onWeekChangeHandler}
             /> : ''}
             <div className={classes.formController}>
-              <p>File accessed from
+              <FormControl className={classes.formControl}>
```
pushyamig: From the UI I feel the types of data need not be a dropdown but I think a cluster of checkboxes would be appropriate in case if we need to have more data sources like `kaltura`/ `runstone` etc. I think we need to check with Jennifer love how the UI should look like and function

assets/src/containers/FilesAccessed.js
```diff
@@ -189,9 +200,23 @@ function FilesAccessed (props) {
...
+                <Select
+                  value={fileFilter}
+                  onChange={onChangeFileHandler}
+                  inputProps={{
+                    name: 'file',
+                    id: 'file-type',
+                  }}
+                >
+                  <MenuItem value="all_files">All</MenuItem>
```
zqian: Can we make this list configurable from settings file? Not only choice values, but also labels.

dashboard/settings.py
zqian: @pushyamig yes. I think it is better to config from the settings file, so that we can generate the list dynamically via configuration.

dashboard/views.py
```diff
@@ -224,6 +232,14 @@ def file_access_within_week(request, course_id=0):
             else:
                 output_df=output_df.drop([i_grade], axis=1)
 
+
+    if (file_type != "all_files"):
+        # drop all other files
+        file_types = [CANVAS_FILE, LECCAP_FILE]
```
zqian: again, you need to consider how to make the file type list dynamic. We do not want to hard code the file type here

dashboard/views.py
```diff
@@ -235,8 +251,8 @@ def file_access_within_week(request, course_id=0):
     output_df.fillna(0, inplace=True) #replace null value with 0
 
     output_df['file_id_part'], output_df['file_name_part'] = output_df['file_id_name'].str.split(';', 1).str
-    output_df['file_name'] = output_df.apply(lambda row: CANVAS_FILE_PREFIX + row.file_id_part + CANVAS_FILE_POSTFIX + CANVAS_FILE_ID_NAME_SEPARATOR + row.file_name_part, axis=1)
-    output_df.drop(columns=['file_id_part', 'file_name_part', 'file_id_name'], inplace=True)
+    # uses canvas url if file type == 0 (cavas file) and leccap url if not
+    output_df['file_name'] = output_df.apply(lambda row: CANVAS_FILE_PREFIX + row.file_id_part + CANVAS_FILE_POSTFIX + CANVAS_FILE_ID_NAME_SEPARATOR + row.file_name_part if row.file_type == 0.0 else LECCAP_FILE_PREFIX + row.file_id_part + CANVAS_FILE_ID_NAME_SEPARATOR + row.file_name_part, axis=1)
```
zqian: I am not sure whether the following pattern work for Lecture Capture file urls. Have you confirmed that?

> LECCAP_FILE_PREFIX + row.file_id_part + CANVAS_FILE_ID_NAME_SEPARATOR + row.file_name_part

vjcao: Yes, I have confirmed that the url format for lecture capture works (at least for the classes I have taken that have lecture captures)

assets/src/containers/FilesAccessed.js
vjcao: I will try to fix this

dashboard/views.py
vjcao: I will try to fix this

dashboard/settings.py
jonespm: I think we're going to have to have a more complex settings file for this feature in the future. The change I mentioned Vincent too had the configuration loaded up from a JSON file. I don't think we'll be able to "easily" do all of the configuration via environment. 

I was even thinking of switching this project to use a JSON configuration rather than the environment so we can have more complex and secure configurations. We could load up the JSON in Openshift via a secret. I just haven't had time to switch over to that.

JSON is the way that configuration is done in Google App Engine.

assets/src/containers/FilesAccessed.js
jonespm: Yeah, I think that would be pretty neat. It looks like there's a few multiselects out there for react
https://jedwatson.github.io/react-select/
https://github.com/Khan/react-multi-select

dashboard/views.py
```diff
@@ -147,7 +154,7 @@ def file_access_within_week(request, course_id=0):
 
     # get time range based on week number passed in via request
 
-    sqlString = f"""SELECT a.file_id as file_id, f.name as file_name, u.current_grade as current_grade, a.user_id as user_id
+    sqlString = f"""SELECT a.file_id as file_id, f.file_type as file_type, f.name as file_name, u.current_grade as current_grade, a.user_id as user_id
```
pushyamig: This query currently throwing an error since `file_type` is not actually a DB column. The file DB schema needs to change  to fit this change. Django migrations need to be applied

assets/src/containers/FilesAccessed.js
pushyamig: React being using Material-UI components. we could leverage this https://material-ui.com/components/checkboxes/

dashboard/settings.py
zqian: Mark this as resolved for now.

PR #545, opened 2019-06-19
Adding copyright text using django flatpages

pushyamig:
solves #523 
- This PR solves the issue of adding Copyright information in the footer using Django [flatpages](https://docs.djangoproject.com/en/1.11/ref/contrib/flatpages/) from Django admin view.
- The requisite seems to agree on a flatpage url currently I have it as `copyright`
- Django Flatpages can provision creating static pages like `/about/` but I took the approach of getting the flatpage content to be populated inside of the template. More info [here](https://docs.djangoproject.com/en/1.11/ref/contrib/flatpages/#getting-a-list-of-flatpage-objects-in-your-templates)
- Currently, I have the instance of this code running on https://dev-myla.tl.it.umich.edu/ for review process.

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,6 +144,12 @@ This will remove everything! (images, containers, volumes)
 
 *Docker stores MySQL data locally in the directory `.data`. If you want to fully clean you'll have to remove this folder.*
 
+## Populating Copyright information in footer
+1. Since MyLA can be used by multiple institution, copyright information needs to be entitled to institutions needs. 
+2. Django Flatpages serves the purpose. The display of the copyright content can be controlled from the Django Admin view. 
+3. The url for configuring copyright info must be `/copyright/` since that is used in the `base.html` for pulling the info
+[More info](https://simpleisbetterthancomplex.com/tutorial/2016/10/04/how-to-use-django-flatpages-app.html)
+
 ## Testing tips!
 
 1. Connect to the docker and edit some files!
--- a/dashboard/settings.py
+++ b/dashboard/settings.py
@@ -31,6 +31,8 @@ LOGIN_URL = '/accounts/login'
 
 # Google Analytics ID
 GA_ID = config('GA_ID', default='')
+
+# This is required by flatpages flow. For Example Copyright information in the footer populated from flatpages
 SITE_ID = 1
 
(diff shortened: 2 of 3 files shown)


dashboard/settings.py
```diff
@@ -31,6 +31,7 @@
 
 # Google Analytics ID
 GA_ID = config('GA_ID', default='')
+SITE_ID = 1
```
jonespm: What's this value? I don't see it used.

pushyamig: The documentation says it is needed. since we don't run multiple apps in Django the default seems to be 1. https://simpleisbetterthancomplex.com/tutorial/2016/10/04/how-to-use-django-flatpages-app.html

dashboard/templates/base.html
```diff
@@ -57,7 +57,13 @@
     {% endblock %}
     {% csrf_token %}
     <footer>
-        <div style="float: left; width: 50%">Copyright © 2018 The Regents of the University of Michigan</div>
+        {% load flatpages %}
+        {% get_flatpages as flatpages %}
```
jonespm: It looks like this could be a bit simpler if you did this
```
{% get_flatpages '/copyright/' as pages %}
<div style="float: left; width: 50%">{{ pages.first.content|safe }}</div>

```
ref: https://stackoverflow.com/questions/33767836/using-a-specific-flatpage-in-a-template

dashboard/settings.py
jonespm: Cool, could you put a comment here that SITE_ID is needed for flatpages so we remember?

pushyamig: Done

README.md
```diff
@@ -144,6 +144,12 @@ This will remove everything! (images, containers, volumes)
 
 *Docker stores MySQL data locally in the directory `.data`. If you want to fully clean you'll have to remove this folder.*
 
+## Populating Copyright information in footer
```
pushyamig: Added the documentation in the README so people will know to wire this piece when running for their institution

dashboard/templates/base.html
pushyamig: Done

PR #551, opened 2019-06-21
Correctly route to index page when app bar name is clicked

justin0022:
When `My Learning Analytics: {Course Name}` is clicked, it routes to `localhost:5001/{course code}` instead of `localhost:5001/courses/{course code}`. This PR resolves this issue.

### DIFF
--- a/assets/src/containers/DashboardAppBar.js
+++ b/assets/src/containers/DashboardAppBar.js
@@ -64,7 +64,7 @@ function DashboardAppBar (props) {
           <Button>
             <Typography variant='h6' color='inherit' className={classes.grow}>
               <Link
-                to={{ pathname: `/${courseId}` }}
+                to={{ pathname: `/courses/${courseId}` }}
                 className={classes.homeButton}>
                   My Learning Analytics: {courseName}
               </Link>


assets/src/containers/DashboardAppBar.js
```diff
@@ -64,7 +64,7 @@ function DashboardAppBar (props) {
           <Button>
             <Typography variant='h6' color='inherit' className={classes.grow}>
               <Link
-                to={{ pathname: `/${courseId}` }}
+                to={{ pathname: `/courses/${courseId}` }}
                 className={classes.homeButton}>
                   My Learning Analytics: {courseName}
```
pushyamig: Just a thought, what do you think of the idea of making `My Learning Analytics` as a another link and clicking on it goes to `/courses/` page ( which is the list of user courses in Myla) and the `{course_name}` part separate link clicking on it goes to `/courses/{courseId}` ( which is the list if views in a course)

justin0022: That could work, although then we should have two buttons, one for the `My Learning Analytics` and another for the course name. 

Alternatively, we could add an option in the Avatar Modal component that has a way for users to switch between courses.

pushyamig: Either way is fine. You can open a separate issue for that fix later. 
LGTM

justin0022: Okay great.

PR #555, opened 2019-06-25
Fixes #546 - Switching environment over to json from environment file

### LINKED ISSUE #546: Change configuration to load via JSON rather than Environment
We've been using Python Decouple to load values from the environment rather than a configuration file. There's three (or more) problems with this

- The python decouple library hasn't been updated in a year
- This library is limited in what we can define and how we define it as environment variables are limited
- There are some [security concerns](http://movingfast.io/articles/environment-variables-considered-harmful/) about storing your values in environment variables

So I think we should move all of these environment into a json based configuration file. This will stay updated (built-in) has less limitations for what we can define and more secure as we can define them on Openshift.

I had a WIP PR where I started to create a configuration for a separate feature that really was easier defined in JSON. I'm thinking now this should be the file for everything.

We noticed doing this would allow us to easier configure the "cron" pod and remove environment variables from this, that would be really nice and prevent accidentally committing any secrets. 

https://github.com/jonespm/student-dashboard-django/commit/f1667f5fec69308f57fdd5219fdce979f1359468

jonespm:


jonespm:
Do you know if any of these environment variables were used directly by the react frontend? I didn't check the react branch. I'm expecting a few new variables were added but hopefully nothing not too many conflicts.

andrew-gardener:
react should only be using variables passed via view template (here: https://github.com/tl-its-umich-edu/my-learning-analytics/blob/react/dashboard/templates/base.html#L33) so should be fine. Also I don't think any new environment variables were added with React

dashboard/common/utils.py
```diff
@@ -1,14 +1,13 @@
...
+from django.conf import settings
 
 def get_build_info():
     logger.debug(get_build_info.__name__)
-    git_commit=config("OPENSHIFT_BUILD_COMMIT",default="")
-    build_namespace=config("OPENSHIFT_BUILD_NAMESPACE",default="")
-    git_branch=config("OPENSHIFT_BUILD_REFERENCE",default="master")
-    build_source=config("OPENSHIFT_BUILD_SOURCE",default="")
-    build_name=config("OPENSHIFT_BUILD_NAME",default="")
+    git_commit=settings.ENV.get("OPENSHIFT_BUILD_COMMIT", "")
```
jonespm: I think these are really in the environment so I'm going to have to switch this to an os.getenv.

dashboard/settings.py
```diff
@@ -338,43 +338,43 @@
     LOGIN_REDIRECT_URL = '/'
 
 # Give an opportunity to disable LTI
-if config('STUDENT_DASHBOARD_LTI', default='False', cast=bool):
+if ENV.get('STUDENT_DASHBOARD_LTI', False):
     INSTALLED_APPS += ('django_lti_auth',)
 
-    PYLTI_CONFIG = {
-        "consumers": json.loads(config("PYLTI_CONFIG_CONSUMERS", default="{}", cast=str)),
+    PYLTI_ENV.get = {
```
pushyamig: My build is failing at this line
```
File "/code/dashboard/settings.py", line 345, in <module>
    "consumers": json.loads(ENV.get("PYLTI_ENV.get_CONSUMERS", {})),
  File "/usr/local/lib/python3.6/json/__init__.py", line 348, in loads
    'not {!r}'.format(s.__class__.__name__))
TypeError: the JSON object must be str, bytes or bytearray, not 'dict'
```

pushyamig: pylint is showing an error on that line
Screen Shot 2019-06-26 at 8 22 32 AM

jonespm: Okay latest update should fix this. Thanks!

jonespm:
FYI I wrote a utility here to hopefully  save you some time if you already had a .env file.

`python env_to_json.py > config/env.json
`

pushyamig:
I see this error now, my `env.json` has an empty value for it 
"PYLTI_CONFIG_CONSUMERS": "",
```
File "/code/dashboard/settings.py", line 345, in <module>
    "consumers": json.loads(ENV.get("PYLTI_CONFIG_CONSUMERS", {})),
  File "/usr/local/lib/python3.6/json/__init__.py", line 354, in loads
    return _default_decoder.decode(s)
  File "/usr/local/lib/python3.6/json/decoder.py", line 339, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/usr/local/lib/python3.6/json/decoder.py", line 357, in raw_decode
    raise JSONDecodeError("Expecting value", s, err.value) from None
```

jonespm:
@pushyamig Did you have `STUDENT_DASHBOARD_LTI=True`? I didn't test this setting. I'll set this and try it.

andrew-gardener:
`"consumers": json.loads(ENV.get("PYLTI_CONFIG_CONSUMERS", {})),` 
should be 
`"consumers": json.loads(ENV.get("PYLTI_CONFIG_CONSUMERS", "{}")),` 
for it to work as a string.

Since env is a json file now we can change it to be `"consumers": ENV.get("PYLTI_CONFIG_CONSUMERS", {}),` and have `PYLTI_CONFIG_CONSUMERS` be the nested json lti settings directly instead of a string though

pushyamig:
@jonespm I have that setting to false

dashboard/settings.py
```diff
@@ -338,43 +338,43 @@
     LOGIN_REDIRECT_URL = '/'
 
 # Give an opportunity to disable LTI
-if config('STUDENT_DASHBOARD_LTI', default='False', cast=bool):
+if ENV.get('STUDENT_DASHBOARD_LTI', False):
     INSTALLED_APPS += ('django_lti_auth',)
 
-    PYLTI_CONFIG = {
-        "consumers": json.loads(config("PYLTI_CONFIG_CONSUMERS", default="{}", cast=str)),
+    PYLTI_CONFIG.get = {
```
andrew-gardener:
```
PYLTI_CONFIG.get = {
        "consumers": json.loads(ENV.get("PYLTI_CONFIG_CONSUMERS", {})),
```
to
```
PYLTI_CONFIG = {
        "consumers": ENV.get("PYLTI_CONFIG_CONSUMERS", {}),
```

config/env_sample.json
```diff
@@ -0,0 +1,127 @@
...
+    "CANVAS_FILE_POSTFIX": "/download?download_frd=1",
+    "/*  Paths to special login credentials (uses SAML)": "*/",
+    "GOOGLE_APPLICATION_CREDENTIALS": "/secrets/bq_cred.json",
+    "/*  ed_app field to look for in Big Query store": "*/",
+    "BIG_QUERY_ED_APP": "http://umich.instructure.com/",
+    "/*  LTI Configuration": "*/",
+    "/*  LTI is disabled by default": "*/",
+    "STUDENT_DASHBOARD_LTI": "False",
+    "/*  list LTI consumers as per https://github.com/rohitjose/django-lti-authusage": "*/",
+    "/*  ex: PYLTI_CONFIG_CONSUMERS={ \"LTI_CONSUMER_KEY_1\": { \"secret\": \"LTI_CONSUMER_SECRET_1\" } }": "*/",
```
andrew-gardener:
```
"/*  ex: PYLTI_CONFIG_CONSUMERS={ \"LTI_CONSUMER_KEY_1\": { \"secret\": \"LTI_CONSUMER_SECRET_1\" } }": "*/",
```
to
```
"/*  ex: "PYLTI_CONFIG_CONSUMERS": { "LTI_CONSUMER_KEY_1": { "secret": "LTI_CONSUMER_SECRET_1" } }": "*/",
```

README.md
```diff
@@ -5,11 +5,15 @@
...
+There is some environment configuration for docker-compose on local host and addtional files needed for this app to run. Most of these files are in config directory.
 
 Configuration:
 
-Copy the file `.env.sample` to `.env` and fill in with the values for local testing. You may also provide a .env file if you have one.
+- Copy the file `.env.sample` to `.env` and fill in with the values for local testing. You may also provide a .env file if you have one.
+
+- You also need to copy the file config/env_sample.json to config/env.json. This is where most of the configuration is done now. 
+
+- There is a utility "env_to_json.py" to help convert your configuration if you were using this prior to JSON configuration.
```
zqian: may be helpful to include the following command line here, too:

python env_to_json.py > config/env.json

dashboard/settings.py
```diff
@@ -25,35 +23,37 @@
...
 # Google Analytics ID
-GA_ID = config('GA_ID', default='')
+GA_ID = ENV.get('GA_ID', '')
 
 # Quick-start development settings - unsuitable for production
 # See https://docs.djangoproject.com/en/1.9/howto/deployment/checklist/
 
 # SECURITY WARNING: keep the secret key used in production secret!
-SECRET_KEY = config('DJANGO_SECRET_KEY')
+SECRET_KEY = ENV.get('DJANGO_SECRET_KEY')
```
zqian: default value needed?

andrew-gardener:
Oh, @jonespm one little problem with the `env_to_json.py` script is that it converts True/False into strings "True"/"False" so checks like `if ENV.get('STUDENT_DASHBOARD_SAML', True):` are true for me even though I had `False` before (might be what happened to @pushyamig )

dashboard/settings.py
jonespm: Well it won't start up unless you have one in the environment file! I think Django failing to start because the secret_key isn't defined is okay.

README.md
jonespm: Thanks, I'll add that to the README. There shouldn't be that many people that need this but it was useful for me to test a few ideas out before going down this path.

config/env_sample.json
jonespm: Thanks @andrew-gardener , I fixed this up.

jonespm:
@andrew-gardener Hmm! I was hoping because of what the [decoder said about JSON](https://docs.python.org/3.4/library/json.html#encoders-and-decoders) that it wouldn't be a problem. 

But I guess I should have done a simple test
```
json.loads('{"abc":"false"}')
{'abc': 'false'}
```

JSON *is* nice in that if you have the correct type (integer, boolean, whatever) than it does come into Python right, but looks like strings aren't being "decoded" automatically. I have the env_to_json.py convert integers but not Booleans. Maybe that's the better fix, to do it there. The sample is wrong then too.

jonespm:
Okay I started it up with the sample now and it works.

Also this json *probably* isn't optimal since it was converted from the environment file. I can see things like the database being a JSON Object instead of individual strings. Probably other things like that.

jonespm:
Also, if people don't like this format and prefer YAML (maybe because the comments aren't ideal in JSON) we could probably easily switch in the future. The work is all done to load this from a file into a dict so that wouldn't be a problem. ;)

andrew-gardener:
> Also, if people don't like this format and prefer YAML (maybe because the comments aren't ideal in JSON) we could probably easily switch in the future. The work is all done to load this from a file into a dict so that wouldn't be a problem. ;)

Oh yes, supporting both json and yaml would be great to have!

dashboard/settings.py
```diff
@@ -25,35 +23,37 @@
 PROJECT_ROOT = os.path.abspath(
     os.path.join(os.path.dirname(__file__), ".."),
 )
+with open(os.getenv("ENV_FILE", "/code/config/env.json")) as f:
```
pushyamig: In OpenShift environment we might need to configure secrets added to that path since the path is Hardcoded

jonespm: Yeah that's what this ENV_FILE environment variable is for. We can change the location of this json file with an environment variable. 

There are still going to be a few environment variables such as the openshift ones (about the build) and MYSQL needs them for the MYSQL Docker image, but not many anymore.

pushyamig: I missed that part, sorry!

PR #559, opened 2019-06-26
Fixes #558 build fails if no env.json set

### LINKED ISSUE #558: Build fails if no env.json set
The build fails if there's no env.json, but this is really only needed for regular operation. However manage.py is run during the build to do a few things and should still work.

jonespm:


jonespm:
Because of this change old env values still work and the app still might start up if you don't even change anything.

Though many future values we're planning on supporting are going to fail.

PR #565, opened 2019-07-08
Resource Access View

vjcao:
Added a checkbox filter to filter resource types in the Resource Access View (File Access). This should close issue #532.

Screen Shot 2019-07-08 at 10 36 52 AM


pushyamig:
One thing that is needed to commit the issue is the `cron.py` needs the query to pull the  Leccap data.  one idea is to get the different media data that is relevant to an institution from a json file. The work has been started during the recent Unizin hackathon at Florinda. If that is committed as part of this PR or separate PR will complete the #532 @jonespm Thoughts??

.env.sample
```diff
@@ -20,6 +20,23 @@ MYSQL_ROOT_PASSWORD=student_dashboard_root_pw
 # This is built into the Docker Image
 TZ=America/Detroit
 
+<<<<<<< HEAD
```
justin0022: I think there's a merge conflict here.

pushyamig:
I see that when no resource_type is selected it is making a backend call
`week_num_start=1&week_num_end=16&grade=90-100&resource_type=` this should not be allowed

assets/src/containers/FilesAccessed.js
```diff
@@ -115,10 +119,24 @@ function FilesAccessed (props) {
...
       } else {
         // setting it to default
         setGradeRangeFilter('All')
+        let tempArray = []
+        resource_values.forEach(function(resource_item) {
+          if (resource_item.disabled === "false") {
+            tempArray.push(resource_item.resource_value)
+          }
+        })
+        setResourceFilter(resourceFilter.concat(tempArray))
```
pushyamig: Code duplication should be avoided. Take the code as a separate function and reuse respective places

vjcao: Ok I will fix the code duplication. For when no resource type is selected, I thought it would make sense to update the chart to show no resources because no resource types are selected.

pushyamig: yes, but if you know from UI they did not select any type then API call should not be allowed as you backend endpoint will fetch no data. We should preserve DB call as much as possible. I think when you see no resource type is selected, display a message saying "please select a resource type for data display"

vjcao: Screen Shot 2019-07-10 at 10 59 27 AM

Should I center the message?

pushyamig: No need to center but need to make it effective for the user to see. Right now it blends in with rest of the text and the info text might miss the attention. Try and see how can you make it effective for viewable

pushyamig:
The design of the view should be such that only Backend cron needs to know what type of data to pull like `Kaltura`, `leccap`, `files` etc. the front should adapt

pushyamig:
@vjcao, I accidentally closed the PR

dashboard/common/utils.py
```diff
@@ -1,6 +1,14 @@
-import logging, os
+import logging, os, json
 logger = logging.getLogger(__name__)
 
+try:
```
jonespm: ENV is already available as settings.ENV, just 

`from django.conf import settings
`
and use settings.ENV. You don't have to re-read this file here.

dashboard/common/utils.py
```diff
@@ -10,3 +18,18 @@ def get_build_info():
...
+    return resource_types
+
+def get_resource_urls():
+    resource_urls = ENV.get("RESOURCE_URLS")
+    return resource_urls
+
+def get_resource_values():
+    resource_types = get_resource_types()
+    resource_values = []
+    for dict_value in resource_types.values():
```
jonespm: You don't need this loop or any of these temporary values.
You can should be able to just return from the first line

return get_resource_types().values()

Because of this is actually used in settings, I'd just move all this code to settings, you don't really need these methods here, just make them variables in settings

Cant just just remove the code here and change move it into settings as:
```
RESOURCE_TYPES = ENV.get("RESOURCE_TYPES")
RESOURCE_URLS = ENV.get("RESOURCE_URLS")
RESOURCE_VALUES = RESOURCE_TYPES.values() 
```

dashboard/models.py
```diff
@@ -219,7 +219,8 @@ def json(self, include_id=True):
 
 
 class File(models.Model):
-    id = models.BigIntegerField(primary_key=True, verbose_name="File Id")
+    resource_type = models.TextField(verbose_name="Resource Type")
```
jonespm: resource_type can probably just be a CharField as it's going to be small. Also will change in the migration.

dashboard/views.py
```diff
@@ -30,6 +34,10 @@
 GRADE_LOW="low_grade"
 NO_GRADE_STRING = "NO_GRADE"
 
+# string for file type
+RESOURCE_TYPE_STRING = "resource_type"
+RESOURCE_TYPES = utils.get_resource_types()
```
jonespm: If you made the settings change I mention above I'd just use the settings.RESOURCE_TYPES directly.

assets/src/containers/FilesAccessed.js
```diff
@@ -41,12 +43,14 @@ const settingNotUpdated = 'Setting not updated'
 
 function FilesAccessed (props) {
   const { classes, courseInfo, courseId } = props
+  const resource_values = RESOURCE_VALUES
```
justin0022: In JS the convention is to favour camelcase, so I think `resourceValues` is a better variable name here.

assets/src/containers/FilesAccessed.js
```diff
@@ -56,6 +60,16 @@ function FilesAccessed (props) {
   const [defaultCheckboxState, setDefaultCheckedState] = useState(true)
   const [defaultLabel, setDefaultLabel] = useState(currentSetting)
 
+  function setDefaultFilterState() {
```
justin0022: Could you explain what this function is doing? The `set` prefix suggests setting some state, but I don't think it is setting any state. It may be worth renaming. 

I also think it can be refactored to be shorter and more readable:

```js
function setDefaultFilterState(resourceValues) {
   return resourceValues.filter(resourceItem => resourceItem.disabled === 'false')
}
```

assets/src/containers/FilesAccessed.js
```diff
@@ -127,8 +143,8 @@ function FilesAccessed (props) {
 
   useEffect(() => {
     // Fetch data once all the setting data is fetched
-    if (dataControllerLoad === 2) {
-      const dataURL = `/api/v1/courses/${courseId}/file_access_within_week?week_num_start=${weekRange[0]}&week_num_end=${weekRange[1]}&grade=${gradeRangeFilter}`
+    if (dataControllerLoad === 2 && resourceFilter != "") {
```
justin0022: `!=` should be changed to `!==`, to avoid [type coercion](https://stackoverflow.com/questions/19915688/what-exactly-is-type-coercion-in-javascript)

assets/src/containers/FilesAccessed.js
```diff
@@ -160,9 +179,27 @@ function FilesAccessed (props) {
     }
   }
 
+  const onChangeFileHandler = event => {
+    const value = event.target.value
+    if (event.target.checked == true && !resourceFilter.includes(value)) {
```
justin0022: `event.target.checked == true` is redundant, it should just be `event.target.checked`

assets/src/containers/FilesAccessed.js
```diff
@@ -160,9 +179,27 @@ function FilesAccessed (props) {
...
+      setResourceFilter([...resourceFilter, value])
+    } 
+    else if (event.target.checked == false) {
```
justin0022: should just be `!event.target.checked`

assets/src/containers/FilesAccessed.js
```diff
@@ -160,9 +179,27 @@ function FilesAccessed (props) {
     }
   }
 
+  const onChangeFileHandler = event => {
+    const value = event.target.value
+    if (event.target.checked == true && !resourceFilter.includes(value)) {
+      setResourceFilter([...resourceFilter, value])
+    } 
+    else if (event.target.checked == false) { 
+      setResourceFilter(resourceFilter.filter(val => {
```
justin0022: should be `resourceFilter.filter(val => val !== value)`

assets/src/containers/FilesAccessed.js
```diff
@@ -160,9 +179,27 @@ function FilesAccessed (props) {
...
+        return val != value
+      }))
+    }
+    
+  }
+
   const FileAccessChartBuilder = (fileData) => {
     if (!fileData || Object.keys(fileData).length === 0) {
-      return (<p>No data provided</p>)
+      if (resourceFilter == "") {
```
justin0022: Empty quotes are considered falsy in JavaScript, so this can simply be `if (!resourceFilter)`

assets/src/containers/FilesAccessed.js
vjcao: Oh, this function creates an array that will be used as the default state of the `resourceFilter` state. Thanks for the suggestions.

assets/src/containers/FilesAccessed.js
vjcao: Didn't work properly

justin0022: Hmm strange. Maybe just use `resourceFilter === ''` instead.

vjcao: I did `if (resourceFilter <= [])`. Is this fine?

pushyamig: less than operator should be used for comparing the number and in this case, it is comparing with an empty list. I don't think that makes sense

vjcao: I can't do `resourceFIlter == []` or `resourceFilter === []` since they always evaluate to false. Is there another way?

pushyamig: Try `resourceFilter.length == 0` ?

vjcao: Thanks

pushyamig:
Some of the naming conventions in Frontend should be renamed to Resource Accessed. For eg., `FileAccessed.js` Doesn't make sense anymore. I think it should be renamed to `ResourceAccessed.js', It is better to use the IDE refactoring feature while doing it as it might refactor at various places. Also, change all the variables names within the `FileAccesses.js' that has "files" reference in them. When you make this change just make sure to test the overall flow so that you still navigate to the resource view as per the website flow

jonespm:
@vjcao If you can make the requested changes, I'll take your changes and put them in a new branch as we talked about this morning with my backend changes and we can work on them there.

data/resource.csv
```diff
@@ -0,0 +1,2 @@
+ID,NAME
+<RESOURCE_ID>,<RESOURCE_NAME>
```
pushyamig: This `data/` folder is redundant you can remove it and all the files inside it. This was never cleaned up when designing the files view

Issue #575, opened 2019-07-15
MyLA home page: Remove "the welcome page" from the page title


jennlove-um:
Change the header on the MyLA home page to read "My Learning Analytics" instead of "My Learning Analytics the welcome page."

jennlove-um:
Testing passes.

PR #583, opened 2019-07-15
Removed "the welcome page"

justin0022:
Addresses #575


pushyamig:
@justin0022 Can you point the PR against master since react branch changes are merged to master? I will be deleting the react branch

Issue #574, opened 2019-07-15
Course home page: Add message when no visualizations are set up for a course


jennlove-um:
On the course home page, if no visualizations at all have been added for a course, display a message to the user: "No data visualizations have been added for this course."

jennlove-um:
Testing passes.

PR #588, opened 2019-07-15
Added error message if no views are active

justin0022:
Addresses #574

### DIFF
--- a/assets/src/containers/IndexPage.js
+++ b/assets/src/containers/IndexPage.js
@@ -3,7 +3,7 @@ import { withStyles } from '@material-ui/core/styles'
 import Grid from '@material-ui/core/Grid'
 import SelectCard from '../components/SelectCard'
 import { Link } from 'react-router-dom'
-import { isObjectEmpty } from '../util/object'
+import { isObjectEmpty, getObjectValues } from '../util/object'
 import Error from './Error'
 import routes from '../routes/routes'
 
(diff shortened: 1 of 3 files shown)


assets/src/containers/IndexPage.js
```diff
@@ -24,6 +26,9 @@ const styles = theme => ({
 function IndexPage (props) {
   const { classes, courseInfo, courseId } = props
 
+  if (isObjectEmpty(courseInfo.course_view_options))
+    return (<Error>No data visualizations have been added for this course.</Error>)
```
pushyamig: the `course_view_options` is not empty when all the views are disabled. This is how the part of Json looks like `"course_view_options": {"fa": 0, "ap": 0, "gd": 0}}`. This info is pulled from the DB.  The code never enter this case
There is error case(when DB look up failed) the Json might looks like this`"course_view_options": ""` 
I think it needs to support both the cases.

justin0022: Ah good point, okay, I'll fix this.

justin0022:
@pushyamig I now check to see if every value of the `course_view_options` are 0 before displaying the message.

pushyamig:
>@pushyamig I now check to see if every value of the course_view_options are 0 before displaying the message.

Really Nice you have added the test. This looks good!
I feel how to run the test from the command line was never added to README. It will be useful to have the info there so that encourages people to add more tests.

I feel the Backend should also start adding testing as Frontend does :(

PR #590, opened 2019-07-16
Fixes #589 switch wait-port over to netcat

### LINKED ISSUE #589: Switch wait-port over to netcat
Wait port isn't working with some npm change, we really don't need it can can just switch to netcat

jonespm:
The wait-port wasn't working anymore to stop the startup until database starts up, seems easier to just drop that and use netcat.


jonespm:
I'm just going to merge this, I don't think it's very contentious. This works and currently without it unless the database is running the app encounters errors.

Issue #552, opened 2019-06-21
Allow users to change course in avatar modal


justin0022:
There should be a way for users to change courses in the dropdown avatar modal. It can simply link to `/courses/`.

pushyamig:
I will remove this from the react project as this can also be worked upon after react branch is merged to master. I will label it to react.

pushyamig:
Testing: inserting multiple users flow
Just select a user of your choice like `select * From user where sis_name='dssss';`
Have an insert command
`insert into user values(null,17700000000111111, 'Test user', 01111311, 'dssss, 1770000000011333, 89.8,90,'StudentEnrollment'`

Delete the row afterword by the id of the rows `delete from user where id = 1223`

jennlove-um:
@pushyamig I get an SQL syntax error when I try to do the insert to test this

jennlove-um:
This works for students in multiple courses.  From within a course visualization, click on the profile gray circle, on the drop-down, click switch courses, then select another course to go to.

PR #597, opened 2019-07-17
Avatar change course

justin0022:
Addresses #552.

To test, the user that is logged in must be in more than one course. Replace the `user` object in `App.js` with the following to test:

```js
const user = Object.freeze({
  admin: myla_globals.is_superuser,
  enrolledCourses: [
    { course_id: 231768, course_name: 'EECS 808 008 FA 2057' },
    { course_id: 123456, course_name: 'Some Other Course' },
    { course_id: 678901, course_name: 'Yet Another Course' }
  ],
  isSuperuser: myla_globals.is_superuser,
  isLoggedIn: !!myla_globals.username
})
```

### DIFF
--- a/assets/src/components/AvatarModal.js
+++ b/assets/src/components/AvatarModal.js
@@ -9,6 +9,8 @@ import List from '@material-ui/core/List'
 import ListItem from '@material-ui/core/ListItem'
 import ListItemIcon from '@material-ui/core/ListItemIcon'
 import ListItemText from '@material-ui/core/ListItemText'
+import DialogTitle from '@material-ui/core/DialogTitle'
+import Dialog from '@material-ui/core/Dialog'
 import LogoutIcon from '@material-ui/icons/ExitToApp'
 import HelpIcon from '@material-ui/icons/HelpOutline'
 import Lock from '@material-ui/icons/Lock'
@@ -25,7 +27,7 @@ const styles = theme => ({
     marginTop: '10px',
     marginBottom: '10px',
     marginLeft: 'auto',
-    marginRight: 'auto',
+    marginRight: 'auto'
   },
   link: {
     textDecoration: 'none',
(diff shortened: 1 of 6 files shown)


justin0022:
Note that changing courses redirects to the Index Page, instead of staying on the current view. For instance, if user is on the grades view for courseID: 1234, and the user tries to switch to courseID: 5678, the user will be on the Index Page instead of the grades view for 5678. 

Preserving the view that the user is on is a little tricky because it's possible that the view for the new course is disabled. If there's a need for allowing users to switch courses quickly between the same view, it'll take some work but could be done.

assets/src/components/AvatarModal.js
```diff
@@ -84,18 +132,12 @@ function AvatarModal(props) {
...
-                      <ListItemText inset primary='Admin' />
-                    </ListItem>
-                  </Link>
-                  <Divider />
-                </>
+                ? Admin
+                : null
+            }
+            {
+              user.enrolledCourses.length > 1
```
justin0022: I only show the `SwitchCourses` component if the user is enrolled in more than one course. However, if they are on course 1234, and they click to switch to course 1234, the entire app reloads instead of simply closing the dialog box. Again, this can be addressed, just takes a bit more work.

assets/src/containers/App.js
```diff
@@ -1,32 +1,45 @@
...
 
+const enrolledCourses = (myla_globals.user_courses_info.length !== 0)
+  ? JSON.parse(myla_globals.user_courses_info)
+  : ''
+
+/*
+Frozen to prevent unintentional changes to this object. This object is strictly readonly.
+myla_globals should ONLY be accessed in App.js, and nowhere else.
+*/
+const user = Object.freeze({
```
justin0022: I decided that it is preferable to access the `myla_globals` variable just from the `App.js` file, and created a frozen object that can't be modified accidentally by some child component.

assets/src/components/AvatarModal.js
pushyamig: Yeah, we might fix it later also what to get feedback from testing this feature. we haven't had a situation where  2 users are in a different courses enabled with Myla tool

pushyamig:
Also you can insert this in DB 
Just select a user of your choice like `select * From user where sis_name='dssss';
Have an insert command
`insert into user values(null,17700000000111111, 'Test user', 01111311, 'dssss, 1770000000011333, 89.8,90,'StudentEnrollment'`

Delete the row afterword by the id of the rows `delete from user where id = 1223`

pushyamig:
>Note that changing courses redirects to the Index Page, instead of staying on the current view. For instance, if user is on the grades view for courseID: 1234, and the user tries to switch to courseID: 5678, the user will be on the Index Page instead of the grades view for 5678.

Preserving the view that the user is on is a little tricky because it's possible that the view for the new course is disabled. If there's a need for allowing users to switch courses quickly between the same view, it'll take some work but could be done.

Let see how testing goes and get feedback from it. I think this should be Ok.

justin0022:
Thanks, good idea for the dB suggestion, that makes much more sense!

assets/src/components/AvatarModal.js
justin0022: Sounds good.

PR #599, opened 2019-07-17
Fixes #598 Retrieve unknown file names from Canvas

### LINKED ISSUE #598: Files branch needs to retrieve data from Canvas
With the new changes to the files branch, the file names coming back from canvas is just "attachment". This change is going to retrieve the correct names from Canvas after getting all of the caliper events, which should be correct for other events now.

jonespm:


pushyamig:
For the course `295532` the some of the files' names are still shown as `attachment` instead of the real name. This is not an issue with all the file with some files and especially with image files for this course.

jonespm:
@pushyamig Thanks! The reason this is happening is because those are hidden files that have caliper data!

So we probably should just be deleting the hidden files from the resources table then . . .

jonespm:
I pushed commit that should address the hidden file. I changed the logic to operate as I mentioned and update every file in canvas from the courses we support.

Issue #543, opened 2019-06-14
Grade Distribution: Add the median grade for the course to the display


jennlove-um:
Add the median grade under the Average Grade on the display.

jennlove-um:
Testing passes

Issue #542, opened 2019-06-14
Grade Distribution - Hide outliers on the grade distribution


jennlove-um:
We need to define the criteria for identifying outliers so we know what to hide.

jennlove-um:
At least the bottom 5 grades are binned and marked with a "< X%" gray dotted line

If all grades are 98% or above, no binning or <X% line displays.

If all binned grades are 2% or below, no <X% line displays.

Testing passes

PR #600, opened 2019-07-18
Added median grade to Grade Distribution view

justin0022:
Addresses #543

### DIFF
--- a/assets/src/containers/GradeDistribution.js
+++ b/assets/src/containers/GradeDistribution.js
@@ -34,7 +34,6 @@ function GradeDistribution (props) {
   if (loaded && isObjectEmpty(gradeData)) return (<Error>No data provided.</Error>)
 
   const buildGradeView = gradeData => {
-    console.log(gradeData)
     return (
       <Grid container>
         <Grid item xs={12} lg={2}>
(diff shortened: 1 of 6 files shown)


assets/src/containers/GradeDistribution.js
```diff
@@ -34,21 +34,30 @@ function GradeDistribution (props) {
...
+              'My Grade', <strong>{gradeData[0].current_user_grade
+                ? `${roundToOneDecimcal(gradeData[0].current_user_grade)}%`
+                : 'There are no grades yet for you in this course'}</strong>
+            ],
+            [
+              'Average Grade',
+              <strong>{roundToOneDecimcal(average(grades))}%</strong>
+            ],
+            [
+              'Median Grade',
```
pushyamig: I did not notice this earlier, BE is sending the average grade already and can also send median. I can round the decimal to one as you are doing it here. This is how it is going to look. I think UI don't need to do the calculation can simply grab the value, what do you say?
```
[  
   {  
      "current_grade":95.9,
      "current_user_grade":null,
      "graph_upper_limit":110,
      "tot_students":35,
      "grade_avg":95.0,
      "grade_median":95.9
   }
]
```

pushyamig: This is a simple change at the BE
```
 average_grade = df['current_grade'].mean().round(1)
 median_grade = df['current_grade'].median().round(1)
 df['grade_avg'] = average_grade
 df['grade_median']=median_grade
```

justin0022: I'm happy with backend doing calculations, however right now for the histogram we need the actual grades sent as well, so it's easy enough to calculate average and median. When we address #542 then we won't be sending all grades, so the backend calculation will be needed.

pushyamig: ok.

Issue #541, opened 2019-06-14
Grade Distribution - Give students the option to view or not view their grade line on the visualization.


jennlove-um:
Default to not showing the grade line
Give them something to click to view their grade

justin0022:
I'll take this on after https://github.com/tl-its-umich-edu/my-learning-analytics/pull/600 is merged.

pushyamig:
@jennlove-um does this setting needs to be persisted in the DB? as files and assignment view has their setting

jonespm:
I'd think it should be. We may need to start to think about a separate preferences page to have all user adjustable settings visible in one place.

jennlove-um:
I agree.  It should work like other student preference settings.  They should be able to turn it on or off and have it persist for the course.  Thinking ahead, this one we may want to allow students to set globally for all their courses so that they can set it once for all courses or set it differently for individual courses.

jonespm:
Yeah sounds good. This is why I think we might have to move toward a preference page. There you could set global defaults and also see site specific overrides. Could mock that up for Winter. But for now I think persisting at the site level would be fine.

pushyamig:
I think For now I agree for course level default having a preference page by click on the Avator model they could set the assignment/files/grades. 
I think this default info can be added to the  course info `/info` endpoint or have a separate endpoint which we have already but currently getting based on the file/assignment type `get_user_default_selection` 
we can generalize this look like this 
{"files":"90-80","assignments":"5","grades":1}

jennlove-um:
Testing passes: 

When a student checks or unchecks the Show My Grade option, the selection persists until they manually change it again.

PR #603, opened 2019-07-19
Show/hide my grade implemented

justin0022:
Addresses #541 

~~Currently the checkbox setting is not persisted to backend - this setting should be saved~~. I'm thinking we should build a general way to store settings for all of the visualizations, and also have a unified UI for the settings to offer a consistent experience.


justin0022:
Screen Shot 2019-07-19 at 2 43 41 PM
Screen Shot 2019-07-19 at 2 43 35 PM

pushyamig:
>I'm thinking we should build a general way to store settings for all of the visualizations, and also have a unified UI for the settings to offer a consistent experience.

currently, we are defaults for file and assignment persisted to DB in the `user_default_selection` table 
Screen Shot 2019-07-22 at 7 39 39 AM

can you elaborate on what you mean by "the general way of storing setting " and have a unified UI for setting  to offer a consistent experience"

pushyamig:
We can get the default setting info  by making this call  to BE 'get_user_default_selection`
generalize this looks like.
```
"user_defaults":{  
      "files":"90-80",
      "assignments":"5",
      "grades":1
   }
```
Or 
I can add this info to the `/info`
```
{  
   "canvas_id":295532,
   "term_id":17700000000000122,
   "name":"NURS 523 001 SS 2019",
   "term":{  
      "id":17700000000000122,
      "canvas_id":122,
      "name":"Spring/Summer 2019",
      "date_start":"2019-05-07 00:00:00+00:00",
      "date_end":"2019-12-31 00:00:00+00:00"
   },
   "current_week_number":11,
   "total_weeks":16,
   "course_view_options":{  
      "fa":1,
      "ap":1,
      "gd":1
   },
   **"user_defaults":{  
      "files":"90-80",
      "assignments":"5",
      "grades":1
   }**
}
```

justin0022:
> > I'm thinking we should build a general way to store settings for all of the visualizations, and also have a unified UI for the settings to offer a consistent experience.
> 
> currently, we are defaults for file and assignment persisted to DB in the `user_default_selection` table
> Screen Shot 2019-07-22 at 7 39 39 AM
> 
> can you elaborate on what you mean by "the general way of storing setting " and have a unified UI for setting to offer a consistent experience"

I thought we could have a "setting" button on each view that would offer a consistent way for users to change the settings.

pushyamig:
Just trying to understand your approach, so will have a setting button on each view "Set defaults" and when the user clicks on it a pop-up appears for allowing the user to set/unset the defaults. For assignment view clicking the button will pop-up drop-down options and for grades view it's just a checkbox. is that what you mean?
or you mean instead of checkboxes just have a button. For assignment whenever the dropdown option user has chosen click on the button will save it to DB.

justin0022:
I haven't figured out what the best UI approach is, but I thought something like a setting icon button on each view which, when clicked, will show the various options for that view. The button can be placed in the same position on each view to keep it consistent.

This can be a future UI enhancement though - it isn't critically important.

pushyamig:
few comments #541 on this page suggested for having separate preference page for setting defaults across all the view that's one approach, yours is another way. Yeah, we don't need to decide on it now could have it as a future enhancement.

pushyamig:
you can try making the call as below to BE now for making the setting for grades persistent 
PUT http://localhost:5001/api/v1/courses/295532/set_user_default_selection payload as {grades:True} , with current design True/False will be stored as string
GET http://localhost:5001/api/v1/courses/295532/get_user_default_selection?default_type=grades

or can design this way https://github.com/tl-its-umich-edu/my-learning-analytics/pull/603#issuecomment-513822217 so we just get all the default option when user `/courses/12344`

justin0022:
Okay I've implemented persisting the user selection for hiding/showing the grade. I'm going to now look into if it's possible to extract the `useEffect` hooks for fetching and persisting user settings so the hooks can be used in different components.

pushyamig:
>  I'm going to now look into if it's possible to extract the useEffect hooks for fetching and persisting user settings so the hooks can be used in different components.

I agree the same code is duplicated across assignment/files and not in grades.

justin0022:
@pushyamig when you have some time could you review this? I added [snackbar](https://material-ui.com/components/snackbars/) notifications and extracted the hooks for getting user settings and setting user settings. 

I think these hooks should be compatible with other views - once you review them, I'll try and modify the Assignment Planning view to use these hooks.

assets/src/containers/GradeDistribution.js
```diff
@@ -40,28 +77,57 @@ function GradeDistribution (props) {
...
+          {userSettingLoaded
+            ? <> {'Show my grade'} <Checkbox
+              checked={showGrade}
+              onChange={() => {
+                setSettingChanged(true)
+                setShowGrade(!showGrade)
+              }} />
+            </>
+            : <Spinner />}
+          <Snackbar
```
justin0022: Ideally I'd like to extract the `snackbar` and the associated hooks above into its own component - in the future once the pattern of usage in other components gets clearer, I'll try to extract it.

justin0022:
Here's a GIF of the snackbar, as well as demonstrating the persistence of preference. 

grades_snackbar

pushyamig:
@justin0022  few observations
1. We Don't want the grades line to be shown when a user did not pick any option, default to never shows the grades line
2. When there are no grades data the response is `{}` then it is throwing an error 
`Invariant Violation: Rendered fewer hooks than expected. This may be caused by an accidental early return statement.` To replicate this in `views.py` on line 286 in the `grade_distribution` function add this line `return HttpResponse(json.dumps({}))`
3. Minor usually default_type pattern for set/get is been `file/assignment` but grades text seems to be `grades`. to be consistent we can make it `grade` or change the others

justin0022:
Thanks @pushyamig I'll work to resolve these issues.

justin0022:
@pushyamig I think all the issues are now resolved.

assets/src/containers/GradeDistribution.js
```diff
@@ -29,39 +35,102 @@ function GradeDistribution (props) {
...
-  if (loaded && isObjectEmpty(gradeData)) return (<Error>No data provided.</Error>)
+  const [gradeLoaded, gradeError, gradeData] = useGradeData(courseId)
+  const [userSettingLoaded, userSetting] = useUserSetting(courseId, 'grade')
+  const [settingChanged, setSettingChanged] = useState(false)
+  const [showGrade, setShowGrade] = useState(false)
 
-  const buildGradeView = gradeData => {
+  useEffect(() => {
+    if (userSettingLoaded) {
+      if (isObjectEmpty(userSetting)) {
```
pushyamig:
```suggestion
      if (isObjectEmpty(userSetting.default)) {
```

pushyamig: when the user doesn't have any set in DB the default is `{default: ""}`

justin0022: Okay, great. Is there ever a case where an empty object is provided?

pushyamig: no

PR #608, opened 2019-07-25
sending the course resource type via info api

pushyamig:
This PR address the flow of sending a course the resource type the course has via `/info` api call
```
{  
   "canvas_id":2922222,
   "term_id":17700000000000111,
   "name":"DENT 225 201 SU 2019",
   "term":{  
      "id":17700000000000111,
      "canvas_id":111,
      "name":"Spring/Summer 2019",
      "date_start":"2019-05-07 00:00:00+00:00",
      "date_end":"2019-12-31 00:00:00+00:00"
   },
   "current_week_number":12,
   "total_weeks":16,
   "course_view_options":{  
      "fa":1,
      "ap":1,
      "gd":1
   },
   "resource_types":[  
      "Files",
      "Videos"
   ]
}
```

### DIFF
--- a/config/env_sample.json
+++ b/config/env_sample.json
@@ -121,9 +121,9 @@
     "/*  DATA_WAREHOUSE_PORT=5432": "*/",
     "/*  Enable/Disable Unizin Date Warehouse specific features/data": "*/",
     "DATA_WAREHOUSE_IS_UNIZIN": true,
-    "RESOURCE_TYPES": {
-        "files": {"resource_label" :"Files", "resource_value":"files", "resources": ["canvas"], "disabled": "false"},
-        "videos": {"resource_label":"Videos", "resource_value":"videos", "resources": ["leccap"], "disabled": "false"}
+    "RESOURCE_VALUES": {
+        "files":  ["canvas"],
+        "videos": ["leccap"]
     },
     "RESOURCE_URLS": {
         "canvas": {"prefix": "https://example.instructure.com/files/", "postfix": "/download?download_frd=1"},
--- a/dashboard/context_processors.py
+++ b/dashboard/context_processors.py
@@ -5,16 +5,6 @@ import logging
 logger = logging.getLogger(__name__)
 
 
-def current_user_course_id(request):
-    logger.info(current_user_course_id.__name__)
-    course_id = str(request.resolver_match.kwargs.get('course_id'))
-    if not course_id:
-        logger.info("Course ID could not be determined from request, attempting to look up for user {}".format(
-            request.user.username))
-        course_id = db_util.get_default_user_course_id(request.user.username)
-    return {'current_user_course_id': course_id}
-
-
 def current_user_courses_info(request):
     logger.info(current_user_courses_info.__name__)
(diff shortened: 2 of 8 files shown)


config/env_sample.json
```diff
@@ -121,9 +121,9 @@
     "/*  DATA_WAREHOUSE_PORT=5432": "*/",
     "/*  Enable/Disable Unizin Date Warehouse specific features/data": "*/",
     "DATA_WAREHOUSE_IS_UNIZIN": true,
-    "RESOURCE_TYPES": {
-        "files": {"resource_label" :"Files", "resource_value":"files", "resources": ["canvas"], "disabled": "false"},
-        "videos": {"resource_label":"Videos", "resource_value":"videos", "resources": ["leccap"], "disabled": "false"}
+    "RESOURCE_VALUES": {
+        "files":  ["canvas"],
+        "videos": ["leccap"]
```
pushyamig: @vjcao This structure should work and is simplified as well

dashboard/context_processors.py
```diff
@@ -5,16 +5,6 @@
...
 
 
-def current_user_course_id(request):
-    logger.info(current_user_course_id.__name__)
-    course_id = str(request.resolver_match.kwargs.get('course_id'))
-    if not course_id:
-        logger.info("Course ID could not be determined from request, attempting to look up for user {}".format(
-            request.user.username))
-        course_id = db_util.get_default_user_course_id(request.user.username)
-    return {'current_user_course_id': course_id}
```
pushyamig: This is redundant as with React flow there is no reload of page React routers control the navigation no BE call to this code is made unless user accesses the /grades view etc from the browser window.

pushyamig:
@jonespm I am merging this code but you can review this code later and will work on changes as separate PR

config/env_sample.json
jonespm: I think this will still need the labels somewhere. Though the better way to do that would be in an i18n message file.

pushyamig: why do we need the labels? what's the purpose?

jonespm: Because it's intended that these will be dynamically inserted in the UI like if we added a new reading configuration.

We could possibly use the keys as the labels on the page. (if we uppercased the keys)

pushyamig: Currently, the API sends something like `['Files', 'Video'] based on the `RESOURCE_VALUES` mapping. I get your point, but I am sending what UI needs they can just grab. I think if we pose a case sensitive property the next person who comes in adds a new value if not aware the casing thing it can be easily missed.

jonespm: css can text-transform: capitalize. 

But as soon as I get some time the things I'd love to focus on are i18n and DRF.

pushyamig: So you want it property as below
```
"RESOURCE_VALUES": {
        "Files":  ["canvas"],
        "Videos": ["leccap"]
    },
```

jonespm: That or the CSS text-transform class would be the easy quick fix for this for now. Then it could still be dynamic in the UI.

PR #609, opened 2019-07-25
Updated Checkbox logic

vjcao:
**One Resource Type**
Screen Shot 2019-07-25 at 10 04 53 AM

**More Than One Resource Type**
Screen Shot 2019-07-25 at 10 06 40 AM

### DIFF
--- a/assets/src/containers/ResourcesAccessed.js
+++ b/assets/src/containers/ResourcesAccessed.js
@@ -51,7 +51,7 @@ function ResourcesAccessed (props) {
   const [curWeek, setCurWeek] = useState(0) // Should be updated from info
   const [weekRange, setWeekRange] = useState([]) // Should be depend on curWeek
   const [gradeRangeFilter, setGradeRangeFilter] = useState('') // Should be fetched from default
-  const [resourceFilter, setResourceFilter] = useState([])
+  const [resourceFilter, setResourceFilter] = useState(resourceTypes)
   const [resourceAccessData, setResourceAccessData] = useState('')
   const [dataControllerLoad, setDataControllerLoad] = useState(0)
   // initial default value that we get from backend and updated value when user save the default setting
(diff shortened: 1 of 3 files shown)


pushyamig:
@vjcao I see that you are adding setTimeout of 3 sec when fetching the data with various parameters 
`/api/v1/courses/${courseId}/resource_access_within_week?week_num_start=${weekRange[0]}&week_num_end=${weekRange[1]}&grade=${gradeRangeFilter}&resource_type=${resourceFilter}` from line 155-173 in the `ResourcesAccessed.js` 

When a user performs any action via slider or dropdown or checkbox option UI just halts for 3 sec and this might confuse the user if nothing changes based on action and will result in bad user experience
I feel it must be removed. what is the purpose for adding the timeout?

vjcao:
@pushyamig I though I already removed the timeout before the API call but the change was not reflected in the files-develop branch and I did not notice that it was still present.

pushyamig:
Whenever the resource data is loading spinner spins once the data is fetched I see `No Data` or `please select a resource type` msg is loaded for a fraction of a second and then view comes up. I feel that should be fixed

assets/src/containers/ResourcesAccessed.js
```diff
@@ -51,7 +51,7 @@ function ResourcesAccessed (props) {
   const [curWeek, setCurWeek] = useState(0) // Should be updated from info
   const [weekRange, setWeekRange] = useState([]) // Should be depend on curWeek
   const [gradeRangeFilter, setGradeRangeFilter] = useState('') // Should be fetched from default
-  const [resourceFilter, setResourceFilter] = useState([])
+  const [resourceFilter, setResourceFilter] = useState(resourceTypes)
```
pushyamig: Because of this, I see resource request call made as below on loading of the Resources view
`resource_access_within_week?week_num_start=10&week…grade=All&resource_type=Files,Videos,Files,Videos`
You can look at this from Chrome Developer tools: Network tab

pushyamig: I think the duplicates are set in the `useEffect` from line 142 - 158. You don't need that anymore if you use `useState(resourceTypes)`

pushyamig: `const [resourceFilter, setResourceFilter] = useState(resourceTypes)`
the `resourceTypes` values can be `[], ['Files', 'Videos'], ['Files'],['Videos']`
i think in case of empty value i feel you can default it to `['Files']`

pushyamig:
Looks Good!

Issue #611, opened 2019-07-29
Resource Accessed:Add default msg when course only have Files data


pushyamig:
When a course only has files ( and not video) data then the user needs to alluded with a message “you are viewing Files data”.

pushyamig:
@jennlove-um Please update message with appropriate wording.

jennlove-um:
Testing passes

PR #612, opened 2019-07-29
Added message that shows up when course only has one resource type

vjcao:
This PR addresses #611.

### DIFF
--- a/assets/src/containers/ResourcesAccessed.js
+++ b/assets/src/containers/ResourcesAccessed.js
@@ -81,6 +81,14 @@ function ResourcesAccessed (props) {
         </div>
       )
     }
+    else if (resourceTypes.length === 1) {
+      let message = "You are viewing " + resourceTypes[0] + " data"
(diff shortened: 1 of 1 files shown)


assets/src/containers/ResourcesAccessed.js
```diff
@@ -81,6 +81,14 @@ function ResourcesAccessed (props) {
         </div>
       )
     }
+    else if (resourceTypes.length === 1) {
+      let message = "You are viewing " + resourceTypes[0] + " data"
```
pushyamig: This is `const`

pushyamig: on Line 48 in `ResourcesAccessed.js`, can you change `resourceTypes` variable from `const` to `let`?

pushyamig:
@jennlove-um I am merging the PR with the wording as in the issues but I will leave the issue open if you want to change the wording

PR #616, opened 2019-07-30
Merging files-develop branch to master

pushyamig:
I have the code running on test-myla for validating this PR, Cron is run with the new code.


pushyamig:
The PR incorporates a way to include resources access data from other sources like lecture recording tools. The setup is made in such a way in future to the Resources view you can include other sources like ebooks accessed.

DB schema changed as part of this change you can find depersonalized data here https://drive.google.com/drive/u/0/folders/1Pj7roNjRPGyumKKal8-h5E6ukUiXTDI9

PR #620, opened 2019-07-31
Fixes #619 resource_type is marked as primary key but doesn't need to be

### LINKED ISSUE #619: resource_type is marked as primary key but doesn't need to be

The field resource_type is marked as a primary key in the model. I don't see why this is a primary key

## Steps to Reproduce :
1. No user visible bug, just a conversion

jonespm:


dashboard/models.py
```diff
@@ -240,7 +240,7 @@ def get_course_resource_type(self, course_id):
 
 
 class Resource(models.Model):
-    resource_type = models.CharField(primary_key=True, max_length=255, verbose_name="Resource Type")
+    resource_type = models.CharField(max_length=255, verbose_name="Resource Type")
```
pushyamig: This looks fine, does this mean we need to run a new set of depersonalized data? if so then create one before merging the PR

jonespm: I don't think it would change the data. It doesn't even look like the SQL made it as a primary key. So I don't think they'll be any noticeable impact. Just wasn't sure why it was marked like that.

PR #621, opened 2019-07-31
#578 fixes showing incorrect graded assignment in view

### LINKED ISSUE #578: Assignment Planning - Data for graded and ungraded assignments is incorrect
In the Assignments Planning view, the assignment grade data is showing up incorrectly.  The display does not match what is in the database.  Sometimes graded assignments are showing up as ungraded.  Sometimes graded assignments are showing up with the wrong grade in the tooltip.

pushyamig:
Fixes #578

### DIFF
--- a/dashboard/cron.py
+++ b/dashboard/cron.py
@@ -357,7 +357,7 @@ class DashboardCronJob(CronJobBase):
                              submission_time as (select sd.id, sd.graded_at from submission_dim sd join sub_fact suf on sd.id=suf.submission_id),
                              assign_fact as (select s.*,a.title from assignment_dim a join sub_with_enroll s on s.assignment_id=a.id where a.course_id='{data_warehouse_course_id}' and a.workflow_state='published' and muted = false),
                              assign_sub_time as (select a.*, t.graded_at from assign_fact a join submission_time t on a.submission_id = t.id),
-                             all_assign_sub as (select submission_id AS id, assignment_id AS assignment_id, course_id, global_canvas_id AS user_id, published_score AS score, graded_at AS graded_date from assign_sub_time order by assignment_id)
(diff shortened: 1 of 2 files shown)


dashboard/cron.py
```diff
@@ -357,7 +357,7 @@ def submission(self):
                              submission_time as (select sd.id, sd.graded_at from submission_dim sd join sub_fact suf on sd.id=suf.submission_id),
                              assign_fact as (select s.*,a.title from assignment_dim a join sub_with_enroll s on s.assignment_id=a.id where a.course_id='{data_warehouse_course_id}' and a.workflow_state='published' and muted = false),
                              assign_sub_time as (select a.*, t.graded_at from assign_fact a join submission_time t on a.submission_id = t.id),
-                             all_assign_sub as (select submission_id AS id, assignment_id AS assignment_id, course_id, global_canvas_id AS user_id, published_score AS score, graded_at AS graded_date from assign_sub_time order by assignment_id)
+                             all_assign_sub as (select submission_id AS id, assignment_id AS assignment_id, course_id, global_canvas_id AS user_id, round(published_score,1) AS score, graded_at AS graded_date from assign_sub_time order by assignment_id)
```
pushyamig: UDW is providing the graded score without rounding sometimes, so  rounding it before adding to our tables

dashboard/views.py
```diff
@@ -480,7 +480,7 @@ def assignments(request, course_id=0):
 def get_course_assignments(course_id):
 
     sql=f"""select assign.*,sub.avg_score from
-            (select assignment_id,name,assign_grp_name,grp_id,due_date,points_possible,group_points,weight,drop_lowest,drop_highest from
+            (select ifnull(assignment_id, 0) as assignment_id ,name,assign_grp_name,grp_id,due_date,points_possible,group_points,weight,drop_lowest,drop_highest from
```
pushyamig: Since Id's are now integers and in case of hidden assignments in the weighted assignment group the assignment id is null and pandas defaults the type to float and when doing the join all the assignment id's are changed. More [info](https://github.com/tl-its-umich-edu/my-learning-analytics/issues/578#issuecomment-512793409)

PR #627, opened 2019-08-01
"the welcome page" snuck in again somehow

justin0022:
It's removed now.


assets/src/containers/CourseList.js
```diff
@@ -52,7 +52,7 @@ function CourseList (props) {
       <AppBar className={classes.root} position='static'>
         <Toolbar>
           <Typography variant='h6' color='inherit' className={classes.grow}>
-            My Learning Analytics the welcome page
```
pushyamig: Probably due bad file merging process when working on Resource view.

justin0022: no big deal

PR #632, opened 2019-08-05
#579 assignments now scrolls to current week row

### LINKED ISSUE #579: Assignment Planning - Week view is not defaulting to current week
In the Assignment Planning view, the visualization on the bottom half of the screen is supposed to center on the current week, displaying the week label in orange.  The UI is not doing this.  It is not centering on the current week, and it is not displaying that week in a different color.

pushyamig:
Fixes #579

### DIFF
--- a/assets/src/containers/AssignmentPlanning.js
+++ b/assets/src/containers/AssignmentPlanning.js
@@ -1,4 +1,4 @@
-import React, { useEffect, useState, useRef } from 'react'
+import React, { useEffect, useState } from 'react'
 import { withStyles } from '@material-ui/core/styles'
 import { renderToString } from 'react-dom/server'
 import Paper from '@material-ui/core/Paper'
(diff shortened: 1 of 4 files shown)


assets/src/containers/AssignmentPlanning.js
```diff
@@ -40,13 +40,15 @@ const styles = theme => ({
...
 
-const getCurrentWeek = assignmentData => {
-  assignmentData.forEach(item => {
+export const getCurrentWeek = assignmentData => {
+  let currentWeek = null
+  assignmentData.some((item) => {
     let weekStatus = item.due_date_items[0].assignment_items[0].current_week
     if (weekStatus) {
-      return item.week
+      return currentWeek = item.week
```
jonespm: Does this assignment do anything?

assets/src/containers/AssignmentPlanning.js
```diff
@@ -40,13 +40,15 @@ const styles = theme => ({
...
     }
   })
+  return currentWeek
```
jonespm: Seems like this change was just `return null` if the check fails.

dashboard/views.py
```diff
@@ -578,7 +578,8 @@ def find_min_week(course_id):
 
 
 def find_current_week(row):
-    current_date = timezone.now()
```
jonespm: I looks like the [recommended](https://docs.djangoproject.com/en/dev/topics/i18n/timezones/#usage) way to do this is
`timezone.localtime(timezone.now())`

assets/src/containers/AssignmentPlanning.js
pushyamig: That is correct and it is fine

assets/src/containers/AssignmentPlanning.js
pushyamig: yes, it returns current week if there is one and since it is going over the loop for finding the current it just exits the loop once found and `ForEach` was not doing it.

Issue #631, opened 2019-08-02
Internal server error for Resource Accessed view


justin0022:
## Expected behavior (A clear and concise description of what you expected to happen) : 
Data for files accessed view should be fetched.

## Describe the bug (Tell us what happens instead of the expected behavior) :
Running the latest code from master, getting a 500 internal server error - no data is being fetched. 

## Steps to Reproduce :
1. Navigate to Resource Accessed view
2. The spinner shows indefinitely.

## Screenshots/Video (If possible, add media to help explain your problem) :
```
[2019-08-02 18:09:09 -0400] [27] [ERROR] Internal Server Error: /api/v1/courses/235420/resource_access_within_week
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/site-packages/django/core/handlers/exception.py", line 41, in inner
    response = get_response(request)
  File "/usr/local/lib/python3.6/site-packages/django/core/handlers/base.py", line 249, in _legacy_get_response
    response = self._get_response(request)
  File "/usr/local/lib/python3.6/site-packages/django/core/handlers/base.py", line 187, in _get_response
    response = self.process_exception_by_middleware(e, request)
  File "/usr/local/lib/python3.6/site-packages/django/core/handlers/base.py", line 185, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/usr/local/lib/python3.6/site-packages/django/contrib/auth/decorators.py", line 23, in _wrapped_view
    return view_func(request, *args, **kwargs)
  File "/usr/local/lib/python3.6/site-packages/rules/contrib/views.py", line 141, in _wrapped_view
    return view_func(request, *args, **kwargs)
  File "/code/dashboard/views.py", line 156, in resource_access_within_week
    filter_list.extend(RESOURCE_VALUES[filter_value.lower()])
KeyError: 'files'
```

pushyamig:
This is happening due to the property is not set. I am working on a fix this but this is the workarond
 ```
"RESOURCE_VALUES": {
        "files":  ["canvas"],
        "videos": ["leccap"]
    },
```

justin0022:
Looks like this is fixed now, thanks!

jennlove-um:
Testing passes: Resources accessed view shows resources for a course.

PR #633, opened 2019-08-05
#631 adding default resources property

pushyamig:
#631 

This error is happening since new properties are added as part of the resources view defaulting to files resources to escape of no configuration error


jonespm:
Looks good, I think we should change the URL to be like demo.instructure.com, even though it doesn't make much of a difference.

Issue #602, opened 2019-07-19
Files view bars are quite large by default


justin0022:
![Screen Shot 2019-07-19 at 2 05 03 PM](https://user-images.githubusercontent.com/8836578/61565561-412a0080-aa2e-11e9-9d9f-03f778e8e26b.png)
We should set a maximum height for the bars so they don't take up quite so much space.

jennlove-um:
Test with #638

Issue #592, opened 2019-07-16
Files Accessed bug when window slider on right is fully collapsed


justin0022:
<img width="1913" alt="Screen Shot 2019-07-16 at 9 57 00 AM" src="https://user-images.githubusercontent.com/8836578/61313879-37936500-a7b0-11e9-98b9-dadcc3e88288.png">

justin0022:
I'm going to hold off on this issue until https://github.com/tl-its-umich-edu/my-learning-analytics/pull/565 gets merged, since there'll be lots of merge conflicts.

jennlove-um:
@justin0022 Hi Justin - what's the expected behavior that I should be testing for on this one?

justin0022:
The window slider on the right, when fully collapsed, shouldn't bug out. You can see that in the left-middle section of the screenshot there's some black squiggly lines caused by the window slider being fully collapsed.

jennlove-um:
Thanks!  It's working for me in testing, now.

Issue #577, opened 2019-07-15
Multiple page reloads when changing the Files filters


jennlove-um:
The Files Viewed page list of files loads and reloads and reloads with different data each time until it finally settles.  This is distracting and confusing for the user.  When the user changes a filter, the page should reload once with the filtered data.

justin0022:
Hi @jennlove-um, I'm unable to reproduce this issue. If you could share a screen recording of the problem, I'd be happy to look into it further!

jennlove-um:
![MyLA_577_GIF](https://user-images.githubusercontent.com/39634416/61301786-a07bdc80-a7b2-11e9-86ee-22a75e9749d1.gif)

jennlove-um:
@justin0022   I uploaded a GIF that shows the flickering.  It's most reproducible when I use the week sliders.  In this case, I change the start week from 1 to 7.  The results flicker and reload several times.

justin0022:
Thanks Jenn! I'll look into it.

justin0022:
I'm going to hold off on this issue until https://github.com/tl-its-umich-edu/my-learning-analytics/pull/565 gets merged, since there'll be lots of merge conflicts.

jennlove-um:
This still occurs for me in testing.  I'm using Chrome.  It happens when I change the sliders.  I'm moving this from the fall release as not high enough priority to hold back the release.  But we should still find out why the screen flickers through several versions of the visualization before settling.

jennlove-um:
I'm closing this issue for now.  The flickering resolves quickly and doesn't seem to be an impediment for students.

PR #638, opened 2019-08-06
Fix Resources view issues

justin0022:
* Resolves https://github.com/tl-its-umich-edu/my-learning-analytics/issues/602 by setting 7 resources to be shown by default. 
* Resolves https://github.com/tl-its-umich-edu/my-learning-analytics/issues/592 by making sure domain of selection never hits `[0, 0]`.
* May have inadvertently resolved https://github.com/tl-its-umich-edu/my-learning-analytics/issues/577.

Also refactored/reorganized code.


assets/src/components/d3/createResourceAccessChart.js
```diff
@@ -169,7 +169,11 @@ function createResourceAccessChart ({ data, width, height, domElement }) {
 
   function brushmove () {
     const fullRange = mainYZoom.range()
-    const selection = d3.event ? d3.event.selection : defaultSelection
+    const selection = d3.event
+      ? d3.event.selection[1] === 0 // prevents [0, 0] from being returned, which causes bug
```
justin0022: The bug was caused when the selection window's domain got to `[0, 0]`. I just check to make sure that the second element of the array is a 0, and if so set it to be small (0.1).

pushyamig:
The UI looks good! Somewhere in the refractor, I see that resources URL link is broken. As part of the flow a user can click the link

justin0022:
> The UI looks good! Somewhere in the refractor, I see that resources URL link is broken. As part of the flow a user can click the link

Ah good catch, I'll look into this.

justin0022:
Links are fixed now @pushyamig

pushyamig:
Screen Shot 2019-08-08 at 3 04 50 PM

The x-axis is showing the wrong %percent in decimal instead of integer.

pushyamig:
Ok, Looks Good to me.

jennlove-um:
Files view looks good.  Default to 7 bars works.  Bars are larger when there are fewer than 7 files, but that's okay.  Links to files work.

PR #645, opened 2019-08-07
Fixes #644 - Name for courses in urls.py wrong

### LINKED ISSUE #644: Fix courses name in the urls.py to be courses

The name for courses in the URL's (code) is named wrong so anything that tries to reverse this isn't going to work. Noticed in LTI and talked to Sam about this today.

jonespm:


jonespm:
@ssciolla I was looking at 
https://docs.djangoproject.com/en/1.11/topics/http/urls/#naming-url-patterns

In the last paragraph is says you can use the same name for multiple patterns if they differ in their kw arguments. Which names were you wanting to change? Django is going to be able to figure out different URLs for these, but before it couldn't between the first courses and "home".

```
reverse('courses')
reverse('courses', kwargs={ 'course_id': 123456 })
```

jonespm:
Yeah, in that example you really don't need the book_detail name (it can overlap books) but I don't think there's any right or wrong way to do it. To me it makes more sense if it's just the same URL with kwargs to keep it the same name, just like we would a method.

I also don't see the prefix being necessary since you can just include based on a namespace now.

Issue #623, opened 2019-07-31
[Discussion View] Implement GraphQL layer


justin0022:
[GraphQL](https://graphql.org) will reduce data payload and number of queries, and make it easier to support the [Discussion view](#564) and future views.

justin0022:
We decided to use [Apollo](https://www.apollographql.com/docs/react/why-apollo/) for the front end GQL client.

PR #651, opened 2019-08-09
Add basic GraphQL endpoint

andrew-gardener:
Login & course permissions are checked in the query root course and courses fields (and therefor propagate down)
Currently this only exposes courses, discussions (topics & messages), and a limit user view (just user_id and name) for discussions.
This will be expanded after discussion NLP is completed to fetch those results

Part of #623

### DIFF
--- a/dashboard/graphql/objects.py
+++ b/dashboard/graphql/objects.py
@@ -1,0 +1,96 @@
+from graphene_django import DjangoObjectType
+import graphene
+
+from django.db.models import Q
+from dashboard.models import DiscussionFlattened, Course, User
+
+import logging
+logger = logging.getLogger(__name__)
+
+class DiscussionUser(DjangoObjectType):
+    user_id = graphene.ID()
+
+    class Meta:
+        model = User
+        only_fields = ('user_id', 'name')
+
+class DiscussionTopicType(DjangoObjectType):
+    topic_id = graphene.ID()
+    course_id = graphene.ID()
+    #assignment_id = graphene.ID()
+    #group_id = graphene.ID()
+    user_id = graphene.ID()
+
+    user = graphene.Field(DiscussionUser)
+
+    class Meta:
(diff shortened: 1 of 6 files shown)


jonespm:
This seems cool, I haven't used this library before. Do you expect that this could/would replace the other api's for the other views too? I had wanted to replace those with DRF but this might make more sense for what we're doing. 

It looks like they have a [plan for testing too](https://docs.graphene-python.org/en/latest/testing/), this is something we should be adding more of.

justin0022:
> This seems cool, I haven't used this library before. Do you expect that this could/would replace the other api's for the other views too? I had wanted to replace those with DRF but this might make more sense for what we're doing.
> 
> It looks like they have a [plan for testing too](https://docs.graphene-python.org/en/latest/testing/), this is something we should be adding more of.

I don't have too much experience using GraphQL, but from what I've read and the experimentation we're doing on a different [visualization project for IMS](https://github.com/patricklinpl/insights-dashboard), I think it's ideally suited for supporting data visualizations. Being able to query specific fields, filter/sort, and aggregate data with one request has been pretty great so far. 

Depending on our experience with developing the Discussion view, I think we can think about how GraphQL can be used for the other views as well.

andrew-gardener:
> Do you expect that this could/would replace the other api's for the other views too?

Yes, but its still a bit on an experiment from my viewpoint so there's no need to rush. 

One of my biggest concerns with GraphQL is the security layer which I tried to mainly resolve by having the query entry points being a course/courses so that course level permissions can be check right away but this means that some queries might not be the most efficient. Also providing access to different fields based on system role might be tricky. 

On the other hand, it will be very easy to fetch all the data needed for visualizations with one query and we can implement some interesting caching strategies on the frontend (as well as caching in the resolve layer for the [backend](https://github.com/graphql-python/graphene/issues/919))

Issue #613, opened 2019-07-29
Can't change course id in Courses admin


jonespm:
Thank you for contributing to this project!

- Make sure to search the [issues](https://github.com/tl-its-umich-edu/my-learning-analytics/issues) for duplicates first!

## Expected behavior (A clear and concise description of what you expected to happen) : 
When I change the ID in the courses admin for a course, it should change the ID

## Describe the bug (Tell us what happens instead of the expected behavior) :
When I change the ID there is a stack trace. It looks like it creates another course with none of the default view options set with this course ID and keeps the original. The error looks like it's in Django.

```
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/site-packages/django/template/base.py", line 903, in _resolve_lookup
    (bit, current))  # missing attribute
django.template.base.VariableDoesNotExist: Failed lookup for key [non_field_errors] in 'None'
```

## Steps to Reproduce :
1. Create a Course as admin in the Django Admin Panel
2. Change the ID number and hit Save 
3. Note that the original is not updated but a new one is created.

## Screenshots/Video (If possible, add media to help explain your problem) :
![image](https://user-images.githubusercontent.com/27447/62060343-f3ae5000-b1f2-11e9-916d-66bfd90040fc.png)

aberatne:
I noticed that saving the course object after converting canvas id to incremented id will update the course view options after saving the course.

aberatne:
I was trying to override the save_model() function using save(force_update=True). Is it possible to do something like this to ensure that the course instance gets the updated id instead of creating a new instance?

jennlove-um:
Testing shows that this behavior still occurs in the most recent release.

jennlove-um:
Testing shows that this behavior still occurs in the most recent release.

ssciolla:
In my initial testing, I did not see the same stack trace that @jonespm highlighted in the initial issue, but I did see that a new course is created upon updating the course ID, with the old one still being there. I believe this occurs because of the `save_model` logic that also updates the primary key.

```
    # When saving the course, update the id based on canvas id
    def save_model(self, request, obj, form, change):
        obj.id = canvas_id_to_incremented_id(obj.canvas_id)
        return super(CourseAdmin, self).save_model(request, obj, form, change)
```

I guess the edited object gets saved as a brand new object because of the new primary key, and it loses the connection to the `course_view_option` data. The backend does in fact throw a 500 error if you try to access the newly created course.

[...]

Issue #343, opened 2019-01-08
Allow admins to set manual start and end dates for the term.


jennlove-um:
Add an option that allows admins to manually set term dates instead of selecting an existing Canvas term.

jennlove-um:
There are also 2 terms in Canvas that don't have start end dates: Default Term and No Term.  If a user picked either of these terms from the list of available terms when adding a course to MyLA, they should also have to enter in term start/end dates.

jennlove-um:
![image](https://user-images.githubusercontent.com/39634416/50911053-10249f80-13fd-11e9-92df-615a62149a41.png)

The 7-week ALA course has override dates in the course itself.  Should we consider pulling course-level override dates into MyLA?  Right now we're relying on term start and end dates, but if instructors override those on a course level, it would be more appropriate to use those override dates in MyLA.

jonespm:
Yeah that last idea seems good. I'd like to rely on Canvas as a source. I see the start_at and conclude_at in the data feed. Currently this table is not being pulled down from Canvas, so we'd have to add some code to the cron to fill in this data (if it exists and isn't populated in MyLA) I think it might be too late to add it for the next release but would be nice for the following one.

I like the temporary idea of putting this course in another separate term, then just deleting that after 7 weeks. This all sounds good, thanks for the extra information.

jonespm:
I created a spring project in github and moved this over to there.

DanielDugan:
Demonstrated to Jennifer Love the UI improvements.
Working through models.py to pass the changed dates successfully.
Going to be tight on the deadline.
Will work to document code well so that review is simple

DanielDugan:
<img width="793" alt="screen shot 2019-01-25 at 11 34 38 am" src="https://user-images.githubusercontent.com/7157421/51759117-7c5a0100-2095-11e9-806b-b0da237282a6.png">

DanielDugan:
<img width="768" alt="screen shot 2019-01-25 at 11 34 23 am" src="https://user-images.githubusercontent.com/7157421/51759128-854ad280-2095-11e9-89e7-68e2e52871ff.png">

DanielDugan:
Please let me know if any changes are requested

pushyamig:
I removed Daniel as assignee and I suppose this is still not done

jennlove-um:
[...]

PR #652, opened 2019-08-09
Adding date_start and date_end to Course model and making other backend adjustments

ssciolla:
This PR makes modifications to `models.py`, `views.py`, and `admin.py` and creates a new migration, in an effort to resolve [Issue 343](https://github.com/tl-its-umich-edu/my-learning-analytics/issues/343). The changes are summarized below:

1) Two new fields -- `date_start` and `date_end` -- were added to the `Course` model in order to allow admins to manually enter a date range for a course. A helper method `get_course_date_range` was also added to the model so that if course dates are not provided, the dates associated with the related`AcademicTerm` are accessed. Additionally, the `term_id` field was changed to `term` to reflect that the field holds the `AcademicTerms` object instance, not an integer or string. The `id` field was also made `editable`, possibly eliminating some unwanted behavior; see [Issue 613](https://github.com/tl-its-umich-edu/my-learning-analytics/issues/613).

2) The `objects` field of the `AcademicTerms` model was removed, as were the `Manager` and `QuerySet` helper classes for the model, as their functionality had been replaced with the changes to `Course` and they were no longer being used. A helper method `get_correct_date_end` was also added to correct far-out end years (usually ten years into the future) to the same year as the start date.

3) A new migration file, `0010_auto_20190812_0924.py`, was added to reflect the changes outlined in 1. and 2. (The original PR included another migration file, `0010_auto_20190809_1625.py`.)

4) Changes were made to `views.py` to rely on the `get_course_date_range` method for determining the course's beginning and end dates. These modifications mean that course dates provided in the admin interface will affect the number of weeks that show up in the front end, for instance, through the Resources Accessed view.

5) Some additional fields were provided in the `list_display` field of the `CourseAdmin` class in `admin.py` to make the list appearance of courses in the admin interface more informative.


dashboard/migrations/0010_auto_20190809_1625.py
```diff
@@ -0,0 +1,45 @@
...
+        ),
+        migrations.AlterField(
+            model_name='academicterms',
+            name='date_start',
+            field=models.DateTimeField(blank=True, null=True, verbose_name='Term Start Date and Time'),
+        ),
+        migrations.AlterField(
+            model_name='course',
+            name='id',
+            field=models.BigIntegerField(db_column='id', primary_key=True, serialize=False, verbose_name='Unizin Course Id'),
```
zqian: Maybe drop the notion of "Unizin" here since MyLA might be used by non-Unizin schools?

dashboard/models.py
```diff
@@ -14,36 +14,28 @@
...
+    date_end = models.DateTimeField(verbose_name="Term End Date and Time", blank=True, null=True)
 
     def __str__(self):
         return self.name
 
+    # # Replace the year in the end date with start date (Hack to get around far out years)
+    # # This should be replaced in the future via an API call so the terms have correct end years, or Canvas data adjusted
+    def get_correct_date_end(self):
+        if (self.date_end.year - self.date_start.year) > 1:
+            logger.debug(f'{self.date_end.year} - {self.date_start.year} greater than 1 so setting end year to match start year.')
```
zqian: Since the date_end field is changed, the logging level will need to be "info" at least, instead of "debug"

dashboard/models.py
```diff
@@ -159,18 +153,31 @@ def get_supported_courses(self):
 
 
 class Course(models.Model):
-    id = models.BigIntegerField(primary_key=True, verbose_name="Unizin Course Id", db_column='id', editable=False)
+    id = models.BigIntegerField(primary_key=True, verbose_name="Unizin Course Id", db_column='id')
```
zqian: Drop the notion of "Unizin"?

dashboard/models.py
```diff
@@ -159,18 +153,31 @@ def get_supported_courses(self):
...
     canvas_id = models.BigIntegerField(verbose_name="Canvas Course Id", db_column='canvas_id')
-    term_id = models.ForeignKey(AcademicTerms, verbose_name="Term Id", on_delete=models.SET_NULL, db_column='term_id', null=True, db_constraint=False)
+    term = models.ForeignKey(AcademicTerms, verbose_name="Term Id", on_delete=models.SET_NULL, db_column="term_id", null=True, db_constraint=False)
```
zqian: Can the "verbose_name" field also be changed to "Term", instead of "Term Id"?

dashboard/views.py
```diff
@@ -593,11 +587,10 @@ def is_weight_considered(course_id):
...
 
 
-def get_term_dates_for_course(course_id):
-    logger.info(get_term_dates_for_course.__name__)
-    sql = "select a.date_start from course c, academic_terms a where c.id = %(course_id)s and c.term_id=a.id"
-    df = pd.read_sql(sql, conn, params={"course_id": course_id}, parse_dates={'date_start': '%Y-%m-%d'})
-    return df['date_start'].iloc[0]
+def get_course_start_date(course_id):
+    logger.info(get_course_start_date.__name__)
+    course_date_start = Course.objects.all().get(id=course_id).get_course_date_range()[0]
```
zqian: Does "Course.objects.all()" part means to retrieve all course objects from database, instead of just a "get" with course_id?

dashboard/migrations/0010_auto_20190809_1625.py
ssciolla: Done! See latest commit.

dashboard/models.py
ssciolla: Thanks for this correction! I'm still getting comfortable with the `logging` module. See latest commit.

dashboard/models.py
ssciolla: Done! See latest commit.

dashboard/models.py
ssciolla: Sure! See latest commit.

dashboard/views.py
ssciolla: From a performance perspective, I think Django would treat this query as the same as `Course.objects.get(id=course_id)`; see [Querysets are lazy](https://docs.djangoproject.com/en/2.2/topics/db/queries/#querysets-are-lazy). Either way though, I think what you suggested is more concise. I made the change in the latest commit.

ssciolla:
FYI, in addition to the changes @zqian requested, I simplified the verbose names for dates on the `AcademicTerms` model, and I modified the name of a method in `views.py` -- `get_course_date_start` -- to be in line with the name of the `Course` model field.

dashboard/views.py
zqian: Good to know!

dashboard/models.py
```diff
@@ -159,18 +153,31 @@ def get_supported_courses(self):
...
+    def get_course_date_range(self):
+        if self.date_start is not None:
+            start = self.date_start
+        else:
+            start = self.term.date_start
+        if self.date_end is not None:
+            end = self.date_end
+        else:
+            end = self.term.get_correct_date_end()
+        return start, end
```
jonespm: A namedtuple would make this easier to read and work with. It's an 'anti-pattern' to use an ambiguous index.

https://docs.quantifiedcode.com/python-anti-patterns/readability/not_using_named_tuples_when_returning_more_than_one_value.html

dashboard/views.py
```diff
@@ -593,11 +587,10 @@ def is_weight_considered(course_id):
...
 
 
-def get_term_dates_for_course(course_id):
-    logger.info(get_term_dates_for_course.__name__)
-    sql = "select a.date_start from course c, academic_terms a where c.id = %(course_id)s and c.term_id=a.id"
-    df = pd.read_sql(sql, conn, params={"course_id": course_id}, parse_dates={'date_start': '%Y-%m-%d'})
-    return df['date_start'].iloc[0]
+def get_course_date_start(course_id):
+    logger.info(get_course_date_start.__name__)
+    course_date_start = Course.objects.get(id=course_id).get_course_date_range()[0]
```
jonespm: Using namedtuple (above) would make this get_course_date_range().date_start

dashboard/models.py
ssciolla: That makes sense to me. See latest commit.

dashboard/views.py
ssciolla: See latest commit.

jonespm:
@ssciolla  Looks good, this was a really good Pycon session where I first saw namedtuple and how cool it was if you ever have time to watch! https://www.youtube.com/watch?v=wf-BqAjZb8M

zqian:
> @ssciolla Looks good, this was a really good Pycon session where I first saw namedtuple and how cool it was if you ever have time to watch! https://www.youtube.com/watch?v=wf-BqAjZb8M

Maybe dev should have a list of "favorite videos to watch when you got time" somewhere...

PR #657, opened 2019-08-13
Fixes #653 map a secrets directory for easier addition of secrets

### LINKED ISSUE #653: Map a secrets directory for easier addition of myla secrets and configs

I think having the secrets directory mapped via docker-compose to a directory outside of the directory would overall be safer and easier for developers. I'd like to map a directory by default in the home directory called "mylasecrets" 

Add this config to docker-compose and document

We have been copying files into MyLA but I think this is easier

jonespm:
I think this is a better idea than the current suggestion of copying in the secrets. I'll update the documentation in another commit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,12 +9,15 @@ These instructions will get a copy of MyLA up and running on your local machine 
 ### Prerequisites
 1. **Install [Docker](https://www.docker.com/)**.
 1. **Install [Git](https://git-scm.com/downloads)**.
-1. Create a new `.env` file and copy the values from `sample.env`, which has the suggested default environment variable settings.
-1. Create a new `env.json` file under the `config` folder and copy the values from `env_sample.json`, which contains most of the suggested defaults for the application.
 
 ### Installation & Setup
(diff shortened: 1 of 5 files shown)


jonespm:
Good idea, that's updated now

README.md
```diff
@@ -47,8 +50,8 @@ These instructions will get a copy of MyLA up and running on your local machine
 ## MyLA Configuring Settings
 - If you were using a prior version of MyLA, there is a utility "env_to_json.py" to help convert your configuration. Running `python env_to_json.py > config/env.json` should create your new config file from your .env file.
 
-### Secrets
-The bq_cred.json is service account for Big Query, it needs to be supplied and put into the /secrets directory and setup in the environment.
+### Secrets (Optional)
+The bq_cred.json defined in the .env file is service account for Big Query, it needs to be supplied and put into the directory defined in the .env file and setup in the environment.
```
zqian: minor correction: BigQuery instead of Big Query

jonespm:
Yeah, there really doesn't even need to be the config folder anymore. I could just move this sample json to the top level or maybe a docs folder?

Issue #646, opened 2019-08-07
Test and update LTI code as needed


jonespm:
Thank you for contributing to this project!

- Make sure to search the [issues](https://github.com/tl-its-umich-edu/my-learning-analytics/issues) for duplicates first!

## Expected behavior (A clear and concise description of what you expected to happen) : 
The LTI launch should be well documented and work as expected.

## Describe the bug (Tell us what happens instead of the expected behavior) :
When testing the code for the LTI there were a few things that could be improved, especially when it comes to launching from Canvas.

## Steps to Reproduce :
1. Try to setup MyLA to launch LTI to Canvas (probably on a test instance with ngrok for instance)
2. See how it works

I can fill in more details about testing this later but it wasn't working well for me and I have a few fixes to make

* Improve the documentation around setting this up
* Improve the exception handling when things aren't setup correctly
* Fix the issue added with the rules backend being added (This throws an error now)
* Add CSP support for iframe embedding

PR #659, opened 2019-08-13
Fixing LTI to explicitly use ModelBackend

andrew-gardener:
Noticed that `rules.permissions.ObjectPermissionBackend` was interfering with the LTI login (ObjectPermissionBackend doesn't have a get_user function) since it is the first auth method listed after the permissions update. Switching to explicitly using the `django.contrib.auth.backends.ModelBackend` backend (and making sure its a backend available if lti is enabled)

### DIFF
--- a/dashboard/lti.py
+++ b/dashboard/lti.py
@@ -23,8 +23,8 @@ def valid_lti_request(user_payload, request):
         except User.DoesNotExist:
             password = ''.join(random.sample(string.ascii_letters, RANDOM_PASSWORD_DEFAULT_LENGTH))
             user_obj = User.objects.create_user(username=username, email=email, password=password)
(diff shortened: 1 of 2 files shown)


pushyamig:
This PR makes sense to me. @jonespm is working getting LTI running #646 he would have more insights.

jonespm:
Thanks Andrew, I was testing this too. I saw a few other issues and I can add them to this PR or just create a new one. Some of the PR's I've been doing lately have been in support of LTI.

jonespm:
Okay, what I was doing is the same thing Django Auth does and just looking for a backend with the get_user method. This seems fine too. I'll commit this and add a few other cleanups on a separate PR.

Issue #457, opened 2019-03-18
Pull course term/dates from data warehouse


jonespm:
We are currently requiring the user to manually enter the term (which is pulled from Canvas after the first run)

This isn't necessary as the term is in the data. However there are two fields start_at and conclude_at which may override the term for a course. These should also be pulled.

All of these values should be utilized wherever the term calculation is being used. (For instance in the get_course_assignments.

jonespm:
Assigning to @ssciolla as he worked on these term dates already. This would be nice to have for Fall but could be pushed back, maybe even to a later Fall release.

jennlove-um:
Testing done in issue #666 .

PR #666, opened 2019-08-16
Adding method to cron.py to update courses

ssciolla:
This PR modifies `cron.py` to use the data for courses already pulled from the data warehouse to update the values for `name`, `term_id`, `date_start`, and `date_end` in course records in the database. This PR aims to resolve issue #457. More details about the changes are provided below.

1) The `verify_course_ids` method in `cron.py` was already running SELECT queries to pull data on each supported course from the data warehouse. The method was modified here to return -- in addition to a list of invalid course ids -- a pandas DataFrame containing all of the course data.

2) A new method was added (and invoked in the `do` method) called `update_course` which takes the DataFrame containing the warehouse course data as a parameter. The method then accesses the existing course records using the Django ORM, and then it loops through the model instances and compares their values to the values in the DataFrame. The method includes logic to only update the values for `name`, `date_start`, and `date_end` when they differ from the values the database is already storing. The related `term` model instance (or `term_id` in the database) is always updated, as the `update_term` method in `cron.py` always deletes the existing records in the `academic_terms` table.

### DIFF
--- a/dashboard/cron.py
+++ b/dashboard/cron.py
@@ -85,6 +85,20 @@ def deleteAllRecordInTable(tableName):
     return f"delete : {tableName}\n"
 
 
+# use Django ORM to compare warehouse and existing data and, if necessary, update DateTime field of model instance
+def update_datetime_field(course_obj, course_field_name, warehouse_dataframe, warehouse_field_name):
+    warehouse_field_value = warehouse_dataframe[warehouse_field_name].iloc[0]
+    if warehouse_field_value is not None:
+        warehouse_field_value = warehouse_field_value.replace(tzinfo=pytz.UTC)
+    course_field_value = getattr(course_obj, course_field_name)
+    if course_field_value != warehouse_field_value:
+        logger.info("{} (course) is not {} (warehouse).".format(str(course_field_value), str(warehouse_field_value)))
+        setattr(course_obj, course_field_name, warehouse_field_value)
+        logger.info(f"{course_field_name} for {course_obj.id} has been updated.")
+        return [course_field_name]
+    return []
+
+
 # cron job to populate course and user tables
 class DashboardCronJob(CronJobBase):
 
@@ -431,7 +445,7 @@ class DashboardCronJob(CronJobBase):
         status = ""
         logger.debug("while using verify_course_ids data to update course table")
 
-        logger.info(warehouse_courses_data.to_json())
+        logger.debug(warehouse_courses_data.to_json())
         courses = Course.objects.all()
         if len(courses) > 0:
             courses_string = ", ".join([str(x) for x in Course.objects.get_supported_courses()])
@@ -455,23 +469,8 @@ class DashboardCronJob(CronJobBase):
             logger.info(f"Term for {course.id} has been updated.")
             updated_fields.append("term")
 
-            warehouse_course_start = warehouse_course_df["start_at"].iloc[0]
-            if warehouse_course_start is not None:
-                warehouse_course_start = warehouse_course_start.replace(tzinfo=pytz.UTC)
-            if course.date_start != warehouse_course_start:
-                logger.info("{} (course) is not {} (warehouse).".format(str(course.date_start), str(warehouse_course_start)))
-                course.date_start = warehouse_course_start
-                logger.info(f"Date Start for {course.id} has been updated.")
-                updated_fields.append("date_start")
-
-            warehouse_course_end = warehouse_course_df["conclude_at"].iloc[0]
-            if warehouse_course_end is not None:
-                warehouse_course_end = warehouse_course_end.replace(tzinfo=pytz.UTC)
-            if course.date_end != warehouse_course_end:
-                logger.info("{} (course) is not {} (warehouse).".format(str(course.date_end), str(warehouse_course_end)))
-                course.date_end = warehouse_course_end
-                logger.info(f"Date End for {course.id} has been updated.")
-                updated_fields.append("date_end")
+            updated_fields += update_datetime_field(course, "date_start", warehouse_course_df, "start_at")
+            updated_fields += update_datetime_field(course, "date_end", warehouse_course_df, "conclude_at")
 
             course.save()
             status += ", ".join(updated_fields) + "\n"
--- a/dashboard/cron.py
+++ b/dashboard/cron.py
@@ -1,5 +1,5 @@
 
-from __future__ import print_function #python 3 support
+from __future__ import print_function  # python 3 support
 
 from django.db import connections as conns
 
@@ -7,11 +7,13 @@ from dashboard.common import db_util
 
 import logging
 import datetime
+import pytz
 
 from sqlalchemy import create_engine
 from django.conf import settings
+from collections import namedtuple
 
-from dashboard.models import Course, Resource
+from dashboard.models import Course, Resource, AcademicTerms
 
 import pandas as pd
 
@@ -33,7 +35,7 @@ logger.debug("db-user:" + db_user)
 
 engine = create_engine("mysql+mysqldb://{user}:{password}@{host}:{port}/{db}?charset=utf8mb4"
                        .format(db = db_name,  # your mysql database name
-                               user = db_user, # your mysql user for the database
+                               user = db_user,  # your mysql user for the database
                                password = db_password, # password for user
                                host = db_host,
                                port = db_port))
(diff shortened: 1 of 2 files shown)


andrew-gardener:
Is it possible to eliminate the `academic_terms` table and just take into account the term when syncing date_start & date_end
ex: when fetching `course_dim`, it can join on `enrollment_term_dim` and get term date_start & date_end and then set the course's date_start & date_end to course or term as needed.

I think it would streamline a bit of code around dates (unless the term name is being needed somewhere).

jonespm:
Yeah, first thinking about it, I'm not sure if there's any usefulness in this application knowing the term. 

Perhaps in the future we could have a feature where we grouped or sorted courses on the main page by term, but this is a new feature for this semester. Currently that isn't anything exposed anywhere though. I'd "probably" leave it in there because Canvas has this table, but it doesn't matter for us either way at the moment.

I think what we'd like to do eventually is have a better data model and move more things away from requiring cron for Canvas schools just going to the warehouse live and caching the results. This seems better for the future if we were supporting all courses. But for schools that don't use Canvas or BigQuery they'd need some structure that mirrors what we were using. I also don't know how efficient (cost wise) it would be to process big query for thousands of courses. We're not there yet though. 

We're almost to the point where if we had a subset of columns (or views) from the Canvas tables we're using that matched closer to Canvas it might be easier? I'm not sure if it would make sense to Model those, but maybe if we did an inspectdb to generate the models it might work out satisfactorily? 

Then we'd have a few myla_ specific tables and a few canvas_ specific tables? 

```
course_fact
course_score_fact
assignment_fact
assignment_group_fact
submission_fact

assignment_dim
assignment_group_dim
assignment_group_rule_dim
course_dim
enrollment_dim
file_dim
pseudonym_dim
submission_dim
user_dim
```

ssciolla:
@andrew-gardener @jonespm Thanks for the feedback. I feel there might be some utility in keeping the academic terms table intact, for instance (as Matt suggested) for enabling a rough classification of courses if we we got to the point where a student who had taken multiple courses where MyLA was deployed wanted to go back and look at past dashboards. Part of me also feels a little uneasy about storing `term`-related dates in a `course` table in the database. They are different entities, and it doesn't seem as transparent as handling the logic with the `get_course_date_range` method, although I could see the alternative improving performance (I'm not an expert in that area). That said, I'll defer to the more senior developers.

I would agree with Matt that creating some kind of MyLA specific table (`MyLACourse`, or `CourseProfile`?) might help clean up the data model, separating the tables determined by the warehouse and the tables we manage for this application. I'm not sure what kind of position we are in to make that kind of major change to the backend, given the approaching fall release.

Regarding this pull request, are we okay with this general approach (using the ORM to update some columns of the course table? Should I change anything to reflect the above discussion?

dashboard/cron.py
```diff
@@ -403,14 +418,66 @@ def update_term(self):
...
         return status
 
+
+    # update course records with data returned from verify_course_ids, only making changes when necessary
+    def update_course(self, warehouse_courses_data):
+        # cron status
+        status = ""
+        logger.debug("while using verify_course_ids data to update course table")
+
+        logger.info(warehouse_courses_data.to_json())
```
zqian: might use debug if the json data gets too big.

dashboard/cron.py
```diff
@@ -403,14 +418,66 @@ def update_term(self):
...
+                course.name = warehouse_course_name
+                logger.info(f"Name for {course.id} has been updated.")
+                updated_fields.append("name")
+
+            # term is always updated because the term records are deleted by the update_term method
+            course.term = AcademicTerms.objects.get(id=warehouse_course_df["enrollment_term_id"])
+            logger.info(f"Term for {course.id} has been updated.")
+            updated_fields.append("term")
+
+            warehouse_course_start = warehouse_course_df["start_at"].iloc[0]
```
zqian: The sections of warehouse_course_start and warehouse_course_end are very similar. Could they be refactored as a util function?

dashboard/cron.py
```diff
@@ -103,13 +110,17 @@ def verify_course_ids(self):
...
 
-            # error out when course id is invalid
+            # error out when course id is invalid, otherwise accumulate dataframes
             if course_df.empty:
                 logger.error(f"""Course {course_id} don't have the entry in data warehouse yet. """)
                 invalid_course_id_list.append(course_id)
+            else:
+                course_dfs.append(course_df)
 
+        courses_data = pd.concat(course_dfs).reset_index()
```
pushyamig: When no course is added to the app then code should handle it, currently throwing an exception. A check should be made on emptiness of the list `course_dfs`
`raise ValueError(\'No objects to concatenate\')\nValueError: No objects to concatenate\n', None`

pushyamig:
Currently, from the admin view for creating a course, the terms data is not loaded I cannot select any term and because of the error mentioned here https://github.com/tl-its-umich-edu/my-learning-analytics/pull/666/files#r316142341 I cannot create a course since terms id is a required field which is empty. This is a bug to reproduce you might need to wipe the DB.

dashboard/cron.py
pushyamig: I will review the code once this error is fixed

ssciolla: Thanks for this catch!  I made some modifications in this [commit](https://github.com/tl-its-umich-edu/my-learning-analytics/pull/666/commits/995204b8ad40d8c1277f014ece456057b54dc6bb) that I believe have resolved this. Now you should be able to run `cron.py` with no courses in the database. Essentially, it will update the `academic_terms` table and nothing else.

dashboard/cron.py
ssciolla: Thanks! I changed to `debug` in this [commit](https://github.com/tl-its-umich-edu/my-learning-analytics/pull/666/commits/a0fa6b6d9f37bdbec3c7cf604f53a9949ab43af4).

dashboard/cron.py
ssciolla: Yep, I gave it a go [here](https://github.com/tl-its-umich-edu/my-learning-analytics/pull/666/commits/a0fa6b6d9f37bdbec3c7cf604f53a9949ab43af4).

ssciolla:
@pushyamig Regarding the issue you raised with not being able to set a term when the database is empty, I resolved this [here](https://github.com/tl-its-umich-edu/my-learning-analytics/pull/666/commits/fa7ff3edc3579f58764b21aacee07422ce01d95c) by making the `term` field read-only in the admin interface. Now you can run add a course when the `academic_terms` table is empty, and then run the cron to update both the terms and the courses. Provided that there is always a term associated with a course in the warehouse, the cron can manage this field going forward. We should discuss, too, whether we can make some of the other fields read-only.

pushyamig:
A couple of things Noticed:
1. When I set a start/end date in Myla to the course during the cron run if the course start or end date is not set in canvas just don't override the dates with warehouse dates instead keep the date from the MyLA DB
2. I like the fact the in spite Term id is read-only during from admin view the cron process populates it. I also like that you are updating the Course name too very handy. 
3. Minor, I feel if there are no courses added to instance after the `course_verification` check is done, just do the `self.update_term()`and you can skip all the other updates to the tables like resources, assignment. 
4. I like that fact the `Course Id` field from the admin interface now don't need to fill in that 16 digit long canvas id but typing something random say 1 after saving gives the correct id. I feel that this field should be hidden from the admin interface so that admin's don't' have to do it, but I can live with that.

ssciolla:
@pushyamig I believe I've addressed the issues you raised in your last comment.

1) There is now logic in the `update_datetime_field` function to only update fields that don't have data provided by a previous cron job run or a user through the admin interface.

2) I simplified the logic in the cron to only run the `update_term` and `update_unizin_metadata` functions when there are no courses in the database.

3) I added `'id` to the `readonly_fields` in the `CourseAdmin`. I believe this resolves issue #666, ensuring that the `AlwaysChangedModelForm` works as expected and creates new instances of `CourseViewOption` when the defaults are untouched in the `Course` form.

pushyamig:
This looks good to me!

jennlove-um:
Here are the conditions that were tested in MyLA-test

1. Added a Fall 2019 course that has override dates in Canvas.  No dates were manually entered in MyLA when I added the course.  Dates in Canvas course correctly populated in the MyLA course along with the correct term.

2. Added a Fall 2019 course that does not have override dates in Canvas.  No dates were manually entered in MyLA when I added the course.  Default dates for Fall term are being used correctly for the resources visualization.

3. Added a Fall 2019 course with no override dates in Canvas.  Manually entered the dates in MyLA.  The dates were correctly maintained and not overwritten when I ran the chron job.  The term correctly populated in MyLA.

4. Added a Fall 2019 course with an overridden start date but no overridden end date in Canvas.  No dates were manually entered in MyLA when I added the course.  Start date correctly came from Canvas course, and no end date came in.  The resources visualization displays the week count correctly based on the fall term end date.

5. Manually set the start date for a Fall 2019 course in MyLA.  When I run the chron job, that date is correctly left alone and not overwritten by the date set up in Canvas.

6. Canvas course names are correctly populating the MyLA courses when I run the chron job.

Testing passes.

Issue #668, opened 2019-08-19
CourseViewOption record with defaults is not created when new Course record is created in Admin


ssciolla:
Thank you for contributing to this project!

- Make sure to search the [issues](https://github.com/tl-its-umich-edu/my-learning-analytics/issues) for duplicates first!

## Expected behavior (A clear and concise description of what you expected to happen) : 

When a new `Course` is created in the Django admin interface, a new `CourseViewOption` record should also be created, regardless of what boxes are checked for the three view inline settings ("Show Resources Accessed View", "Show Assignment Planning View," and "Show Grade Distribution View").

## Describe the bug (Tell us what happens instead of the expected behavior) :

When a new `Course` is created through the Admin with the default options selected for the `CourseViewOption` inline integer fields, a new `CourseViewOption` record is not created. If you uncheck any of the boxes (diverging from the default), a `CourseViewOption` record is created.

## Steps to Reproduce :
1. In the Django admin interface, click the "Add Course" button.
2. Fill out the form without modifying the three CourseViewOption checkboxes
3. Look at the Courses Dashboard, where there will be now views listed under "COURSE VIEW OPTION(S)" for the new course. Alternatively, look at the `course_view_option` table in the database, noting the absence of a new record.

## Additional context (Add any other context about the problem here) :

This bug was previously documented and fixed by @jonespm (see PR #452). I believe it was reintroduced when I removed the `editable=false` option from the `id` field in `Course` in `models.py` as part of PR #652 (thanks to @pushyamig for alerting me to the issue!). This bug could be fixed by re-adding the `editable=false` option. However, this problem and Issue #613 might be fixed by adding `AutoField` primary keys to both the `Course` and `CourseViewOption` model and making the table relationship dependent on those ids rather than field values we may want to change. I will investigate the latter option and report back. If that is a dead end, I will re-add the `editable=false` option.

jonespm:
It should still have editable=false. 

The save_model override of CourseAdmin is intended for creating and saving the id based on the canvas_id. This isn't a field the user should need to enter. It's possible this isn't executing correctly if this isn't in there. 

```
    def save_model(self, request, obj, form, change):
        obj.id = canvas_id_to_incremented_id(obj.canvas_id)
        return super(CourseAdmin, self).save_model(request, obj, form, change)
```

ssciolla:
[...]

PR #669, opened 2019-08-20
Adding AutoField primary keys to Course and CourseViewOption models

ssciolla:
This PR introduces changes to the `Course` and `CourseViewOption` models, seeking to simplify their relationship by adding `Autofield` primary keys. The PR aims to resolve issues #668 and #613, enabling changes to course ids. This PR is tentative, and feedback is welcome. The changes are summarized below.

1) `AutoField` primary keys were added to both `Course` and `CourseViewOption`, and the former `id` primary key of `Course` was changed to `warehouse_id`. As the application references the `id` field of the `course` table and model throughout, this latter change meant that some other changes were necessary in `views.py` and `db_util.py`. 

2) Three migration files were added that sequentially walk through the process of creating temp models (`CourseTemp` and `CourseViewOptionTemp`), dropping the old models, and renaming the temp models to the old names.

3) Changes were made to `admin.py` to make the new `warehouse_id` read-only, so that it appears in the admin interface but is only updated by entry to the `canvas_id` form field.


dashboard/common/db_util.py
```diff
@@ -36,26 +36,31 @@ def get_course_name_from_id(course_id):
...
+            cursor.execute("SELECT name FROM course WHERE warehouse_id = %s", [warehouse_id])
             row = cursor.fetchone()
             if (row != None):
                 course_name = row[0]
     return course_name
 
 
-def get_course_view_options (course_id):
-
+def get_course_view_options(course_id):
```
ssciolla: This function seems to be no longer in use. I turned up no other occurrences through a search, and in the primary place where it would be used (see `get_course_info` method in `views.py`, line 120), a QuerySet is used. Should we remove it?

ssciolla:
I’m closing this for the moment while we discuss the need for the change and work on other updates to the Course table.

PR #671, opened 2019-08-21
Fixes #646 - Fix some issues related with LTI integration

### LINKED ISSUE #646: Test and update LTI code as needed

jonespm:
These are the things fixed

- Improved documentation for setting this up in Canvas
- Sample settings for configuring for embedding in Canvas
- Added some more logging on failure and returned LTIException when there is an exception
- CSP values not set should not be set in the settings (they'll fall back to defaults based on the spec)
- Updated rules to check either the sis_id or sis_name as Canvas returns an id
- Added CSRF settings for Django as it didn't seem to work without these


README.md
```diff
@@ -69,13 +69,19 @@ Note that by default all views are enabled when a course is added.
...
+`Public: Various identifying information (name, email, Canvas ID, SIS ID of the course, SIS ID of the user, etc.) is sent to the vendor.`
+
+You also need to configure CSP value in the environment, specifically the FRAME_SRC. (See next section) In addtiion you make sure you are using https and CSRF_COOKIE_SECURE is true with your domain (or instructure.com) in trusted origins.
 
 Environment variables:
 
 `STUDENT_DASHBOARD_LTI`: Set to True to enable LTI (default false).
 
-`PYLTI_CONFIG_CONSUMERS`: JSON string of supported LTI Consumers (default none). Formated `{ "LTI_CONSUMER_KEY_1": { "secret": "LTI_CONSUMER_SECRET_1" } }`.
+`PYLTI_CONFIG_CONSUMERS`: JSON string of supported LTI Consumers (default none). Formated `{ "LTI_CONSUMER_KEY_1": { "secret": "LTI_CONSUMER_SECRET_1" } }`. These are the actual key and secret you put into your launch (like in Canvas).
```
andrew-gardener: can you delete the second `LTI_EMAIL_FIELD` listed just below this (just noticed it. not sure how it got there, might have been my bad)

dashboard/lti.py
```diff
@@ -27,13 +32,13 @@ def valid_lti_request(user_payload, request):
...
     if canvas_course_id:
         url = reverse('courses', kwargs={ 'course_id': canvas_course_id })
 
     return url
 
-# def invalid_lti_request(user_payload):
-#     pass
\ No newline at end of file
+def invalid_lti_request(user_payload):
+    logger.info(f"Invalid LTI request for {user_payload}")
```
andrew-gardener: I think you need to add `HttpResponseRedirect(get_reverse('django_lti_auth:denied'))` in here if your using this function (default action taken by django-lti-auth https://github.com/rohitjose/django-lti-auth/blob/master/django_lti_auth/views.py#L56)

README.md
jonespm: Great, this will be fixed in the morning!

dashboard/lti.py
jonespm: Thanks, I'll call this. The LTI integration is looking really cool, happy how it's working for us. A few other schools have really been asking so I wanted to try it all out!

dashboard/rules.py
```diff
@@ -38,7 +39,7 @@ def is_instructor_in_course(user, course):
 def is_above_student_in_course(user, course):
     try:
         enrolled = User.objects.get(
-            sis_name=user.get_username(),
+            sis_name=Q(user.get_username()) | Q(sis_id=user.get_username()),
```
zqian: use "Q(sis_name=user.get_username()) | Q(sis_id=user.get_username())" instead?

In fact, there are multiple occurrence of "Q(sis_name=user.get_username()) | Q(sis_id=user.get_username())" form, maybe that could be generalized.

dashboard/settings.py
```diff
@@ -410,7 +410,11 @@
 
 # Method to show the user, if they're authenticated and superuser
 def show_debug_toolbar(request):
-    return DEBUG and request.user and request.user.is_authenticated and request.user.is_superuser
+    try:
+        return DEBUG and request.user and request.user.is_authenticated and request.user.is_superuser
+    except:
+        # If there's an exception here just don't show it
```
zqian: Should the exception be logged at least?

dashboard/settings.py
```diff
@@ -438,11 +442,20 @@ def show_debug_toolbar(request):
...
+        # If there's a value set for this CSP config, set it as a global
+        if (csp_val):
+            globals()["CSP_"+csp_key] = csp_val
 # If CSP not set, add in XFrameOptionsMiddleware
 else:
     MIDDLEWARE_CLASSES += ['django.middleware.clickjacking.XFrameOptionsMiddleware',]
 
+# These are mostly needed by Canvas but it should also be in on general 
+CSRF_COOKIE_SECURE = ENV.get("CSRF_COOKIE_SECURE", True)
+if CSRF_COOKIE_SECURE:
```
zqian: Does those variables get default values set if CSRF_COOKIE_SECURE is false?

dashboard/lti.py
jonespm: It looks like from that link the code you're referencing will still run and return fine after the invalid_login_method_hook is executed. I tested and still getting the same "Sorry, you are not allowed to access this app" template from the module. That seems "fine" as you'd generally only see it if the key or secret are wrong.

dashboard/settings.py
jonespm: Yeah we could, I don't 100% remember what problem I saw here was but there was an exception being thrown when the user wasn't logged in from somewhere in Django and I just couldn't get it to go away. I'll except it but still log it for now. It will only come up when DEBUG is enabled, which should only be on our test servers.

dashboard/settings.py
jonespm: They just aren't set to anything. They [have defaults as part of Django](https://docs.djangoproject.com/en/1.11/ref/settings/).
[CSRF_TRUSTED_ORIGINS](https://docs.djangoproject.com/en/1.11/ref/settings/#std:setting-CSRF_TRUSTED_ORIGINS) = [] 
[SESSION_COOKIE_SECURE](https://docs.djangoproject.com/en/1.11/ref/settings/#session-cookie-secure)=False
[SECURE_PROXY_SSL_HEADER](https://docs.djangoproject.com/en/1.11/ref/settings/#secure-proxy-ssl-header) = None

dashboard/rules.py
jonespm: Yeah this is a good idea, I think we can put these in the model and clean this up.

dashboard/settings.py
jonespm: I'll just remove this, I don't think it's necessary to have in here and I don't remember why I added it.

dashboard/rules.py
jonespm: I moved these into the model and removed an unused method

Issue #674, opened 2019-08-26
Slightly re-org Dockerfile to improve layer caching


jonespm:
Thank you for contributing to this project!

- Make sure to search the [issues](https://github.com/tl-its-umich-edu/my-learning-analytics/issues) for duplicates first!

## Describe the bug (Tell us what happens instead of the expected behavior) :
I noticed in the docker file it does some of the copying of node deps after it copies the project. We should be able to to do that library copy first to save a few seconds on each build.

Also the TZ in never changes and that can just go to the top.

jennlove-um:
@jonespm Is this still needed?

PR #675, opened 2019-08-26
Issue #674 Slightly re-org Dockefile to improve layer caching

jonespm:


Dockerfile
```diff
@@ -25,6 +22,11 @@ RUN npm prune --production && \
 # FROM directive instructing base image to build upon
 FROM python:3.6 AS app
 
+# Sets the local timezone of the docker image
+ARG TZ
+ENV TZ ${TZ:-America/Detroit}
+RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone
+
```
pushyamig: what is the benefit of moving this `TZ` setting here?

Dockerfile
```diff
@@ -36,21 +38,17 @@ RUN apt-get update && \
...
-COPY . /code/
 # copy built react and node libraries for production mode
 COPY --from=node-prod-deps /usr/src/app/package-lock.json /code/package-lock.json
 COPY --from=node-prod-deps /usr/src/app/webpack-stats.json /code/webpack-stats.json
 COPY --from=node-prod-deps /usr/src/app/assets /code/assets
 COPY --from=node-prod-deps /usr/src/app/node_modules /code/node_modules
 
+# NOTE: project files likely to change between dev builds
+COPY . /code/
+
```
pushyamig: Why this line needed to be moved?

pushyamig:
This looks Good.

Dockerfile
jonespm: I think moving these up higher is actually wrong, I'll move it back down where it was.
https://medium.com/@esotericmeans/optimizing-your-dockerfile-dc4b7b527756

Dockerfile
jonespm: The main part of this was moving the COPY of code down below the copy of the node dependencies. That takes a little time to copy that and those don't change as much as the code. You can look at this as move the 4 COPY lines for --from=node-prod-deps above this one. ;)

PR #678, opened 2019-08-28
Changing default for CSRF_COOKIE_SECURE

jonespm:


jennlove-um:
How can we test for this?

PR #686, opened 2019-09-06
Remove remaining code referencing files accessed view

pushyamig:
This issue fixes the `files access` text that remained after the rename to `resource access`. 

I see that My Learning Analytics Help is still hardcoded to umich one(https://sites.google.com/umich.edu/my-learning-analytics-help/home), not sure if different universities want to have their own Help doc. But the content present in the doc is suitable for all universities.


jonespm:
Thanks! I'll create a new feature request to make the help URL configurable.

Issue #683, opened 2019-09-05
Demo mentioned in README no longer works


jonespm:
Thank you for contributing to this project!

- Make sure to search the [issues](https://github.com/tl-its-umich-edu/my-learning-analytics/issues) for duplicates first!

## Expected behavior (A clear and concise description of what you expected to happen) : 
When running the "Populate initial demo terms and courses" something should still work with this process.

`docker exec -it student_dashboard /bin/bash ./demo_init.sh`

## Describe the bug (Tell us what happens instead of the expected behavior) :
In the current version running this it gives a lot of errors, probably because of the AcademicTerm changes. We may want to just change this demo process to load up the sample SQL instead but if we do this and remove this old demo then that process should be documented instead.

jonespm:
I think it would still be nice to have some mock "demo" data. I'm not sure of the best way to do it. 

I was thinking we could just re-generate de-personalized data whenever we made large changes to the data model but that requires keeping up with that process. 

We could create [fixtures](https://docs.djangoproject.com/en/2.2/howto/initial-data/) instead of the SQL file (based on the SQL file). I think I was looking at one time at [dumpdata](https://docs.djangoproject.com/en/dev/ref/django-admin/#dumpdata) (to dump the personalized data) and just keeping the fixtures updated from that point forward.

jennlove-um:
@jonespm  Any advice on testing this one?

jonespm:
Ok I tested it, I'm going to create a new issue to add demo_init.sh to the git ignore otherwise it comes up as an untracked file.

PR #692, opened 2019-09-11
Shifting to demo_init.sh.sample and aligning commands with updated data model

ssciolla:
This PR fixes and adjusts the `course` and `term` commands in `dashboard/management/commands` to reflect the changes to the data model introduced in PR #652. It also shifts the demo approach to using a `.sample` version of the file with fabricated data. The PR aims to resolve issue #683. Specific changes are discussed in more detail below.

1) To correct `course.py`, I changed the name of the term field when creating the course instance to `term` (rather than `term_id`). The new fields `date_start` and `date_end` were added as arguments, and I modified the `required` parameter values to reflect which values are `nullable` in `models.py`. In some cases, I needed to introduce some conditions checking whether fields had non-null values, so no invalid operations would be performed using `None`.

2) In both `course.py` and `term.py`, I added functionality to ensure that DateTime objects created were being stored with the UTC timezone, which is in line with how Unizin and Canvas store DateTimes. This also eliminated `RuntimeWarning`s indicating that `naive` DateTime objects were being stored in Django fields.

3) Per discussions with @zqian and @jonespm, I changed the current `demo_init.sh` to a `.sample` version of the file with fabricated values that users can amend for their particular application. This makes sense for a couple reasons: the existing values were specific to one institution (meaning it would not work in concert with `cron.py` for all contributing developers), and while course and term info are not particularly sensitive, it seems a generally good practice to leave such data out of a public repository. The new `demo_init.sh.sample` includes some updated comments and commands for one fake course and one fake term; it will run successfully if you rename the file to `demo_init.sh`. The README has been updated accordingly to explain how we expect the demo might be used.


jonespm:
Ok I tested it out. I think we should add demo_init.sh to the .gitignore now though otherwise someone might check in their demo file.

pushyamig:
We can do it as a separate issue can be included in the `2019.02.02 `project

PR #696, opened 2019-09-13
Update PTVSD library to support current VSCode

jonespm:


jennlove-um:
How can we test for this?

PR #697, opened 2019-09-13
Configurable Help URL Through Django Flat Pages

### LINKED ISSUE #688: Help URL should be configurable

The URL to help should be able to be configured via the json settings. Currently it's hardcoded to [Michigan's Help](https://github.com/tl-its-umich-edu/my-learning-analytics/blob/106d8ab585ba4479fb54d0624af9e33347d271eb/assets/src/components/AvatarModal.js#L45). Other schools may want to provide they own custom help.

Probably just have the frontend read some value from the environment or pass this URL though the myla_globals.

vjcao:
The help URL is now configurable through the Django admin page. Help URLs can now be changed by adding a flat page with the url set as /help_url/ and the content set as the desired url for the help page.

Closes #688.

### DIFF
--- a/assets/src/components/AvatarModal.js
+++ b/assets/src/components/AvatarModal.js
@@ -37,13 +37,12 @@ const styles = theme => ({
 })
 
 function AvatarModal (props) {
-  let globals = myla_globals
   const { classes, user } = props
 
   const url = window.location.href
   const logoutURL = '/accounts/logout'
 
-  const [helpURL, setHelpURL] = useState(globals.help_url)
+  const [helpURL, setHelpURL] = useState(user.helpURL)
   const [openChangeCourseDialog, setOpenChangeCourseDialog] = useState(false)
 
   const Admin = () => (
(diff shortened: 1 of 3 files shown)


assets/src/components/AvatarModal.js
```diff
@@ -37,12 +37,13 @@ const styles = theme => ({
 })
 
 function AvatarModal (props) {
+  let globals = myla_globals
```
justin0022: Hi @vjcao, just one suggestion. Instead of accessing `myla_globals` from within `AvatarModal`, could you please add the `help_url` property to the frozen `user` object in `App`? 

Something like this:
```js
const user = Object.freeze({
  admin: myla_globals.is_superuser,
  enrolledCourses,
  isSuperuser: myla_globals.is_superuser,
  isLoggedIn: !!myla_globals.username,
  helpURL: myla_globals.help_url
})
```

Since the `user` object is passed to `AvatarModal` you'll be able to access the helpURL.

The reason I'm asking for this change is because I'm trying to reduce the number of components that directly access `myla_globals`. Since objects in JavaScript are passed by reference, there's a chance that some component could change some property in `myla_globals`, which could then cause widespread problems throughout the application, and it would be pretty tricky to debug as well.

vjcao: Thanks! I will change this.

Issue #687, opened 2019-09-06
make the `course_enabled` api as configurable option for making it public/protected API


pushyamig:
At Umich we have made `course_enabled API (list courses view options enabled/disabled) as Public. The purpose of the API was to enable the Myla visuals Links( likes /assignments) in canvas. This approach might not be True with other universities, so making this a configuration Defaulting to False via `env.json` and removing it from Django-rules set.

jennlove-um:
@pushyamig How can I test this one?

jennlove-um:
Testing passes with configuration turned on and off.

PR #699, opened 2019-09-16
Issue 687

vjcao:
`COURSES_ENABLED` is a environmental variable in `env.json` that can be set as `true` or `false` to enable or disable the api for public view. I also made it so that superusers will always be able to view what the api returns.

Addresses issue #687


dashboard/settings.py
```diff
@@ -63,6 +63,9 @@
 # Only report on the default database
 WATCHMAN_DATABASES = ('default',)
 
+# courses_enabled api
+COURSES_ENABLED = ENV.get('COURSES_ENABLED', True)
```
pushyamig: default to False

dashboard/views.py
```diff
@@ -639,19 +642,23 @@ def logout(request):
...
-    for cvo in CourseViewOption.objects.all():
-        data.update(cvo.json())
-
-    callback = request.GET.get('callback')
-    # Return json
-    if callback is None:
-        return HttpResponse(json.dumps(data), content_type='application/json')
-    # Return jsonp
+    current_user_is_admin = request.user.is_superuser
+    if COURSES_ENABLED or current_user_is_admin:
```
pushyamig: I think you can remove the admin user check. It is not needed.

PR #701, opened 2019-09-17
Update CONTRIBUTING.md

justin0022:
Fixed typo


pushyamig:
@justin0022 I am merging this now, but in future can you create PR from your Forked branch rather than creating branch on Main repo? Branches to main repo should be created more like a feature request like Discussion view, for minor issues like this we don't clutter up the branches here if we forget to remove it.

justin0022:
Oh good point, sorry @pushyamig. I used the editor built into Github to edit the readme, and I guess since I have write permissions to this repo it defaults to branching from this repo instead of forking to mine.

PR #703, opened 2019-09-18
I542 outlier detection

### LINKED ISSUE #542: Grade Distribution - Hide outliers on the grade distribution

pushyamig:
Fixes #542 
The Idea is to bin the least 5 scores so that low-grade student grades are not exposed.
histogram binning is by 2 [ [0,2], [2,4], [4,6], …..] each item in the list starting number is inclusive and second is exclusive.
  1.  This is the case where the lowest 5 will be binned together in one bin
Actual distribution: `[69.79, 80.0, 80.5, 88.21, 88.79, 92.71, 92.71, 92.71, 93.14, 94.43, 95.71, 95.71, 96.57, 97.79, 97.86]`
Fifth item: `88.79`
Next of Fifth item: `92.71`
Binning Distribution: `[88.79, 88.79, 88.79, 88.79, 88.79, 92.71, 92.71, 92.71, 93.14, 94.43, 95.71, 95.71, 96.57, 97.79, 97.86]`
2. This is the case where last 5 grades are not the binned number instead it will be binned until 95.89 this is due to histogram binning is by count of 2
Actual  Distribution: `[90.77, 93.09, 93.42, 94.85, 94.87, 94.88, 94.9, 95.55, 95.89, 96.28, 96.4, 96.47, 96.49, 96.68]`
Fifth item: `94.87`
Next of Fifth item: `94.88`
Actual binned number: `95.89`
Binning Distribution: `[95.89, 95.89, 95.89, 95.89, 95.89, 95.89, 95.89, 95.89, 95.89,96.28, 96.4, 96.47, 96.49, 96.68 ]`
Screen Shot 2019-09-18 at 3 55 35 PM
Screen Shot 2019-09-18 at 3 56 18 PM

3.  The default is to show the MyGrade line and moved the `Show my grade` to top right corner, but when user toggles checkbox Snackbar is still showing toward the bottom left. I don't know if we need to move that. Thoughts?

4.   Defaulting to not to show student count on the bars and provided a configurable option per course to enable that. Ideally, we are starting to channel away for AB testing. If we prefer AB testing on lot of views with different features then these can be separated to its own table. But Keeping it simple to `course` table
Screen Shot 2019-09-18 at 4 35 20 PM.

5.  Event logging if admin enabled showing bar counts in Grades distribution. From the Django-admin under Eventlong for action `VIEW_GRADE_DISTRIBUTION`. 
```
{
  "course_id":17700000000122343,
  "show_number_on_bars":false
}
```

6. Event logging toggling of `show my grade` option. Can be viewed as action `VIEW_SET_DEFAULT`
```
{
  "course_id":17700000000122332,
  "default_type":"grade",
  "default_value":true
}
```


assets/src/__tests__/testData/d3TestData.js
```diff
@@ -1,108 +1,9 @@
...
-      'data': 4
-    },
-    {
-      'label': 'South Korea',
-      'data': 9
-    }
-  ]
-)
 
-export const scatterplotData = Object.freeze(
```
pushyamig: Removing unused one

assets/src/containers/GradeDistribution.js
```diff
@@ -10,7 +10,7 @@ import Histogram from '../components/Histogram'
 import Spinner from '../components/Spinner'
 import Table from '../components/Table'
 import UserSettingSnackbar from '../components/UserSettingSnackbar'
-import { average, roundToOneDecimal, median } from '../util/math'
```
pushyamig: Mean and average grade is provided from BE hence removing

assets/src/containers/GradeDistribution.js
```diff
@@ -37,14 +37,14 @@ function GradeDistribution (props) {
...
+  const [showGrade, setShowGrade] = useState(true)
 
   useEffect(() => {
     if (userSettingLoaded) {
       if (isObjectEmpty(userSetting.default)) {
-        setShowGrade(false)
+        setShowGrade(true)
       } else {
-        setShowGrade(userSetting.default !== 'False')
+        setShowGrade(userSetting.default === 'False' ? false : true)
```
pushyamig: logic changed to show My grades by default

assets/src/containers/GradeDistribution.js
```diff
@@ -62,51 +62,52 @@ function GradeDistribution (props) {
...
               'There are no grades yet for you in this course'
           }</strong>
-        ] : []
+        ] : [],
     ]
 
     const gradeCheckbox = !user.admin ?
       <> {userSettingLoaded ?
-        <> {'Show my grade'}
+        <> <Typography align='right'>{'Show my grade'}
```
pushyamig: This is how I am right aligning the `show my grade` but this takes added space and pushes the histogram little down. Any better way of doing this?

dashboard/admin.py
```diff
@@ -48,7 +48,7 @@ def clean(self):
 class CourseAdmin(admin.ModelAdmin):
     inlines = [CourseViewOptionInline, ]
     form = CourseForm
-    list_display = ('id', 'canvas_id', 'name', 'term', '_courseviewoption')
+    list_display = ('canvas_id', 'name', 'ab_test_course', 'term', '_courseviewoption')
```
pushyamig: Removed the Canvas long id from the admin display list as it was not needed

dashboard/models.py
```diff
@@ -165,6 +165,8 @@ class Course(models.Model):
     name = models.CharField(max_length=255, verbose_name="Name")
     date_start = models.DateTimeField(verbose_name="Start Date and Time", null=True, blank=True)
     date_end = models.DateTimeField(verbose_name="End Date and Time", null=True, blank=True)
+    ab_test_course = models.BooleanField(blank=False, null=False, default=False, verbose_name=
+                                         "AB test course (For showing grades count on bars)")
```
ssciolla: I have a couple thoughts on the `ab_test_course` field. Could we name it something that isn't about the AB test? It seems like that is a temporary description of what the field does; I'd prefer something like `show_grade_counts`.

Also, how would you feel about moving this field to the `CourseViewOption` model? It seems like this is more about how a MyLA course is presented (or viewed) than about the course itself. Thoughts?

dashboard/admin.py
ssciolla: This is still showing up in the Django Admin, I think. I believe we also need to remove `id` from the `readonly_fields`.

ssciolla: Hm, scratch that. When you remove `id` from `readonly`, it starts showing up as an editable field. There must be another way to remove it entirely...

ssciolla: Yeah, I think the way to do it would be to make `id` on the `Course` model `editable=False` and then run a migration. Then you can remove `id` from `readonly` and it doesn't show up.

assets/src/components/d3/createHistogram.js
```diff
@@ -2,10 +2,34 @@ import * as d3 from 'd3'
 import { adjustViewport } from '../../util/chart'
 import { roundToOneDecimal } from '../../util/math'
 
-function createHistogram ({ data, width, height, domElement, xAxisLabel, yAxisLabel, myGrade, maxGrade = 100 }) {
+function createHistogram ({ data, width, height, domElement, xAxisLabel, yAxisLabel, myGrade, maxGrade = 100,
+                            showNumberOnBars = false}) {
   const margin = { top: 20, right: 20, bottom: 50, left: 40 }
   const [aWidth, aHeight] = adjustViewport(width, height, margin)
 
+  // data usually will be [50.6, 50.6, 50.6, 50.6, 50.6, 74.28, 74.52, 75.89, 76.69,,.,.,.,.] the first 5 or few will be binning grades
```
ssciolla: The comment here could be clearer.

assets/src/containers/GradeDistribution.js
ssciolla: Did we decide we didn't like it on the left hand side under the grade statistics?

dashboard/views.py
```diff
@@ -625,6 +641,57 @@ def are_weighted_assignments_hidden(course_id, df):
             return True
```
ssciolla: Variable naming could be more consistent and clear in `find_binning_grade_value`. If `item` and `element` are interchangeable, then stick to one?

dashboard/views.py
```diff
@@ -625,6 +641,57 @@ def are_weighted_assignments_hidden(course_id, df):
...
 
+def find_binning_grade_value(grades):
+    fifth_item = grades[4]
+    next_to_fifth_element = grades[5]
+    if next_to_fifth_element - fifth_item > 2:
+        binning_grade_value, index = fifth_item, 4
+    else:
+        binning_grade_value, index = binning_logic(grades, fifth_item)
+    return binning_grade_value, index
+
```
ssciolla: @jonespm  had me start using `namedtuple` when returning tuples from functions in [this PR](https://github.com/tl-its-umich-edu/my-learning-analytics/pull/652).

dashboard/views.py
```diff
@@ -625,6 +641,57 @@ def are_weighted_assignments_hidden(course_id, df):
...
+    :param grades: sorted in asc
+    :param fifth_item_in_list:
+    :return: max grade in the binned list, length of binned grades
+    """
+    binning_list=grades[:5]
+    for grade in grades[5:]:
+        if check_if_grade_qualify_for_binning(grade, fifth_item_in_list):
+            binning_list.append(grade)
+    return max(binning_list), len(binning_list)
```
ssciolla: Seems like it's unnecessary to check all grades above the fifth; if the grades are sorted by value, couldn't you just use a `return` once `check_if_grade_qualify_for_binning` returns `False`?

dashboard/views.py
```diff
@@ -625,6 +641,57 @@ def are_weighted_assignments_hidden(course_id, df):
...
+
+
+def is_odd(num):
+    if num % 2 == 0:
+        return False
+    else:
+        return True
+
+
+def check_if_grade_qualify_for_binning(grade, fifthElement):
```
ssciolla: Maybe rename `check_if_grade_qualifies_for_binning`?

assets/src/components/d3/createHistogram.js
```diff
@@ -84,21 +122,41 @@ function createHistogram ({ data, width, height, domElement, xAxisLabel, yAxisLa
...
       .attr('y2', 0)
       .attr('stroke', 'darkorange')
       .attr('stroke-width', '2')
     svg.append('text')
-      .attr('x', x(myGrade) - 110)
-      .attr('y', margin.top-5)
-      .text(`My Grade: ${roundToOneDecimal(myGrade)}%`)
+      .attr('x', myGrade < 5 ? x(myGrade) + 3 : x(myGrade) - 110)
+      .attr('y', margin.top + 15)
+      .text(`My Grade:${roundToOneDecimal(myGrade)}%`)
```
ssciolla: Add a space so it's "My Grade: "? Also, sometimes the label covers part of the histogram. Would it be possible to add logic such that based on where the line is in the distribution, it would appear on the left or right of the line? Maybe that's a feature request for the future.

dashboard/views.py
```diff
@@ -301,27 +301,43 @@ def grade_distribution(request, course_id=0):
...
+
+    df['graph_lower_limit'] = df['current_grade'].min()
 
     # json for eventlog
     data = {
-        "course_id": course_id
+        "course_id": course_id,
+        "show_number_on_bars": df['show_number_on_bars'].values[0]
     }
     eventlog(request.user, EventLogTypes.EVENT_VIEW_GRADE_DISTRIBUTION.value, extra=data)
```
ssciolla: I think we could simplify what data is passed to the front end, rather than passing the whole DataFrame. Looking in the browser on what's returned from this view, there are a number of columns that are always the same value. Maybe we could remodel the Response to use something like the following structure:

```
{
   "summary": {
      "current_user_grade": XX.X,
      "tot_students": XX,
      "show_number_on_bars": true,
      "grade_avg": XX.X,
      "median_grade": XX.X,
      "grade_upper_limit": XXX,
      "grade_lower_limit": XX.X
   },
   "grades": [XX.X, XX.X, ...]
}
```

justin0022:
Very cool! I'll review this by end of day today.

assets/src/containers/GradeDistribution.js
justin0022: I think `setShowGrade(userSetting.default === 'False' ? false : true)` and `setShowGrade(userSetting.default === 'False' )` are logically equivalent.

justin0022:
Screen Shot 2019-09-19 at 9 59 26 AM

Here the `56%` isn't correctly aligned with the x-axis (it's more aligned to `54%`).

justin0022:
> 1. The default is to show the MyGrade line and moved the `Show my grade` to top right corner, but when user toggles checkbox Snackbar is still showing toward the bottom left. I don't know if we need to move that. Thoughts?

I think we should keep the Snackbar showing consistently to one side - the bottom left I think works well.

pushyamig:
> Here the 56% isn't correctly aligned with the x-axis (it's more aligned to 54%).

The number on the dashed line shown is based on the First item in the normal distribution list, in this case, the distribution would be[4,5,10,20,54, 56.8, 75, 80] and the binned distribution is [54,54,54,54,54, 56.8, 75, 80,....]. I am trying to say the binned distribution is < than 56 ( which is normal distribution). It is not based on ticks but based on grade value. That's what asked to be shown.

dashboard/views.py
pushyamig: I like this one too... But just did not put the effort to do it

dashboard/views.py
pushyamig: Yes, I agree. I will return on First False.

justin0022:
Oh I see. Would it be clearer to align the dashed line to the 56% on the x-axis?

dashboard/models.py
pushyamig: I think @jennlove-um also asked it to be like that but I felt if other universities are looking at it from admin felt won't make sense to then why it is placed so I have name as AB test and extra description what it is.
`CourseViewOption` is more for enabling or disabling the views. It's at higher level. I think We might create separate table saying `Grade_distribution_feature_testing` and have columns like `show_grade_count`, 'Show_y-axis' etc. 
But now I will change it to `show_grade_counts` since you are the second person to ask I should listen

assets/src/containers/GradeDistribution.js
pushyamig: We decided on top right last time, but will find out today if team like the current positioning

assets/src/containers/GradeDistribution.js
pushyamig: Changed it.

dashboard/admin.py
pushyamig: Thanks @ssciolla . Removed the long canvas id from the Django admin view

dashboard/models.py
pushyamig: I have changed it to `show_grade_counts`

dashboard/views.py

assets/src/components/d3/createHistogram.js
pushyamig: Done, Although I just simplified did not put in more details here.

dashboard/views.py

dashboard/views.py

assets/src/components/d3/createHistogram.js
pushyamig: have added spacing, But your idea for changing the `Mygrade` position based on bar height, let park it for now.

dashboard/views.py
pushyamig: I have created the issue https://github.com/tl-its-umich-edu/my-learning-analytics/issues/707. This can be taken up separetely

pushyamig:
> Oh I see. Would it be clearer to align the dashed line to the 56% on the x-axis?

I think the design team here is ok with it, The problem is it is based on grades distribution for a class and it is difficult to have one solution fits everybody. We are thinking put it to students and get review and go from there.

pushyamig:
I think I resolved all the issues and replied for all the comments.

PR #709, opened 2019-09-20
Fixes #708 Only support gunicorn, remove runserver

### LINKED ISSUE #708: Switch to only supporting gunicorn

Currently the service supports both Gunicorn and Django Runserver because of issues supporting PTVSD debugging. Recently [issues were fixed in VSCode](https://github.com/microsoft/vscode-python/issues/2138) to support Gunicorn. So we should just switch to using it.

Remove the code for runserver, only have the Gunicorn.

jonespm:


jonespm:
I need to close this for now, it isn't as reliable as the old method

Issue #722, opened 2019-09-26
Migrate settings.py settings over to a live-settings package


jonespm:
Thank you for contributing to this project!

## Describe your problem or feature you'd like added
The settings.py is starting to get pretty unwieldy. It's hard to know when new settings come in and having to restart to set these isn't always ideal. We should move to a global settings package.

## Describe the solution you'd like
There are a number of packages listed here that support this. We'd want it to be in the Django Admin

https://djangopackages.org/grids/g/live-setting/

This one is used the most, has the most stars and updated recently: https://github.com/jazzband/django-constance

This one also looks good
https://github.com/EliotBerriot/django-dynamic-preferences

jonespm:
Constance was added in #1176, an issue #1200 is made to follow up on this.

PR #723, opened 2019-09-26
Add ENV_JSON settings support

andrew-gardener:
Allow loading settings from a JSON encoded environment variables instead of only by json file. This helps support certain deployment situations where it might be hard/impossible to mount a file into the container

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,8 @@ These instructions will get a copy of MyLA up and running on your local machine 
 1. Download the latest SQL file from this link: https://drive.google.com/drive/u/0/folders/1Pj7roNjRPGyumKKal8-h5E6ukUiXTDI9.
 1. Load database with data. `docker exec -i student_dashboard_mysql mysql -u student_dashboard_user --password=student_dashboard_pw student_dashboard < {name of sql file}`
 
+You may also optionally place the json settings directly into the `ENV_JSON` environment variable if your deployment environment doesn't easily support mounting the `env.json` file into container. When using `ENV_JSON` put the entire contents of `env.json` into it as single line string.
+
 #### Logging in as admininstrator
 1. Navigate to http://localhost:5001/ and log in as:
     ```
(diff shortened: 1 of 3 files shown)


jonespm:
Can you add a sample env file, README or Wiki about how someone might use ENV_JSON? We'd like to work on #722 this summer and get rid of most of file based settings as they worked out for the last year but are getting cumbersome.

andrew-gardener:
Just to explain our situation a bit, we're planning on deploying on [AWS Fargate](https://aws.amazon.com/fargate/) which currently doesn't allow persistent volume mounting (though it is on the [roadmap](https://github.com/aws/containers-roadmap/issues/53)). 

This means that it will need a bit of a workaround to load the json file settings easily. 

loading it is relatively easy, just dump the expected file contents into an environment variable. ex:

`docker-compose.yml`
```
...
  web:
    build:
      context: .
      args:
        TZ: ${TZ}
    command: bash -c "./start.sh"
    volumes:
      - .:/code
      - /code/node_modules/
      - /code/static/
      - ${HOME}/mylasecrets:/secrets
    ports:
      - "5001:5000"
      - "3000:3000"
      - "3001:3001"
    depends_on:
      - mysql
      - webpack_watcher
    env_file:
      - .env
    environment:
      - GUNICORN_RELOAD=True
      - >-
        ENV_JSON={
            "DJANGO_SECRET_KEY": "<Generate Secret Key>",
            ...
            "NLP_MODEL_DIRECTORY": "/nlp_models"
        }
    container_name: student_dashboard
...
```

I'll add something to the readme

andrew-gardener:
in regards to #722, I could also just drop this PR and use this as a stopgap for our deployment until its #722 completed

jonespm:
Ah yeah, Openshift has a YAML for deployment too so we've got a setting like below and a persistent volume (ConfigMap) with this json.
```
        - env:
            - name: ENV_FILE
              value: /secrets/app/env_test.json
```

This might be useful to other schools that aren't able to create the json file. Though we'd like to move most of the contents of that json over to Django admin via one of those settings packages. I think we'll get that resolved this semester.

jonespm:
Yeah that makes sense, I wouldn't expect #722 would be done until November or so, I just thought about it today and I personally won't have time for a few weeks to work on it but someone else here might pick it up.

Issue #725, opened 2019-09-27
Resources Accessed: Message displays saying no data is available when the user changes their setting


jennlove-um:
# Expected behavior (A clear and concise description of what you expected to happen) : 
User can select a new setting from the filter drop-down and the filtered results and week timeline continue to display correctly when the setting is saved.

## Describe the bug (Tell us what happens instead of the expected behavior) :
After clicking Save My Setting, 2 unexpected results are happening: 
1. The timeline disappears from the top of the page  
2. The filtered list of results disappears and a message displays "Resource data for your selections is not available.  (even though results are available)

Workaround: Refresh the page.

## Steps to Reproduce :
1.  User goes to Resources page
2. User selects a filter on the Resources page (make sure resources still show in the filtered list)
3. User clicks Save My Setting

jennlove-um:
@justin0022  I assigned this to @zqian, but if you want to look at it, let me know.  Thanks!

jennlove-um:
![MyLA_RA_SettingError](https://user-images.githubusercontent.com/39634416/65805057-2d97c600-e152-11e9-8e16-e250b238309f.gif)

justin0022:
> @justin0022 I assigned this to @zqian, but if you want to look at it, let me know. Thanks!

I can take a look but it may take me a while - the `AssignmentPlanning` redesign is taking up a large chunk of my time! If you want me to prioritize this just let me know.

jennlove-um:
Thanks, @justin0022  - no need to look at it.  Zhen is taking this one one.

pushyamig:
Test Plan:
I have refactored the login to the resource, so test it as a whole functionality of the view not just the bug.

jennlove-um:
Resources Accessed functionality passes all tests in test plan.

Issue #642, opened 2019-08-07
Refactor Resources Accessed to use userSetting hooks


justin0022:
This will keep the way that user settings are fetched and stored consistent with the Grade Distribution view.

jennlove-um:
This can be closed.  The user settings hooks are being used in Resources Accessed now.

PR #730, opened 2019-09-30
issue #725 changes to ResourceAccessed.js

zqian:
the current approach uses the useSetUserSetting.js, similar to other two views.

related the [issues-642](https://github.com/tl-its-umich-edu/my-learning-analytics/issues/642)

### DIFF
--- a/assets/src/containers/ResourcesAccessed.js
+++ b/assets/src/containers/ResourcesAccessed.js
@@ -1,5 +1,7 @@
 import React, { useEffect, useState } from 'react'
 import { withStyles } from '@material-ui/core/styles'
+import useSetUserSetting from '../hooks/useSetUserSetting'
+import useUserSetting from '../hooks/useUserSetting'
 import Paper from '@material-ui/core/Paper'
 import Grid from '@material-ui/core/Grid'
 import Typography from '@material-ui/core/Typography'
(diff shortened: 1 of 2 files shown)


assets/src/containers/ResourcesAccessed.js
```diff
@@ -104,32 +104,25 @@ function ResourcesAccessed (props) {
...
+          },
+          credentials: 'include',
+          body: JSON.stringify({resource: gradeRangeFilter}),
+          method: 'PUT'
+        }).then(handleError)
+          .then(res => res.json())
+          .then(res => {
+            setResponse(res)
+            setLoaded(true)
+        }).catch(error => setError(error.message))
```
ssciolla: Is `setError` available here? It seems to be defined in `useFetch.js` and then not defined or imported within this file.

ssciolla: Yeah, I may be missing something; I'm not super knowledgeable about hooks. I guess I have a similar question about `setResponse` and `setLoaded`.

justin0022: None of them are imported or available - they are all undefined. 

I recommend using [standard](https://standardjs.com/), a JavaScript linter. It'll reveal any undefined functions/variables.

zqian: OK. I made another iteration to use useSetUserSetting.js. Please review.

justin0022:
I'm not seeing the week range slider in my view:
Screen Shot 2019-10-01 at 9 03 24 AM
 
I was actually working on the refactor of this view a couple of weeks ago, but I switched to working on the Assignment Planning redesign. For what it's worth, here's the progress I made to refactoring this view: https://github.com/justin0022/my-learning-analytics/blob/refactor-resources/assets/src/containers/ResourcesAccessed.js

pushyamig:
@justin0022 The problem seems to be caused `get_user_default_selection` called each time after the set_user_default_selection. I am not sure what changed in between it was not the behavior before. I am taking a look at it. this may be due to the way `useFetch` is implemented now not sure. This is not happening with Assignmentview since the refactored it

pushyamig:
Closing as this is fixed  in #735

PR #732, opened 2019-09-30
Fixes #731 - LTI launch should use custom_canvas_user_login_id and

### LINKED ISSUE #731: LTI Launch should use custom_canvas_user_login_id for ID and support first/last name

By default the LTI launch is using `lis_person_sourcedid` and this *works* but I think it should be using `custom_canvas_user_login_id` which is the actual userid and works better. 

We also should support setting the first/last name if provided.

jonespm:
support first/last name


jonespm:
Thanks, I put extra code in to use either the lis_person_sourcedid or this one, but this one works better with how the UI is implemented.

PR #735, opened 2019-10-02
Fixing the disappearing slider when defaults set

### LINKED ISSUE #725: Resources Accessed: Message displays saying no data is available when the user changes their setting

pushyamig:
Fixes #725
1. I have refactored the code and used the same naming convention for hook as in AssignmentPlanning view for making it easier to understand login in both views.
2. Added the Snackbar, separated the get/set default logic outside of `ResourcesAccessed.js` as in `AssignmentPlanning.js`
3. once thing I did not do is the content call is still inline in `useEffect` instead of `api.js`. I felt it can be handled separately as the logic is different from Assignment view

### DIFF
--- a/assets/src/containers/AssignmentPlanning.js
+++ b/assets/src/containers/AssignmentPlanning.js
@@ -84,6 +84,7 @@ function AssignmentPlanning (props) {
     if (userSettingLoaded) {
       if (isObjectEmpty(userSetting.default)) {
         setAssignmentGradeFilter(0)
+        setUserSavedFilterSetting(0)
       } else {
         setAssignmentGradeFilter(Number(userSetting.default))
         setUserSavedFilterSetting(Number(userSetting.default))
--- a/assets/src/containers/ResourcesAccessed.js
+++ b/assets/src/containers/ResourcesAccessed.js
@@ -9,16 +9,16 @@ import FormGroup from '@material-ui/core/FormGroup'
 import FormControlLabel from '@material-ui/core/FormControlLabel'
 import Select from '@material-ui/core/Select'
 import MenuItem from '@material-ui/core/MenuItem'
-import { useUserSettingData } from '../service/api'
 import { handleError, defaultFetchOptions } from '../util/data'
-import Cookie from 'js-cookie'
 import AlertBanner from '../components/AlertBanner'
-import WarningBanner from '../components/WarningBanner'
 import RangeSlider from '../components/RangeSlider'
 import ResourceAccessChart from '../components/ResourceAccessChart'
 import Spinner from '../components/Spinner'
 
-import { type } from 'os';
+import useSetUserSetting from '../hooks/useSetUserSetting'
+import { isObjectEmpty } from '../util/object'
+import useUserSetting from '../hooks/useUserSetting'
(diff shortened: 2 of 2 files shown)


assets/src/containers/ResourcesAccessed.js
```diff
@@ -236,7 +229,6 @@ function ResourcesAccessed (props) {
       )
     }
   }
-  if (error) return (<WarningBanner/>)
```
pushyamig: Removed this as it is not needed, when we put the content call in `api.js` then we can handle this. The BE always returns empty response so logic is baked in to display appropriate message to user

assets/src/containers/AssignmentPlanning.js
```diff
@@ -84,6 +84,7 @@ function AssignmentPlanning (props) {
     if (userSettingLoaded) {
       if (isObjectEmpty(userSetting.default)) {
         setAssignmentGradeFilter(0)
+        setUserSavedFilterSetting(0)
```
pushyamig: During the initial load of the assignment view, the display needs to say `My Current Setting` instead it was showing ` `Remember My Setting`, this fixes that

assets/src/containers/ResourcesAccessed.js
```diff
@@ -99,41 +106,28 @@ function ResourcesAccessed (props) {
...
-            setDefaultValue(gradeRangeFilter)
-            return
-          }
-          setDefaultLabel(settingNotUpdated)
-        }).catch(err => {
-        setDefaultLabel(settingNotUpdated)
-      })
+  // if user setting is saved, don't show checkbox and sync userSavedFilterSetting with assignmentGradeFilter
+  useEffect(() => {
+    console.log()
```
justin0022: Other than the `console.log()` looks good to me

assets/src/containers/ResourcesAccessed.js
```diff
@@ -99,41 +106,28 @@ function ResourcesAccessed (props) {
...
+    if (userSettingSaved) {
+      setShowSaveSetting(false)
+      setUserSavedFilterSetting(resourceGradeFilter)
     }
-  }
+  }, [userSettingSaved])
 
   useEffect(() => {
+    console.log("Setting the slider range")
```
justin0022: Other than the `console.log()` looks good to me

assets/src/containers/ResourcesAccessed.js
pushyamig: Done

assets/src/containers/ResourcesAccessed.js
pushyamig: Done

Issue #728, opened 2019-09-27
Assignment Planning view is showing grades for assignments that should be hidden


jennlove-um:
## Expected behavior (A clear and concise description of what you expected to happen) : 
Student goes to Assignment Planning view and assignments that have been released in Canvas have grades showing in MyLA.  Assignments that have not had grades released in Canvas do not show up in MyLA. 

## Describe the bug (Tell us what happens instead of the expected behavior) :
Grades for assignments that have not been released to the student in Canvas are showing up as graded in MyLA.  

Grades for assignments that have been released to the student in Canvas are showing up as ungraded in MyLA.

## Steps to Reproduce :
1.  Select a course that has assignments that have been released to all students and assignments that have not had grades posted for all students.
2. Login to MyLA and go to the Assignment Planning page
3. Review graded and ungraded assignments.  No grades for assignments should display for the student if that student can't see the grade in Canvas.

pushyamig:
fixed by using `posted_by` #727 .

jennlove-um:
Testing passes.  Students are correctly seeing grades on assignments that they can see as graded in Canvas.

Testing was done for at least one student in all courses scheduled for fall 2019 in addition to test courses.

jennlove-um:
since testing passes and this issue was resolved by #727, I'm closing this one as well

PR #743, opened 2019-10-04
#727 student submission only pulled when posted_At not null

### LINKED ISSUE #727: Use the Posted date in the Submission table to determine whether to show assignment grade
Use the posted_at attribute on the submissions_dim to determine whether the student should see an assignment grade in MyLA.

If there is a posted_at date, then the grade has been released to the student and can be displayed in MyLA's Assignment Planning page.

If there is no posted_at date, then the grade has not been released to the student and should not be in MyLA for the assignment.

pushyamig:
Fixes #727, #728
1. Only Pulling student submission grades when `posted_at not null and posted_at < current_date `. This is the change that got introduced with the new Canvas grade book previously used to determine this by `muted` column in the `assignment_dim` table which seems to be False when the instructor grades the assignment and did not release to students. More info #728 
2. I have added a new column `grade_posted_local_date` for tracking purposes, after discussion with @jennlove the `posted_at` grade is not displayed from Gradebook UI from the instructor and student view. So we don't have a way to test when was grade posted. So I am adding it to our DB for debugging. I also felt if it is in local date it might be easier for testing purposes, thoughts??
3. The `posted_at` is not shown in the Assignment view. Although with Assignment redesign the BE has a different setup so might need to check if everything works.
4. Current code doesn't support if schools use old canvas grade book, I don't know how important is that with participating MyLA pilot school.


dashboard/cron.py
```diff
@@ -385,14 +385,14 @@ def submission(self):
...
-                             all_assign_sub as (select submission_id AS id, assignment_id AS assignment_id, course_id, global_canvas_id AS user_id, round(published_score,1) AS score, graded_at AS graded_date from assign_sub_time order by assignment_id)
-                             select f.*, f1.avg_score from all_assign_sub f join (select assignment_id, round(avg(score),1) as avg_score from all_assign_sub group by assignment_id) as f1 on f.assignment_id = f1.assignment_id
-                             """
+                enrollment as (select  distinct(user_id) from enrollment_dim where course_id = '{data_warehouse_course_id}' and workflow_state='active' and type = 'StudentEnrollment'),
+                sub_with_enroll as (select sf.* from sub_fact sf join enrollment e on e.user_id = sf.user_id),
+                submission_time as (select sd.id, sd.graded_at, sd.posted_at at time zone 'utc' at time zone '{settings.TIME_ZONE}' as grade_posted_local_date from submission_dim sd join sub_fact suf on sd.id=suf.submission_id where posted_at is not null and posted_at < getdate()),
```
zqian: >>> "sd.posted_at at time zone 'utc' at time zone '{settings.TIME_ZONE}' as grade_posted_local_date"

why there is a repeated "at time zone" phrase?

pushyamig: Here is [Documentation](https://popsql.com/learn-sql/postgresql/how-to-convert-utc-to-local-time-zone-in-postgresql/) for it. I just took the approach from already code [here](https://github.com/tl-its-umich-edu/my-learning-analytics/blob/master/dashboard/cron.py#L363), Probably implemented by me in past

jonespm: It looks like you're changing this timezone from UTC to local time? Is there any reason you're doing this? It seems better just to keep everything in the database in UTC.

pushyamig: That time has no use from the assignment view. The time when the grade was released seems to be not visible from Canvas UI for instructor/student. So I added that for tracking. I felt it is easier for @jennlove-um to test it time in local date rather than UTC.

pushyamig: for `graded_at` we don't need it local time format. I used that as a clue if assignment is graded or not for color representation in UI

jonespm: Ok, I just suggested adding a comment to the model that the field was shifted to local time as part of the cron. If anyone used the model and tried to get the time it's going to be incorrect. 

We could store it as UTC (which would match the UDW) but then anyone using it would need to convert it in their queries.

dashboard/models.py
```diff
@@ -281,7 +281,8 @@ class Submission(models.Model):
     user_id = models.BigIntegerField(verbose_name="User Id")
     score = models.FloatField(blank=True, null=True, verbose_name="Score")
     graded_date = models.DateTimeField(blank=True, null=True, verbose_name="Graded DateTime")
-    grade_posted_local_date = models.DateTimeField(blank=True, null=True, verbose_name="Posted Grade in local DateTime")
+    # This is used for tracking of grade posted date and not used in Assignment view hence making it CharField
+    grade_posted_local_date = models.CharField(max_length=255,blank=True, null=True, verbose_name="Posted Grade in local DateTime")
```
pushyamig: Added the comment purpose of this field and make it `CharField` instead of `DateTime`

dashboard/cron.py
pushyamig: Done as per suggestion

pushyamig:
merging it now

PR #745, opened 2019-10-04
Fixes #744 Files created within the last day are just attachment

### LINKED ISSUE #744: Files created within the last day just are "attachment"

The current process for pulling files in the cron is
- Get all resources for the course from BigQuery
- Update the file names from UDW

This was because previously the Caliper events didn't contain the file names. The problem is that UDW is delayed so new files will come over with the name "attachment"

However recently Canvas is adding the file names in the event, so we can use this as part of the query. This is in 

`“$.object.extensions[‘com.instructure.canvas’].asset_name”
`
This is just a documentation update

jonespm:
Just changes the field in the Caliper event used for the name


config/env_sample.json
```diff
@@ -178,7 +178,7 @@
...
+            "query": "SELECT 'canvas' AS resource_type, CAST(SUBSTR(JSON_EXTRACT_SCALAR(event, '$.object.id'), 35) AS STRING) AS resource_id, CAST(SUBSTR(JSON_EXTRACT_SCALAR(event, '$.membership.member.id'), 29) AS INT64) AS user_id, CAST(SUBSTR(JSON_EXTRACT_SCALAR(event, '$.group.id'),31) AS INT64) AS course_id, JSON_EXTRACT_SCALAR(event, '$.object.extensions[\\'com.instructure.canvas\\'].asset_name') as name, datetime(EVENT_TIME) as access_time FROM event_store.events where JSON_EXTRACT_SCALAR(event, '$.edApp.id') IN UNNEST(['http://m.canvas.umich.edu/' , 'http://umich.instructure.com/' ]) and type = 'NavigationEvent' and SUBSTR(JSON_EXTRACT_SCALAR(event, '$.object.id'),24,10) = 'attachment' and JSON_EXTRACT_SCALAR(event, '$.action') = 'NavigatedTo' and JSON_EXTRACT_SCALAR(event, '$.membership.member.id') is not null and SUBSTR(JSON_EXTRACT_SCALAR(event, '$.group.id'),31) IN UNNEST(@course_ids)",
```
zqian: Does this approach with previous records of file access?

jonespm: For previous caliper records this value will be **null** but still not cause any other issues.

This value will be updated by `update_canvas_resource`. There's nothing that checks that this value is set or not, that method just updates the value to match Canvas.

PR #756, opened 2019-10-08
Upgraded MUI to v4 and react-router to v5

### LINKED ISSUE #736: Upgrade material-ui to V4 from V3
Looks like a major upgrade with better theming support: https://medium.com/material-ui/material-ui-v4-is-out-4b7587d1e701

I'm going to test and see if there are any major breaking changes. If there are no major breaking changes I'll send in a PR.

justin0022:
Resolves #736.

Required a few changes, nothing major. I didn't notice any UI changes after the upgrade. It also removes the deprecation React warnings that showed up on console.

After this gets merged, I'm going to modify the way styles are added to components from the old `withStyles` to the new [`makeStyles` hooks api](https://material-ui.com/styles/basics/#hook-api).


pushyamig:
I have added this branch on https://dev-myla.tl.it.umich.edu/ for easier review. Quickly notices one difference the gap b/w icon and text is more.
Screen Shot 2019-10-09 at 8 52 20 AM

What is the major difference between v3 and v4?

justin0022:
Here's a blog post explaining the major differences: https://medium.com/material-ui/material-ui-v4-is-out-4b7587d1e701.

The big one is theming. It also supports Hooks, reducing the need for Higher Order Components, which cause a lot of DOM nesting. 

I've fixed the styling issues on Avatar now.

assets/src/containers/DashboardAppBar.js
```diff
@@ -16,7 +16,7 @@ const styles = theme => ({
     flexGrow: 1
   },
   button: {
-    margin: theme.spacing.unit
+    margin: theme.spacing()
```
ssciolla: What's the default here? 1 (times the 8 multiplier)? If it's 1, it may be more transparent to add the 1 as input. I was looking at [this doc page](https://material-ui.com/customization/spacing/#spacing), it wasn't clear to me.

pushyamig:
All the title header for the views have color contract issues
V4
Screen Shot 2019-10-14 at 9 24 31 AM
V3
Screen Shot 2019-10-14 at 9 24 43 AM
I don't know if this may be flagged.

To me this PR seems to be fine, reviewing this PR is like testing it I feel that is exhaustive.

ssciolla:
I think I agree with @pushyamig, in that there do seem to be small differences throughout but it may be difficult to find, review, and test all of them right now. I think if we fix the couple that we've identified here, we can merge and continue reviewing as we work on other features.

assets/src/containers/DashboardAppBar.js
justin0022: > All the title header for the views have color contract issues

I wonder if we need to import some typography: https://material-ui.com/components/typography/

justin0022:
> All the title header for the views have color contract issues

I wonder if we need to import some typography: https://material-ui.com/components/typography/

assets/src/containers/DashboardAppBar.js
justin0022: I'm not sure, I put in `1` and it didn't make a visual difference - I've committed that now.

ssciolla:
> > All the title header for the views have color contract issues
> 
> I wonder if we need to import some typography: https://material-ui.com/components/typography/

Typography can be managed on the theme-level; we may want to revisit this later in relation to theme work. For consistency, I feel we should try to set font color/size/typeface at as global a level as possible.

ssciolla:
@justin0022, I think @pushyamig and I agree this is ready to be merged. Were you still thinking about something?

justin0022:
HI @ssciolla, apologies for the late reply (I was away on vacation). I'm happy to have this merged in if you guys think it's ready to go.

ssciolla:
@justin0022, no worries, yeah, I think we should merge and start working with the new dependencies and file issues if we find bugs or problems.

PR #757, opened 2019-10-09
Fix snackbar sidedrawer issue

### LINKED ISSUE #705: Snackbar in grade distribution is triggered when sidebar is opened
## Expected behavior: 
Snackbar should only display when `Show my grade` is clicked. 

Snackbar is triggered when sidebar is opened/closed if the `Show my grade` is changed _before_ the sidebar is opened. 

## Steps to Reproduce :
1. Navigate to grade distribution.
2. Click `Show my grade`
3. Click sidebar (top left button)
4. Snackbar will open (but state will not change of the `Show my grade`).

justin0022:
Resolves #705 

The cause was subtle, caused by putting the `UserSettingSnackbar` component within the `buildGradeView` function. This means that every time GradeDistribution rerenders (including any time that the side drawer is opened), a _brand new_ instance of the  `UserSettingSnackbar` component is created, which causes the `useEffect` function within the component to lose track of state, making it render every single time instead of checking the dependency array. 

Took forever to debug 😂

### DIFF
--- a/assets/src/containers/GradeDistribution.js
+++ b/assets/src/containers/GradeDistribution.js
@@ -35,8 +35,7 @@ function GradeDistribution (props) {
     classes,
     disabled,
     courseId,
-    user,
-    sideDrawerState
+    user
   } = props
   if (disabled) return (<AlertBanner>The Grade Distribution view is hidden for this course.</AlertBanner>)
 
@@ -61,8 +60,6 @@ function GradeDistribution (props) {
     settingChanged,
     [showGrade]
   )
-
-  console.log(userSettingSaved)
 
   if (gradeError) return (<WarningBanner />)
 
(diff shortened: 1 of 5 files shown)


assets/src/containers/GradeDistribution.js
```diff
@@ -56,75 +61,87 @@ function GradeDistribution (props) {
...
-            {content}
+            {
+              gradeLoaded
+                ? <BuildGradeView />
+                : <Spinner />
+            }
+            <UserSettingSnackbar
+              saved={userSettingSaved}
+              response={userSettingResponse}
+            />
```
justin0022: Such a simple fix!

assets/src/containers/GradeDistribution.js
```diff
@@ -31,7 +31,12 @@ const styles = theme => ({
 })
 
 function GradeDistribution (props) {
-  const { classes, disabled, courseId, user } = props
+  const {
```
jonespm: This change seems like inconsistent style with the other components and containers. Is there a reason this style is changed/preferred? I know doesn't like multiple imports on the same line but this project ignores the [E401](https://lintlyci.github.io/Flake8Rules/rules/E401.html) lint suggestion in the backend. ;) Most examples of destructuring seem to keep everything on one line (as long as the line is short) and appears slightly easier to read for me.

justin0022: Good point, I'm reverting it back.

PR #760, opened 2019-10-14
Fixes #759 - Re-do Dockerfile to avoid directly pulling in previous

### LINKED ISSUE #759: Re-do Dockerfile to avoid directly pulling in previous stage

There's an issue with Openshift where the syntax of 

`FROM ... AS ... 
`
doesn't work if the AS target is built in a previous stage. We need to re-write the Dockerfile so it copies instead from the previous stage.

jonespm:
stages

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,8 +13,9 @@ RUN npm run prod && \
     rm -rf /usr/src/app/assets/src
 
 # build node libraries for production mode
-FROM node-webpack AS node-prod-deps
+FROM node:11.10-alpine AS node-prod-deps
 
+COPY --from=node-webpack /usr/src/app /usr/src/app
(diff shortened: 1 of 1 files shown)


jonespm:
This looks like it does the same thing and I think it will work for our build process. Just verifying a little more and going to test it. Previously it was just copying over everything from the previous stage/layer but I think we really only need the one directory.

jonespm:
Once Alpine 3.11 is released I'd like to switch over to using that. I've been waiting for pandas to be available as a package and it's only been in edge all summer. Then we can really look at trying to reduce the sizes here, over 300MB for just the base image. It's possible this could be slimmed down a little before that but I figured why bother until we can do that.

Issue #684, opened 2019-09-05
Add link to course to Courses admin page


jonespm:
Thank you for contributing to this project!

## Describe your problem or feature you'd like added
I feel like we've talked about this but I don't see the issue for this one. It would be nice to have a link that goes to the course somewhere in the admin UI. Currently we have to change this around. I remember having to change around the courses URL to retrieve this a while ago. The URL can be retrieved with

`url = reverse('courses', kwargs={ 'course_id': canvas_course_id })`

## Describe the solution you'd like
Maybe [something like this answer](https://stackoverflow.com/a/2156277) to add a custom URL to the Admin table for Courses? A button next to "History" after clicking on a Course would also be good, I'm not entirely sure how to add that.

jonespm:
@ssciolla Did you do any work on this one? I'd thought it was implemented for some reason? :)

ssciolla:
We talked about this, yeah, but I was sidetracked by some other things. The ideal solution is to use `get_absolute_url`, since I believe it will automatically create a button similar to the one we have now for "Login As" when logging in as a user. I remember encountering some difficulty that made me start thinking about the Django version upgrade, but I will return to this.

ssciolla:
So adding the following code to the `Course` model works (i.e. adds a "VIEW ON SITE" button to the Course's edit page in the Admin). 

```
    def get_absolute_url(self):
        return reverse('courses', kwargs={'course_id': self.canvas_id})
```

However, you have to change the one `Site` record (referred to in `settings.py` by index in the `SITE_ID` variable) so its domain is `localhost:5001`, or the appropriate domain for a different environment. Does this "solve" the issue, or should we try to find some way to dynamically set the domain? It seems like there might be a way to do it, but I haven't found too much online (which makes me think I'm searching with the wrong terms).

PR #761, opened 2019-10-14
Adding links to courses in the Django admin (#684)

ssciolla:
This PR makes changes to the Django back-end and `start.sh` in order to provide convenient links in the Django admin to course landing pages in the MyLA front-end. Besides making changes to `models.py` and `admin.py`, I created a new environment variable `LOCALHOST_PORT` and added a new management command `site` so that the domain could be set automatically through `start.sh`. I have tested this locally and on [the MyLA development instance](https://dev-myla.tl.it.umich.edu), where the changes are still deployed (as of the morning of 10/14). More details on the changes are provided below. This PR aims to resolve issue #684.

1) Adding a "View on Site" link to the Course edit page and a field link in the Course list view was relatively straightforward. The former was created automatically once `get_absolute_url` was added as a method to the `Course` model. I then created the list view link by adding a `course_link` method to the `CourseAdmin` model, which could then be included in the `list_display` link. Using [`format_html`](https://docs.djangoproject.com/en/2.2/ref/utils/) seems to be a best practice for preparing small HTML fragments.

2) However, because calling `reverse` pulls in the value for the domain from the `django_site` table in the MySQL database (which was previously `example.com`), I needed a flexible way to set the domain that would work in all the environments the application is run in. The solution I arrived at (there may be others) was to create a new `site` management command that could be invoked from `start.sh`. The new command allows for either the creation of a new site record (using the `--new` flag) or the updating of the default site, which is identified by importing the `SITE_ID` variable from `settings.py`.

3) To isolate the inputs for the new `site` command within `start.sh`, I needed to create a new environment variable `LOCALHOST_PORT` and filter the existing `ALLOWED_HOSTS` array in the `env.json` file. I could not find a place where the port used with `localhost` was specified except in `docker-compose.yml`, and it seemed strange to pull the value from there. Note: for the links to work, reviewers will need to set `LOCALHOST_PORT` if their port differs from the default value of `5001`. I used a [jq](https://stedolan.github.io/jq/manual/v1.5/) query to filter `ALLOWED_HOSTS` down to either a default `localhost` or whatever domain is being used in a deployed instance. Then the `site` command is invoked using either `localhost` and the specified port or simply the provided domain.

Feedback is welcome, and to ensure this works for other institutions, it would be great if someone from UBC could review and test this.

### DIFF
--- a/config/env_sample.json
+++ b/config/env_sample.json
@@ -9,10 +9,9 @@
         "127.0.0.1",
         "localhost"
     ],
-    "/* Port the server will use on localhost": "*/",
+    "/* Port in use when the application is running on localhost": "*/",
     "LOCALHOST_PORT": 5001,
     "/* Configuration of CSP see https://django-csp.readthedocs.io/en/latest/configuration.html": "*/",
-    
     "/* To enable this at all you must set have the CSP key set": "*/",
     "/* and REPORT_ONLY should be be false, otherwise it will just console Warn": "*/",
     "CSP": {
--- a/config/env_sample.json
+++ b/config/env_sample.json
@@ -9,6 +9,8 @@
         "127.0.0.1",
         "localhost"
     ],
+    "/* Port the server will use on localhost": "*/",
+    "LOCALHOST_PORT": 5001,
     "/* Configuration of CSP see https://django-csp.readthedocs.io/en/latest/configuration.html": "*/",
     
     "/* To enable this at all you must set have the CSP key set": "*/",
--- a/config/env_sample.json
+++ b/config/env_sample.json
@@ -9,8 +9,6 @@
         "127.0.0.1",
         "localhost"
     ],
-    "/* Port in use when the application is running on localhost": "*/",
-    "LOCALHOST_PORT": 5001,
     "/* Configuration of CSP see https://django-csp.readthedocs.io/en/latest/configuration.html": "*/",
     "/* To enable this at all you must set have the CSP key set": "*/",
     "/* and REPORT_ONLY should be be false, otherwise it will just console Warn": "*/",
--- a/dashboard/management/commands/site.py
+++ b/dashboard/management/commands/site.py
@@ -1,0 +1,25 @@
+from django.core.management.base import BaseCommand
+from django.contrib.sites.models import Site
+from django.conf import settings
+
+
+class Command(BaseCommand):
+    def add_arguments(self, parser):
+        parser.add_argument('--domain', dest='domain', type=str, required=True)
+        parser.add_argument('--name', dest='name', type=str, required=True)
+        parser.add_argument('--new', dest='new', type=bool, required=False)
+
+    def handle(self, *args, **options):
+        domain = options.get('domain')
+        name = options.get('name')
+        create_new_site = options.get('new')
+
+        if create_new_site:
+            new_site = Site.objects.create(domain=domain, name=name)
+            new_site.save()
+        else:
+            site_id = settings.SITE_ID
+            default_site = Site.objects.get(id=site_id)
+            default_site.domain = domain
+            default_site.name = name
+            default_site.save()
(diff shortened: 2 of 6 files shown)


start.sh
```diff
@@ -27,20 +27,26 @@ else
...
+
 if [ -z "${ENV_JSON}" ]; then
     MYSQL_HOST=$(jq -r -c ".MYSQL_HOST | values" ${ENV_FILE})
     MYSQL_PORT=$(jq -r -c ".MYSQL_PORT | values" ${ENV_FILE})
     IS_CRON_POD=$(jq -r -c ".IS_CRON_POD | values" ${ENV_FILE})
     PTVSD_ENABLE=$(jq -r -c ".PTVSD_ENABLE | values" ${ENV_FILE})
     CRONTAB_SCHEDULE=$(jq -r -c ".CRONTAB_SCHEDULE | values" ${ENV_FILE})
     RUN_AT_TIMES=$(jq -r -c ".RUN_AT_TIMES | values" ${ENV_FILE})
+    DOMAIN=$(jq -r -c "${DOMAIN_JQ} | values" ${ENV_FILE})
+    PORT=$(jq -r -c ".LOCALHOST_PORT | values" ${ENV_FILE})
```
jonespm: I'd probably set a default for this PORT incase it isn't defined in the configuration file. Just put it up above where the other defaults are. It's only used in the localhost case anyway. I guess you could put it in the jquery too.

ssciolla: How does [this commit](https://github.com/tl-its-umich-edu/my-learning-analytics/pull/761/commits/2d63c9fb54edb95c31ee13411acd8fa1992fa954) look? I tried putting the `if fi` block above this, but it was resulting in `PORT` having an empty string value.

dashboard/management/commands/site.py
```diff
@@ -0,0 +1,25 @@
...
+class Command(BaseCommand):
+    def add_arguments(self, parser):
+        parser.add_argument('--domain', dest='domain', type=str, required=True)
+        parser.add_argument('--name', dest='name', type=str, required=True)
+        parser.add_argument('--new', dest='new', type=bool, required=False)
+
+    def handle(self, *args, **options):
+        domain = options.get('domain')
+        name = options.get('name')
+        new = options.get('new')
```
zqian: when will the "new" option be used? Did not see an example in start.sh. May need documentation here.

config/env_sample.json
```diff
@@ -9,8 +9,9 @@
         "127.0.0.1",
         "localhost"
     ],
+    "/* Port in use when the application is running on localhost": "*/",
+    "LOCALHOST_PORT": 5001,
```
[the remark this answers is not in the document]
jonespm: Inside the container this is running on port 5000 by default. However docker-compose.yml is exposing this as 5001. We *could* switch docker-compose to be 5000 and then not have to have this extra environment variable. However if someone switched it, they'd have to manually update the port somehow. The container doesn't know which port is being exposed on the host. 

We *could* add it to the docker-compose.yml environment instead I think, maybe that's a better idea. Nobody probably needs to configure this in production.

start.sh
jonespm: I think we can forget about this, if we move this from the environment to the docker-compose is should always be set on the localhost.

dashboard/management/commands/site.py
ssciolla: It isn't being used, but it seemed like a site command (that could be run from the container shell) should be a bit more flexible than just updating the default site. Do you want me to add a comment?

ssciolla:
@jonespm, @zqian, okay, I've moved the variable to `docker-compose.yml`. I removed the lines that handle setting that variable from `start.sh`, since the variable will now always be set in the same place and won't need to be configured. Please let me know if you think those deletions are okay.

ssciolla:
I just added a new commit trying to address @zqian's feedback that includes comments explaining how the `--new` flag works. I also simplified that argument (using the `--superuser` flag from `createuser` as a model) so that you only need to provide `--new` (as opposed to `--new=True`).

dashboard/management/commands/site.py
ssciolla: See latest commit, and let me know if that addresses your concern.

zqian: Perfect!

Issue #758, opened 2019-10-10
Bring in all terms from Canvas to MyLA


jennlove-um:
#164  Expected behavior : 
All course terms, including those with no start and end dates, get added to MyLA to support Canvas courses using MyLA.  Courses with any term in Canvas should be allowed to be added to MyLA.

## Describe the issue :
The current cron query for terms errors when trying to get terms for Canvas courses using "No Term" or "Default Term."  Both of these terms have no start or end date.  

The query currently looks for terms after 2016-11-15.  It should bring in all terms, instead.

jonespm:
The issue was here
https://github.com/tl-its-umich-edu/my-learning-analytics/blob/0837829e955e83fce99de87f7f34f8c25d83497b/dashboard/cron.py#L434

I think we can remove this EARLIEST_TERM_DATE setting now as we are pulling the term and syncing it up now. We used to have a dropdown people were forced to select from.

ssciolla:
I've started work on this issue, which you can see on [this commit from the issue-758-fetch-all-terms branch](https://github.com/ssciolla/my-learning-analytics/commit/be86c875adf3e036860cd834730840f9b5c32f16#diff-64b16f0fa7d7ac8585eaf90c3df4346cR435) of my fork.

@jonespm, so as you said, we can just modify the term query to pull in all the terms by removing the `WHERE` clause. If we do this, we should also probably modify the `get_course_date_range` method of the `Course` model to use the fallback dates (now, two weeks from now) if `date_start` and `date_end` are null. I also did this in the above linked commit, as well as added `logger.warning` messages to alert users that the fallback dates are in use.

What I am less sure about is how this will affect the rest of the application. It seems the main use case for this is to handle courses that go beyond a single term, but then the `MAX_DEFAULT_WEEKS` setting used in `views.py` may need to be changed or removed:

https://github.com/tl-its-umich-edu/my-learning-analytics/blob/11dd6f55333b40e6c863f7ccef077b068dcc070a/dashboard/views.py#L120-L123

Thoughts? I was testing with a course id @jennlove-um gave me, which was useful for testing the term and date changes but seemed to have not have substantial resource access, assignment, or grade data to work with.

jonespm:
@ssciolla Aside from the print I think your WIP looks good, can you create that as a PR? Also remove the EARLIEST_TERM_DATE setting from the settings and the sample config.

PR #771, opened 2019-10-25
Fetching all terms using cron (#758)

ssciolla:
This PR includes a modification to the `update_term` method of `cron.py` to bring in all term records available from the Unizin Data Warehouse, as well as some changes to the handling of term dates in the `get_course_date_range` method. See below for additional explanation and details. The PR aims to resolve issue #758.

1) Previously, the terms fetched were limited (using a `WHERE` clause) to those before an `EARLIEST_TERM_DATE` specified as an environment variable. Removing this clause brings in terms earlier than the date previously specified, as well as terms without dates, (e.g. "No Term" or "Default Term") to the local database.

2) Because some of the new terms pulled using `cron.py` do not have dates, I added some additional handling to the `get_course_date_range` method of the `Course` model. When `self.term.date_start` or `self.term.date_end` are `None`, the fallback dates set in PR #690 are used. I also added some `warning` logs that are triggered in all cases when the fallback dates are used to help with application maintenance.

3) Because `EARLIEST_TERM_DATE` is no longer used, I removed the references to this variable from `settings.py` and `env_sample.json`.

Note: This PR brings in all the term values, but more investigation would need to be done to ensure  proper handling of courses that have dates that span beyond the limits of a single term (based on dates set at the course level). When supporting these courses, we would need to look at modifying the `MAX_DEFAULT_WEEKS` variable.

### DIFF
--- a/config/env_sample.json
+++ b/config/env_sample.json
@@ -113,8 +113,6 @@
     "DJANGO_WATCHMAN_TOKEN": "",
     "/*  A single default IDP to use if multiple are present in the SAML metadata (Get from Metadata)": "*/",
     "DJANGO_SAML2_DEFAULT_IDP": "",
-    "/*  Earliest term date from database that we'll show and support academic terms from (default is 2016-11-15 which the first date of 'our' data)": "*/",
-    "EARLIEST_TERM_DATE": "2016-11-15",
     "/*  CRON POD SETTINGS": "*/",
     "/* ": "*/",
     "/*  All of these settings go together and need to be enabled for cron job": "*/",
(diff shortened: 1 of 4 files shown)


ssciolla:
@zqian, to respond to your question, I had a similar concern about how well the application will work for courses longer than a typical term length. However, I think we have to assume that if we're supporting (deploying with) those courses, they would have dates specified at the course level (`Course.date_start` and `Course.date_end`). The fallback dates (today, two weeks from today) are really just to ensure that an error isn't thrown when no dates are available. I added the logging messages here to help bring attention to the fact that fallback dates are in use, so admins will notice and address the issue.

PR #774, opened 2019-10-28
Fixes #763 - Invalid symlinks in Docker when copying

### LINKED ISSUE #763: Invalid symlinks in Docker when copying

We're getting an error like this from certain versions of Docker
```
-> COPY --from=node-webpack /usr/src/app /usr/src/app
error: build error: API error (500): {“message”:“Error processing tar file(exit status 1): invalid symlink \“/usr/src/app/node_modules/.bin/acorn\” -> \“../acorn/bin/acorn\“”}
```

It looks like it's similar to this issue here.
https://stackoverflow.com/questions/35787702/docker-cp-a-folder-with-a-relative-symlink-invalid-symlink

It looks like a workaround is to copy symlink directories with tar. It's also possible that this directory isn't needed at all as these binaries aren't used in the production application. They may only be used by the node-webpack stage.

jonespm:


jonespm:
This was confirmed by our operations to successfully build, and I've locally tested that we don't need the node symlinks though I still have the file that we can uncompress in-case we do. Note that this file is hidden locally because of docker-compose but I verified it's there and looks correct

Dockerfile
```diff
@@ -28,20 +33,21 @@ EXPOSE 5000
...
-COPY --from=node-prod-deps /usr/src/app/node_modules /code/node_modules
+COPY --from=node-prod-deps /usr/src/app/package-lock.json package-lock.json
+COPY --from=node-prod-deps /usr/src/app/webpack-stats.json webpack-stats.json
+COPY --from=node-prod-deps /usr/src/app/assets assets
+COPY --from=node-prod-deps /usr/src/app/node_modules node_modules
+COPY --from=node-prod-deps /usr/src/app/all_symlinks.tgz .
 
 # NOTE: project files likely to change between dev builds
-COPY . /code/
+COPY . .
```
ssciolla: Couple questions here: 1) What made this change necessary? and 2) Is this line even necessary now with the change?

Dockerfile
```diff
@@ -28,20 +33,21 @@ EXPOSE 5000
 WORKDIR /code
 
 # NOTE: requirements.txt not likely to change between dev builds
-COPY requirements.txt /code/requirements.txt
+COPY requirements.txt .
```
ssciolla: See question below. Is this necessary if you're just copying it to the same place?

Dockerfile
jonespm: This was kind of a separate issue and just a cleanup. Since the WORKDIR was /code it wasn't necessary to have /code/ everywhere. I can just remove it and put it in a separate issue.

Dockerfile
jonespm: Yeah, basically does the same thing

Dockerfile
ssciolla: I think it's fine to keep this simplification in this issue. I guess I was looking for some clarity on why this line is necessary at all, if you're copying a directory contents to the same place.

Dockerfile
jonespm: Yeah this Docker syntax is actually <src> <dest>. The <src> is on the localhost and <dst> is inside the container image. This syntax copies the current working directory into the working directory set in the container. Confusing to look at but it isn't the same place. ;)

ssciolla: Okay, that makes sense to me, sorry for my ignorance. You learn something every day!

ssciolla: Should have read the documentation before commenting. I think I understood this at one point, since I've written a `Dockerfile`, but I guess I forgot.

Issue #748, opened 2019-10-07
Poor color contrast on files bars, inconsistency with assignment bars


jonespm:
## Expected behavior (A clear and concise description of what you expected to happen) : 

On the files page, the bars have a poor color contrast making the number very difficult to read. 
The background is #808080 and text is #0F0F0F, which has a 4.5:1 ratio, barely passing.

Files:
![image](https://user-images.githubusercontent.com/27447/66310139-6e71a680-e8d9-11e9-8429-fb3ffabe025f.png)

Assignment has a background of #E1E1E1 which has a 14:1 ratio.
![image](https://user-images.githubusercontent.com/27447/66310156-7af5ff00-e8d9-11e9-923f-19ead0fac54f.png)

ssciolla:
This is related to issue #673, which I'm working on and hope to have a pull request soon.

jennlove-um:
The contrast between the gray bar and the black text is now 10:1 using #B5B5B5 for the gray.

jennlove-um:
Testing passes.

gsilver:
@pushyamig @ssciolla - nice.

Issue #673, opened 2019-08-22
Allow for custom themes/styles


jonespm:
Thank you for contributing to this project!

## Describe your problem or feature you'd like added
We were thinking of easy ways that this tool could be customized for styling. If it's launched via LTI from Canvas you can use
`canvas.css.common = $Canvas.css.common
`
To get the URL to the Canvas common CSS for your theme. This is probably good, but minimally we should have a way to change the colors, fonts and some other things for each installation manually.

## Describe the solution you'd like
We could do theming but that seems like a heavy solution. Maybe [MUICSS](https://www.muicss.com/) would work? Have you considered anything for this? We'd love to at *least* have fonts and the top bar color changed for the Fall release.

ssciolla:
I've made some progress on this issue, which you can see on my [673 issue branch](https://github.com/ssciolla/my-learning-analytics/tree/issue-673-enable-custom-themes). I just wanted to let people know to avoid duplication of effort and so @justin0022 can keep this in mind while working on PR #702. Next steps will be the following: 

1) Think about other colors used in the UI and whether they should be controlled by a theme.
2) If necessary, brainstorm ways to refactor theme use in d3 components and `RangeSlider` so that I'm not just importing `theme.js` directly. I can probably pass `theme` down through `props` or another mechanism.

PR #775, opened 2019-10-29
Creating theme with default palette and consolidating color setting (#673 and #748)

ssciolla:
In order to increase consistency and enable easier experimentation and customization, this PR creates a `theme.js` module that defines a default color palette and then provides those values to the rest of the application (in place of hardcoded values) through a `ThemeProvider` component and imports for d3 and other custom components. See below for additional description of the changes (more detail may be added). This PR aims to resolve (or at least work towards resolution of) issues #673 and #748.

1) I added a `theme.js` module that uses the `createMuiTheme` function, and then employed the `ThemeProvider` in `App.js` to pass the theme down to child components. The `theme` module specifies a palette with a `primary` color (a Michigan blue, currently), a `secondary` color (the steel blue indicating a positive, viewed, graded, completed state), and a `negative` color (the gray indicating a negative, un-viewed, un-graded, un-completed state). The names `primary` and `secondary` are used as they are supported by Material UI specifically, which automatically produces `light`, `dark`, and `textContrast` versions (see [ThemeProvider](https://material-ui.com/customization/theming/#theme-provider)). Grays had been inconsistent previously in the application (see issue #748); I opted for a lighter gray with `negative`, but I'm open to other suggestions.

2) In the three view containers, the `rangeSlider` component, and the d3 components, I removed hardcoded color values and replaced them with theme references. In the React/Material UI components, color values could be accessed by name using the `color` property, while I needed to import `theme.js` to the others to make the object accessible (there may be a better to accomplish this; let me know if you have ideas). I decided to make the `rangeSlider` styles for the selected and un-selected portions consistent with the positive/negative scheme elsewhere, which required the use of some new properties.

Note: This is the first stage of ongoing theme-related work I've been undertaking, with the next being changes to allow users to update the application's primary color through the Django admin. You can see my progress on the next stage at my [issue-673-add-institution-and-primary-color-hooks branch](https://github.com/ssciolla/my-learning-analytics/tree/issue-673-add-institution-and-primary-color-hooks).

### DIFF
--- a/assets/src/components/RangeSlider.js
+++ b/assets/src/components/RangeSlider.js
@@ -6,18 +6,18 @@ import theme from "../theme";
 
 const rangeSlider = props => {
     const activeDotStyle = {
-        borderColor: theme.palette.primary.light,
+        borderColor: theme.palette.secondary.light,
     };
     const dotStyle = {
-        borderColor: theme.palette.secondary.main,
+        borderColor: theme.palette.uncompleted.main,
     }
     const unselectedStyle = {
-        borderColor: theme.palette.secondary.main,
-        backgroundColor: theme.palette.secondary.main,
+        borderColor: theme.palette.uncompleted.main,
+        backgroundColor: theme.palette.uncompleted.main,
     };
     const selectedStyle = {
-        borderColor: theme.palette.primary.light,
-        backgroundColor: theme.palette.primary.light,
+        borderColor: theme.palette.secondary.light,
+        backgroundColor: theme.palette.secondary.light,
     };
     const wrapperStyle = { width: "70%", margin: "0 auto" }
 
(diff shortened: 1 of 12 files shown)


ssciolla:
> This looks good. My only comment would be the name of the file/export "theme" seems a bit too generic. Maybe it should be something like myla-theme or site-theme?

Sure, good idea! I think I'll go with `siteTheme.js`.

ssciolla:
@justin0022 and @pushyamig, do you think one of you might be able to review this soon? I'd like to keep moving this work forward, before it becomes outdated.

justin0022:
> @justin0022 and @pushyamig, do you think one of you might be able to review this soon? I'd like to keep moving this work forward, before it becomes outdated.

Hey sorry @ssciolla I'll review right now

ssciolla:
@justin0022, thanks for bringing up the style question, since I've wondered in the past what standards we were following and neglected to ask. I am happy to follow your lead on this. Please take a look at my latest commit, which should correct some of those issues. I've set up my local dev environment (I use IntelliJ currently) to alert me of the Standard rules.

justin0022:
Thanks @ssciolla, I'll add in some notes in the `CONTRIBUTING` file about linting.

PR #781, opened 2019-11-07
Fixes #780 - Webpack watcher won't start

### LINKED ISSUE #780: Error with fresh install of MyLA on localhost for webpack-stats.json

Fresh install of MyLA gaves this error `at localhost:5001: Error reading /code/webpack-stats.json. Are you sure webpack has generated the file and the path is correct?`

This was related to a recent change in symlinks on #763. We're going to need to uncompress these when starting up the container locally.

jonespm:


Dockerfile
```diff
@@ -12,9 +12,10 @@ RUN npm run prod
 
 # This is to find and remove symlinks that break some Docker builds.
 # We we need these later we'll just uncompress them
```
ssciolla: "We" is repeated here.

docker-compose.yml
```diff
@@ -43,7 +43,8 @@ services:
     build:
       context: .
       target: node-webpack
-    command: npm run watch
+    command: sh -c "tar zxvf node_modules/all_symlinks.tgz && npm run watch"
+
```
ssciolla: Extra line seems unnecessary here?

Dockerfile
jonespm: Oui, Oui!

PR #784, opened 2019-11-08
Moving myla_globals construction to views.py (#779)

### LINKED ISSUE #779: Frontend doesn't load when user is affiliated with a course with an apostrophe in the name

The application should be able to support courses with any name, including if the name has an apostrophe (e.g. "Sam's Course on Silliness").

@jennlove added a couple of courses to the test instance of MyLA that she was enrolled in and discovered that she could only see the footer information when she visited those course pages. After some investigation, we discovered that an apostrophe included in one of the courses was causing a JavaScript syntax error when the `myla_globals` object was being created. We'll need to look at the `base.html` file in the backend and the `App.js` file in the frontend to ensure single quotes are properly escaped or handled.

## Steps to Reproduce :
1. Add a new course with an apostrophe in the name and run the cron (or edit an existing course)
2. Log in as a user with some affiliation to the course (teacher or student)
3. Try to navigate to any page in the frontend.

ssciolla:
This PR moves the construction of the `myla_globals` object from the `base.html` template to the `get_home_template` function view of `views.py` in order to fix a bug involving improper handling of apostrophes and streamline the appearance of `base.html`. Some other changes were made to simplify or clarify the process of generating a list of courses a user is enrolled in. More details will be forthcoming. The PR aims to resolve #779.

1) The majority of this PR is focused on moving the code and logic constructing a `myla_globals` dictionary or object from `base.html` to the `get_home_template` view. The dictionary can be constructed within the function view, serialized to a JSON string via `json.dumps`, and then passed via `render`'s `context` parameter to the template (see [this StackOverflow page](https://stackoverflow.com/questions/31151229/django-passing-json-from-view-to-template) and [the Django documentation](https://docs.djangoproject.com/en/2.2/topics/http/shortcuts/#render) for reference). Once available to the template, we can initialize the variable within the `script` element, using the `safe` template to ensure the object is passed as is to the React frontend. JavaScript will then see the JSON string as an object, even without calling `JSON.parse`.

2) As part of the process in (1), I updated the `get_user_courses_info` function in `db_utils.py` to return a dictionary of a student's related courses rather than JSON, so it could be more straightforwardly incorporated and nested within `myla_globals`. I believe the change also enabled me to remove some conditional logic from the frontend's `App.js`, since the property's value now is always a list or array (even if it's empty). Additionally, I changed the name of `get_user_courses_info`'s parameter, since `username` communicated a string better to me than `user_id` (which I initially thought was an ID number). I also added some newlines to `db_utils.py` to make the formatting consistent.

3) Because the `get_home_template` view now calls `get_user_courses_info` directly, I removed the `current_user_courses_info` context processor and a reference to it in `settings.py` (the processor is not used elsewhere).

3) Lastly, I changed the variable name of `enrolledCourses` to `relatedCourses` in the frontend, since "enrolled" seemed to be describing the courses, which seemed confusing; "related" seemed to be more clearly bi-directional.

### DIFF
--- a/dashboard/common/db_util.py
+++ b/dashboard/common/db_util.py
@@ -85,7 +85,7 @@ def get_default_user_course_id(user_id):
 def get_user_courses_info(user_id):
     logger.info(get_user_courses_info.__name__)
     course_list = []
-    course_info={}
+    course_info = []
     with django.db.connection.cursor() as cursor:
         cursor.execute("SELECT course_id FROM user WHERE sis_name= %s", [user_id])
         courses = cursor.fetchall()
@@ -99,9 +99,9 @@ def get_user_courses_info(user_id):
             cursor.execute("select canvas_id, name from course where canvas_id in %s",[course_tuple])
             course_names = cursor.fetchall()
             df = pd.DataFrame(list(course_names))
-            df.columns=["course_id", "course_name"]
-            course_info=df.to_json(orient='records')
-            logger.info(f"User {user_id} is enrolled in courses {course_info}")
+            df.columns = ["course_id", "course_name"]
+            course_info = df.to_dict(orient='records')
+            logger.info(f"User {user_id} is affiliated with these courses {df.to_json(orient='records')}")
     return course_info
 
 
--- a/dashboard/common/db_util.py
+++ b/dashboard/common/db_util.py
@@ -103,7 +103,7 @@ def get_user_courses_info(username):
             df = pd.DataFrame(list(course_names))
             df.columns = ["course_id", "course_name"]
             course_info = df.to_dict(orient='records')
-            logger.info(f"User {username} is enrolled in these courses {df.to_json(orient='records')}")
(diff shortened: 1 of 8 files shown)


dashboard/common/db_util.py
```diff
@@ -82,12 +84,12 @@ def get_default_user_course_id(user_id):
...
 
-def get_user_courses_info(user_id):
+def get_user_courses_info(username):
     logger.info(get_user_courses_info.__name__)
     course_list = []
-    course_info={}
+    course_info = []
     with django.db.connection.cursor() as cursor:
-        cursor.execute("SELECT course_id FROM user WHERE sis_name= %s", [user_id])
+        cursor.execute("SELECT course_id FROM user WHERE sis_name= %s", [username])
```
zqian: so the user_id should have been username prior to the change? Nice catch!

jonespm: Yeah, looks like it was just renaming the value local to this method.

Issue #707, opened 2019-09-20
Simplify the Grade_distribution API format 


pushyamig:
As per [discussion](https://github.com/tl-its-umich-edu/my-learning-analytics/pull/703#discussion_r326229366) we could remodel the Response to use something like the following structure:
```
{
   "summary": {
      "current_user_grade": XX.X,
      "tot_students": XX,
      "show_number_on_bars": true,
      "grade_avg": XX.X,
      "median_grade": XX.X,
      "grade_upper_limit": XXX,
        },
   "grades": [XX.X, XX.X, ...]
}
````

PR #793, opened 2019-11-15
Simplified API Format for Grade Distribution View (#707)

vjcao:
The grade distribution API used to send a full dataframe. This PR changes it so that a simplified JSON string is sent instead of a full dataframe. This JSON string follows this format:
```
{ 
   "summary":{ 
      "current_user_grade": XX.X,
      "tot_students": XX,
      "grade_avg": XX.X,
      "median_grade": XX.X,
      "show_number_on_bars": false,
      "show_dash_line": true,
      "graph_upper_limit": XX.X
   },
   "grades":[XX.X, XX.X, ...]
}
```
This PR addresses and closes issue #707.

### DIFF
--- a/dashboard/views.py
+++ b/dashboard/views.py
@@ -350,12 +350,17 @@ def grade_distribution(request, course_id=0):
     if df.empty or df['current_grade'].isnull().all():
         return HttpResponse(json.dumps({}), content_type='application/json')
 
-    df['tot_students'] = df.shape[0]
+    result = {}
+    result['summary'] = {}
+    result['summary']['current_user_grade'] = df['current_user_grade'].values[0]
+    result['summary']['tot_students'] = df.shape[0]
     df = df[df['current_grade'].notnull()]
     df['current_grade'] = df['current_grade'].astype(float)
-    df['grade_avg'] = df['current_grade'].mean().round(2)
-    df['median_grade'] = df['current_grade'].median().round(2)
-    df['show_number_on_bars'] = df['show_number_on_bars'].apply(lambda x: True if x == 1 else False)
+    result['summary']['grade_avg'] = df['current_grade'].mean().round(2)
+    result['summary']['median_grade'] = df['current_grade'].median().round(2)
+    result['summary']['show_number_on_bars'] = False
+    if df['show_number_on_bars'].values[0] == 1:
+        result['summary']['show_number_on_bar'] = True
 
     df.sort_values(by=['current_grade'], inplace=True)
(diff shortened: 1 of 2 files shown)


dashboard/views.py
```diff
@@ -350,12 +350,17 @@ def grade_distribution(request, course_id=0):
     if df.empty or df['current_grade'].isnull().all():
         return HttpResponse(json.dumps({}), content_type='application/json')
 
-    df['tot_students'] = df.shape[0]
+    result = {}
```
pushyamig: I feel `result` can be renamed better, variable naming should say the purpose of it. `result` doesn't fullfill that

dashboard/views.py
```diff
@@ -350,12 +350,17 @@ def grade_distribution(request, course_id=0):
...
+    result['summary'] = {}
+    result['summary']['current_user_grade'] = df['current_user_grade'].values[0]
```
pushyamig: using the double array can be confusing when somebody is reading the code. Can this be simplified like this?

```grades = [23,34,45,55,20,10]
summary = dict()
summary['tot_stu']= 80
summary['median']=30
sum = {}
sum['summary']=summary
sum['grades']=grades
print(sum)
print(json.dumps(sum))

Result : {"summary": {"tot_stu": 80, "median": 30}, "grades": [23, 34, 45, 55, 20, 10]}
```

dashboard/views.py
pushyamig: can this is inialized like `result = dict()` instead its more clean

dashboard/views.py
ssciolla: I would agree about the double array, though having `sum` and `summary` is confusing to me. Like @pushyamig is suggesting, try first building the summary dictionary and then assign it all at once to a key in the parent dictionary. I would probably name the outer dictionary something like `grades_dist_data` and the summary dictionary something like `grades_summary`.

dashboard/views.py
ssciolla: I agree about the name of the `result` variable; see my comments below. I don't particularly mind either way whether you use `dict()` or `{}`. I was taught to use the latter. @pushyamig, I'd be curious to hear why you like `dict()` better.

pushyamig: My IDE somehow warns if I declare like`{}` when I change it to `dict()` it is quiet. 
Screen Shot 2019-12-09 at 5 33 37 PM

pushyamig:
This Looks Good!

Issue #806, opened 2019-11-19
Snackbar may not be visible long enough for some users


ssciolla:
Thank you for contributing to this project!

- Make sure to search the [issues](https://github.com/tl-its-umich-edu/my-learning-analytics/issues) for duplicates first!

## Expected behavior (A clear and concise description of what you expected to happen) : 

When a user makes a change that triggers a Snackbar, the Snackbar should be visible long enough that the majority of users can read or perceive the message before it disappears.

## Describe the bug (Tell us what happens instead of the expected behavior) :

From the 10/28/19 accessibility review:

"Toasts/Snackbars may display for too short a time to be perceivable by some users. May also be unperceivable to others. ... Some users may not benefit from confirmation that their action was successful."

## Steps to reproduce :

1. On any of the views, choose an option from a dropdown menu or click a checkbox.
2. Then, click the checkbox labeled with "Remember my setting" that appears.
3. Observe the dark box (the Snackbar) that appears in the bottom left (and how long it is present) telling the user that the setting has been saved.

![Screen Shot 2019-11-27 at 9 42 15 AM](https://user-images.githubusercontent.com/35741256/69732990-9f60a180-10fa-11ea-9aa7-1c9ff055f9fd.png)

## Additional context (Add any other context about the problem here) :

The review suggests the following regarding the duration: "Fade in (or slide in) for 200-400ms for saccade time for the eye to refocus, then show for 50ms for each of the characters in the message."

The review also notes that the Snackbar's HTML element container should have the `role=alert` attribute-value pair.

Fixing this will likely involve making changes to the `UserSettingSnackbar` component or its usages in the views.

jpyan:
Hey I'm trying to look into this bug yet I can't trigger any snackbars. Would it be possible that you give a more specific reproduce steps (screenshots are helpful too)? Thanks!

ssciolla:
@jpyan, sure, I'll update the description with a few more details momentarily.

justin0022:
> Hey I'm trying to look into this bug yet I can't trigger any snackbars. Would it be possible that you give a more specific reproduce steps (screenshots are helpful too)? Thanks!

@jpyan if you go to the GradeDistribution view, you should be able to trigger the snackbar by clicking on the `Show my grades` checkbox.

jpyan:
Thanks, I can see now.

justin0022:
[...]

PR #811, opened 2019-11-27
update read.me (remove unaccesible links)

jpyan:
remove links in the README to specific courses  for issue #810

Issue #719, opened 2019-09-25
Update README for latest version of data (9-11)


jonespm:
Thank you for contributing to this project!

- Make sure to search the [issues](https://github.com/tl-its-umich-edu/my-learning-analytics/issues) for duplicates first!

## Expected behavior (A clear and concise description of what you expected to happen) : 
It was noted on the call that myla_test_data_2019_09_11 file is not correct with the current README. We need to update the README for the data in this file. 

I'd like to figure out a process for the depersonalizer that will keep some id's static so we don't have to make so many documentation updates.

ssciolla:
We had an EECS student try to update the README to better reflect the changing course id values. In thinking about those changes (which I had them remove), I had the following idea:

I feel like with the changes introduced in PR #761, we probably shouldn't be listing any hardcoded course values and just instruct people to use the admin interface. So instead, under `Logging in as an administrator`, we'd have the following:

1. As you are now logged in as `root`, there are no courses listed. To view the Django admin interface, navigate to http://localhost:5001/admin
1. Click on "Courses" under "Dashboard" to view a list of sample courses.
1. Click on the link in the "Course Link" column for the sample course you wish to view.

@jonespm, thoughts?

ssciolla:
Wups, I guess some of this has already been addressed by PR #811. I still think relying on the Django admin might be a good idea.

ssciolla:
@lsloan, we talked about updating 5. under "Logging in as administrator" so that it doesn't reference specific course ids but just gives instructions on finding and viewing the course via the Admin. This is to eliminate the need to repeatedly update time-related information.

jennlove-um:
@ssciolla Is this addressed in #661 ?  Can we close this issue?

ssciolla:
Looks like this has been handled by my PR or another earlier on. The `README.md` doesn't reference a specific file, just the Google Drive directory. I'll close it.

PR #815, opened 2019-12-03
Update snackbar visualization time for issue #806

jpyan:
Change according to the suggestion : "Fade in (or slide in) for 200-400ms for saccade time for the eye to refocus, then show for 50ms for each of the characters in the message." for issue #806. Also added role=alert

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,17 @@ You may also optionally place the json settings directly into the `ENV_JSON` env
     username: root
     password: root
     ```
-1. As you are now logged in as `root`, there are no courses listed. Go through the latest SQL file you downloaded from the link in the setup and search for the line "INSERT INTO `course` VALUES". You can find the course_id in the second column of the tuples. And then nagivate to http://localhost:5001/courses/{course_id} with the course_id you found. (For example, with SQL file myla_test_data_2019_10_16.sql, on line 323, you can find course_id = 235420 or course_id = 362855. And then nagivate to http://localhost:5001/courses/235420 or http://localhost:5001/courses/362855 and you can view the course page as an admin.)
+1. As you are now logged in as `root`, there are no courses listed. Next 3 steps will help you view a sample course.
+1. Connect to MySQL database.
+    ```
+    Host: 127.0.0.1
+    Username: student_dashboard_user
+    Password: student_dashboard_pw
+    Database: student_dashboard
+    Port: 5306
+    ```
+1. Navigate to `course` table and select canvas_id `canvas_id`("SELECT canvas_id From course"), which will be used in the next step.
+1. Nagivate to http://localhost:5001/courses/{canvas_id} with the canvas_id you found. (For example, with SQL file myla_test_data_2019_10_16.sql loaded in, nagivate to http://localhost:5001/courses/235420 or http://localhost:5001/courses/362855 and you can view the course page as an admin.) 
 1. To get to the Django admin panel, click on the Avator in the top right, then click `Admin`, or go here: http://localhost:5001/admin.
 
 #### Logging in as a student
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ You may also optionally place the json settings directly into the `ENV_JSON` env
     username: root
     password: root
     ```
-1. As you are now logged in as `root`, there are no courses listed. 
+1. As you are now logged in as `root`, there are no courses listed. Go through the latest SQL file you downloaded from the link in the setup and search for the line "INSERT INTO `course` VALUES". You can find the course_id in the second column of the tuples. And then nagivate to http://localhost:5001/courses/{course_id} with the course_id you found. (For example, with SQL file myla_test_data_2019_10_16.sql, on line 323, you can find course_id = 235420 or course_id = 362855. And then nagivate to http://localhost:5001/courses/235420 or http://localhost:5001/courses/362855 and you can view the course page as an admin.)
 1. To get to the Django admin panel, click on the Avator in the top right, then click `Admin`, or go here: http://localhost:5001/admin.
 
 #### Logging in as a student
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ You may also optionally place the json settings directly into the `ENV_JSON` env
     username: root
     password: root
     ```
-1. As you are now logged in as `root`, there are no courses listed. Navigate to http://localhost:5001/courses/231768 or http://localhost:5001/courses/430174 to view a sample course as a superuser.
+1. As you are now logged in as `root`, there are no courses listed. 
 1. To get to the Django admin panel, click on the Avator in the top right, then click `Admin`, or go here: http://localhost:5001/admin.
 
 #### Logging in as a student
--- a/assets/src/components/UserSettingSnackbar.js
+++ b/assets/src/components/UserSettingSnackbar.js
@@ -4,7 +4,7 @@ import IconButton from '@material-ui/core/IconButton'
 import CloseIcon from '@material-ui/icons/Close'
 import Slide from '@material-ui/core/Slide'
 
-function SlideTransition(props) {
+function SlideTransition (props) {
   return <Slide {...props} direction='up' />
 }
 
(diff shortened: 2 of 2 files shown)


jpyan:
Update to show for 200ms for each of the characters in the message

jpyan:
Add role =  alertdialog value using the ContentProps property according to "If a Snackbar requires focus to close it, then content authors should use the role of alertdialog."  from https://material-ui.com/zh/components/snackbars/

ssciolla:
Thanks, I'll review this tomorrow.

jonespm:
I think this is a [great recommendation here](https://ux.stackexchange.com/a/85898) which is consistent with the accessibility suggestions.

I'd probably change it to this to give it a minimum. I'm not sure about having a maximum as suggested there though. Though 7000ms works out to 140 characters which is also the twitter maximum and hopefully the messages aren't longer than that.

`Math.max("{snackbarMessage}" * 50, 2000)`

assets/src/components/UserSettingSnackbar.js
```diff
@@ -31,8 +37,16 @@ function UserSettingSnackbar (props) {
...
         horizontal: 'left'
       }}
+      // If a Snackbar requires focus to close it, then content authors should use the role of alertdialog.
+      ContentProps={{
+        'role': 'alertdialog',
+      }}
       open={savedSnackbarOpen}
-      autoHideDuration={3000}
+      // show for 200ms for each of the characters in the message
+      autoHideDuration={"{snackbarMessage}".length * 200}
```
ssciolla: I think (but am not 100 percent sure) this is a bug. I believe this is not pulling in the `snackbarMessage` message string, just using the string literal, which will always be 17. You actually don't need to have `snackbarMessage` wrapped in quotation marks or curly braces; you can just use `snackbarMessage`.

I like to calculate values outside of the returned component so that I can use `console.log` more easily to see what's actually happening before it's returned. Here that would mean defining a new variable like `snackbarDuration` above `return` and then referencing it below with `autoHideDuration`.

assets/src/components/UserSettingSnackbar.js
```diff
@@ -2,6 +2,12 @@ import React, { useState, useEffect } from 'react'
 import Snackbar from '@material-ui/core/Snackbar'
 import IconButton from '@material-ui/core/IconButton'
 import CloseIcon from '@material-ui/icons/Close'
+import Fade from '@material-ui/core/Fade';
```
ssciolla: You should remove this since you aren't using the `Fade` component.

README.md
```diff
@@ -30,7 +30,17 @@ You may also optionally place the json settings directly into the `ENV_JSON` env
...
+1. Connect to MySQL database.
+    ```
+    Host: 127.0.0.1
+    Username: student_dashboard_user
+    Password: student_dashboard_pw
+    Database: student_dashboard
+    Port: 5306
+    ```
+1. Navigate to `course` table and select canvas_id `canvas_id`("SELECT canvas_id From course"), which will be used in the next step.
+1. Nagivate to http://localhost:5001/courses/{canvas_id} with the canvas_id you found. (For example, with SQL file myla_test_data_2019_10_16.sql loaded in, nagivate to http://localhost:5001/courses/235420 or http://localhost:5001/courses/362855 and you can view the course page as an admin.)
```
ssciolla: I feel like with the changes introduced in PR #761, we probably shouldn't be listing any hardcoded course values and just instruct people to use the admin. So 

```
1. As you are now logged in as `root`, there are no courses listed. To view the Django admin interface, navigate to http://localhost:5001/admin
1. Click on "Courses" under "Dashboard" to view a list of sample courses.
1. Click on the link in the "Course Link" column for the sample course you wish to view.
```

@jonespm, thoughts?

assets/src/components/UserSettingSnackbar.js
jpyan: You're right. I'll fix it

assets/src/components/UserSettingSnackbar.js
jpyan: Sure

ssciolla:
> I think this is a [great recommendation here](https://ux.stackexchange.com/a/85898) which is consistent with the accessibility suggestions.
> 
> I'd probably change it to this to give it a minimum. I'm not sure about having a maximum as suggested there though. Though 7000ms works out to 140 characters which is also the twitter maximum and hopefully the messages aren't longer than that.
> 
> `Math.max("{snackbarMessage}" * 50, 2000)`

@jonespm, I like the idea of having a minimum, following these guidelines would result in a shorter duration then before (see issue #806). Material Design suggests a minimum of four seconds and a maximum of 10 seconds. None of messages are particularly long (I think they are all less than 30 characters).

Maybe we could use this same approach but stretched out? Perhaps something like this:
`Math.max(snackbarMessage * 200, 4000)`

jpyan:
Hi please review my newest commit: 

I get rid of the Fade component. 

I create a variable snackbarDuration calculating values outside of the returned component and you can use console.log("snackbarDuration: ", snackbarDuration) to see the  duration on the console. Now it's changed according to the length of messages. I also use the idea of minimum and I set it as 4000ms which makes more sense to me. 

Besides, I get rid of the semicolon and use only single quotes.

Thanks for your feedback and let me know if you have other concern!

assets/src/components/UserSettingSnackbar.js
```diff
@@ -25,14 +30,21 @@ function UserSettingSnackbar (props) {
...
+  const snackbarDuration = Math.max(snackbarMessage.length * 200, 4000)
+  
   return (
     <Snackbar
       anchorOrigin={{
         vertical: 'bottom',
         horizontal: 'left'
       }}
+      ContentProps={{
+        'role': 'alertdialog',
```
ssciolla: This can be just `role: 'alertdialog`, since you're creating an object; I think `standard` prefers unquoted property names.

assets/src/components/UserSettingSnackbar.js
```diff
@@ -2,6 +2,11 @@ import React, { useState, useEffect } from 'react'
 import Snackbar from '@material-ui/core/Snackbar'
 import IconButton from '@material-ui/core/IconButton'
 import CloseIcon from '@material-ui/icons/Close'
+import Slide from '@material-ui/core/Slide'
+
+function SlideTransition(props) {
```
ssciolla: Add a space so it's `SlideTransition (props)`.

assets/src/components/UserSettingSnackbar.js
```diff
@@ -25,14 +30,21 @@ function UserSettingSnackbar (props) {
...
         horizontal: 'left'
       }}
+      ContentProps={{
+        'role': 'alertdialog',
+      }}
       open={savedSnackbarOpen}
-      autoHideDuration={3000}
+      autoHideDuration={snackbarDuration}
+      TransitionComponent={SlideTransition}
+      transitionDuration={{exit: 400, enter: 400}}
```
ssciolla: Add spaces so it's `{{ exit: 400, enter: 400 }}`.

jpyan: fixed

assets/src/components/UserSettingSnackbar.js
jpyan: fixed

assets/src/components/UserSettingSnackbar.js
jpyan: fixed

README.md
ssciolla: @jpyan, thank you for trying to improve the README, but because this isn't really related to the Snackbar issue, let's remove these changes for now. I'm going to move my comment to issue #719, where it's more directly related and our team will address this separately.

jpyan: For sure. It was actually related to closed PR 811 https://github.com/tl-its-umich-edu/my-learning-analytics/pull/811. I guess I'm not familiar with github enough to separate my commits to different PR. I can just git revert those few commits related to README. Does that sound good to you? The change is already being merged into the master.

ssciolla: Oh, I see. Thanks for filling me in on this. To avoid this in the future, you might create separate branches for your work and then submit PRs from those branches. So for instance, you could have submitted this PR from an `issue-806-fix-snackbar-duration` branch` and the other from an `update-readme` branch.

I think we'll be okay merging this, if the other commits are in there. Let me consider for a moment.

jpyan: Thanks a lot for future suggestion! Let me know if anything needed to be changed.

jpyan:
Hi I've created a new branch and sent PR #823. Please take a look. Thank you!

ssciolla:
Since PR #823 was created and merged, I am closing this one.

Issue #670, opened 2019-08-21
Supply a better message if the course doesn't exist


jonespm:
Thank you for contributing to this project!

- Make sure to search the [issues](https://github.com/tl-its-umich-edu/my-learning-analytics/issues) for duplicates first!

## Expected behavior (A clear and concise description of what you expected to happen) : 
If you go to a course that doesn't exist for instance
/courses/99999999 it should display a message that the course doesn't exist.
## Describe the bug (Tell us what happens instead of the expected behavior) :
Currently the UI just says "Something went wrong" which is sometimes hard to find what the error is
![image](https://user-images.githubusercontent.com/27447/63454759-6937c580-c419-11e9-9514-3e0b78f6674b.png)

## Steps to Reproduce :
1. Just go to a URL that doesn't exist in the system
2. See the error in the UI

jonespm:
There's also an issue if the user doesn't have permission to the course. The api just throws a 403 but the frontend displays the same message. Need to figure out a good way of dealing with these.

luctiger98:
Hello, myself and a partner are currently students enrolled in EECS 481. We recently saw a presentation regarding MyLA in class and wanted to contribute to fixing an open source issue. We just wanted to let you know that we will be attempting to tackle this bug. We will return with further developments and questions if necessary.

justin0022:
Thanks @luctiger98, we appreciate your help! Don't hesitate to ask us if you have any questions.

jacobmal:
Hello, me and @luctiger98 have been working on this bug and we think we've solved it. Now, if a user accesses a course that is not found (404) they will see:
![404-Not_Found](https://user-images.githubusercontent.com/46611530/69679999-a1cde780-1077-11ea-8b55-cd415adf8a76.png)


And if they access a course that they do not have permission to view (403) they will see:
![403-Forbidden](https://user-images.githubusercontent.com/46611530/69680881-4ea96400-107a-11ea-9641-92bf553eaaf4.png)


Before we submitted a pull request with the changes we had a couple of questions. First, are these descriptions given for the error messages descriptive enough?

Also, while searching for a solution for this bug, I found that there is no 404 handler for links that do not exist. This is what shows up when I try to open a non existent link with Django Debug options off: 
![Not-Found-Link](https://user-images.githubusercontent.com/46611530/69686086-93d69180-108c-11ea-8026-15037478d522.png)

[...]

PR #817, opened 2019-12-04
Supplies better error message if the course doesn't exist (#670)

jacobmal:
Hi, me and @luctiger have been working on this bug fix so that we could make an open source contribution for our 481 class. The proposed change makes it so that a user will see helpful error messages if they try to access a course that does not exist (404) or a course that they do not have permission to view (403). A different message is displayed for each error. Here are the error messages:
404-Not_Found
403-Forbidden

### DIFF
--- a/assets/src/containers/Course.js
+++ b/assets/src/containers/Course.js
@@ -16,8 +16,8 @@ function Course (props) {
   const [loaded, error, courseInfo] = useCourseInfo(courseId)
   const [sideDrawerState, setSideDrawerState] = useState(false)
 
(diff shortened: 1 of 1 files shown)


pushyamig:
Functionally this works!
@jennlove-um suggests if an alternate error message needs to be provided other than showed in the screenshot

jennlove-um:
Let's make the error text a bit shorter.  It feels like we're giving them the general then the specific error when all users really need is the specific information.  For example, "Course 12345 does not exist" and "You do not have access to Course 12345"

Issue #769, opened 2019-10-24
Research Canvas data - Points data instead of percentages 


jennlove-um:
## Describe your problem or feature you'd like added
For more gamified courses, it would be better to show the Grade Distribution and Assignment Planning views.  Instructors for these types of courses emphasize points rather than grading by percentages earned.

## Describe the solution you'd like
Get grade and assignment data by points instead of by percentages from Canvas data.  Design the Grade Distribution and Assignment Planning alternative views to use points instead of percentages.  This would need to be a parameter that could be configured on the course level.

The first step is to research whether Canvas provides the data that we would need.

jonespm:
It actually *looks* like most courses are graded on points. I wonder why we went with percentages. I think this option sounds like a good one. Canvas provides the data of what the type is displayed but I feel like internally they still use points. . .

|grading_type|% of courses|
|---|---|
|letter_grade|2.06|
|percent|2.56|
|points|81.91|
|gpa_scale|0.03|
|not_graded|5.12|
|pass_fail|8.33|

pushyamig:
Yes, I think we should ask Canvas to make the point grade available in UDW. since it is right there sitting in one of their tables.

jonespm:
I'm not sure what you mean, submission_dim has score in points. We're calculating the percent toward final grade locally. 

What I think this is describing is instead of having percent toward final grade we'd have points attained toward maximum points.

jonespm:
For instance in a gamified course, you want to show something like

600/5000 (points)
This is basically the same as showing
12% of 100% (which is what we're doing now)

But it's showing the points instead of percent on the bar.

simsmat116:
Hello, I am currently in EECS481 and would like to contribute to working on this enhancement for my contribution project. Would anyone be able to give some guidance for this task?

justin0022:
Hi @simsmat116, I think this is a fairly large feature with many different underlying components. Is there a specific component you'd be interested in implementing? I imagine there is considerable work to be done both in the front and back-end - do you have a preference?

simsmat116:
Hi @justin0022 ,  would I be able to work on the back-end relating the Grade Distribution part of this issue?  I could also work on the front-end component for this as well depending on how long it takes to work on the back-end. I am looking to spend at least 12 hours of designing the changes, implementing them and testing them. Would you say this would be reasonable? Thanks!

justin0022:
[...]

PR #818, opened 2019-12-04
Grade Distribution View: Enhancement that allows for Points data to be displayed

simsmat116:
This pull request is intended to solve part of Issue #769: Research Canvas data - Points data instead of percentages. To achieve this, a new column (myla_grading_type) was added into the course table in the database that is used to specify whether the grades displayed in the Grade Distribution should be in Points or Percentages. This field is only configurable from the admin portal, but future features could be added so that this information is drawn from Canvas or is configurable by the Instructor. I have also configured the back-end and front-end to use the grading type to retrieve and display the correct information. I have provided an overview of the changes / additions made to the codebase below:

- models.py - added in myla_grading_type field in the Course class as a CharField and choices as Percentage or Point. This was done because of Zhen’s comment - https://github.com/tl-its-umich-edu/my-learning-analytics/issues/769#issuecomment-560127646

- admin.py - added in the new myla_grading_type field into the List display in the admin portal. This was done to allow for admins to configure the grading type for a course. 

- **views.py**
1. Added new method, get_course_grading_type, that queries the course table to find the grading type. This was done to avoid code duplication when finding the course grading type in the Assignment Planning and Grade Distribution views.
2. Added new method, get_user_id, that determines the user’s id by querying the use table. UserID is necessary when getting the user’s current score in points in a class via querying the submission table. It seemed appropriate to separate this query of the user table into its own function.
3. Grade Distribution View - added a new SQL query, grade_points_sql, into the backend. This query runs if the grading type is points, and the old percentage query runs if the grading type is percentage. The fields in the dataframe return from the query are the same in both cases, and this allowed for less code refactoring having to be done and for the data to integrate nicely with the front-end without many changes being made. 

- **GradeDistribution.js** - gradingType is provided as argument to the Histogram element

- **createHistogram.js** - now the ‘%’ symbol is only displayed when the gradingType is percentage. This was the only change that had to be made as the points data integrated well with the creation of the Histogram. 

Along with changes to these files, I created test cases for the following scenarios:
- Histogram with Grading Type == Points - this is located at /assets/src/__tests__/createHistogramPoints.test.js - there was already a test for creating a histogram with percentage data


jonespm:


```diff

Issues
======
- Added 1
           

Complexity increasing per file
==============================
- assets/src/__tests__/d3/createHistogramPoints.test.js  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/app/student-dashboard-django/student-dashboard-django/pullRequest?prid=4617723&bid=15486266)

pushyamig:
@simsmat116, thanks for the contribution. A couple of things
1. The SQL Query from L352-L356, ideally we want that to be `cron.py` querying the UDW tables (Myla pulls all the course data and store it locally in the MySQL DB) and store it in the `user` table by adding a new column called `points_grade`. Your approach works but not feasible if a course has more students like 500-2000 as a call to the view can get stuck for a bit when trying to calculate all the total points.
2. The front end is missing the showing the dashed line which obfuscates the lower performers on the grade scale. More info on this view can be found [here](https://sites.google.com/umich.edu/my-learning-analytics-help/home/grade-distribution?authuser=0) 
3. Usually, when you make a model change you commit the migration file so that the change to Database tables is picked up automatically. Here is the [guide](https://sixfeetup.com/blog/django-migrate-or-makemigrations) to run the migration after making a model change. Can you also update the READme with make migration details?
4. Your changes to the course table adding a new column are pretty good, You can make a separate PR just with that change. It's ok if this change might not be used but your change can be used for the future. One change I want is to have the text as `Percent` and 'Point` in Django admin UI and DB instead of a list like this `[('PERC', 'Percentage'), ('PT', 'Point'), ]`, what is the rationale for this approach?

simsmat116:
@pushyamig Thanks for the feedback! I will work on creating a separate pull request for the configuration of the 'myla_grading_type' field in the database (i.e the migration file) as well as the admin view of the course data. Would it be possible to get this reviewed by sometime early tomorrow? 

As for the other changes,  I have a one question:
For adding the SQL query to the cron.py, how should I go about testing this? I've had trouble in the past with getting the cron job to execute successfully. Is there any documentation relating to configuring this so that it works correctly?

The other changes seem relatively straightforward and I should be able to get these completed in the next 2-3 weeks.

pushyamig:
>For adding the SQL query to the cron.py, how should I go about testing this? I've had trouble in the past with getting the cron job to execute successfully. Is there any documentation relating to configuring this so that it works correctly?

To execute `cron.py` you need access to Unizin Dataware House (UDW) which you cannot have access since it has student data in it. the reason that is not working for you is you need to add UDW credentials and then it is pretty straightforward to run cron.

pushyamig:
>The other changes seem relatively straightforward and I should be able to get these completed in the next 2-3 weeks.

I have merged your changes with model change with the course table, thanks for the contribution. This issue is a research issue and it served the purpose that is intended to. We will take it from here.

PR #820, opened 2019-12-05
Provide a way to override variables with environment Issue 787

emilychee:
Modified settings.py to take in user-defined environmental variables over the default environmental variables by adding several if statements.


jonespm:
First, I removed the SQL file that was included

Second, thanks for the contribution but I really don't want to change this entire file to fulfill this feature request with if/else statements everywhere. This is just way too error-prone and not something I'd like to do. I believe this will also cause a problem for booleans and strings unless we're casting to the correct types. We used to use a library in the past version of this code [decouple](https://pypi.org/project/python-decouple/) that did this casting but I don't want to add that back again. 

I feel like for this to work as individual variables we'd have to either have a library that does type inference which is going to be a little more work than this (unless there's one out there).

Probably the *easier* way to fulfill the initial request I'd think is just to use this special ENV called
ENV_JSON

The thing with that is it's either or. You either have the entire env defined in ENV_JSON or you have it defined in the file. What I'd do instead of this.

- Try to load The env from the file into ENV (on lines 32-38)
- Whether or not this passes or fails get the value ENV_JSON from the environment as ENV_JSON (lines 27-29)
- Update the values from ENV_JSON into ENV. (You can just do this with ENV.update(ENV_JSON)
- If ENV is still empty after all this, just set it to os.environ so it still runs.

This wouldn't allow someone to define individual values in the ENV and have them override but they could define a JSON in the env and it would override. I feel like this would be sufficient for the request and avoid modifying this entire file. Sorry as it looks like you did a a lot of work here.

jonespm:
I'm closing this since it's out of date and we need a more programmatic way of supporting this feature.

Issue #661, opened 2019-08-14
Reorganize docs and configs into a docs folder


jonespm:
Most large python projects seem to organize their documentation into a docs folder. This would probably make sense to do.

Like: https://github.com/django/django

We could move the config/env_sample.json, our existing docs and create something nice like 
https://docs.readthedocs.io/en/stable/intro/getting-started-with-sphinx.html

Then as we documented the code better it would also generate better documentation. #

_Originally posted by @jonespm in https://github.com/tl-its-umich-edu/my-learning-analytics/pull/657#issuecomment-521266024_

justin0022:
Great @huynhdom let us know if you need a hand

jennlove-um:
The separate pages of the README.md look good.  I verified and updated a couple of links and added navigation links at the bottom of pages to let users move from one section to the next without going back to the README page if they want.

PR #825, opened 2019-12-10
Add support for sphinx documentation

huynhdom:
Feature request https://github.com/tl-its-umich-edu/my-learning-analytics/issues/661

Sphinx documentation and setup. There are still somethings left fairly bare bones as I don't know what Is actually wanted for the index page or the general setup or what Documentation files are wanted to be included, so I just did a general setup with the README, LICENSE, and CONTRIBUTING and how they can be linked to each other and how to link a document to the index.txt.

Most of the features I chose for Sphinx when setting up can be modified, disabled, or removed in the conf.py file located in the source folder in the docs folder. Django can also be set up in that file if desired, but I couldn't really test anything as Djano is not something I know about, and there was no provided file for me to use. HTML theme, CSS. Pygments, and other visual elements can also be handled there. 

From what I can tell, Sphinx seems to only be able to handle .txt or .rst or .md files. The current set default is .txt, which can be modified. It seems that you can link documents that aren't .txt in the index file (they're all set to .txt currently). For example, .md files work just fine, but it will throw a fit if you have no extension like the LICENSE file. 

The files also have to have a title. As seen in the modified LICENSE file. For example,
LICENSE needs to be followed by a new line of ===============, 
Or something of similar format must be there, otherwise errors are thrown. You also need a newline before you try linking documents within each other or to place them in the given index file

This is the current final product 

image4

This is a past iteration. Just proof that the format can be changed fairly easily 
image6

First iteration with no styling
image7

It should also be noted that there's no build folder currently. There should be, but I hat to remove it so I could commit the changes. That's where the html is located after you build it. In order to generate the desired html, you just need to type "make html" in the doc folder and it should generate. From there, you can go to the build/html/index.html to see the results.


zqian:
@huynhdom is there an existing Github issue for this? If so, could you please mention/link to the issue number in this pull request?

huynhdom:
Ah right, sorry to include that. 
The issue  link is here: https://github.com/tl-its-umich-edu/my-learning-analytics/issues/661

ssciolla:
@huynhdom, thank you for your contribution and effort on this. We finally are revisiting this issue, and we've decided for the time being to not adopt Sphinx. I'm going to close this PR, but we will keep your work in mind if we decide to go down this path in the future. Best wishes

PR #831, opened 2019-12-18
issue_829 added support for a null asset_name in BQ query

zqian:
added the check for  not_null values of asset_name.

Long term solution will be also add checks for sqlalchemy.exc.OperationalError for every pandas dataframe to_sql inserts.


jonespm:
I'm looking at the data and not sure avoiding nulls is right. It seems like something may have changed in the caliper events? I'm seeing a missing asset_name but the event has both filename and display_name. I think it's fine if nulls get in this column because the process is still to re-set the filename from the UDW so whether they're set or not it should fix them up, this was changed in #744.

- It seems like Prior to around October object.name was attachment and we were using that
- However they provided around this time a new object.extensions[\'com.instructure.canvas\'].asset_name with the actual name
- But some events don't have asset_name, only filename and display_name

Looking at our count there's 25145086 events with a null asset_name and 89785394 with a not null asset name. However there's 11992205 with a null display_name and filename. 

I think what we should do is a COALESCE in the SQL to try to find a value here, this will for sure return some string. 

`COALESCE(JSON_EXTRACT_SCALAR(event, '$.object.extensions[\'com.instructure.canvas\'].asset_name'), JSON_EXTRACT_SCALAR(event, '$.object.extensions[\'com.instructure.canvas\'].filename'), JSON_EXTRACT_SCALAR(event, '$.object.name'), 'attachment')`

jonespm:
I checked the data it looks like there are no null asset_names for NavigationEvent attachment objects after 2019-12-06, but they did exist all the way up till then. But the code is still there to fix up everything anyway.

jonespm:
FYI: if you update this to have in the config you'll have to double escape the single quotes in there.

zqian:
COALESCE statement is added to combine all sources of  file names. Please check again.

jonespm:
Looks good to me!

Issue #485, opened 2019-04-18
Update to Django 2.2 LTS


jonespm:
Django 2.2 LTS [has just been released](https://www.djangoproject.com/download/). We should update to this version over the summer.

All other django-* libraries in the requirements should also be tested and updated as we move to this new version

jonespm:
@ssciolla was talking about this so I'm assigning to him to also look at what this would take.

ssciolla:
I made some good progress on this today and worked with @jonespm to debug a couple issues. With the latest commits in my [issue branch here](https://github.com/ssciolla/my-learning-analytics/commits/issue-485-django-update), I was able to get the application to run (and collect data via the cron) using Python 3.7 and Django 2.2.6. We'll need to do more testing to make sure all the dependencies are still working as expected and that this also works in OpenShift, but hopefully I'll have a PR soon.

ssciolla:
It's looking like issue #484 would also be addressed by my work on this.

ssciolla:
@jonespm, speaking of SAML, I'm going to deploy this to test, and I'm pretty sure I'll find that the djangosaml2 library we're using will break, since it doesn't appear to support Django 2.2 (it doesn't appear to have been rewritten using Django 2 at least).

If or when that happens, we'll need to have an exit strategy for this dependency and a replacement. [`social-app-django`](https://github.com/python-social-auth/social-app-django) seems to be alternative, but perhaps you've already done a survey of what's out there? Thoughts?

ssciolla:
Update: I deployed to the test instance, and it didn't seem to break the SAML authentication. I was seeing a couple tracebacks (I think related to the Django Debug Toolbar and Watchman?). I'll need to do some more testing; it still seems like it might be a good idea to move away from `djangosaml2`.

jonespm:
I haven't really looked at anything else. For next year I'd probably prefer
just to be using LTI instead rather than the SAML integration. This was
just more convenient for us to setup at the time. It's also possible the
LTI module will need to switch as this is the only 1.3 library I can find.

https://pypi.org/project/PyLTI1p3/

ssciolla:
Hm... okay, but that would only solve MyLA; we may need to solve this problem for Student Explorer.

jonespm:
Student Explorer is on holiday after this release. ;) 🍹

jonespm:
Yeah, if it doesn't work we'll have to look at something else. This library wasn't my choice it was just already being used.

[...]

PR #842, opened 2020-01-09
Update Python, Django, other dependencies, and streamline paths, views (#484, #485, #843)

### LINKED ISSUE #484: Update SQLAlchemy and database libraries
SQLAlechemy should be updated to 1.3.x. 
We also should update MySQL client to 1.4.x and other database libraries as needed.

ssciolla:
This PR updates a number of project dependencies, including bumping the Python version to 3.7, the Django version to 2.2.9, and other packages to more current releases. It also makes changes to fix bugs either caused by dependency changes or discovered during the upgrade process. The major changes include the following:

- [x] Updating `Dockerfile` and `requirements.txt`
- [x] Replacing uses of `url()` with `path()` and `re_path()` in `urls.py`
- [x] Updating `settings.py` to remove unneeded middleware and add SameSite cookie settings
- [x] Fixing the `myla_globals` implementation (see #843)
- [x] Removing `get_course_template` to streamline existing `urlpatterns` and views
 
Some details are provided below to explain noteworthy, unexpected, or arbitrary decisions. The PR aims to resolve #484, #485, and #843.

1) The majority of the changes needed for the move to Django 2.2 involved changing uses of `url()` for `urlpatterns` to `path()` and `re_path()`. In addition, we removed`django.contrib.auth.middleware.SessionAuthenticationMiddleware` (since [similar functionality is now baked into Django](https://docs.djangoproject.com/en/3.0/releases/2.0/#miscellaneous)) and changed the `accounts/logout` pattern to use the class-based `LogoutView`.

2) To make MyLA work as an LTI tool in an iframe, we had to set two new `settings.py` variables that were introduced in Django 2.1: [`SESSION_COOKIE_SAMESITE`](https://docs.djangoproject.com/en/2.2/ref/settings/#session-cookie-samesite) and [`CSRF_COOKIE_SAMESITE`](https://docs.djangoproject.com/en/2.2/ref/settings/#csrf-cookie-samesite). Both of these values were made configurable and set to `None` to enable full functionality in an iframe. While this makes LTI integration work temporarily, we expect that we will need to change the defaults as popular Internet browsers roll out changes in 2020 that require this value to be set to the string literal `"None"`.

2) While implementing and testing changing the uses of `url` to `path` in `urls.py`, I realized that some views that use `base.html` may not have access to `myla_globals`. PR #784 had moved the construction of `myla_globals` into the `get_home_template` view, making it only available to the context in that view. With this PR's code, the only other template that extends `base.html` and needs access to `myla_globals` besides `frontend/index.html` would be `registration/login.html`. Though the application does not rely overmuch on templates, I thought it made sense to find an application-wide solution to prevent possible issues in the future. While one possibility would have been to move `myla_globals` back to the template, I opted for a solution that would keep that logic in Python (rather than the template language): a context processor. I followed existing patterns to update `context_processors.py`, `util.py` and `settings.py`. One developer I discussed this with had possible performance-related concerns with this approach. It would be great to have assistance testing this, and I'm open to other alternatives.

3) You'll notice that I also changed how `myla_globals` is passed to the frontend. I found a [new template tag](https://docs.djangoproject.com/en/2.2/ref/templates/builtins/#json-script) called `json_script` introduced in Django 2.1 that seemed well-suited to what we were trying to accomplish, and I followed the documentation's approach in implementing this in `base.html` and `App.js`.

4) While updating `urlpatterns`, I realized that the `get_course_template` view wasn't actually seeing much use in the application. The `courses/` catch-all pattern was matching most requests, and the React frontend was handling the routing from that point. As a result, the `courses/{some_int}` pattern, which used the `get_course_template` view, was only triggered by `Course.get_absolute_url()` in `models.py` and by the `reverse` in `lti.py`. Furthermore, the `get_course_template` view was very close in form to `get_home_template`. To simplify the code, I decided to change the view referenced by the `courses/{some_int}` pattern to `get_home_template` and remove the code related to `get_course_template`. The previously mentioned uses of the `courses/{some_int}` pattern still work as expected. I believe this more clearly aligns the code with how the patterns were already being used, but I'd be interested to hear what others think.


jonespm:


```diff

Issues
======
+ Solved 2
- Added 1
           

Complexity increasing per file
==============================
- dashboard/common/utils.py  2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/ITS_Teaching_And_Learning/student-dashboard-django/pullRequest?prid=4793979&bid=15958692)

jonespm:


```diff

Issues
======
+ Solved 10
           

Complexity increasing per file
==============================
- dashboard/common/utils.py  2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/ITS_Teaching_And_Learning/student-dashboard-django/pullRequest?prid=4793979&bid=15958692)

assets/src/containers/App.js
```diff
@@ -1,4 +1,3 @@
-/* global myla_globals */
```
jonespm: This comment was for Codacy but I guess it didn't need it anymore with the change of variable.

ssciolla: Yeah, because I opted to use the `json_script` tag and the `document.getElementById` call, I don't think we need it anymore?

Issue #462, opened 2019-03-22
[React] Customize styling and theme


Adonais0:
Not a high priority issue, but better to spend some time think of. 

Currently, we are using the default material styling. In the future, we'll want to add customized themes to our design to better fit [UofM color branding guidlines](http://vpcomm.umich.edu/brand/style-guide/design-principles/colors) and provide better visual accessibility.

Reference: https://material-ui.com/customization/themes/

pushyamig:
This is for later as well, just important for this react branch getting merged master

ssciolla:
We have a newer issue about this, but I want to establish the link so we close both. It's issue #673.

PR #845, opened 2020-01-16
Make primary color configurable and improve globals use by FE (#821)

### LINKED ISSUE #821: Make primary UI color configurable

Different institutions should be able to modify the `primary` structural color, which is currently a UM blue (#40658F).

To keep this simple for now, I think I will make this color configurable through `env.json` (`env.json` -> `settings.py` -> `views.py` (`myla_globals`) -> `base.html` -> `App.js`.

I have also looked into modifying the data model and having the primary color be a field in an institution table, so that the color can be changed via the Django admin interface. Some of our discussions indicated that this would be unnecessary currently and bloat the project with extra code, given that such a value would only be changed a couple times at most in one deployed instance.

I also looked into using a hook and creating an API endpoint to pass data to the frontend. However, it makes sense to leverage `myla_globals` for this application-wide setting.

ssciolla:
As a follow-up to my work on PR #775, this PR makes the primary color of the theme's palette in the front end able to be configured through `env.json`. To accomplish this, I needed to make some changes to `index.js` and `App.js`, rename `siteTheme.js` to `defaultPalette.js`, and create a `globals.js`. My approach will be discussed below. This PR aims to resolve #821 and finishes resolving #462 and #673.

Some of these changes are straightforward, but I want to explain in a bit more detail why I created `globals.js` (and the changes to other files associated with that). Previously, I had used `createMuiTheme` within the `siteTheme.js` module and then imported the resulting theme object to `App.js`. However, when I made the `primary` color configurable, I needed to find a way to update the palette using `mylaGlobals` (which was previously processed in `App.js`) before passing it to `createMuiTheme`. I tried first to create the theme within `App.js`, but I also needed to export `siteTheme` so that I could make it available to the `d3` and `rangeSlider` components. My attempts to do this from `App.js` ran up against some perplexing errors, with my best guess being that they introduced circular dependencies. 

To solve this, I decided to extract the `mylaGlobals` processing to a new module, `globals.js`, where I could import a `defaultPalette` object,  create a `siteTheme`, and then export it along with other processed globals. To keep `App.js` as self-contained as possible, I also elected to import these globals to `index.js`, wrap `App` with `ThemeProvider` there, and pass the globals `App` needed as `props`. With this arrangement, I was also able to cleanly import `siteTheme` from `globals` to the non-React components.


assets/src/globals.js
```diff
@@ -0,0 +1,28 @@
...
+const user = Object.freeze({
+  username: mylaGlobals.username,
+  admin: mylaGlobals.is_superuser,
+  relatedCourses: mylaGlobals.user_courses_info,
+  isSuperuser: mylaGlobals.is_superuser,
+  isLoggedIn: !!mylaGlobals.username,
+  helpURL: mylaGlobals.help_url
+})
+
+const palette = JSON.parse(JSON.stringify(defaultPalette))
```
ssciolla: @justin0022, I know I probably isn't critical to clone the object here (since we don't really need a stable `defaultPalette` for any reason I can think of), but it seemed like a best practice. Do you think it's fine to use `JSON.stringify` here? I also looked at `_.cloneDeep`, but it seemed somewhat heavyweight for this purpose. We don't list `lodash` as a dependency in `package.json` (though it's available, I believe because it's a dependency of many of the packages we use) and it's not used anywhere else.

justin0022: Hmm I'm not sure if this is necessary, maybe you could simply freeze it?

ssciolla: How would you approach that? I couldn't freeze `defaultPalette`, since then I couldn't update that object (stored in a new variable) with a specified primary color. I'm also not sure I can freeze `palette`, since I'm guessing `createMuiTheme` does some magic to fill in things you didn't specify.

I think the question is, should I bother cloning at all? It's not much code and leaves open some more possibilities, so I guess I leaned in that direction.

justin0022: Ya that's fair, I think you can just leave it as is for now.

Issue #301, opened 2018-11-15
Implement PEP 3107 for function annotations


jonespm:
We should improve the code quality and implement function annotations for return types and parameter types to MyLA. 

This is described here as reference

https://www.python.org/dev/peps/pep-3107/
https://python-future.org/func_annotations.html

jacobmal:
Hi, my EECS 481 partner (luctiger98) were looking at adding this enhancement too. We were added to the google sheet for tracking the 481 student assignments. Should I add a row to the sheet with this issue and my uniqname?

jonespm:
Yeah, sure go ahead and keep the sheet updated so we can better track who's working on what. We plan to assign people have committed to issues to the issues by the end of the week.

jonespm:
It might be too much work to do every python class for this. If you want to break it up into a subtask and only get to views, models or some other class that would still be a nice improvement for us. 

Something like this might be of assistance: https://stackoverflow.com/questions/44057003/how-to-annotate-django-views-methods

jonespm:
We had a contribution to do this but decided it would be better to implement this as needed rather than all at once. Might revisit this.

PR #848, opened 2020-01-22
Issue #301 added timestamp to BigQuery statements

zqian:
There are two steps in this fix:

- get the earliest start date for all courses on file
- add the start date parameter into the BigQuery statement


config/env_sample.json
```diff
@@ -176,13 +176,13 @@
...
+            "query": "select 'leccap' AS resource_type, CAST(SUBSTR(JSON_EXTRACT_SCALAR(event, '$.object.id'), 48) AS STRING) AS resource_id, @canvas_data_id_increment + CAST(JSON_EXTRACT_SCALAR(event, '$.federatedSession.messageParameters.custom_canvas_user_id') AS INT64) AS user_id, @canvas_data_id_increment + CAST(JSON_EXTRACT_SCALAR(event, '$.federatedSession.messageParameters.custom_canvas_course_id') AS INT64) AS course_id, JSON_EXTRACT_SCALAR(event, '$.object.name') as name, datetime(EVENT_TIME) as access_time FROM event_store.events where JSON_EXTRACT_SCALAR(event, '$.edApp') = 'https://leccap.engin.umich.edu/#applicationName=Lecture+Capture' and type = 'MediaEvent' and JSON_EXTRACT_SCALAR(event, '$.action') = 'Started' and JSON_EXTRACT_SCALAR(event, '$.federatedSession.messageParameters.custom_canvas_course_id') is not null and JSON_EXTRACT_SCALAR(event, '$.federatedSession.messageParameters.custom_canvas_course_id') IN UNNEST(@course_ids_short) and event_time > @course_start_time",
```
ssciolla: It would be great to see these queries re-formatted for better readability!

dashboard/cron.py
```diff
@@ -261,12 +261,14 @@ def update_with_bq_access(self):
...
                 bigquery.ArrayQueryParameter('course_ids_short', 'STRING', data_warehouse_course_ids_short),
-                bigquery.ScalarQueryParameter('canvas_data_id_increment', 'INT64', settings.CANVAS_DATA_ID_INCREMENT)
+                bigquery.ScalarQueryParameter('canvas_data_id_increment', 'INT64', settings.CANVAS_DATA_ID_INCREMENT),
+                bigquery.ScalarQueryParameter('course_start_time', 'TIMESTAMP', utils.find_earliest_start_datetime_of_courses())
             ]
             job_config = bigquery.QueryJobConfig()
             job_config.query_parameters = query_params
 
             # Location must match that of the dataset(s) referenced in the query.
+            logger.info(final_bq_query)
```
ssciolla: Is the `info` level intentional here? Maybe `debug` makes more sense?

config/env_sample.json
jonespm: I don't believe JSON allows for multiline strings. It would have to be some hacky solution where we maybe had an array that had each pice on a different line then the code combined it. :/ 

I think these queries can be simplified in other ways though by pulling the patterns out of them but that's a fix I haven't gotten to yet.

For example:
```
SELECT 'canvas' AS resource_type, CAST(SUBSTR(JSON_EXTRACT_SCALAR(event, '$.object.id'), 35) AS STRING) AS resource_id, CAST(SUBSTR(JSON_EXTRACT_SCALAR(event, '$.membership.member.id'), 29) AS INT64) AS user_id, CAST(SUBSTR(JSON_EXTRACT_SCALAR(event, '$.group.id'),31) AS INT64) AS course_id, COALESCE(JSON_EXTRACT_SCALAR(event, '$.object.extensions[\\'com.instructure.canvas\\'].asset_name'), JSON_EXTRACT_SCALAR(event, '$.object.extensions[\\'com.instructure.canvas\\'].filename'), JSON_EXTRACT_SCALAR(event, '$.object.name'), 'attachment') as name, datetime(EVENT_TIME) as access_time`
````
Could probably be something like
```
resource_type: "canvas",
resource_id: "SUBSTR(JSON_EXTRACT_SCALAR(event, '$.object.id'), 35)",
user_id: "SUBSTR(JSON_EXTRACT_SCALAR(event, '$.membership.member.id'), 29)",
course_id: "SUBSTR(JSON_EXTRACT_SCALAR(event, '$.group.id'),31)",
```

This could probably be improved further since many of these are pulling substrings out of the json event so that could be special too. 

We could probably improve on all other parts of this query too by finding and replacing the patterns, or using some "methods" to do that. I really don't like how this is either but this is the real query. ;)

File: dashboard/common/utils.py [hunk 3]
```diff
@@ -4,10 +4,21 @@
...
 
 logger = logging.getLogger(__name__)
 
 
+def find_earliest_start_datetime_of_courses():
+    course_start_datetimes = []
+    for course in Course.objects.all():
+        course_start_datetimes.append(course.get_course_date_range().start)
+    logger.debug(course_start_datetimes)
+    earliest_start = sorted(course_start_datetimes)[0]
```
jonespm: This should *probably* do something if course_start_datetimes is still empty (like if there's no courses)

Though the cron won't run if there's not any courses and this is currently only called from the cron.

config/env_sample.json
```diff
@@ -176,13 +176,13 @@
...
+            "query": "SELECT 'canvas' AS resource_type, CAST(SUBSTR(JSON_EXTRACT_SCALAR(event, '$.object.id'), 35) AS STRING) AS resource_id, CAST(SUBSTR(JSON_EXTRACT_SCALAR(event, '$.membership.member.id'), 29) AS INT64) AS user_id, CAST(SUBSTR(JSON_EXTRACT_SCALAR(event, '$.group.id'),31) AS INT64) AS course_id, COALESCE(JSON_EXTRACT_SCALAR(event, '$.object.extensions[\\'com.instructure.canvas\\'].asset_name'), JSON_EXTRACT_SCALAR(event, '$.object.extensions[\\'com.instructure.canvas\\'].filename'), JSON_EXTRACT_SCALAR(event, '$.object.name'), 'attachment') as name, datetime(EVENT_TIME) as access_time FROM event_store.events where JSON_EXTRACT_SCALAR(event, '$.edApp.id') IN UNNEST(['http://m.canvas.umich.edu/' , 'http://umich.instructure.com/' ]) and type = 'NavigationEvent' and SUBSTR(JSON_EXTRACT_SCALAR(event, '$.object.id'),24,10) = 'attachment' and JSON_EXTRACT_SCALAR(event, '$.action') = 'NavigatedTo' and JSON_EXTRACT_SCALAR(event, '$.membership.member.id') is not null and SUBSTR(JSON_EXTRACT_SCALAR(event, '$.group.id'),31) IN UNNEST(@course_ids) and event_time > @course_start_time",
```
jonespm: I feel like we'd want this where clause for every query. I'd suggest putting it in the cron in the for loop around line 254 and just append the and event_time > @course_start_time. Then we don't have to change this config. 

There might be times when we want to run everything. Maybe just add a new key here and check that? The default can just be true?

`'limit_to_course_start': False`

config/env_sample.json
zqian: I have hacked the way via turning the long sql statement into array (multi lined) and the array items are joined later to form one string again.

As for the query simplification, I think it is worth another issue (issue #856)

config/env_sample.json
zqian: I moved the cause into cron.py

I will refrain from adding another param 'limit_to_course_start', until we see the need of this. For now, all BigQuery query will have course start time as one of the where clause.

dashboard/common/utils.py
zqian: addressed in the latest code.

dashboard/cron.py
```diff
@@ -251,7 +254,14 @@ def update_with_bq_access(self):
 
             final_bq_query = []
             for k, query_obj in settings.RESOURCE_ACCESS_CONFIG.items():
-                final_bq_query.append(query_obj['query'])
+                # concatenate the multi-line presentation of query into one single string
+                query = " ".join(query_obj['query'])
+
+                if (course_start_time is not None):
+                    # insert the start time parameter for query
```
ssciolla: "concatenate" over "insert"?

dashboard/cron.py
```diff
@@ -251,7 +254,14 @@ def update_with_bq_access(self):
...
+                    query = query + " and event_time > @course_start_time"
```
ssciolla: `query += " and event_time > @course_start_time"`?

dashboard/cron.py
```diff
@@ -263,10 +273,15 @@ def update_with_bq_access(self):
...
             ]
+            if (course_start_time is not None):
+                # insert the start time parameter for query
+                query_params.append(bigquery.ScalarQueryParameter('course_start_time', 'TIMESTAMP', course_start_time))
+
             job_config = bigquery.QueryJobConfig()
             job_config.query_parameters = query_params
 
             # Location must match that of the dataset(s) referenced in the query.
+            logger.debug(final_bq_query)
```
ssciolla: It looks like `final_bq_query` will be the same here as when it is logged above `query_params`. Can we remove one of these?

config/env_sample.json
```diff
@@ -176,13 +176,45 @@
...
+                "datetime(EVENT_TIME) as access_time ",
+                "FROM event_store.events ",
+                "where JSON_EXTRACT_SCALAR(event, '$.edApp.id') IN ",
+                "UNNEST(['http://m.canvas.umich.edu/' , 'http://umich.instructure.com/' ]) ",
+                "and type = 'NavigationEvent' ",
+                "and SUBSTR(JSON_EXTRACT_SCALAR(event, '$.object.id'),24,10) = 'attachment' ",
+                "and JSON_EXTRACT_SCALAR(event, '$.action') = 'NavigatedTo' ",
+                "and JSON_EXTRACT_SCALAR(event, '$.membership.member.id') is not null ",
+                "and SUBSTR(JSON_EXTRACT_SCALAR(event, '$.group.id'),31) IN UNNEST(@course_ids) "
+            ],
```
ssciolla: I think I like this. Ultimately it might make sense to put these queries in separate files, but this makes it a lot more readable! Glad you added the comment to `cron.py` about the multi-line presentation, as it could be confusing initially why the join is necessary.

config/env_sample.json
```diff
@@ -176,13 +176,45 @@
...
+            "query": [
+                "SELECT 'canvas' AS resource_type, ",
+                "CAST(SUBSTR(JSON_EXTRACT_SCALAR(event, '$.object.id'), 35) AS STRING) AS resource_id, ",
+                "CAST(SUBSTR(JSON_EXTRACT_SCALAR(event, '$.membership.member.id'), 29) AS INT64) AS user_id, ",
+                "CAST(SUBSTR(JSON_EXTRACT_SCALAR(event, '$.group.id'),31) AS INT64) AS course_id, ",
```
ssciolla: Add a space before `31`?

ssciolla: Saw a couple other places where you could add a space before integers. Up to you whether it's necessary.

jonespm:


```diff

Issues
======
- Added 1
           
```


See the complete overview on [Codacy](https://app.codacy.com/manual/ITS_Teaching_And_Learning/my-learning-analytics/pullRequest?prid=4877958&bid=16165502)

Issue #837, opened 2020-01-03
Truncating to 7 characters feels a little short, should make this value larger


jonespm:
## Expected behavior (A clear and concise description of what you expected to happen) : 
The new resources view is truncating to 7 characters with no way to change from https://github.com/tl-its-umich-edu/my-learning-analytics/blob/177816663c337f83740e3a75f3ead15a5c509ca3/assets/src/components/d3/createResourceAccessChart.js#L233

This seems a little short. It looks like file names in Canvas get about 40 characters until they're truncated. I think we should work to at least fit 20 characters and seems like there's more space.

![image](https://user-images.githubusercontent.com/27447/71747537-ec444000-2e3d-11ea-8166-1df06d056896.png)

jonespm:
I wonder if we'd have to use a fixed width font here for these file names? 23 characters all caps looks like it overlaps at about character 18 now. Also I don't know why we have 4 dots for some of these? 

I think we either need to re-open this or still do something to fix this. Fixed width font, reduce the characters a few more or move everything more left. (There seems to be a lot of space in the margins)

![image](https://user-images.githubusercontent.com/27447/72260549-17b6ef80-35e1-11ea-8899-b8b26cd8104f.png)

jennlove-um:
@pushyamig Testing is okay in myla-test.  Do you want to close this issue with the limited fix in place so far and open a new issue for improvements in the future?

pushyamig:
We will close this issue, I will open another one and put it in winter.02.02

PR #854, opened 2020-01-27
#837- adjusting the resource name too long

pushyamig:
#837 
This PR addresses some of the issues that are raised by #837 
1. decreased the size between icon and resource text name
2. The resource will capital case (like "This is the sentence") overflow is addressed with this PR but if the file name has Upper case then an overflow is expected. 
Screen Shot 2020-01-27 at 12 32 44 PM
3. address the Some of the JS Lint errors.

The solution to the overflow is not fully addressed but this is workable for now.


jonespm:
This looks like it's able to use "[getComputedTextLength](https://stackoverflow.com/a/42718947/3708872)" to figure out exactly how far to make the text.

It adds the ellipses to the x-axis rows instead of the y-axis. I was looking at what it would take to convert it to our code for a bit yesterday but didn't finish. https://stackoverflow.com/a/39848896/3708872

There is the truncate method and it might need to have the length (or node) passed in as a parameter.

pushyamig:
Let me give it another try.

jonespm:
This is related to #854

PR #860, opened 2020-01-31
#857 insufficient grades count won't shown the view

### LINKED ISSUE #857: Traceback from Grade Distribution view with less than five grades

The Grades Distribution view should provide a meaningful result, even when there are no grades or only a couple of them. The Django view should also execute without tracebacks.

When I visited the Grades Distribution page for a course with only a couple of grades, I saw the standard "Something went wrong; please try again later" message. In the logs, there was a traceback from the `grade_distribution` view:

```
File "/code/dashboard/views.py", line 335, in grade_distribution
BinningGrade = find_binning_grade_value(grades)
File "/code/dashboard/views.py", line 658, in find_binning_grade_value
fifth_item = grades[4]
IndexError: list index out of range
```

## Steps to Reproduce :
1. Set up MyLA.
2. Run the cron for a course with less than five grades for students
3. Visit the Grades Distribution view for that course.

pushyamig:
Fixes #857

### DIFF
--- a/dashboard/views.py
+++ b/dashboard/views.py
@@ -313,8 +313,8 @@ def grade_distribution(request, course_id=0):
         from user where course_id=%(course_id)s and enrollment_type='StudentEnrollment';
                     """
     df = pd.read_sql(grade_score_sql, conn, params={"current_user": current_user, 'course_id': course_id})
-    if df.empty or df['current_grade'].isnull().all() or df.count().current_grade < 6:
-        logger.info(f"Not enough students grades {df.count().current_grade} in a course {course_id} to show the view")
+    if df.empty or df.count().current_grade < 6:
+        logger.info(f"Not enough students grades (only {df.count().current_grade}) in a course {course_id} to show the view")
         return HttpResponse(json.dumps({}), content_type='application/json')
 
     grade_view_data = dict()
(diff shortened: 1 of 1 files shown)


dashboard/views.py
```diff
@@ -313,7 +313,8 @@ def grade_distribution(request, course_id=0):
         from user where course_id=%(course_id)s and enrollment_type='StudentEnrollment';
                     """
     df = pd.read_sql(grade_score_sql, conn, params={"current_user": current_user, 'course_id': course_id})
-    if df.empty or df['current_grade'].isnull().all():
+    if df.empty or df['current_grade'].isnull().all() or df.count().current_grade < 6:
```
ssciolla: Could we simplify this condition to just `df['current_grade'].count() < 6`? If we always have a `current_grade` column in the DataFrame we get back from the SQL query, this should catch any of the cases, I think?

ssciolla: Also, do we want the threshold to be higher than 6? I can imagine in the short window where some students have grades and others don't, we could have similar privacy concerns to what we had with outliers (i.e. bars with 1 student at the top or bottom)? @jennlove-um, any thoughts?

dashboard/views.py
```diff
@@ -313,7 +313,8 @@ def grade_distribution(request, course_id=0):
...
+        logger.info(f"Not enough students grades {df.count().current_grade} in a course {course_id} to show the view")
```
ssciolla: Maybe modify the comment so the number of grades are in parentheses? Something like this:
`"Not enough students (only {df['current_grade'].count()}) in a course ..."`

PR #866, opened 2020-02-03
Fixes #865 - Cron except should be logged as exception

### LINKED ISSUE #865: Cron except should be logged as exception

When the cron has an exception we should log this as an exception rather than info, this would print more information. Also I'd like to add the line number and file to the log messages to help track down issues easier

jonespm:
I removed the processid from the log message because it doesn't seem to be useful to me. We could add it back in. I'm testing this out really quick.

### DIFF
--- a/dashboard/cron.py
+++ b/dashboard/cron.py
@@ -280,6 +280,7 @@ class DashboardCronJob(CronJobBase):
             job_config = bigquery.QueryJobConfig()
             job_config.query_parameters = query_params
 
+            logger.debug(final_bq_query)
             # Location must match that of the dataset(s) referenced in the query.
             bq_query = bigquery_client.query(final_bq_query, location='US', job_config=job_config)
             #bq_query.result()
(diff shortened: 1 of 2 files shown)


dashboard/cron.py
```diff
@@ -280,6 +280,7 @@ def update_with_bq_access(self):
             job_config = bigquery.QueryJobConfig()
             job_config.query_parameters = query_params
 
+            logger.debug(final_bq_query)
```
ssciolla: We have this on line 269, and I don't think the value changes after that point, correct?

dashboard/settings.py
```diff
@@ -241,7 +241,7 @@
     # Gunicorns logging format https://github.com/benoitc/gunicorn/blob/19.x/gunicorn/glogging.py
     'formatters': {
         "generic": {
-            "format": "%(asctime)s [%(process)d] [%(levelname)s] %(message)s",
+            "format": "%(asctime)s [%(levelname)s] [%(filename)s:%(lineno)d] %(message)s",
```
ssciolla: This is cool. Are these variables defined by Gunicorn?

jonespm: These are defined by python logging [LogRecord](https://docs.python.org/3/library/logging.html#logrecord-objects)

I think I copied the format from Gunicorn originally because it was better than what we had, but looking at some of these and running this locally now having the file and line number is pretty nice.

jonespm:
Just run the cron locally to test it! I've ran the cron locally and it works as expected. I'll look at that duplicate line and merge this.

Issue #870, opened 2020-02-04
When DataFrames from some queries are concatenated, NaT is replaced with None in datetime fields


ssciolla:
Thank you for contributing to this project!

- Make sure to search the [issues](https://github.com/tl-its-umich-edu/my-learning-analytics/issues) for duplicates first!

## Expected behavior (A clear and concise description of what you expected to happen) : 

The code that updates the datetime fields in the course table should be able to run without error.

## Describe the bug (Tell us what happens instead of the expected behavior) :

After the changes in PR #868, we encountered errors when running the cron that suggested the null handling was no longer working. After some investigation, I determined that when we use pd.concat() on line 136, `None` values appears to be converted to `NaT`. This was not occurring before when we used `SELECT * FROM ...`, but is now happening when we specify fields. The `NaT`s are not properly handled by `update_datetime_field`, leading to errors further on in the code.

## Steps to Reproduce :
1. Add a course to MyLA that you know doesn't have a course start or end date, and one that you know does have course start and end dates.
2. Run the cron.
3. Observe the errors and premature ending in the log message.

jonespm:
Fixed by #871

jonespm:
I verified that the cron ran after this change. I believe it's good!

PR #871, opened 2020-02-04
Replace not None check with pd.notna in update_datetime_field

ssciolla:
This PR fixes an error we were encountering after merging PR #868 that was caused by unexpected conversion of `None` values to `NaT` values when concatenating DataFrames in the `verify_course_ids` method in `cron.py`. The change to the condition in `update_datetime_field`, with the use of `pd.isnull`, should handle cases of `None` and `NaT`. The PR aims to resolve issue #870.

### DIFF
--- a/dashboard/cron.py
+++ b/dashboard/cron.py
@@ -91,7 +91,7 @@ def update_datetime_field(course_obj, course_field_name, warehouse_dataframe, wa
         logger.info(f"Update of {course_field_name} skipped; existing value was found.")
     else:
         warehouse_field_value = warehouse_dataframe[warehouse_field_name].iloc[0]
-        if not pd.isnull(warehouse_field_value):
+        if pd.notna(warehouse_field_value):
             warehouse_field_value = warehouse_field_value.replace(tzinfo=pytz.UTC)
             setattr(course_obj, course_field_name, warehouse_field_value)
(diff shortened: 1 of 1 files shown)


dashboard/cron.py
```diff
@@ -91,7 +91,7 @@ def update_datetime_field(course_obj, course_field_name, warehouse_dataframe, wa
         logger.info(f"Update of {course_field_name} skipped; existing value was found.")
     else:
         warehouse_field_value = warehouse_dataframe[warehouse_field_name].iloc[0]
-        if warehouse_field_value is not None:
+        if not pd.isnull(warehouse_field_value):
```
jonespm: Minor but I believe this is an alias for pd.isna. Probably doesn't matter which one to use.

Also I believe Pandas has a [pd.notna](https://github.com/pandas-dev/pandas/blob/v0.23.4/pandas/core/dtypes/missing.py#L259-L336) (and pd.notnull) rather than using this not. The logic here is slightly more than just calling "not" and probably should be used.

jonespm:
```suggestion
        if pd.notna(warehouse_field_value):
```

ssciolla: Yeah, that looks right, and your suggestion makes sense. I can commit that in a second.

Issue #874, opened 2020-02-05
createHistogram.test.js test fails after recent mylaGlobals changes


ssciolla:
Thank you for contributing to this project!

- Make sure to search the [issues](https://github.com/tl-its-umich-edu/my-learning-analytics/issues) for duplicates first!

## Expected behavior (A clear and concise description of what you expected to happen) : 

All frontend tests that are implemented should run successfully and have access to any variables they need, even if they are `null`.

## Describe the bug (Tell us what happens instead of the expected behavior) :

When running the frontend tests currently using `docker exec -it webpack_watcher npm test`, the `createHitogram.test.js` suite fails because of a new line in `globals.js` (which the `d3` component, `createHistogram` relies on because of its use of `siteTheme.js`). `globals.js` tries to get the `textContent` from a nonexistent element (`null`), which then makes the test fail. Note: `globals.js` was introduced in PR #845.

## Steps to Reproduce :
1. Start MyLA locally.
2. Run `docker exec -it webpack_watcher npm test`

zqian:
verified the tests all pass. 

![image](https://user-images.githubusercontent.com/359944/73877144-254d4700-4826-11ea-8f5a-c55748dd6e0a.png)

Logged in as root and verified nothing break.

PR #853, opened 2020-01-23
Modify palette colors and visualization text colors to improve contrast (#748, #805)

ssciolla:
This PR makes some changes to the color palette used by the front end's `siteTheme`, based on work done by @jennlove-um. The  default `primary` color  was changed to better match UM's integration with Canvas (note that the `primary` after PR #845 is configurable through `env.json`). The `secondary` and `negative` colors were modified to maximize contrast between them (which is still below where we'd like it to be, at 3.03) while also maintaining adequate contrast between the `negative` gray and the white background. I also modified the `d3` components for the visualizations in Resources Accessed and (the old) Assignment Planning so that text on bars is white when the background color is `secondary` (blue) and black when the background color is `negative` (gray). This PR aims to resolve issues #748 and #805.

PR #875, opened 2020-02-05
Add fallback empty object for mylaGlobals and update createHistogram test (#874)

ssciolla:
Responding to changes from PR #845, this PR adjusts `globals.js` to use an empty object if it is unable to find the`myla_globals` JSON element in the DOM. Along with an extra condition checking whether `mylaGlobals.primary_color` is `undefined` allows tests to import and consume `siteTheme` without an error. I also updated the hex values in the snapshot for `createHistogram.test.js` to reflect the new secondary color from PR #853. This PR aims to resolve issue #874.

### DIFF
--- a/assets/src/globals.js
+++ b/assets/src/globals.js
@@ -1,9 +1,9 @@
 import { createMuiTheme } from '@material-ui/core/styles'
 import defaultPalette from './defaultPalette'
 
-// The fallback to an empty object is needed to not break createHitogram.test.js.
-const mylaGlobalsEl = JSON.parse(document.getElementById('myla_globals'))
-const mylaGlobals = mylaGlobalsEl !== null ? mylaGlobalsEl.textContent : {}
+// The fallback to an empty object is needed to not break tests using siteTheme.
+const mylaGlobalsEl = document.getElementById('myla_globals')
+const mylaGlobals = mylaGlobalsEl !== null ? JSON.parse(mylaGlobalsEl.textContent) : {}
 
 /*
 Frozen to prevent unintentional changes to this object. This object is strictly readonly.
--- a/assets/src/globals.js
+++ b/assets/src/globals.js
@@ -1,7 +1,9 @@
 import { createMuiTheme } from '@material-ui/core/styles'
 import defaultPalette from './defaultPalette'
 
-const mylaGlobals = JSON.parse(document.getElementById('myla_globals').textContent)
+// The fallback to an empty object is needed to not break createHitogram.test.js.
+const mylaGlobalsEl = JSON.parse(document.getElementById('myla_globals'))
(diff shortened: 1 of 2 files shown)


assets/src/globals.js
```diff
@@ -17,7 +19,7 @@ const user = Object.freeze({
 })
 
 const palette = JSON.parse(JSON.stringify(defaultPalette))
-if (mylaGlobals.primary_ui_color !== null) {
+if (mylaGlobals.primary_ui_color !== null && mylaGlobals.primary_ui_color !== undefined) {
```
ssciolla: I think we could use `mylaGlobals.primary_ui_color != null` here instead, but this is a bit more explicit.

pushyamig: I think you can use it as it makes the check simpler, you can add a comment saying checking for both null and undefined

pushyamig:
All the Unit test are passing

assets/src/globals.js
```diff
@@ -1,7 +1,9 @@
 import { createMuiTheme } from '@material-ui/core/styles'
 import defaultPalette from './defaultPalette'
 
-const mylaGlobals = JSON.parse(document.getElementById('myla_globals').textContent)
+// The fallback to an empty object is needed to not break tests using siteTheme.
+const mylaGlobalsEl = document.getElementById('myla_globals')
+const mylaGlobals = mylaGlobalsEl !== null ? JSON.parse(mylaGlobalsEl.textContent) : {}
```
justin0022: @ssciolla I think you can replace these two lines with:

```js
const mylaGlobals = document.getElementById('myla_globals') || {}
```
I think this is called JavasScript short circuiting: https://codeburst.io/javascript-what-is-short-circuit-evaluation-ff22b2f5608c

ssciolla: Hm, I hadn't really ever tried this. Thanks for pointing it out. We still need to parse the JSON and use `.textContent`. Because the null check needs to happen before we use `.textContent`, I'm not sure how to really get this on one line. Thoughts?

justin0022: Good point, your implementation is totally fine. I think I would have done something like:

```js
const mylaGlobals = document.getElementById('myla_globals')
   ? JSON.parse(document.getElementById('myla_globals').textContent)
   : {}
```

justin0022: very similar, at the end of the day, I'm happy with either

ssciolla: Thanks! I don't love the repetition of `document.getElementByID`, but I could change the beginning of the second line of my version to just `mylaGlobalsEl`, basically like you suggested. I'll try that.

justin0022: 👍 otherwise LGTM!

assets/src/globals.js
ssciolla: Done! Let me know what you think.

PR #891, opened 2020-02-06
Fixes #890 - Add favicon

### LINKED ISSUE #890: Add favicon to application
We do have an icon for this tool, we should use this as the favicon so it will be used as an icon in the tabs. 

The favicon is a separate issue and should probably be split out. We don't actually have an icon for this tool but that could be created and placed.

To Test, open MyLA, look at the tab. It should be the icon rather than a generic globe.

image

_Originally posted by @jonespm in https://github.com/tl-its-umich-edu/my-learning-analytics/issues/783#issuecomment-563428232_

jonespm:
Also updates the static links to use the static template tag


dashboard/templates/base.html
```diff
@@ -13,9 +13,10 @@
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
     <meta name="viewport" content="width=device-width, initial-scale=1">
     <title>{% use_macro title %}</title>
-    <link rel="stylesheet" type="text/css" href="/static/@fortawesome/fontawesome-free/css/all.css">
-    <link rel="stylesheet" type="text/css" href="/static/@fortawesome/fontawesome-free/css/fontawesome.css">
-    <link rel="stylesheet" type="text/css" href="/static/@fortawesome/fontawesome-free/css/fontawesome.min.css">
+    <link rel="stylesheet" type="text/css" href="{% static '@fortawesome/fontawesome-free/css/all.css' %}">
+    <link rel="stylesheet" type="text/css" href="{% static '@fortawesome/fontawesome-free/css/fontawesome.css' %}">
+    <link rel="stylesheet" type="text/css" href="{% static '@fortawesome/fontawesome-free/css/fontawesome.min.css' %}">
+    <link rel="shortcut icon" href="{% static '/images/favicon.ico' %}"/>
```
ssciolla: I think the `/` in front of `images` is a bug. When I start up the app, it works initially, but when I click around and log in and out as a regular user, I see a `SuspiciousFileOperation` exception.

jonespm: Yeah it's a bug, SuspiciousFileOperation! Wow!

jonespm:
```suggestion
    <link rel="shortcut icon" href="{% static 'images/favicon.ico' %}"/>
```

jonespm: Also <link> doesn't have an end tag in HTML, so fixing that.

Issue #897, opened 2020-02-12
Version ranges in requirements.txt should allow for minor releases from trusted dependencies


ssciolla:
Thank you for contributing to this project!

- Make sure to search the [issues](https://github.com/tl-its-umich-edu/my-learning-analytics/issues) for duplicates first!

## Expected behavior (A clear and concise description of what you expected to happen) : 

For trusted and established dependencies, we should use version ranges that allow for minor releases without code changes. Also, versions should only include a max of two decimals, and specified minor versions should only use two decimal places (e.g. `.99`).

## Describe the bug (Tell us what happens instead of the expected behavior) :

Currently, the Django version range in `requirements.txt` is `Django>=2.2.9,<2.2.9.99`. While this successfully includes Django 2.2.9 in the project, it 1) does not allow for minor releases and 2) should really be 2.2.99, since there are no releases with multiple decimals.

Other dependencies (`SQLAlchemy`, `psycopg2`, `mysqclient`, and `pinax-eventlog`) use unusual ranges.

PR #898, opened 2020-02-12
Correct version ranges in requirements.txt (#897)

ssciolla:
This PR modifies some version ranges in `requirements.txt` to ensure minor version release updates are allowed for trusted dependencies (specifically, with Django) and that version ranges are normalized (using realistic and consistent patterns). The PR aims to resolve issue #897.

The changes here resulted in only these differences from the test instance (which is built off the `2020.01.x` branch).

```
< Django==2.2.9
---
> Django==2.2.10
35c35
< jsonfield==2.0.2
---
> jsonfield==2.1.1
```

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,8 +1,8 @@
 gunicorn>=19.9.0,<19.9.99
 
 # We're waiting on Django 3 until it's further into the lifecycle.
-Django>=2.2.9,<2.2.9.99
-whitenoise>=4.1.4,<4.1.499
+Django>=2.2.9,<2.2.99
+whitenoise>=4.1.4,<4.1.99
 
 django-registration>=3.0,<3.0.99
 django-cron>=0.5.1,<0.5.99
@@ -30,12 +30,12 @@ protobuf>=3.10.0,<3.10.99
 djangosaml2==0.17.2
 pysaml2==4.8.0
 
-SQLAlchemy>=1.3.10,<1.3.199
-psycopg2>=2.8.4,<2.8.499 --no-binary psycopg2
-mysqlclient>=1.4.4,<1.4.499
+SQLAlchemy>=1.3.10,<1.3.99
+psycopg2>=2.8.4,<2.8.99 --no-binary psycopg2
+mysqlclient>=1.4.4,<1.4.99
 google-cloud-bigquery>=1.21.0,<1.21.99
 
-pinax-eventlog>=2.0.3,<2.99
+pinax-eventlog>=2.0.3,<2.0.99
 ptvsd>=4.3.2,<4.3.99
 django-ptvsd-debug==1.0.3
 django-lti-auth==1.0.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,7 @@
 gunicorn>=19.9.0,<19.9.99
 
 # We're waiting on Django 3 until it's further into the lifecycle.
-Django>=2.2.9,<2.2.99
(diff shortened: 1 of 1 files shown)


requirements.txt
```diff
@@ -1,8 +1,8 @@
 gunicorn>=19.9.0,<19.9.99
 
 # We're waiting on Django 3 until it's further into the lifecycle.
-Django>=2.2.9,<2.2.9.99
-whitenoise>=4.1.4,<4.1.499
+Django>=2.2.10,<2.2.99
```
ssciolla: I figured I'd just bump this now. Interesting that in the time I've been working on this, the `jsonfield` sub-dependency had another minor release ('2.1.0 -> 2.1.1`). The `diff` in the description has been updated.

requirements.txt
```diff
@@ -30,12 +30,12 @@ protobuf>=3.10.0,<3.10.99
...
-SQLAlchemy>=1.3.10,<1.3.199
-psycopg2>=2.8.4,<2.8.499 --no-binary psycopg2
-mysqlclient>=1.4.4,<1.4.499
+SQLAlchemy>=1.3.10,<1.3.99
+psycopg2>=2.8.4,<2.8.99 --no-binary psycopg2
+mysqlclient>=1.4.4,<1.4.99
 google-cloud-bigquery>=1.21.0,<1.21.99
 
-pinax-eventlog>=2.0.3,<2.99
+pinax-eventlog>=2.0.3,<2.0.99
```
pushyamig: The newest version of `pinax-eventlog` available is 3.0.0, https://pypi.org/project/pinax-eventlog/#history. Why haven't we upgraded to that?

requirements.txt
```diff
@@ -1,8 +1,8 @@
 gunicorn>=19.9.0,<19.9.99
 
 # We're waiting on Django 3 until it's further into the lifecycle.
-Django>=2.2.9,<2.2.9.99
-whitenoise>=4.1.4,<4.1.499
+Django>=2.2.10,<2.2.99
+whitenoise>=4.1.4,<4.1.99
```
pushyamig: Even the latest version of whitenoise has latest https://pypi.org/project/whitenoise/5.0.1/

requirements.txt
ssciolla: We can. Truthfully, I started working on updating the BE dependencies in October, and the major releases for `pinax-eventlog` and `whitenoise` had not come out yet (release dates in December). I suppose I should have done another check when I was finishing PR #842 in January. There are likely others that have had releases in Q4 of 2019 and Q1 of 2020. I'm not too afraid of updating these, but there is some consideration for waiting for a couple minor releases to come out so initial bugs are worked out. I consider this a separate issue. Would you like me to create an issue?

requirements.txt
ssciolla: See comment above.

requirements.txt
ssciolla: Yes, I just did a scan, and there are about 14 libraries with newer versions (some major) that we could update.

pushyamig: Ya... I know it's bit of more testing involved what change in the major release. Also, we need to have a full understanding of how the library used in the project for testing parts of it. Since not all of the code written by one person and it will be impractical to document the purpose of the library being used

ssciolla: Still, it seems like a good practice to include an issue for each release cycle checking the BE dependencies and assessing whether we should upgrade or not.

Issue #895, opened 2020-02-11
All the API URL's should include a trailing slash for better fetching data


pushyamig:
The [path](https://docs.djangoproject.com/en/2.2/ref/urls/) function URL pattern should include trailing slash to enable fetching results if API is called with or without trailing slash. According to the documentation that's what is recommended. This was [fixed](https://github.com/tl-its-umich-edu/my-learning-analytics/pull/894) for `course_enabled` URL but need to apply the same pattern to other API URL's as well

zqian:
@pushyamig @ssciolla  new URL endpoints got introduced lately without trailing slashes, e.g. "graphql", etc. Should we address those urls, too?

ssciolla:
@zqian, please read through the comments on PR #899. Or the comments in `urls.py`. There were a couple cases, including `graphql`.

Also, I don't think any new URLs have been added since this was merged.

zqian:
@ssciolla sorry for missing the comment line!

Issue #893, opened 2020-02-11
course_enabled url is not fetching any results


pushyamig:
Thank you for contributing to this project!

- Make sure to search the [issues](https://github.com/tl-its-umich-edu/my-learning-analytics/issues) for duplicates first!

## Expected behavior (A clear and concise description of what you expected to happen) : 
The course_enabled url `/api/v1/course_enabled` should fetch result regardless with or without trailing slash

## Steps to Reproduce :
1. Go to `$HOST/api/v1/course_enabled` this is not returning any results
2.
3.

pushyamig:
Test_plan:
1. Go to canvas test instance for the courses that are enabled Myla
2. Change to student view
3. Links should show up for /assignments, files, grades

PR #899, opened 2020-02-12
Add trailing slashes to URL routes (#895)

ssciolla:
This PR seeks to establish a consistent approach for defining URL routes by adding trailing forward slashes (where possible) to existing routes. We discovered that missing slashes can cause problems during testing of the Canvas integration (see #893). Trailing forward slashes are included in examples in the [Django documentation](https://docs.djangoproject.com/en/2.2/ref/urls/). With these added, the [`APPEND_SLASH` setting](https://docs.djangoproject.com/en/3.0/ref/settings/#append-slash) (part of the `CommonMiddleware`, it defaults to `True`) should redirect users if they try to visit one of these routes without the trailing slash. However, redirects for `PUT` or `POST` requests will not retain data, so I needed to change the `useSetUserSetting` hook to use the trailing slash (Alternatively, we could not use a trailing slash here, but it seems nice to be consistent). The PR aims to resolve issue #895.


pushyamig:
Looks good, just need to resolve the conflicts before merging

dashboard/urls.py
```diff
@@ -33,8 +33,9 @@
...
     path('', views.get_home_template, name = 'home'),
     path('status/', include('watchman.urls')),
-    path('status/bare_status', watchman.views.bare_status),
-    path('admin', admin.site.urls),
+    path('status/bare_status/', watchman.views.bare_status),
+
+    path('admin/', admin.site.urls),
 
     # graphql access
     path('graphql', DashboardGraphQLView.as_view(graphiql=True)),
```
ssciolla: I tried adding a `/` here but that seemed to break the GraphQL implementation. Part of the problem seems to be that `PUT` calls will lose data when the slash is appended and a redirect happens. I was hoping there would be a way to add `/` to calls in the front-end -- like we did with `userSetUserSetting` -- but I wasn't able to find anything easily. @justin0022, any suggestions? I'm fine keeping it without the `/` though, it'd be nice to be consistent.

ssciolla:
I still need to deploy to the test environment to ensure the SAML routes work. I'll try to do that today or later this week.

dashboard/urls.py
justin0022: Hey sorry @ssciolla I'll look into this. Didn't see this message till now.

ssciolla:
Okay, I hit a hiccup when testing the SAML integration. Logging in worked, but I encountered a strange error page when trying to access `/accounts/logout/` (something like `'Connection aborted.', OSError("(104, 'ECONNRESET')"`. For now, I have rolled back the change to the override in the SAML block (`accounts/logout`), which seems to have fixed this issue.

When talking with @jonespm about this, we also noticed that the FE's `logout` button was not using the URL from `settings.py`, despite it being passed as part of `myla_globals`. I made a few changes to the FE (mainly moving the `loginURL` and `logoutURL` into `user`) to fix this.

dashboard/urls.py
justin0022: I'm not sure what the best way to address this is

ssciolla:
@pushyamig, @zqian, and @jonespm, do you think it's okay to merge this? Note that I've elected to leave the SAML and GraphQL routes alone for the moment.

dashboard/urls.py
ssciolla: K, let's leave it alone for now.

ssciolla:
> Yeah, could you just add a comment to the code by these two endpoints you didn't change so if someone comes in and wonders why we didn't do slashes here why we left them as is? Then I think it's good.

Okay @jonespm, let me know if the latest commit works for you.

jonespm:
Looks great thanks @ssciolla

ssciolla:
FYI, I just added another comment to `settings.py` explaining why `LOGOUT_URL` doesn't have the trailing slash. Merging now.

Issue #907, opened 2020-02-28
Keyboard Focus Behavior on IndexPage Buttons is Clunky and Redundant


vjcao:
## Expected behavior (A clear and concise description of what you expected to happen) : 
Tabbing on the index page of a class should smoothly traverse the buttons on the page.
## Describe the bug (Tell us what happens instead of the expected behavior) :
Currently, each button has two keyboard focusable elements forcing the user to tab twice to move to the next button
## Steps to Reproduce :
1. Click below the AppBar element.
2. Tab through the buttons.

## Screenshots/Video (If possible, add media to help explain your problem) :
![Feb-28-2020 16-21-51](https://user-images.githubusercontent.com/33735083/75588636-12a1e880-5a47-11ea-99c8-8a56be5bd4e2.gif)

yuandrew:
Hi, I would like to take a cack at this issue if no one else is currently working on it!

vjcao:
@yuandrew I already have a pull request open for this issue. Thank you for your interest in this project!

jennlove-um:
testing passes with SpringSummer-2020.02.01 release.

gsilver:
@jennlove-um @pushyamig - initial issue resolved, it is clear now what the links are to (the viz, some explanation of the viz). I would open a new one that adds aria-label="same value as the title of the 'i' link"

PR #908, opened 2020-02-28
Smoother Keyboard Traversing of Buttons on Index Page

vjcao:
Tabbing through the Index Page will not make users press tab twice to traverse to next button. This PR addresses issue #907.

Feb-28-2020 16-46-57


vjcao:
@ssciolla This behavior is not present in Chrome-based browsers; however, I agree. The previous implementation is a little more annoying than the highlights.

Issue #694, opened 2019-09-11
Resource Accessed responsive bug


justin0022:
## Expected behavior (A clear and concise description of what you expected to happen) : 
The slider on the right overflows and exits the display when the browser window is resized. 

## Screenshots/Video (If possible, add media to help explain your problem) :
![Screen Shot 2019-09-11 at 4 27 20 PM](https://user-images.githubusercontent.com/8836578/64742520-5f9dfc80-d4b1-11e9-8192-e91a97f3e518.png)

naeunk:
Hello, I am currently a student enrolled in EECS 481. My partner and I were wondering if it is possible to contribute to fixing this issue. Thanks.

vjcao:
Thank you @naeunk , let us know if you run into any difficulty.

ssciolla:
I have been working on this issue after studying and getting at-least-a-little clued in to `d3`. Right now I have two solutions which have somewhat different behavior. 

### Solution 1: Calculate chart and margin sizes using percents, don't adjust font sizes

Code: [issue-694-fix-responsiveness-bug-with-percents](https://github.com/ssciolla/my-learning-analytics/tree/issue-694-fix-responsiveness-bug-with-percents)

Pro(s):
* Text sizes stay the same, making content more readable on smaller windows.

Con(s):
* Content gets jumbled and some cut off on the smallest window sizes.

### Solution 2: Use an SVG `viewbox` attribute on the component's wrapper

* [issue-694-fix-responsiveness-bug-with-svg-viewbox](https://github.com/ssciolla/my-learning-analytics/tree/issue-694-fix-responsiveness-bug-with-svg-viewbox)

Pro(s):
* Everything scales well and looks neat.

Con(s):
* Text become very difficult to read on smaller window sizes.

ssciolla:
Seems like the researchers prefer having the font sizes of text stay fixed or at least at a more legible size, i.e. Solution 1. I'll proceed under that assumption.

jonespm:
@ssciolla Is this fixed with #910?

ssciolla:
Yes, I'll close, but should still be in QA column for project.

jennlove-um:
Testing passes.  Tested in Firefox and Chrome.  When the window is resized, the slider displays correctly and the font sizes don't change.

Issue #858, opened 2020-01-30
[Resources Accessed] Remove the brush element and instead use a simple slider


jennlove-um:
Replace the brush navigation element with a simple slider to allow students to navigate up and down the list of resources.  It would be better for accessibility.

jennlove-um:
![GD_Text_change](https://user-images.githubusercontent.com/39634416/73487936-ef1b4d80-4375-11ea-90df-3d47aa1991b8.png)

jennlove-um:
@ssciolla Here's the updated version related to the discussion today.

PR #910, opened 2020-03-04
Make Resources Accessed resize gracefully, re-work spacing (#694)

ssciolla:
This PR makes significant changes to the Resources Accessed view to make spacing more consistent and the charts and text respond appropriately to changing window sizes. It accomplishes this by introducing percent-based calculations of component sizes and coordinates, removing the use of a `Grid` component wrapper, and truncating text based on pixel length. The task list and commentary below provide more details on the changes. The PR aims to resolve issue #694.

- [x] Remove `Grid` wrapper for chart from the `ResourcesAccessed` container.
- [x] Determine width, height, and margin sizes for the `main` and `mini` charts based on decimals; update coordinates, translations, and references accordingly.
- [x] Modify `useResponsiveness` hook to allow for an optional `minHeight` to help prevent text from getting jumbled in smaller windows.
- [x] Rewrite `truncate` function so it uses `getCompetedTextLength`; thanks to @pushyamig for putting me on the right track.

Note that this PR does not fix all current issues with the view, but hopefully it makes strides toward improving the user experience in general and making it easier for developers to change things quickly with less potential for bugs.

Regarding the new `truncate` method, I have seen some discussion on forums and in blogs suggesting that `getComputedTextLength` can be time-intensive. Please let me know if you think the view has become more laggy to the point where it is disruptive; it seems OK to me at the moment, but more testing may be needed. I chose this route because it seemed the simplest to integrate with the current implementation.

### DIFF
--- a/assets/src/components/d3/createResourceAccessChart.js
+++ b/assets/src/components/d3/createResourceAccessChart.js
@@ -71,7 +71,7 @@ function appendLegend (svg) {
 function createResourceAccessChart ({ data, width, height, domElement }) {
   const resourceData = data.sort((a, b) => b.total_percent - a.total_percent)
 
-  const sideMarginSize = width * 0.05
+  const sideMarginSize = width * 0.075
   const margin = { top: 0, right: sideMarginSize, bottom: 0, left: sideMarginSize }
 
   const [availWidth, availHeight] = adjustViewport(width, height, margin)
@@ -115,8 +115,9 @@ function createResourceAccessChart ({ data, width, height, domElement }) {
   // Build the chart
   const svg = d3.select(domElement).append('svg')
     .attr('class', 'svgWrapper')
-    .attr('width', mainMargin.left + mainWidth + mainMargin.right + miniMargin.left + miniWidth + miniMargin.right)
-    .attr('height', mainMargin.top + mainHeight + mainMargin.bottom)
+    .attr('width', availWidth)
+    .attr('height', availHeight)
+    .attr('transform', `translate(${margin.left}, ${margin.top})`)
     .on('wheel.zoom', scroll)
     .on('mousedown.zoom', null) // Override the center selection
     .on('touchstart.zoom', null)
@@ -153,11 +154,10 @@ function createResourceAccessChart ({ data, width, height, domElement }) {
       .enter()
       .append('text')
       .attr('class', 'label')
-      .attr('x', d => mainXScale(d.total_percent) + 3 + margin.left + mainMargin.left)
-      .attr('y', d => mainYScale(d.resource_name) + mainYScale.bandwidth() / 2 + margin.top + mainMargin.top)
+      .attr('x', d => mainXScale(d.total_percent) + 3 + mainMargin.left)
+      .attr('y', d => mainYScale(d.resource_name) + mainYScale.bandwidth() / 2 + mainMargin.top)
       .attr('dx', -10)
       .attr('dy', '.35em')
-      .style('font-size', 10)
       .style('fill', d => d.self_access_count > 0 ? 'white' : 'black')
       .attr('text-anchor', 'end')
       .text(d => (
@@ -259,7 +259,7 @@ function createResourceAccessChart ({ data, width, height, domElement }) {
   // Main chart group
   const mainGroup = svg.append('g')
     .attr('class', 'mainGroupWrapper')
-    .attr('transform', `translate(${margin.left + mainMargin.left}, ${margin.top + mainMargin.top})`)
+    .attr('transform', `translate(${mainMargin.left}, ${mainMargin.top})`)
     .append('g')
     .attr('clip-path', 'url(#clip)')
     .style('clip-path', 'url(#clip)')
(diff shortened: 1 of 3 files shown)


assets/src/components/d3/createResourceAccessChart.js
```diff
@@ -16,7 +16,7 @@ import { siteTheme } from '../../globals'
 const accessedResourceColor = siteTheme.palette.secondary.main
 const notAccessedResourceColor = siteTheme.palette.negative.main
 const linkColor = siteTheme.palette.link.main
-const mainMargin = { top: 50, right: 10, bottom: 50, left: 200 }
+const foriegnObjSide = 24
```
justin0022:
```suggestion
const foreignObjSide = 24
```

ssciolla: Thanks! Gosh, must have been looking at this code for too long. Should be fixed now.

justin0022:
This is probably unrelated to the PR, but I wonder if we should set a minimum height for the brush: 
Screen Shot 2020-03-05 at 10 03 16 AM

ssciolla:
> This is probably unrelated to the PR, but I wonder if we should set a minimum height for the brush:

Yep, this is a problem, which I've been considering out of scope for this PR. I'm gonna try to tackle issue #858 next, which this should fall under. Thanks for the review!

jonespm:


```diff

Issues
======
+ Solved 1
           
```


See the complete overview on [Codacy](https://app.codacy.com/manual/ITS_Teaching_And_Learning/my-learning-analytics/pullRequest?prid=5147347&bid=16907439)

assets/src/components/d3/createResourceAccessChart.js
```diff
@@ -16,7 +16,7 @@ import { siteTheme } from '../../globals'
 const accessedResourceColor = siteTheme.palette.secondary.main
 const notAccessedResourceColor = siteTheme.palette.negative.main
 const linkColor = siteTheme.palette.link.main
-const mainMargin = { top: 50, right: 10, bottom: 50, left: 200 }
+const foreignObjSide = 24
```
jonespm: How do you come up with this number? Is it based on anything else? What would happen if someone changed it? Could you add a comment here as it's not obvious to me what this does from this value, the name of the variable or the code.

assets/src/components/d3/createResourceAccessChart.js
```diff
@@ -66,36 +65,78 @@ function appendLegend (svg) {
...
+  */
+  const mainWidth = availWidth * 0.55
+  const mainHeight = availHeight * 0.7
+  const mainMargin = {
+    top: availHeight * 0.15,
+    right: 0,
+    bottom: availHeight * 0.15,
+    left: availWidth * 0.225
+  }
+  const resourceLabelWidth = mainMargin.left * 0.889 - foreignObjSide
```
jonespm: I understand the comment for the values above but not sure about this calclulation. 
`mainMargin.left * 0.889 - 24 ` (currently defined somewhere else). What's the 0.889?

assets/src/components/d3/createResourceAccessChart.js
```diff
@@ -66,36 +65,78 @@ function appendLegend (svg) {
...
-  const [mainWidth, miniHeight] = adjustViewport(width, height, mainMargin)
+  const sideMarginSize = width * 0.075
+  const margin = { top: 0, right: sideMarginSize, bottom: 0, left: sideMarginSize }
+
+  const [availWidth, availHeight] = adjustViewport(width, height, margin)
+
+  /*
+  The decimal multipliers for mainMargin.left, mainWidth, mainMargin.right, miniMargin.left,
+  miniWidth, and miniMargin.right should add up to 1.0. In other words, the sum of those values
+  should be equal to the availWidth.
```
jonespm: I'm guessing this comment is also applicable to availHeight?

assets/src/components/d3/createResourceAccessChart.js
ssciolla: The number is arbitrary. It specifies the side length of the square foreign object element, which contains the resource icon and the spacing to its right. Originally you'll see on the old lines 376 and 377 that it was 32. I think I made it a little smaller since the spacing between the icon and resource name was fairly large. This was created and moved up since the value is also needed for determining the placement of the resource label. I will add a comment.

assets/src/components/d3/createResourceAccessChart.js
ssciolla: Yeah, I will update that. Thanks!

assets/src/components/d3/createResourceAccessChart.js
ssciolla: Yeah, this isn't ideal. Essentially, I want to find 8/9's of `mainMargin.left`, since that would be `0.20`, leaving the same margin size as `miniMargin.left`. It's a bit awkward because the mainMargin.left is really acting like it's own section. I can fix this so that I define the variable after I define the `mini` stuff, and then use `const resourceLabelWidth = mainMargin.left - miniMargin.left - foreignObjSide.` Hopefully, that will be clearer.

ssciolla:
Thanks all for reviews! @jonespm, if you could glance at the updates I made in response to your comments and give me the OK, I'd appreciate it!

PR #911, opened 2020-03-06
Updated README and Wiki with Keyboard Navigation Instructions

vjcao:


jonespm:
I suppose it's fine to put this here, but I think this should actually go into the user facing help somewhere if this is something people will need to do. @jennlove-um ? End users are unlikely to see this here.

jennlove-um:
Agreed - I can get this added to the user-facing help.

Issue #896, opened 2020-02-11
Hamburger Menu Button Focus Ambiguity


vjcao:
## Describe the bug (Tell us what happens instead of the expected behavior) :
After PR #863, when using the keyboard to navigate the hamburger menu buttons, the currently selected view button looks the same as which button currently has keyboard focus. This introduces some ambiguity to which button has focus in the hamburger menu.

## Screenshots/Video (If possible, add media to help explain your problem) :
![Screen Shot 2020-02-11 at 4 51 21 PM](https://user-images.githubusercontent.com/33735083/74282818-cc325700-4cee-11ea-9cd3-6688d4592920.png)
Which button currently has keyboard focus? Your guess is as good as mine.

jennlove-um:
Testing passes with SpringSummer  2020.02.01 release

gsilver:
@pushyamig @jennlove-um - mark as done. Grey background clearly shows what menu item has focus.

PR #912, opened 2020-03-06
Changed Selected Button Behavior in Sidebar

vjcao:
This PR aims to remove the ambiguity between selected and current view button in the sidebar. Fixes issue #896.

### Current:
Screen Shot 2020-03-06 at 2 38 02 PM

### This PR:
Screen Shot 2020-03-06 at 2 28 00 PM

### DIFF
--- a/assets/src/containers/SideDrawer.js
+++ b/assets/src/containers/SideDrawer.js
@@ -54,7 +54,7 @@ function SideDrawer (props) {
             <ListItemIcon>
               {selectedIndex === key
                 ? <props.icon color='secondary' />
-                : <props.icon style={{color: '#E1E1E1'}}/>
+                : <props.icon color='negative'/>
               }
             </ListItemIcon>
             <ListItemText primary={props.title} className={classes.text} />
--- a/assets/src/containers/SideDrawer.js
+++ b/assets/src/containers/SideDrawer.js
@@ -49,10 +49,14 @@ function SideDrawer (props) {
             to={props.path}
             key={key}
             className={classes.sideDrawerLinks}
-            selected={selectedIndex === key}
             onClick={() => setSelectedIndex(key)}
           >
-            <ListItemIcon><props.icon /></ListItemIcon>
+            <ListItemIcon>
+              {selectedIndex === key
+                ? <props.icon color='secondary' />
+                : <props.icon style={{color: '#E1E1E1'}}/>
+              }
+            </ListItemIcon>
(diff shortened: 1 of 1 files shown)


assets/src/containers/SideDrawer.js
```diff
@@ -49,10 +49,14 @@ function SideDrawer (props) {
...
             key={key}
             className={classes.sideDrawerLinks}
-            selected={selectedIndex === key}
             onClick={() => setSelectedIndex(key)}
           >
-            <ListItemIcon><props.icon /></ListItemIcon>
+            <ListItemIcon>
+              {selectedIndex === key
+                ? <props.icon color='secondary' />
+                : <props.icon style={{color: '#E1E1E1'}}/>
```
ssciolla: Maybe we can use the `negative` color from the palette here instead? That would keep it consistent with the rest of the application.

I would also probably format it like this, but that's a minor issue:
```
              {
                selectedIndex === key
                  ? <props.icon color='secondary' />
                  : <props.icon color='negative' />
              }
```

vjcao:
@ssciolla This behavior is consistent with previous versions and is not unique to the changes I made. However, I agree that the SideBar should behave the way you described.

assets/src/containers/SideDrawer.js
vjcao: I tried using the negative color but thought that there wasn't enough contrast between secondary and negative colors. I attached a screenshot of the SideBar using the secondary and negative as the colors. Let me know if you think that there is enough contrast and that I should change it back. Thanks.

Screen Shot 2020-03-13 at 10 21 03 AM

ssciolla: @jennlove-um, do you think the contrast is worse here between `secondary` (the blue) and `negative` (the gray). I kind of see what @vjcao is saying, but I'd rather stick to the same colors throughout if we can.

jennlove-um:
I agree with Sam about using the consistent colors throughout.  Vincent is right that there is less contrast between them than with the colors he picked.  But we did test the negative color and made sure it as one that meets accessibility contrast requirements.

pushyamig:
@ssciolla  this PR is ready to merge. Do you still have any concerns?

ssciolla:
> @ssciolla this PR is ready to merge. Do you still have any concerns?

Yes, @pushyamig, I think so. I'll run it once more to be sure, and then approve and merge.

PR #915, opened 2020-03-11
Fixes #873 - Display course in cron where earliest found

### LINKED ISSUE #873: Find Earliest Start Date Time of course should display course where earliest date found
In the utils.py the line to log the earliest start date should include the course id that this date was found in to make it easier to track these early courses down.
https://github.com/tl-its-umich-edu/my-learning-analytics/blob/8723b0b79458a42c141f84ec98236820f295d8a8/dashboard/common/utils.py#L20

## Steps to Reproduce :
1. Run the cron
2. Log for the line 

> Earliest start datetime for all courses:

jonespm:
* This changes the two methods in the model for Course to a property so they can be more directly accessed.

* It also sorts the objects by this value so the object id can be re-used later.

* It also has a rename of dashboard in the cron as we don't need to see that in the logs anymore

### DIFF
--- a/dashboard/common/utils.py
+++ b/dashboard/common/utils.py
@@ -9,15 +9,13 @@ logger = logging.getLogger(__name__)
 
 
 def find_earliest_start_datetime_of_courses():
-    course_start_datetimes = []
-    for course in Course.objects.all():
-        course_start_datetimes.append(course.get_course_date_range().start)
-    logger.debug(course_start_datetimes)
+    sorted_courses = sorted(Course.objects.all(), key=lambda course: course.course_date_range.start)
 
-    earliest_start = None
-    if len(course_start_datetimes) > 0:
-        earliest_start = sorted(course_start_datetimes)[0]
-        logger.info(f"Earliest start datetime for all courses: {earliest_start.isoformat()}")
+    earliest_course = None
+    if len(sorted_courses) > 0:
+        earliest_course = sorted_courses[0]
+        earliest_start = earliest_course.course_date_range.start
+        logger.info(f"Earliest start datetime for all courses: {earliest_start.isoformat()} found in course {earliest_course.canvas_id}")
     else:
         logger.info(f"No course listed. Return None as the earliest_start_datetime_of_course. ")
     return earliest_start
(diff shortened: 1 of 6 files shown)


dashboard/common/utils.py
```diff
@@ -9,15 +9,13 @@
...
-        course_start_datetimes.append(course.get_course_date_range().start)
-    logger.debug(course_start_datetimes)
-
-    earliest_start = None
-    if len(course_start_datetimes) > 0:
-        earliest_start = sorted(course_start_datetimes)[0]
-        logger.info(f"Earliest start datetime for all courses: {earliest_start.isoformat()}")
+    sorted_courses = sorted(Course.objects.all(), key=lambda course: course.course_date_range.start)
+
+    earliest_course = None
```
ssciolla: I think this needs to be `earliest_start` like it was before.`

ssciolla: This WOULD cause an error I think if line 138  in `cron.py` wasn't already checking whether the number of courses was zero. It actually makes this `if/else` arrangement look unnecessary, but I suppose more precautionary null handling isn't a huge deal.

jonespm: Yeah, you're right about changing the wrong variable. The if/else is mostly in support of the logging rather than the null handling I think.

.vscode/settings.json
```diff
@@ -5,5 +5,8 @@
     "python.linting.pylintArgs": ["--load-plugins=pylint_django"],
     "[javascript]": {
         "editor.tabSize": 2
-    }
+    },
+    "python.linting.banditEnabled": true,
+    "python.linting.pycodestyleEnabled": true,
+    "python.linting.pylintEnabled": true
```
ssciolla: Sneaking this in, eh?

dashboard/models.py
```diff
@@ -366,3 +366,26 @@ def __str__(self):
...
+class Example3(   object ):
+    def __init__    ( self, bar ):
+     #Comments should have a space after the hash.
+     if bar : bar+=1;  bar=bar* bar   ; return bar
+     else:
+                    some_string = """
+                       Indentation in multiline strings should not be touched.
+Only actual code should be reindented.
+"""
+                    return (sys.path, some_string)
```
ssciolla: What's going on here?

.vscode/settings.json
ssciolla: That's okay with me.

jonespm:
Yeah oops!!! That was a work in progress, I didn't mean to commit any of that.

dashboard/models.py
jonespm: There's nothing going on here! :)

PR #919, opened 2020-03-12
#918 same resources(id) used in different course fail the cron

### LINKED ISSUE #918: Cron fails when same resource is used in separate courses

Cron is failing when accessing resources data( like lecture video )with the same id used on multiple courses. This is probably occurring due to the Primary key on the resource id which is the resource id of the file. 

The better solution to fix this will be use auto-increment id and a separate column called `resource_id` 

 By Default, cron is only checking if course exists in the Canvas Dataware house(CDH) and if some of the courses added to Myla don't have a presence in CDH then wiping of the data don't happen for those courses. 

## Steps to Reproduce :
1. 
2.
3.

pushyamig:
Fixes #918


dashboard/models.py
```diff
@@ -273,10 +273,10 @@ def get_queryset(self):
     def get_course_resource_type(self, course_id):
         return self.get_queryset().get_course_resource_type(course_id)
 
-
 class Resource(models.Model):
+    id = models.AutoField(primary_key=True, verbose_name="Table Id")
     resource_type = models.CharField(max_length=255, verbose_name="Resource Type")
-    id = models.CharField(primary_key=True, max_length=255, verbose_name="Resource Id")
+    resource_id = models.CharField(blank=True, max_length=255, null=False, verbose_name="Resource Id")
```
ssciolla: I think @jonespm  was talking about using `db_index=True` here, too, to help with performance. Probably makes sense to do this on `course_id`, too.

ssciolla: https://docs.djangoproject.com/en/3.0/ref/models/fields/#db-index

ssciolla: Do we need to allow `blank=True` for some reason?

jonespm:
Yeah, I don't understand the migrations here either. One adds a new table but this one isn't even used because it's removed by the next migration. I'm going to need to pull down this PR to try it out.

jonespm:
I re-wrote the the migration file manually and it's working for me with data prior to this change as well as an empty database. Will test out the problematic courses.

dashboard/migrations/0015_auto_20200312_1323.py
```diff
@@ -0,0 +1,28 @@
+# Generated by Django 2.2.10 on 2020-03-12 17:23
```
ssciolla: I think it would make sense to add a comment saying you modified the auto-generated version. We should also probably change the migration file name so it doesn't say 'auto'.

PR #920, opened 2020-03-12
Initial Cypress.io Testing

mcrwong:
This partially addresses issue 879 by using cypress as a tool for automating some tests on the frontend. Mainly This initial set of tests checks a user's ability to login and a user's ability to access and use each of the current analytics tools. This uses the cypress-plugin-snapshots for checking snapshots of the DOM similar to jest.


justin0022:
Excellent! I'll take a look.

justin0022:
Actually before I do, could you merge the latest master into your branch first?

ssciolla:
Glad we're working on this. It would be nice if others' commits weren't included here.

justin0022:
To run Cypress:
`npm install`
`npm run cypress:open`

All my tests fail because I don't have `donald07` added as authorized user. 

After adding this user, 2/5 tests pass.

Screen Shot 2020-03-12 at 9 24 48 AM

mcrwong:
> To run Cypress:
> `npm install`
> `npm run cypress:open`
> 
> All my tests fail because I don't have `donald07` added as authorized user.
> 
> After adding this user, 2/5 tests pass.
> 
> Screen Shot 2020-03-12 at 9 24 48 AM

Let me see what I can do. Hopefully it isn't a problem that requires waiting for cypress to be updated.

mcrwong:
I'm going to close this pull request for now since I think there are a couple major changes I need to make before this will be ready.

PR #924, opened 2020-04-21
reverse scrolling direction for resource view

### LINKED ISSUE #913: reverse resource view scrolling direction
### Expected behavior

Resource view scrolling should follow same direction as page scrolling.

### Describe the bug

The action to scroll the resource view is currently the opposite of that to scroll the page.

### Steps to Reproduce

1. Try scrolling a long web page.
1. Try scrolling the resource view.
1. Notice the different action required.

### Environment:

* **OS**: macOS 10.14.5
* **Browser**: Google Chrome 80.0.3987.163
* **Scroll setting**: macOS "Trackpad" preference panel's "Scroll Direction: Natural" setting is enabled.

yuandrew:
This PR fixes #913 by changing multiplying the `dy` value by -1 in the `scroll` function when creating the chart in `createResourceAccessChart.js`.

### DIFF
--- a/assets/src/components/d3/createResourceAccessChart.js
+++ b/assets/src/components/d3/createResourceAccessChart.js
@@ -212,7 +212,7 @@ function createResourceAccessChart ({ data, width, height, domElement }) {
     const range = miniYScale.range()
     const y0 = d3.min(range)
     const y1 = d3.max(range) + miniYScale.bandwidth()
-    const dy = d3.event.deltaY
+    const dy = d3.event.deltaY * -1
     const topSection = selection[0] - dy < y0
       ? y0
       : selection[1] - dy > y1
--- a/assets/src/components/d3/createResourceAccessChart.js
+++ b/assets/src/components/d3/createResourceAccessChart.js
@@ -205,14 +205,18 @@ function createResourceAccessChart ({ data, width, height, domElement }) {
     bar.exit().remove()
   }
 
-  function scroll () {
+  function scroll (e) {
     // Mouse scroll on the chart
     const selection = d3.brushSelection(gBrush.node())
     const size = selection[1] - selection[0]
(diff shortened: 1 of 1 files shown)


lsloan:
Clever.  I'll try to make some time soon to test this change.

jonespm:
When this is tested we should test it on Windows, and Mac both with scroll direction natural checked and unchecked. Also in Safari and at least one other browser.

I'd just wonder if we need to detect anything else in the Javascript.
https://stackoverflow.com/a/23668035/3708872

image

pushyamig:
We have Browser Guideline list in wiki page https://github.com/tl-its-umich-edu/my-learning-analytics/wiki/MyLA-Browser-Support

lsloan:
Good point.  Wasn't asked about in the issue comments.  I'll add some info to the issue description.

jonespm:
Yeah, I didn't think about it till I saw what the fix was! :) I know on Mac that's one of the first options I have to uncheck. So I wonder if it is different only for users that have that option checked. Didn't really consider it!

lsloan:
I kinda like the fix you referred to on SO, @jonespm.  I was going to suggest to @yuandrew that a simple negative sign rather than the multiplication would suffice.  I didn't think of what it would do for people who use "unnatural" scrolling.  (Supernatural scrolling?  Freak of nature scrolling?)

If `event.webkitDirectionInvertedFromDevice` is usually available, I would use that to determine whether to negate the scrolling rather than the way the SO fix negates it, then conditionally negates it again (unnegating it).

yuandrew:
I added a condition with `event.webkitDirectionInvertedFromDevice` to check scrolling direction and match the direction! I tested this on MacOS with Chrome, Safari, and Firefox, and they all seemed to work.

Issue #936, opened 2020-05-26
Lost "View on Site" button with change to course.absolute_url


ssciolla:
Thank you for contributing to this project!

- Make sure to search the [issues](https://github.com/tl-its-umich-edu/my-learning-analytics/issues) for duplicates first!

## Expected behavior (A clear and concise description of what you expected to happen) : 

The admin page used to have a "View on Site" button, which is automatically generated when a model has a `get_absolute_url` method.

## Describe the bug (Tell us what happens instead of the expected behavior) :

It looks like in PR #915 when we changed to using a property (i.e. `course.absolute_url`), we lost this button, meaning a link is no longer available from the edit page, only from the course list.

## Screenshots/Video (If possible, add media to help explain your problem) :

![Screen Shot 2020-05-26 at 4 34 30 PM](https://user-images.githubusercontent.com/35741256/82949653-f6441800-9f71-11ea-8683-4180a586dc06.png)

The course referenced above is fabricated.

## Additional context (Add any other context about the problem here) :

I think we can keep the property implementation and just add back a `get_absolute_url` method that uses the property. Something like this:

```
def get_absolute_url(self):
   return self.absolute_url
```

jennlove-um:
Testing passes: The button displays and works as expected.

PR #937, opened 2020-05-26
Bring back "View on Site" on "Change Course" page (#936)

ssciolla:
This PR makes a minor modification to `dashboard/models.py` to re-enable a "View on Site" button available from the "Change Course" page in the Django admin. This seems to have been lost in an earlier PR when `get_absolute_url` (which has special meaning to Django) was changed to a `@property` implementation with `absolute_url`. The fix here merely re-creates the `get_absolute_url` method and leverages the aforementioned property. The PR aims to resolve issue #936 

"Change Course" admin page with change:

Screen Shot 2020-05-26 at 4 48 17 PM

The course referenced above is fabricated.

### DIFF
--- a/dashboard/models.py
+++ b/dashboard/models.py
@@ -200,6 +200,9 @@ class Course(models.Model):
     def absolute_url(self):
         return reverse('courses', kwargs={'course_id': self.canvas_id})
 
+    def get_absolute_url(self):
+        return self.absolute_url
+
     class Meta:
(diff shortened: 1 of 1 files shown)


ssciolla:
> I think this could have also been reversed to be
> 
> But it probably doesn't really matter.

Yeah, I was trying to be minimally disruptive. Like you said, these should have the same result.

PR #938, opened 2020-05-27
Fixed rounding bug

### LINKED ISSUE #881: [Redesign assignment] Weird rounding behaviour
https://www.youtube.com/watch?v=S2fRNzXuGPs&feature=youtu.be

When lock grade is checked/unchecked, the goal grade changes sometimes.

justin0022:
Resolves #881.

I was able to reproduce the bug - it's caused by rounding as you suspected @ssciolla. 

I fixed by rounding to one decimal when the numbers are displayed in `AssignmentTable`. I also had to round in `assignment` because every time the state changes, GraphQL needs to store the floats and I was getting an error from GraphQL that the floats were too large - I decided to round to 3 decimal places for the internal state, which resolves the rounding error.

### DIFF
--- a/assets/src/util/assignment.js
+++ b/assets/src/util/assignment.js
@@ -1,5 +1,5 @@
 import { calculateWeekOffset, dateToMonthDay } from './date'
-import { sum, roundToXDecimals, getDecimalPlaceOfFloat } from './math'
+import { sum, roundToXDecimals } from './math'
 
 const clearGoals = assignments => assignments
   .map(a => {
@@ -21,15 +21,11 @@ const setAssigmentGoalInputState = (key, assignments, inputFocus) => {
 }
 
 const setAssignmentGoalGrade = (key, assignments, goalGrade) => {
-  // Use decimal place of pointsPossible if it's a decimal; otherwise, round to nearest tenth
-  const placeToRoundTo = (String(assignments[key].pointsPossible).includes('.'))
-    ? getDecimalPlaceOfFloat(assignments[key].pointsPossible) : 1
-
   return [
     ...assignments.slice(0, key),
     {
       ...assignments[key],
-      goalGrade: goalGrade === '' ? '' : roundToXDecimals(Number(goalGrade), placeToRoundTo),
+      goalGrade: goalGrade === '' ? '' : roundToXDecimals(Number(goalGrade), 3),
       goalGradeSetByUser: goalGrade
     },
     ...assignments.slice(key + 1)
@@ -95,7 +91,7 @@ const calculateAssignmentGoalsFromCourseGoal = (
 
   return assignments.map(a => {
     if (notGradedOrGoalGradeSetByUser(a) && a.inputBlur) {
-      a.goalGrade = roundToXDecimals(requiredGrade * a.pointsPossible, 1) || ''
+      a.goalGrade = roundToXDecimals(requiredGrade * a.pointsPossible, 3) || ''
     }
     return a
   })
--- a/assets/src/components/AssignmentTable.js
+++ b/assets/src/components/AssignmentTable.js
@@ -14,6 +14,7 @@ import PopupMessage from './PopupMessage'
 import ConditionalWrapper from './ConditionalWrapper'
 import StyledTextField from './StyledTextField'
 import { calculateWeekOffset } from '../util/date'
+import { roundToXDecimals } from '../util/math'
 
 const styles = theme => ({
   root: {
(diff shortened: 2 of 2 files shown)


assets/src/util/assignment.js
```diff
@@ -21,15 +21,11 @@ const setAssigmentGoalInputState = (key, assignments, inputFocus) => {
 }
 
 const setAssignmentGoalGrade = (key, assignments, goalGrade) => {
-  // Use decimal place of pointsPossible if it's a decimal; otherwise, round to nearest tenth
-  const placeToRoundTo = (String(assignments[key].pointsPossible).includes('.'))
-    ? getDecimalPlaceOfFloat(assignments[key].pointsPossible) : 1
-
```
ssciolla, line 27: I think you are right that we need to remove this here and store values at some pre-determined precision and then instead round right before displaying. However, I'm concerned that removing this entirely might re-introduce the bug that I was trying to resolve with this code in the first place . See this comment from your Assignment Planning PR: https://github.com/tl-its-umich-edu/my-learning-analytics/pull/702#issuecomment-580877948

I think we still need some way to sometimes show decimal values with greater precision than a tenth in the UI. Maybe this logic that you removed could also get pushed into `AssignmentTable.js` somewhere, just with like the rounding.

assets/src/util/assignment.js
```diff
@@ -1,5 +1,5 @@
 import { calculateWeekOffset, dateToMonthDay } from './date'
-import { sum, roundToXDecimals, getDecimalPlaceOfFloat } from './math'
```
ssciolla, line 2: You may still want this function to help with what I described in the other comment, but if you don't end up using it, we should delete the definition in `math.js`.

assets/src/util/assignment.js
ssciolla: You can re-produce the bug I describe in that comment in this issue branch by modifying the points for an assignment to be `25.75`, or something similar.

justin0022: Okay I'll try to take a look at this next week.

jonespm:


```diff

Issues
======
+ Solved 2
- Added 3
           
```


See the complete overview on [Codacy](https://app.codacy.com/manual/ITS_Teaching_And_Learning/my-learning-analytics/pullRequest?prid=5608581&bid=18376833)

assets/src/util/assignment.js
justin0022: @ssciolla I think I fixed this now - could you review again?

assets/src/util/assignment.js
justin0022: I ended up using it to fix your prior review comment.

assets/src/util/assignment.js
ssciolla: Hi @justin0022, sure, I'll try to get to this today.

assets/src/util/assignment.js
```diff
@@ -21,15 +21,11 @@ const setAssigmentGoalInputState = (key, assignments, inputFocus) => {
...
-  // Use decimal place of pointsPossible if it's a decimal; otherwise, round to nearest tenth
-  const placeToRoundTo = (String(assignments[key].pointsPossible).includes('.'))
-    ? getDecimalPlaceOfFloat(assignments[key].pointsPossible) : 1
-
   return [
     ...assignments.slice(0, key),
     {
       ...assignments[key],
-      goalGrade: goalGrade === '' ? '' : roundToXDecimals(Number(goalGrade), placeToRoundTo),
+      goalGrade: goalGrade === '' ? '' : roundToXDecimals(Number(goalGrade), 3),
```
ssciolla, line 28: Just thinking/writing through this: if we only store goals to the the thousandth, but in some wacky case, the `pointsPossible` is ten thousandths, nothing would break because the `roundToXDecimals(some_num, 5)` call would still return a valid result. I think thousandths will be enough to store, since I've never seen that before (only hundredths, I think).

justin0022: Ya that's right. It wouldn't hurt to store a couple more decimals worth of precision - if this becomes an issue in the future we can revisit.

justin0022:
> Regarding the Codacy items about the blocks and integer, are these legitimate concerns or do you think we can ignore those particular patterns?

I just took a look, I think we need all the blocks that Codacy flagged - we use those imports. I think the comment about the integer we can ignore as well - I don't see any floating point issues.

ssciolla:
> I just took a look, I think we need all the blocks that Codacy flagged - we use those imports. I think the comment about the integer we can ignore as well - I don't see any floating point issues.

Okay, I think maybe sometime in the future I'd like to work with you and the other JS contributors (definitely @pushyamig ) to go through Codacy's JS rules and create our own curated set so we know anything that's flagged should be addressed.

Feel free to merge this PR.

justin0022:
That's a good idea

PR #940, opened 2020-05-28
Refactor mouse-over PopupMessage, fill in goal and assignment details (#876)

ssciolla:
This PR modifies the existing `PopupMessage` component used by the Assignment Planning Redesign view so that -- if the assignment/submission has been graded -- it includes the student's last goal and the class average. Goals are only shown if they were set by a user (either by entering a value or checking the "Lock Goal" checkbox). The PR aims to resolve issue #876.

Note: I still want to do some testing of the assignment group rules. Also, the rounding to the nearest hundredth for the average score is somewhat arbitrary, but I'm hoping we might soon have a more flexible solution that comes out of PR #938. Still, I think this is more or less what a solution to the issue will look like, so I thought I'd start getting feedback.

Screenshot of graded assignment/submission (data shown is from a test course, with fake students, prepared by @jennlove-um):

Screen Shot 2020-05-28 at 12 57 40 PM


jonespm:


```diff

Complexity increasing per file
==============================
- assets/src/components/PopupMessage.js  2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/ITS_Teaching_And_Learning/my-learning-analytics/pullRequest?prid=5608580&bid=18376832)

jennlove-um:
I think we still need to have "Your grade" and "Points possible" included in the mouse-over because the student can't see that information when the pop-up displays.  They would have to remember it in order to compare it to the goal and average grade when they're looking at the pop-up.

Also, can we simplify "previously set goal" to "Your goal"?

ssciolla:
> I think we still need to have "Your grade" and "Points possible" included in the mouse-over because the student can't see that information when the pop-up displays. They would have to remember it in order to compare it to the goal and average grade when they're looking at the pop-up.
> 
> Also, can we simplify "previously set goal" to "Your goal"?

Sure, I'll make those changes.

ssciolla:
Okay, @jennlove-um, how about something like this?

Screen Shot 2020-05-29 at 3 16 53 PM

Note that now if the assignment/submission has not been graded yet, it will not show "Your grade" or "Class average"; "Your goal", "Points possible", and "Rules" will always be shown, even if they have a null-like value.

jennlove-um:
Thanks, Sam.  This looks great.

pushyamig:
I feel the Pop message list should be consistence, In cases where the assignment is still not graded then the pop up should at least display `Your Grade: N/A` and `Class Average: N/A`
Currently, this is how it is showing
Screen Shot 2020-06-01 at 8 53 50 AM
Screen Shot 2020-06-01 at 8 53 54 AM

VS

Screen Shot 2020-06-01 at 8 59 17 AM

Thoughts??? @jennlove-um

assets/src/components/PopupMessage.js
```diff
@@ -1,5 +1,6 @@
 import React from 'react'
 import { Typography } from '@material-ui/core'
+import { roundToXDecimals } from '../util/math'
 
 function PopupMessage ({ a, assignmentGroups }) {
   const getAssignmentRules = (a, assignmentGroups) => {
```
pushyamig, line 6: @ssciolla On line#7 there is a typo in the const ` const assignmenGroup = assignmentGroups.find(ag => ag.id === a.assignmentGroupId)`. Can you fix that as well since your PR is modifying that file? and it is also a small change to fix

assets/src/components/PopupMessage.js
```diff
@@ -17,36 +18,34 @@ function PopupMessage ({ a, assignmentGroups }) {
...
   if (dropHighest !== 0 && dropLowest !== 0) {
-    return (
-      <Typography>
-        {
-          `The highest ${dropHighest} scores will be dropped from this assignment group
-              The lowest ${dropLowest} scores will be dropped from this assignment group`
-        }
-      </Typography>
-    )
+    rulesMessage = `Both the highest ${dropHighest} scores and lowest ${dropLowest} scores will be dropped from this assignment group`
```
pushyamig, line 23: Just a thought, this line is exceeding max character limit that IDE suggests, standard doesn't seem to see this a problem. I wonder we need to add any custom rules. But not relevant to this PR.

pushyamig:
This PR looks good.

assets/src/components/PopupMessage.js
ssciolla: Should be fixed now, @pushyamig .

assets/src/components/PopupMessage.js
ssciolla: I fixed this using some re-formatting. Yeah, not sure if `standard` has a line length rule...

ssciolla:
@jennlove-um , do you want to weigh in on @pushyamig 's question about whether "Your grade" and "Class average" should be conditionally hidden or just shown with "None" or N/A" when the submission has not been graded yet? I feel that showing a bunch of nulls isn't very helpful, and that it will be clear to the user that it hasn't been graded because they can enter a goal.

jonespm:


```diff

Issues
======
- Added 1
           

Complexity increasing per file
==============================
- assets/src/components/PopupMessage.js  2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/ITS_Teaching_And_Learning/my-learning-analytics/pullRequest?prid=5608580&bid=18376832)

ssciolla:
Hi @pushyamig and @justin0022, I know you already approved this, but I added some rounding behavior similar to what Justin did in #938. If you could OK my latest commit, I'll then merge this.

ssciolla:
Thanks, @justin0022! Merging now...

Issue #801, opened 2019-11-19
Views lack prominent headers or titles


ssciolla:
Thank you for contributing to this project!

- Make sure to search the [issues](https://github.com/tl-its-umich-edu/my-learning-analytics/issues) for duplicates first!

## Expected behavior (A clear and concise description of what you expected to happen) : 

Each view should have a clear heading or title so that all users will clearly understand when a view or route change has occurred.

## Describe the bug (Tell us what happens instead of the expected behavior) :

From the 10/28/19 accessibility review:

"There is no visible or programmatically discoverable way to tell in what view of this [single page application] the user is currently in. This is important for the blind persona."

## Additional context (Add any other context about the problem here) :

The Resources Accessed and Grades Distribution views have headers, but they are nested within Material UI Paper components that contain the rest of the content. Assignment Planning does not have a header (hopefully this will be addressed as part of work on issue #691 and PR #702). Perhaps headers or titles should live in their own Paper component; creating a new custom component might be appropriate.

The review provides these suggested solutions: "Change the title on view/route change. Or pass the focus to a heading that announces the view change."

justin0022:
I appreciate all of these accessibility issues. It'll take time to implement but well worth the effort.

jennlove-um:
testing passes

gsilver:
@jennlove-um - testing passes indeed!

Issue #807, opened 2019-11-19
HTML containers and headers are not consistently structured or labeled


ssciolla:
## Expected behavior

The structuring and labeling of container and heading HTML element should be consistent across all pages of the application.

## Describe the issue

From the 10/28/19 accessibility review:

<details>
<summary>(See review documents…)</summary>

> * MyLA Accessibility Review 10/28/18
>   * Google original: https://docs.google.com/document/d/1IFBNWHvzY1l8YT7G1TX9jCrOyU9yAxF1-KazNbb4kc0
>   * GitHub PDF copy: https://github.com/tl-its-umich-edu/my-learning-analytics/files/5502938/MyLA.Accessibility.Review.10_28_18.pdf
> * MyLA Accessibility Table 10/28/18 - Issue Notes
>   * Google original: https://docs.google.com/spreadsheets/d/1ZWB6Ip4I9H8718G4uMn6VcRDrpS7dkS5BCR7nB0Q8ao
>   * GitHub PDF copy: https://github.com/tl-its-umich-edu/my-learning-analytics/files/5502940/MyLA.Accessibility.Table.10_28_18.-.Issue.Notes.pdf

</details>

"Content is indifferently structured. This becomes important for the blind persona. … Header and footer are identified programmatically. The main content is not, however. … In addition - heading structure is a bit off. It starts with an h6 (we assume because it is embedded in Canvas) - but it has h5 subsections - at least the Progress toward Final Grade view does. This needs to be normalized."

## Additional context

~~Regarding the main container, the review suggests adding the `role="main"` attribute-value pair to the HTML element serving as the main container. The header problem might be solved simultaneously to issue #801 about the view headers.~~

WRT roles, new issue #1165 has been opened for that.

lsloan:
@ssciolla, am I correct in understanding that the main points of this issue are:

* h5 sections as children of h6 ones
* add `role="main"` to the main HTML element

ssciolla:
@lsloan, for the second bullet point, I think so? May want to read the W3C notes about that `role` type and how it should be used.

[...]

PR #949, opened 2020-06-03
Implement ViewHeader component, make view structure consistent (#801)

ssciolla:
This PR works to make the titles consistent across all the views and to announce the view title programmatically by sending focus to that element. More details are provided below in the task list. The PR aims to resolve issue #801 and work toward resolving issue #807.

- [x] Create `ViewHeader` component that sends focus to the element on initial render.
   - [x] Use `onBlur` event handler to trigger removal of `tabIndex` attribute when focus moves away from header.
   - [x] Add custom CSS class for modifying focus offset (Note: don't think modification of `focus` is possible without CSS or some other helper library; we can also consider modifying the focus style more or remove it entirely if we think it's too distracting).
- [x] Implement `ViewHeader` in "ResourcesAccessed" (required minor edits).
- [x] Implement `ViewHeader` in "Grade Distribution".
   - [x] Use `ViewHeader`. 
   - [x] Refactor slightly so the header will always be visible, even when the `AlertBanner` indicates there is no grade data available.
- [x] Implement `ViewHeader` in "Assignment Planning" (version 1).
   - [x] Refactor so that all page elements are within a single `Paper` component` (they were split between two before).
   - [x] Use `ViewHeader`.
   - [x] Add a new `section` class with top and bottom padding.
   - [x] Refactor to add `div`s that can use the `section` class to ensure proper spacing between sections ("Progress toward Final Grade" and "Assignments Due by Date").
- [x] Implement `ViewHeader` in "Assignment Planning" (version 2).
   - [x] Use `ViewHeader`.
   - [x] Add a new `section` class with top and bottom padding.
   - [x] Refactor to add `div`s that can use the `section` class to ensure proper spacing between sections ("Grade Progress" and "Assignments by Due Date").
- [x] Add `variantMapping` to the theme in `globals.js` so that `h5` and `h6` appear as `h1` and `h2` respectively in HTML.


ssciolla:
I am hoping to chat with an ITS accessibility expert about the focusing part of these changes, but I think the other changes to view structure and spacing will increase consistency. Unfortunately some of these changes meant I had to large chunks of the JSX for these views (mostly adding `div` tags and indenting), so some significant testing may be required.

pushyamig:
@ssciolla , I think you can put your code on the dev instance and ask for a review from Gonzalo(@gsilver)when this PR opens. He can give you some insights into it. He has been removed from our github account but if it is useful I can add him as colloborator

ssciolla:
> @ssciolla , I think you can put your code on the dev instance and ask for a review from Gonzalo(@gsilver)when this PR opens. He can give you some insights into it. He has been removed from our github account but if it is useful I can add him as colloborator

Thanks, @pushyamig, I already reached out to Gonzalo by email. I will plan to deploy this to the dev instance if I meet with him, or at some point during the review process.

assets/src/components/ViewHeader.js
```diff
@@ -0,0 +1,25 @@
...
+  // Focuses the page header only when the domElement changes.
+  // This should only be when it changes from null to something on initial render.
+  useEffect(() => {
+    if (domElement !== null) {
+      domElement.focus()
+    }
+  }, [domElement])
+
+  return (
+    <div tabIndex={0} ref={domElement => setDomElement(domElement)}>
```
ssciolla, line 17: If we don't want this `div` to be in the Tab order, but there is some benefit when using a screen reader to focus on this element, we could consider changing this to `-1`. See [MDN's tabindex page](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex).

justin0022:
@ssciolla I'm just taking an initial look, and I know the intent is for the title to be "focused", but does this mean that the title must have the "blue box" around it? 

Screen Shot 2020-06-03 at 10 12 43 AM

ssciolla:
> @ssciolla I'm just taking an initial look, and I know the intent is for the title to be "focused", but does this mean that the title must have the "blue box" around it?

Yeah, that's why I'm unsure about the current solution. If we change the `tabIndex` to `-1`, I think it would be focused but not available in the Tab order. However, without knowing how to use a screen reader myself, I don't know if that solves the issue raised by the accessibility review. That's why I'm reaching out to the ITS expert.

jonespm:
This doesn't even need a tabindex attribute if there's no action that can be taken on it.

It would have a tabindex=-1 if it should be skipped by keyboard navigation but there should be javascript actions available to it. (Which is rare)

ssciolla:
> Unless it's a link or a form control it shouldn't be tabbable and we should just remove the index. It doesn't _look_ like either of those but I haven't run the code.

@jonespm, we need to send focus to this element, I think, and everything I've read suggests that you need a `tabIndex` to focus. The proposed solution to the original issue was "Change the title on view/route change. **Or pass the focus to a heading that announces the view change.**" 

I don't think this needs to be available via the Tab order, so I've changed it to `-1`, which is what Gonzalo suggested in my initial communications with him. I'm still figuring out how to manage the prominent focus on the title, which feels a little distracting and unnecessary.

ssciolla:
FYI, I've deployed my latest code to the dev MyLA instance if you want to take a look. I haven't made any styling changes yet.

assets/src/components/ViewHeader.js
```diff
@@ -1,43 +1,33 @@
 import React, { useEffect, useState } from 'react'
 import Typography from '@material-ui/core/Typography'
-import useTimeout from '../hooks/useTimeout'
```
pushyamig: This is exactly what I was thinking, why do we need a timeout with all that hook fancy code, looks impressive but is that needed really. Thank for keeping it simple

ssciolla: Yeah, Gonzalo said the timeout wasn't really necessary, I think the use of `onBlur` makes more sense.

assets/src/components/ViewHeader.js
ssciolla: I've changed it to `-1` but also I'm removing the `tabIndex` when the focus moves away now.

ssciolla:
K, I think this is ready for review. I'm rebuilding an image for MyLA dev so should be available later today or tomorrow.

assets/src/containers/ResourcesAccessed.js
```diff
@@ -236,7 +237,7 @@ function ResourcesAccessed (props) {
       <Grid container spacing={2}>
         <Grid item xs={12}>
           <Paper className={classes.paper}>
-            <Typography variant='h5' gutterBottom className='title'>Resources Accessed</Typography>
```
pushyamig, line 239: Since you removed the Typography can you also remove the import as well?

assets/src/components/ViewHeader.js
```diff
@@ -0,0 +1,38 @@
...
+  // This should only be when it changes from null to something on initial render.
+  useEffect(() => {
+    if (domElement !== null) {
+      domElement.focus()
+    }
+  }, [domElement])
+
+  const headerProps = {
+    className: 'view-header',
+    variant: 'h5',
```
pushyamig, line 21: Since the variant mapping to appears as h5->h1 and h6->h2, does it make sense to have this variant as `h5` instead of `h1`?

assets/src/globals.js
```diff
@@ -20,13 +20,27 @@ const user = Object.freeze({
...
 
-const siteTheme = createMuiTheme({ palette })
+// Variant mapping: https://material-ui.com/components/typography/#changing-the-semantic-element
+// This will make uses of h5 appear as h1 and h6 appear as h2, without changing styling.
+const componentSettings = {
+  MuiTypography: {
+    variantMapping: {
+      h5: 'h1',
+      h6: 'h2'
+    }
```
pushyamig, line 36: Why is this change needed?

pushyamig:
I looked at the color around the title header and it is a little bit different than the Myla color palette. is there a reason for this color?
Screen Shot 2020-06-05 at 8 28 21 AM

assets/src/containers/ResourcesAccessed.js
ssciolla: Oh, yeah, I'll check to see if we can remove that, thanks.

assets/src/globals.js
ssciolla: If this has been implemented correctly, all uses of `h5` should appear as `h1`, and `h6` as `h2` in the page's HTML. According to Gonzalo, [Material UI](https://material-ui.com/components/typography/#changing-the-semantic-element), and [W3C](https://www.w3.org/WAI/tutorials/page-structure/headings/), it is important for accessibility that a page's headings start at `h1` and descend without skipping levels. Previously, we were using `h5` and `h6` as the preferred styles for headings, so I wanted to maintain those styles while making the heading numbers appear correctly in the HTML. Does that make sense?

assets/src/components/ViewHeader.js
ssciolla: This has been confusing to me, but I think we have to use the heading with the style we want (`h5`) and `variantMapping` will convert the actual tag to the right number.

ssciolla:
> I looked at the color around the title header and it is a little bit different than the Myla color palette. is there a reason for this color?

The only modification I am making right now to the focus style is to add an offset of `5px` to the `outline` (see the changes to `index.css`), so what you're seeing should be your browser's (or maybe Material UI's?) focus treatment. It doesn't seem like the offset is being registered in the image you showed me for some reason.

@pushyamig, what style would you like this to be? Would you like me to make invisible even? I can style the outline in terms of pattern or color in various different ways, or remove it, just let me know what you think would be best.

jonespm:


```diff

Issues
======
+ Solved 1
- Added 2
           

Complexity increasing per file
==============================
- assets/src/components/ViewHeader.js  2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/ITS_Teaching_And_Learning/my-learning-analytics/pullRequest?prid=5646924&bid=18492831)

assets/src/globals.js
pushyamig: Indeed make sense, this is a kind global change for this inline version
`<Typography variant="h1" component="h5">`
This is reflected correctly in the UI and it doesn't have the same effect of `<h2>` header tag
Screen Shot 2020-06-05 at 10 41 55 AM

Screen Shot 2020-06-05 at 10 42 55 AM

assets/src/containers/ResourcesAccessed.js
ssciolla: This should be resolved now.

assets/src/index.css
```diff
@@ -26,6 +26,11 @@ code {
   animation: sk-scaleout 1.0s infinite ease-in-out;
 }
 
+/* Modifies default focus style for header tag in ViewHeader component */
+.view-header:focus {
+  outline-offset: 5px;
```
pushyamig, line 31: I don't see the effect of this styling in the UI. I have changed the outline-offset setting am I missing something?
Screen Shot 2020-06-05 at 11 04 41 AM

pushyamig: The `outline-offset: 5px;` property doesn't see the effect the chrome, but only in FF. I am not sure why is this case? Need to support all prime browser as per this [list](https://github.com/tl-its-umich-edu/my-learning-analytics/wiki/MyLA-Browser-Support)

pushyamig:
> what style would you like this to be? Would you like me to make the focus invisible even? I can style the outline in terms of pattern or color in various different ways, or remove it. Just let me know what you think would be best.

Invisible is better, if that is not the expected behavior then add color to the header the blend in with current style guide recommendation

assets/src/index.css
pushyamig:
```
outline-offset: 5px;
  outline-style: solid;
```
This combination seem to work in Chrome to add little spacing around the box

ssciolla: Okay, I think I added a specific style for the focus so it is consistent across browsers. Below is a screenshot of how it looks in Chrome. I opted to use the `secondary` color (`#2C6496`), since that indicates a selected state in the rest of the application.
Screen Shot 2020-06-05 at 1 24 48 PM

ssciolla: @jennlove-um, I would like your input on how this looks. What I've seen online suggests that [focus should be visible](https://www.w3.org/TR/UNDERSTANDING-WCAG20/navigation-mechanisms-focus-visible.html), but this could be considered a slightly different case since it is not for an element that accepts user interaction. Gonzalo said that, if we are concerned about the appearance, it would be okay to remove the focus style entirely. It will still focus when the page loads, but it just won't be visible. Note that the focus box (seen above) will disappear if the user presses Tab or otherwise moves the focus elsewhere.

jonespm:


```diff

Issues
======
+ Solved 1
- Added 3
           

Complexity increasing per file
==============================
- assets/src/components/ViewHeader.js  2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/ITS_Teaching_And_Learning/my-learning-analytics/pullRequest?prid=5646924&bid=18492831)

jonespm:
On the[ ReactJS accessibility page](https://reactjs.org/docs/accessibility.html) they mentioned the [DocumentTitle](https://github.com/gaearon/react-document-title) and [React Helmet package](https://github.com/nfl/react-helmet) that sound like they do updates directly to the header. Did you try either of these?

ssciolla:
> On the[ ReactJS accessibility page](https://reactjs.org/docs/accessibility.html) they mentioned the [DocumentTitle](https://github.com/gaearon/react-document-title) and [React Helmet package](https://github.com/nfl/react-helmet) that sound like they do updates directly to the header. Did you try either of these?

I looked into `react-helmet` and raised the possibility of changing the document title this way with Gonzalo. He replied "Changing the title is a tall proposition and may not be useful to all screen reading software. I think the best solution is to send focus to the heading, and remove the tabindex when that loses focus."

I don't think using something like this is a bad idea (although I'm not exactly sure how I would change the title, since it's currently "My Learning Analytics" -- maybe adding something after a colon), but from Gonzalo's perspective, it seems the header focus might be more helpful to users.

jonespm:
Before single page apps, the user noticed that they changed pages by the title changes. I feel like updating the title would have benefit both to visual users as well as screen reader users as this is what's displayed in the tab. 

Apps like Gmail are a SPA and the title updates when you folders something from 

`Inbox - jonespm@gmail.com - Gmail`
`Starred - jonespm@gmail.com - Gmail`
or more specifically

`<Folder Name> <email> - Gmail`

It seems like our style should be of the format
`<Page Name> - <username> - My Learning Analytics`

Sorry for being a day late to the comments here. It's also possible the username is not needed.

ssciolla:
> Before single page apps, the user noticed that they changed pages by the title changes. I feel like updating the title would have benefit both to visual users as well as screen reader users as this is what's displayed in the tab.

@jonespm, yeah, I think this is a good idea. However, with the way you are commenting (without reviewing or requesting changes), I am very unclear about how you want me to proceed. Are you suggesting what I've done with focus should be removed and replaced with something like a `react-helmet` implementation? Are you saying this would be good to in addition? Is it something that has to be in this PR, or could it be resolved in a separate issue? If you -- as a developer senior to myself on this project, and someone whose opinion I value -- could give me clear direction, that would be appreciated.

jonespm:
@ssciolla I don't consider myself to be very experienced with React. I also haven't installed a screenreader to test this out. I've done training and read a bit more and done work on accessibility in general and just making observations based on other pages and also Django Admin on how they change the title.

I don't know anything about those two libraries I mentioned or what they do. I'd estimate would take a few hours of effort to try them out myself. Perhaps they could replace some of this effort, perhaps they could supplement it. I was just wondering if you'd considered them as it looked a little similar to what you were writing with this new component. 

I usually just figure if there's a seemingly common problem that someone else already has a good solution, this didn't feel very unique to this project. 

The code here could probably be swapped out later too for something else.

Sorry there's really no clear direction, as that direction would be:
- Try out these libraries and let us know how they work
- Try out all 3 solutions with a screenreader (VoiceOver on OSX works, NVAccess is free and popular), let us know what works best. I know JAWS is popular but don't know if what is most used anymore.

I tested this with VoiceOver. This solution works better than changing the Title as even on Gmail the title updates don't get Voiceover to read them. I don't know  about other screenreaders. I feel *both* solutions might be useful but not sure of the effort.

ssciolla:
Okay, @jonespm, thanks for explaining a bit more. I am feeling the pressure to at least partially resolve this issue/PR because we are trying to release next week. Because there is research involved in deciding on a library and figuring out it works, I feel it would make sense to tackle the title updating separately. I will create an issue and assign it to myself. It would still be great to have your review on what I have here -- it sounds like it will be, based on your testing and Gonzalo's expertise.

pushyamig:
> I also haven't installed a screenreader to test this out.
I don't want to make the conversations anymore longer than what it is but do we want to make testing with screen readers as a standard process for resolving the accessibility issue? It will be hard to reach out to Accessibility experts for each and every issue.

ssciolla:
FWIW, I just tried implementing title changes with `react-helmet`, and it was quite easy: https://github.com/ssciolla/my-learning-analytics/commit/4c4e6bd6168b8651ee26067da4c71702a860d36c

I still think it might be worth investigating this more, and it would likely make sense to adjust the title for other routes or views, like the list of courses or views. I'm also not sure the best place for `Helmet` is in `ViewHeader`; maybe it should be wrapped around the instances of `Route`. Thus I'd still like to try to tackle this in another issue/PR and limit the scope of this one, since I'm already making a good number of changes.

pushyamig:
We should merge this what @ssciolla has done here is valuable and we are all learning with accessibility stuff with this project. It is separate Question how will you mark this issue as done

pushyamig:
> Thus I'd still like to try to tackle this in another issue/PR and limit the scope of this one, since I'm already making a good number of changes.

I agree

ssciolla:
I spoke with @jennlove-um about the focus styling, and we opted to remove the focus outline (i.e. change the CSS class to only use `outline: none`. I inspected the HTML and used VoiceOver, and the title still properly receives focus.

jonespm:


```diff

Issues
======
+ Solved 1
- Added 2
           

Complexity increasing per file
==============================
- assets/src/components/ViewHeader.js  2
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/ITS_Teaching_And_Learning/my-learning-analytics/pullRequest?prid=5646924&bid=18492831)

ssciolla:
I'm merging now, thanks everyone for the feedback.

Issue #953, opened 2020-06-04
goalGradeSetByUser stored in DB with inconsistent data types


ssciolla:
Thank you for contributing to this project!

- Make sure to search the [issues](https://github.com/tl-its-umich-edu/my-learning-analytics/issues) for duplicates first!

## Expected behavior (A clear and concise description of what you expected to happen) : 

The data stored as JSON in the database for `default_view_value` as part of the Assignment Planning Redesign via a GraphQL mutation should be stored consistently, with consistent data types. (In an ideal world with infinite time, this data would atomized into one or more tables so we could use SQL data types to enforce some of this.)

## Describe the bug (Tell us what happens instead of the expected behavior) :

While testing PR #940, I noticed that sometimes the value of the `goalGradeSetByUser` property was sometimes stored as a boolean (`true` or `false`) and other times stored as the goal text input. I believe this is supposed to always be a boolean, since there is an alternative `goalGrade` property in the data.

## Steps to Reproduce :
1. Visit the Assignment Planning Redesign view.
2. Add a goal manually via a text input field.
3. Add another goal by clicking the Lock checkbox next to an auto-generated goal.
4. Look at the `user_default_selection` table in the database.

## Additional context (Add any other context about the problem here) :

I think (but am not positive) the reason for behavior described above is the following function, where it appears `goalGrade` is assigned to `goalGradeSetByUser`.

https://github.com/tl-its-umich-edu/my-learning-analytics/blob/f55ced5cf4d6ba16a07f19f058bc4840fd24a5af/assets/src/util/assignment.js#L23-L33=

ssciolla:
Hi @justin0022, any chance you could take a look at this? I'm not exactly sure of what your intentions were with these properties. It might be a one-line fix.

justin0022:
Hi Sam, yes I think this is a bug. I'll look into this, thanks for pointing it out

pushyamig:
goalGradeSetByUser property is logged a boolean.
Test passes

PR #955, opened 2020-06-04
Fixed goalGradeSetByUser type

justin0022:
Resolves https://github.com/tl-its-umich-edu/my-learning-analytics/issues/953.

Instead of `null` as the default value, I've switched it to `false`.


ssciolla:
Hi Justin, I haven’t tried running this yet, but I think maybe I didn’t communicate the issue clear enough. I’m seeing sometimes goals (e.g. 92.5) stored in `goalGradeSetByUser`, but it seems like this should always have a Boolean value, right? I can follow up with more details tomorrow. Thanks!

assets/src/util/assignment.js
```diff
@@ -26,7 +26,7 @@ const setAssignmentGoalGrade = (key, assignments, goalGrade) => {
     {
       ...assignments[key],
       goalGrade: goalGrade === '' ? '' : roundToXDecimals(Number(goalGrade), 3),
-      goalGradeSetByUser: goalGrade
+      goalGradeSetByUser: !!goalGrade
```
justin0022, line 29: Sam I think this is the change that should fix this issue? Sorry I think I may have incorrectly merged the master in the previous commit - this one should work.

ssciolla: Cool, thanks, I'll try this in the next couple hours and review.

ssciolla:
@pushyamig, I'm reviewing this, but I wanted you to be aware of the issue/changes.

pushyamig:
@ssciolla OK I will also look into it

justin0022:
> in my testing, I never saw a case where `goalGradeSetByUser` was stored as `false`. 

If you uncheck the "Lock Goal", the associated assignment `goalGradeSetByUser` should be set to `false`. I use this as a way to filter for `gradedAssignments` and `ungradedAssignments`. 

The idea is that `gradedAssignments` can be considered as assignments that either are graded or the goal grade is set by the user, which are the "fixed" grades, and that `ungradedAssignments` are the inverse, which are the "unfixed" grades that can be calculated to match the overall course goal grade.

ssciolla:
> If you uncheck the "Lock Goal", the associated assignment `goalGradeSetByUser` should be set to `false`. I use this as a way to filter for `gradedAssignments` and `ungradedAssignments`.

Okay, but I'm when I uncheck "Lock Goal", I'm not seeing anything stored at all for that assignment.

With one goal, it looks like this: 

``` 
{"goalGrade": "90", "assignments": [{"assignmentId": "1770000000XXXXXX", "goalGradeSetByUser": true, "goalGrade": 9}]}
```

And after I un-check, it looks like this:

```
{"goalGrade": "90", "assignments": []}
```

This doesn't mean it's not working, I'm just trying to understand the logic of what's stored in the DB.

justin0022:
> Okay, but when I uncheck "Lock Goal", I'm not seeing anything stored at all for that assignment.

The user settings are not set unless "Lock Goal" is checked, since when this is unchecked, it means that the grade calculation is done automatically by the application. 

The added benefit is that by not saving the automatically calculated grades, if there are additional assignments added to the course in the future, the calculated grades can freely adjust. If we stored these grades, and new assignments get added, then the old stored grades will need to be discarded, recalculated, and stored. 

If there's a need to store the calculated grades, it can definitely be done!

ssciolla:
> The user settings are not set unless "Lock Goal" is checked, since when this is unchecked, it means that the grade calculation is done automatically by the application.

Not storing the system-generated (un-locked) goals is fine with me. I was responding to your earlier assertion that "If you uncheck the "Lock Goal", the associated assignment `goalGradeSetByUser` should be set to `false`." I have yet to see `false` be stored in the database. Maybe you meant that because it's undefined, it's `falsy` and the system knows that.  But then you get back to my question of why this field is needed in the first place. You probably could just check whether `goalGrade` is set rather than I relying on this boolean to decide whether it should be included in `gradedAssignments`. I understand the convenience of having a separate flag, and to be clear, I'm not asking you to change it.

justin0022:
Sorry I meant it was stored on the front end in the AssignmentTable component - not in the back-end. There was a reason I had to store `goalGradeSetByUser` independently from `checkboxState` - I can't exactly recall now why I added it, but I do remember thinking it was too bad that this additional state was necessary. 

If everything works as expected without `goalGradeSetByUser`, I'm all for removing it.

ssciolla:
Thanks for the explanation, @justin0022. If you want to explore removing it, I'm okay with that, but I'm also totally fine with keeping it as is for now. As noted elsewhere, we are trying to release this week, so merging and creating a separate issue would be a fine way to proceed.

justin0022:
Thanks I'll keep it on my backlog, I'll merge this in for now.

PR #978, opened 2020-07-01
#950 upgrade to python3.8

### LINKED ISSUE #950: Update Python from 3.7 to 3.8 look at new images

Back at one time we looked at Alpine but it [sounds like just using the slim](https://pythonspeed.com/articles/base-image-python-docker-images/) is fine.

We should update to 3.8 since it's been out for over 3 months now and probably use the 3.8-slim Docker image as a base.

pushyamig:
Fixes #950 

I tested the cron and application launches fine.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -27,7 +27,7 @@ RUN npm prune --production && \
     find /usr/src/app/node_modules -type d -name "examples" -print0 | xargs -0 rm -rf
 
 # FROM directive instructing base image to build upon
-FROM python:3.7 AS app
+FROM python:3.8-slim AS app
 
 # EXPOSE port 5000 to allow communication to/from server
 EXPOSE 5000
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -36,7 +36,8 @@ WORKDIR /code
 # NOTE: requirements.txt not likely to change between dev builds
 COPY requirements.txt .
 RUN apt-get update && \
-    apt-get install -y --no-install-recommends netcat vim-tiny jq python3-dev xmlsec1 cron && \
+    apt-get install -y --no-install-recommends \
+    build-essential default-libmysqlclient-dev  libpq-dev netcat vim-tiny jq python3-dev xmlsec1 cron && \
```
jonespm, line 40: This reminds me, I'd like to create an issue where we remove the cron. I wonder how Unizin is running this though.

Issue #952, opened 2020-06-04
Upgrade to Django 3 and update other pip packages supporting it


pushyamig:
New version of the Django helps improve performance for Myla that we are hoping to achieve for Fall 2020. This has support for some of the features like asyncio.

Also update the pip packages in support of Django 3 or latest package updates from PyPi that directly not dependency on django

This is mentioned in #950

pushyamig:
The [stable](https://pypi.org/project/Django/#history) release version for Django is `3.0.8`, The next version 3.1 is [scheduled](https://www.djangoproject.com/weblog/2020/jul/20/django-31-release-candidate-1-released/) to for a release is Aug. I will stick with 3.0.8 for Fall, thoughts???

pushyamig:
I have been locally testing with 3.0.8 since June

pushyamig:
I see this issue when upgrading to [Pixan-eventlog](https://github.com/pinax/pinax-eventlog/issues/32) library. I reported the issue

pushyamig:
I also see this issues when upgrading the new version of [promise 2.3](https://github.com/tl-its-umich-edu/my-learning-analytics/issues/994)

jonespm:
This looks like it's being incorporated by #982

ssciolla:
I will QA.

ssciolla:
I confirmed that the package updates are reflected in the images created by `docker-compose build`, everything seems to be still functioning with the changes -- with no glaring issues in the logs during build or performing regular application tasks. Will move to Done.

PR #981, opened 2020-07-08
Supporting upserting resources and inserting new resource_accessed

### LINKED ISSUE #973: Change the Resources ETL to use a daily-delta approach
Update the ETL to use a delta approach instead of dropping all data and reloading.

jonespm:
resources/resources_accessed tables.

Fixes #973

Here's the specific things fixed here:
* Removed the deprecated inplace from all to_sql
* Using `pangres.upsert` here to update the resources table on the resource_id/course_id combo key
* Find if the latest record in the resources_accessed, only query from that date/time and insert after that, otherwise use the earliest course start date

There was also some model changes to help facilitate this.
* Foreign key from resource_access.resource_id to resource.resource_id. I'd like to do one between user_id but maybe in another fix.
* Foreign key from resource.course_id to course.id (This was semi-optional but would help with any cleanup and references)
* Made resource_id and course_id unique_together so they can be up-serted

TODO: 

- [x] Test that migration 16 can run on a populated database, I dumped and rebuild the database multiple times so I haven't tested this
- [x] Fully test the UI. I didn't change any columns so everything should still work.


jonespm:


```diff

Complexity increasing per file
==============================
- dashboard/migrations/0016_auto_20200707_1718.py  1
- dashboard/cron.py  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/ITS_Teaching_And_Learning/my-learning-analytics/pullRequest?prid=5829273&bid=19085658)

jonespm:
So I ran into the problem of composite foreign keys.

Currently resource_access (populated from BigQuery) is
```
id (auto generated)
resource_id
user_id
access_time
```

Currently resource (populated from BigQuery and updated from Canvas) is
```
id (auto generated)
resource_type (comes from query)
resource_id
name
course_id
```

The problem is the code does a query in the view to link these tables on resource_id and user_id. But we've had the bug in the past as the unique columns are resource_id and course_id.

`WHERE a.resource_id = r.resource_id and a.user_id = u.user_id`

I'm trying to figure out how to make these tables work together. What we need is
- Only insert new records into resource_access
- Upsert into resource

resource_access can't do a FK on id because it would have to look-up the course_id and resource_id. I'd like to do a composite FK but Django doesn't support that directly. There is this package for **composite-fk** I found but it's been 2 years since it's been last updated.

Maybe resource_access doesn't need to be composite. Here's a thought. 

- Remove course_id from resource.
- Add course_id to resource_access (As a FK to course)
- FK Link resource_access.resource_id to resource.resource_id
- Change some of the queries in the views to use course_id from resource_access

This might work. Then this means that a resource is unique in the resource table by resource_id again. This would change some of what was in #918.

ssciolla:
I think the approach you described sounds pretty good, and simpler than what I had suggested before (where there were separate tables for `resource_access` and `resource_course`). Essentially you're making `resource_access` a M:M table similar to `enrollment` in the `course-inventory` project, where it's connecting three entities (there it was `course`, `enrollment`, and `section`; here it's `resource`, `course`, and `user`). It makes sense to me that a `resource_access` event would require IDs for these three entities.

I'd recommend using EER diagrams to communicate what you're talking about. Happy to talk to you about your thinking outside GitHub.

jonespm:
Yeah I wanted to talk about this at the meeting today but other topics dominated. I've got a short week next week and wanted to get it done before I left but don't think that will happen. Aiming for the end of July now.

jonespm:


```diff

Issues
======
- Added 1
           

Complexity increasing per file
==============================
- dashboard/migrations/0016_auto_20200720_1203.py  1
- dashboard/cron.py  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/ITS_Teaching_And_Learning/my-learning-analytics/pullRequest?prid=5829273&bid=19085658)

dashboard/models.py
```diff
@@ -13,6 +13,8 @@
 from django.conf import settings
 from django.core.exceptions import ObjectDoesNotExist
 from django.urls import reverse
+from compositefk.fields import CompositeForeignKey
```
ssciolla, line 16: You're not using this anymore, right? Let's remove it here and in `requirements.txt` if so.

requirements.txt
```diff
@@ -48,3 +49,4 @@ django-ptvsd-debug==1.0.3
 django-lti-auth==1.0.0
 django-webpack-loader==0.6.0
 django-csp==3.5
+django-composite-foreignkey==1.1.0
```
ssciolla, line 52: See above note.

dashboard/models.py
```diff
@@ -356,17 +369,4 @@ def __str__(self):
...
-    resource_id = models.CharField(blank=True, max_length=255, null=False, verbose_name='Resource Id')
-    user_id = models.BigIntegerField(blank=True, null=False, verbose_name='User Id')
-    access_time = models.DateTimeField(verbose_name="Access Time")
-
-    def __str__(self):
-        return f"Resource {self.resource_id} accessed by {self.user_id}"
-
-    class Meta:
-        db_table = 'resource_access'
+        unique_together = (('id', 'course_id'),)
```
ssciolla, line 372: Nit: This isn't related to what you're doing, but I think `unique_together` will be deprecated in the near future.

https://docs.djangoproject.com/en/3.0/ref/models/options/#unique-together

dashboard/cron.py
```diff
@@ -247,6 +245,15 @@ def update_with_bq_access(self):
...
         course_start_time = utils.find_earliest_start_datetime_of_courses()
 
+        # the earliest latest date of all resources_accessed
+        try:
+            latest_resource = ResourceAccess.objects.latest('access_time')
+            latest_resource_time = latest_resource.access_time
+        except ResourceAccess.DoesNotExist:
+            latest_resource_time = course_start_time
+
+        logger.info(f"Earliest Start: {course_start_time} Latest resource: {latest_resource_time}")
```
ssciolla, line 255: I find this log message somewhat confusing. I think if it were me, I'd nest log messages in the `try/except` and log something different depending on what source you use (courses or resource access).

pushyamig:
I was trying to run this locally, just this PR not the #989 and I started with empty DB and I see this error
```
[2020-07-27 12:20:52 -0400] [ERROR] [cron.py:545] Exception running BigQuery update
  File "/code/dashboard/cron.py", line 542, in do
    status += self.update_with_bq_access()
  File "/code/dashboard/cron.py", line 314, in update_with_bq_access
    resource_df = resource_df.set_index(["resource_id", "course_id"])
  File "/usr/local/lib/python3.7/site-packages/pandas/core/frame.py", line 4411, in set_index
    raise KeyError("None of {} are in the columns".format(missing))
KeyError: "None of ['course_id'] are in the columns"
```

I see none of the resources tables are populated

dashboard/cron.py
```diff
@@ -298,13 +305,16 @@ def update_with_bq_access(self):
             # Because we're pulling all the data down into one query we need to manipulate it a little bit
             # Make a copy of the access dataframe
             resource_df = resource_access_df.copy(deep=True)
-            # Drop out the columns user and access time from resource data frame
-            resource_df.drop(["user_id", "access_time"], axis=1, inplace=True)
+            # Drop out the columns user course_id and access time from resource data frame
```
ssciolla, line 308:
```suggestion
            # Drop out the columns user_id, course_id, and access time from resource data frame
```

ssciolla: Not really sure if this comment adds anything.

dashboard/cron.py
```diff
@@ -298,13 +305,16 @@ def update_with_bq_access(self):
...
-            resource_df.drop_duplicates(["resource_id", "course_id"], inplace=True)
+            resource_df = resource_df.drop_duplicates(["resource_id",])
+
+            # Set a dual index for upsert
+            resource_df = resource_df.set_index(["resource_id", "course_id"])
 
-            # Drop out the columns resource_type, course_id, name from the resource_access
-            resource_access_df.drop(["resource_type","name", "course_id"], axis=1, inplace=True)
+            # Drop out the columns resource_type, name from the resource_access
+            resource_access_df = resource_access_df.drop(["resource_type","name"], axis=1)
```
ssciolla, line 317:
```suggestion
            # Drop out the columns resource_type and name from the resource_access
            resource_access_df = resource_access_df.drop(["resource_type", "name"], axis=1)
```

dashboard/migrations/0017_auto_20200720_1203.py
```diff
@@ -0,0 +1,38 @@
...
+    dependencies = [
+        ('dashboard', '0015_auto_20200116_1408'),
+    ]
+
+    operations = [
+        # We have to remove all rows from resource and resource access before performing this migration
+        # As there is no way of knowing which course_id a resource access was linked to
+        # This is the current process for the cron anyway
+        migrations.RunSQL("DELETE FROM resource_access;"),
+        migrations.RunSQL("DELETE FROM resource;"),
```
ssciolla, line 18: Have we typically worried about doing data manipulation in migrations in the past for MyLA? Seems like usually we just determine whether the database needs to be emptied before running migrations. Maybe this is smart, maybe it's unnecessary. Might be kind of nice to just have this migration be fully auto-generated. I'm okay either way, just thought I'd like to hear your thinking.

dashboard/models.py
```diff
@@ -280,9 +282,8 @@ def get_course_resource_type(self, course_id):
 class Resource(models.Model):
     id = models.AutoField(primary_key=True, verbose_name="Table Id")
     resource_type = models.CharField(max_length=255, verbose_name="Resource Type")
-    resource_id = models.CharField(blank=True, db_index=True, max_length=255, null=False, verbose_name="Resource Id")
+    resource_id = models.CharField(unique=True, blank=False, db_index=True, max_length=255, null=False, verbose_name="Resource Id")
```
ssciolla, line 285: `blank=False` and `null=False` maybe unnecessary here; those are defaults.
https://docs.djangoproject.com/en/3.0/ref/models/fields/#null

dashboard/cron.py
ssciolla: Similar note about the comment here; the code basically says exactly what the comment does.

dashboard/cron.py
```diff
@@ -298,13 +305,16 @@ def update_with_bq_access(self):
...
-            # Drop out the columns user and access time from resource data frame
-            resource_df.drop(["user_id", "access_time"], axis=1, inplace=True)
+            # Drop out the columns user course_id and access time from resource data frame
+            resource_df = resource_df.drop(["user_id", "course_id", "access_time"], axis=1)
             # Drop out the duplicates
-            resource_df.drop_duplicates(["resource_id", "course_id"], inplace=True)
+            resource_df = resource_df.drop_duplicates(["resource_id",])
+
+            # Set a dual index for upsert
+            resource_df = resource_df.set_index(["resource_id", "course_id"])
```
ssciolla, line 314: Yep, as noted by @pushyamig, this is throwing an error. Your data model no longer has a `course_id` on `Resource`, so I tried removing `"course_id"` and things seemed to work. Maybe this line and the comment above are an artifact of a previous iteration.

Also, how have you been testing the updating? Do you have a test course set up?

jonespm: Thanks Sam, I'll check and see if I just missed pushing an update for this.

I was testing this by manually manipulating the database in the local destination database rather than waiting for the source to change.

I'd delete some rows from resource_access to make sure the new ones get inserted and change some values in resource. I think otherwise we'll have to use a test course like you suggest and make changes over a few days.

jonespm:


```diff

Issues
======
- Added 1
           

Complexity increasing per file
==============================
- dashboard/migrations/0017_auto_20200720_1203.py  1
- dashboard/cron.py  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/ITS_Teaching_And_Learning/my-learning-analytics/pullRequest?prid=5829273&bid=19085658)

dashboard/models.py
jonespm: Thanks, no I didn't end up using this will remove!

dashboard/models.py
jonespm: It looks like this was just a notice for Django 3 and UniqueConstraint is only in Django 3. 

#952 is still a work in progress. I feel like this shouldn't wait on the update and we'd replace this a future Django update.

dashboard/cron.py
jonespm: Okay, that makes sense. I'll have 2 custom messages.

dashboard/models.py
ssciolla: I'm seeing this was introduced in Django 2.2, but yeah, it doesn't need to be resolved now.

https://docs.djangoproject.com/en/2.2/ref/models/options/#unique-together

dashboard/models.py
jonespm: Okay, I didn't notice what the defaults were I can remove these.

dashboard/migrations/0017_auto_20200720_1203.py
jonespm: I'm sure we could have handled it, but we are also not the only school exclusively running this anymore so I figured we need to have everything automated. Having it like this just seemed to make it easier.

If these tables aren't emptied the alternative is to UPDATE the course_id in resource_access from resource.

However resource_access before only had `resource_id + user_id` so there wasn't 100% chance that the match would work. I think it was initially implemented on the wrong table. This is likely a *very* low case (just lecture capture really) so we could have updated and retained the data but I figured since the process as-is currently deletes and re-populates, one more time would be fine.

jonespm:


```diff

Complexity increasing per file
==============================
- dashboard/migrations/0017_auto_20200720_1203.py  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/ITS_Teaching_And_Learning/my-learning-analytics/pullRequest?prid=5829273&bid=19085658)

jonespm:
Okay, I think I've got all the issues and conflicts resolved in here.

pushyamig:
I was testing this locally here are steps and things noticed
1. clean DB, migrations applied fine.
2. add a course to site and ran migration. I see data loaded
3. I added another course to the site, data loaded. I don't feel the data got deleted one of the message said `0/0 deleted` I saw all the resources got updated
4. I launched the resource view and I don't see the data loaded. Once of the course 335188 i usually test with and it has right start and end date taken from the test-myla instance. This is the case for both the courses i tested with
5. I see resource access table course_id is Null. Probably this is the reason for not loading the data. I don't see any error when a GET call is made. Looks to me like the query did not match anything from DB

jonespm:
Thanks for testing, looks like I missed a line that was still dropping course_id from resource_access. Pushed a commit to fix this.

dashboard/migrations/0017_auto_20200720_1203.py
```diff
@@ -0,0 +1,38 @@
...
+        migrations.RemoveField(
+            model_name='resource',
+            name='course_id',
+        ),
+        migrations.AddField(
+            model_name='resourceaccess',
```
pushyamig, line 24: The model name is spelled wrong here must be `resource_access'?

pushyamig: Probably this is django generated. I see migrations applied correctly

jonespm:


```diff

Complexity increasing per file
==============================
- dashboard/migrations/0017_auto_20200720_1203.py  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/ITS_Teaching_And_Learning/my-learning-analytics/pullRequest?prid=5829273&bid=19085658)

requirements.txt
```diff
@@ -27,6 +27,7 @@ django-model-utils>=3.2.0,<3.2.99
 # These should be okay to update minors
 numpy>=1.17,<1.17.99
 pandas>=0.25.1,<0.25.99
+pangres>=2.1,<2.99
```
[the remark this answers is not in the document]
jonespm: I'm not seeing anywhere definitely if it does or doesn't. All it uses is dataframe which seems like it should be compatible and in the requirements it has

`pandas>=0.25.3`

It seems like the big Pandas 1.0.x requirement is only supporting Python 3.6+ and they'll start breaking things in Pandas 2.0.x.

dashboard/migrations/0017_auto_20200720_1203.py
jonespm: Yeah it was generated by Django. It looks like it's the name of the class, case insensitive.

dashboard/migrations/0017_auto_20200720_1203.py
```diff
@@ -0,0 +1,38 @@
+# Generated by Django 2.2.14 on 2020-07-20 16:03
+
+from django.db import migrations, models
+import django.db.models.deletion
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('dashboard', '0015_auto_20200116_1408'),
```
jonespm, line 10: Needs to go after the utf8 conversion now.

```suggestion
        ('dashboard', '0016_utf8_conversion'),
```

Issue #962, opened 2020-06-11
LTI integration & SameSite readiness


jennlove-um:
Instructors need to be able to add MyLA to courses as an LTI

This is the highest priority for Fall 2020.

jennlove-um:
- [ ] Instructor admin page to configure MyLA for their course (#849 )
- [ ] Manage course end dates (#963 )

jennlove-um:
Notes from discussion: 
Can we upgrade to New LTI version for Myla instead of 1.1? (We can but it might either require writing a new Django Middleware or finding one, I haven't seen one yet)

1.3 library: new library available other than django-tsugi. PyLTI1.3 as of March 2020.
https://github.com/dmitry-viskov/pylti1.3-django-example

pushyamig:
How are we planning to add Myla as LTI tool? are we adding it by default in the side navigation or make it available in the Canvas App at the account level so that instructor can add it to their course if they prefer.

pushyamig:
Since we are planning on a on demand cron approach i.e if an instructor enables it from the App list and (Hopefully) launches the tools then we create a his course entry in the course table. Later when the cron run we can load the data he needs.

I see that as part of the launch we get passed in different roles the user has in Canvas, not just single role.

For example,
['http://purl.imsglobal.org/vocab/lis/v2/institution/person#Student'
'http://purl.imsglobal.org/vocab/lis/v2/institution/person#Instructor'
'http://purl.imsglobal.org/vocab/lis/v2/institution/person#Student'
'http://purl.imsglobal.org/vocab/lis/v2/system/person#User'
'http://purl.imsglobal.org/vocab/lis/v2/membership#Learner'
'http://purl.imsglobal.org/vocab/lis/v2/system/person#User']

membership is the one that actually give info about his role in the course.
I see that for 
`instructor: http://purl.imsglobal.org/vocab/lis/v2/membership#Instructor`
`Assistance: http://purl.imsglobal.org/vocab/lis/v2/membership#Instructor`
`TA: [http://purl.imsglobal.org/vocab/lis/v2/membership#Instructor,http://purl.imsglobal.org/vocab/lis/v2/membership/Instructor#TeachingAssistant]`

Some time a TA is instructor in certain section and just a TA in certain sections, I am assuming TA will never be a instructor in the actual course but can be instructor in a section and TA in another course section

[...]

PR #982, opened 2020-07-09
LTI 1.3 integration

pushyamig:
Fixes issues #962 and #952 

This is a basic workflow the launches the Tools and redirects to the Myla tools to `/course/:courseId` page. Here are the steps to adding LTI tool in [canvas](https://community.canvaslms.com/docs/DOC-16794-canvas-release-lti-13-and-lti-advantage-2019-06-22). I have introduced Database caching with Mysql for storing launch specific info in cache for speeding up the [validation process](https://github.com/dmitry-viskov/pylti1.3#cookies-issue-in-the-iframes) and for accessing [LTI advantage service](https://github.com/dmitry-viskov/pylti1.3#accessing-cached-launch-requests) incase we use it in Myla. This is something recommend by PyLTI1.3 Library
1. Some background on[ LTI1.3](https://www.youtube.com/watch?v=f_6pWiQpg5s&list=PLb5mG7w3UZkPKHODmz5YCkIqnWQEsjMkd)
2. configuration parameter for LTI is stored in the [box](https://umich.app.box.com/folder/117602651100) folder 
3. Updated [Wiki](https://github.com/tl-its-umich-edu/my-learning-analytics/wiki/Configuration-for-LTI-1.3) for LTI configurations
2. Current implementation of LTI both the SAML and LTI code can co-exist. 

4. Put the Public and Private key `/secrets/ directory in Docker same place as `env.json`
 "/secrets/private.key",
 "/secrets/public.key",
5. Need to enable these 2 custom parameters
```
user_username=$User.username
canvas_course_id=$Canvas.course.id
```

Things to solve:
1. - [x] Fix the launch loading time currently ~15 sec
2. - [x] Send the `launch_id` to Frontend so that When need to [look](https://github.com/dmitry-viskov/pylti1.3#accessing-cached-launch-requests) up launch data this can be used. or in other words coming up with strategies for storing the launch id
3. - [x] Enabling course viewing option for admin
4. - [x] Clean up the logic around showing multiple courses the user enrolled with LTI
5. - [x] Support both Standalone and LTI tool 
4. - [ ] ~~Support the one `iss` and multiple `client_id`Tool [config](https://github.com/dmitry-viskov/pylti1.3#configuration)~~
5. - [ ] ~~Testing become user feature for admin after enabling course viewing option~~
6. - [x] Adding the step for LTI launch for READMe or Wiki


jonespm:


```diff

Issues
======
- Added 14
           

Complexity increasing per file
==============================
- dashboard/middleware/middleware.py  9
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/ITS_Teaching_And_Learning/my-learning-analytics/pullRequest?prid=5829272&bid=19085657)

pushyamig:
Regarding the First issue. This is what the Django debug Toolbar say about the SQL Query time
Screen Shot 2020-07-10 at 10 07 06 AM
These are the SQL queries that are happening across multiple separate launches it will max to 20ms. I Don't see this being as issue at all.

dashboard/ltiv1p3.py
```diff
@@ -0,0 +1,144 @@
...
+    launch_data_storage = DjangoCacheDataStorage(cache_name='default')
+    message_launch = DjangoMessageLaunch(request, tool_conf, launch_data_storage=launch_data_storage)
+    lti_globals = validate_lti_1_3_request(request, message_launch)
+    template = loader.get_template('frontend/index.html')
+    context = {
+        'lti_globals': lti_globals
+    }
+    # return redirect(url)
+    logger.info({'lti_globals': lti_globals })
+    return HttpResponse(template.render(context, request))
```
pushyamig, line 143: Instead of redirect I opted for rending the `index.html` and I am passing all the launch related info all the globals such as username, user_image url etc.

Up until now context_processors.get_myla_globals ` was sending that info. But now that is not needed anymore

pushyamig:
Here is the Idea that's going on, after launch happens all the we store bunch of variables about `course/user/launch` in the globals. The workflow in UI is done purely by React routers and if we need any data that is specific to the view we get it via API call

lsloan:
I'm working on my review of this PR now.  Done soon…

dashboard/ltiv1p3.py
```diff
@@ -0,0 +1,144 @@
...
+    try:
+        course_details = Course.objects.get(canvas_id=course_id)
+    except ObjectDoesNotExist:
+        if is_instructor:
+            canvas_long_id = canvas_id_to_incremented_id(course_id)
+            course_details = Course.objects.create(id=canvas_long_id, canvas_id=course_id, name=course_name)
+            CourseViewOption.objects.create(course_id=canvas_long_id)
+
+    if course_details.term_id is None:
+        logger.info(f"Course {course_id} don't have a cron run yet")
```
lsloan, line 102: I'd be a little less specific.  That is, don't mention `cron` specifically.  If the job is run by Jenkins, that doesn't use `cron` at all.  Maybe something like…

```suggestion
        logger.info(f'Course ({course_id}) data loading job has not run yet.')
```

Yeah, I know it's just a log statement, but it could be used as a user-facing message, too.

I see a lot of mixing of single and double quoting in the code.  Best practices say to be consistent.  My advice is to use single quotes in most places and double quotes only when they are necessary.

dashboard/ltiv1p3.py
```diff
@@ -0,0 +1,144 @@
...
+        "primary_ui_color": settings.PRIMARY_UI_COLOR,
+        "help_url": settings.HELP_URL,
+        "google_analytics_id": settings.GA_ID
+    }
+    return lti_globals
+
+
+@csrf_exempt
+def login(request):
+    logger.info("LTI Login called...........")
```
lsloan, line 120: Not a very useful log message.  Try:

```suggestion
    logger.debug(f'request data: {request}')
```

The logging configuration should automatically include the module and method name in the log message.

dashboard/ltiv1p3.py
```diff
@@ -0,0 +1,144 @@
...
+    oidc_login = DjangoOIDCLogin(request, tool_conf, launch_data_storage=launch_data_storage)
+    target_link_uri = get_launch_url(request)
+    return oidc_login.enable_check_cookies().redirect(target_link_uri)
+
+
+@require_POST
+@method_decorator(xframe_options_exempt, name='dispatch')
+@method_decorator(csrf_exempt, name='dispatch')
+def launch(request):
+    logger.info("LTI Launch of the tool")
```
lsloan, line 132: See comment about logging in `login()` function above.

dashboard/templates/base.html
```diff
@@ -32,7 +32,8 @@
             z-index: 5000;
         }
     </style>
-    {{ myla_globals|json_script:"myla_globals" }}
+<!--    {{ myla_globals|json_script:"myla_globals" }}-->
```
lsloan, line 35: Remove.

dashboard/urls.py
```diff
@@ -88,10 +93,10 @@
...
      )
 
-if apps.is_installed('django_lti_auth'):
-    urlpatterns += (
-        path('lti/', include('django_lti_auth.urls')),
-    )
+# if apps.is_installed('django_lti_auth'):
+#     urlpatterns += (
+#         path('lti/', include('django_lti_auth.urls')),
+#     )
```
lsloan, line 99: Remove.

pushyamig:
I did some initial testing with Become user that Django admin has, It doesn't go well with LTI launch. I am not sure if we need to support the become user functionality from Myla, instead we can use it from canvas. Ofcourse the assumption that you have admin privileges in canvas. Thoughts???

pushyamig:
@lsloan Thank you for your review, Sorry that you are seeing comment code as I mentioned this is still work in progress. I wanted an early feedback so that people can look at it and have a sense of what's going on

I will go through all your comments and reply back again.

My Main concern with this PR is LTI launch time is talking lot of time. Once I figure out where are the bottle necks then I will fix all the concerns you raised with your comments.

If LTI 1.3 code gives us confidence then We might want to remove all the LTI1.1 and SAML code.

lsloan:
> …Thank you for your review, Sorry that you are seeing comment code as I mentioned this is still work in progress. I wanted an early feedback so that people can look at it and have a sense of what's going on

I see.  So this should be a draft PR.  I would not have given such a detailed review if I had known it was still in progress.  I saw the message "pushyamig marked this pull request as ready for review 5 days ago" [above](982#event-3534161685), so I thought it was time to give the PR a complete review.

Maybe this PR was started as draft, but it was accidentally marked as "ready". I'll change this PR's status back to draft.

> If LTI 1.3 code gives us confidence then We might want to remove all the LTI1.1 and SAML code.

Would it be too complicated to support LTI 1.1 and 1.3 together?  If LTI 1.1 support is to be removed, please update the descriptions of this PR and #962 to explicitly state that it will  no longer be supported.

pushyamig:
> Would it be too complicated to support LTI 1.1 and 1.3 together? If LTI 1.1 support is to be removed, please update the descriptions of this PR and #962 to explicitly state that it will no longer be supported.

It might be possible to make it as configurable option, that might complicate things

lsloan:
> I did some initial testing with Become user support from Django admin has, It doesn't go well with LTI launch. I am not sure if we need to support the become user functionality from Myla, instead we can use it from canvas. Ofcourse the assumption that you have admin privileges in canvas. Thoughts???

I don't know anything about the workflow for using the Django admin UI for this purpose.  I'd like to see a video (live or recorded) demonstration of how you want that to work.

Will MyLA be used in U-M production environments **_without_** LTI launch?  If no, then supporting user-proxy any other way is low priority.  As long as we can user-proxy in Canvas and see the correct results when MyLA is launched via LTI, that's good enough.  Anything else should be a new issue.

pushyamig:
The reason for load time primarily seems to be due to the Huge JS file that webpack is generating that has all the source code and dependency

Screen Shot 2020-07-15 at 9 54 36 AM

Screen Shot 2020-07-15 at 9 54 56 AM

I wonder why we did not notice this long load time before.

This doesn't seems to be easy problem to solve with out fine tuning the loading resources based on need of the app

lsloan:
> The reason for load time primarily seems to be due to the Huge JS file that webpack is generating that has all the source code and dependency

Is this because it is running in development debug mode?  Will it go away when that mode is disabled?

It's not exactly the same problem, but when I was using MyLA's LTI 1.1 interface, the ngrok inspection tool told me that it wouldn't show me some resources because they were too big.  I think that was one of them.

pushyamig:
> Is this because it is running in development debug mode? Will it go away when that mode is disabled?

I took out the webpack from the docker-compose and the JS bundle build is purely based on the Dockerfile  `RUN npm run prod ` and also I have Browser react [plugin](https://chrome.google.com/webstore/detail/react-developer-tools/fmkadmapgofadopljbjfkapdkoienihi?hl=en) installed that will tell if it is running in prod or dev mode

Screen Shot 2020-07-15 at 10 30 48 AM

I enabled the prod mode i React for testing this delay part

lsloan:
> I enabled the prod mode in React for testing this delay part

So, those results were with all dev. modes disabled?  If so, it seems like this is a browser caching problem.

If this is running on your local machine, why did the first screenshot show that the JS is being loaded from the cloudfront.net domain?  If browser caching is not being allowed, then going all the way out to some other domain means that your computer is effectively uploading that whole JS file to the server, then downloading it again for every request.

pushyamig:
@lsloan When I was trying to diagnose the problem I have used Chome Dev tool for doing it. I connect My Local Myla instance configured via Ngrok to Canvas test instance, Those cloud front url are from Canvas. Since Myla is loading inside canvas so the report generated all that is part of that URL which is `https://canvas.test.instructure.com/courses/333338/external_tools/332323`

The problem purely seems to be loading of static resource mainly JS file ( source code + npm libraries). It takes time to load `main.js` for the first time if the user never visited the site before once after it is cached in browser and the 2nd time visiting the page it doesn't take that much time. 

The load time is pretty pre-dominant for the first time, which is annoying for user visiting the first time and even for development as well when we keep on changing the Frontend when working on an issue

The LTI 1.3 have more back and forth between Platform(Canvas) and Tool (Myla), There is a delay that is imparted as part of the Launch itself and on top of it the static asset is taking significant amount of time.

Here is the screen shot what is happening in LTI 1.3 launch
Screen Shot 2020-06-19 at 12 06 16 PM

pushyamig:
> How do the people that developed the LTI 1.3 specification test it in their local environments?

I don't know people who implemented LTI 1.3

lsloan:
> I don't know people who implemented LTI 1.3

Should probably contact the IMS Global working group that covers LTI.

lsloan:
Attaching the Lighthouse report @pushyamig sent me in Google Chat room [tl-dev](https://chat.google.com/room/AAAArw3kwxU)…

[Lighthouse Report.pdf](https://github.com/tl-its-umich-edu/my-learning-analytics/files/4926437/Lighthouse.Report.pdf)

jonespm:
Interesting this file (main-*.js) is relatively tiny in prod though. Why is it suddenly so big as part of this PR? I don't think it has anything to do with LTI or the LTI exchange, but a file this large is a problem.

lsloan:
> I don't know people who implemented LTI 1.3

The screenshot you included [above](982#issuecomment-658845771) is from this video by Claude Vervoort:  https://youtu.be/g3y4vwtP6vQ?t=280

I met Claude at a couple of IMS Global conferences (LILI and a quarterly).  He's friendly and helpful.  He's probably willing to talk about how to develop LTI locally.  Claude has a GH repo called [claudevervoort/ltibootcamp](/claudevervoort/ltibootcamp).

I checked with IMS Global and it was suggested we talk to Bracken Mosbacker.  Bracken contributes to the [IMSGlobal/ltibootcamp](/IMSGlobal/ltibootcamp) repo (as does Claude) and has some interesting personal repos like [bracken/lti_messaging](/bracken/lti_messaging) and  [bracken/lti_example](/bracken/lti_example) (the latter he forked from Brian Whitmer, a founder of Instructure).

Those seem like good people to ask about local development of LTI support.

jonespm:
So this has nothing to do with LTI and the launch pattern. 

This is because webpack watcher is starting up in watch (or development) mode and ngrok on your local internet is a little slow. 

If you start up webpack with prod mode the file goes from 32MB down to 800KB. This still results in an uncached initial launch of a few seconds but it shouldn't be noticeable once this is on openshift.

We should [look into these tips](https://medium.com/@rajaraodv/two-quick-ways-to-reduce-react-apps-size-in-production-82226605771a) to possibly reduce the size of this file even further but this should be on a separate issue.

```
diff --git a/docker-compose.yml b/docker-compose.yml
index 6286c99..06daa09 100644
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -43,7 +43,7 @@ services:
     build:
       context: .
       target: node-webpack
-    command: sh -c "tar zxvf node_modules/all_symlinks.tgz && npm run watch"
+    command: sh -c "tar zxvf node_modules/all_symlinks.tgz && npm run prod"
     volumes:
       - .:/usr/src/app:delegated
```

pushyamig:
@jonespm Thanks! That was the first thing that occurred to me that React Dev mode may be is the trouble for slowness, But My attempt to have a local Prod mode was not rightly done I guess that resulted in the huge .js file. Even the browser plugin for checking the react prod vs dev was deceiving. May be canvas is written in React that's why it showed me.  

> If you start up webpack with prod mode the file goes from 32MB down to 800KB. This still results in an uncached initial launch of a few seconds but it shouldn't be noticeable once this is on openshift.

This is good news and I can mark this issue with slowness here as check.

> command: sh -c "tar zxvf node_modules/all_symlinks.tgz && npm run prod"

This is a nice trick!

jonespm:


```diff

Issues
======
- Added 2
           

Complexity increasing per file
==============================
- assets/src/containers/App.js  1
- dashboard/migrations/0016_adding_mysql_cache.py  1
- dashboard/ltiv1p3.py  8
- dashboard/middleware/samesite.py  7
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/ITS_Teaching_And_Learning/my-learning-analytics/pullRequest?prid=5829272&bid=19085657)

pushyamig:
> Misleading comment: Correct or remove it.
Screen Shot 2020-07-23 at 9 57 54 AM

This PR did not introduce the change, Sam added that comment for whatever reason. So i am not going to change it here

pushyamig:
> Not sure I understand the logic here: why the first round of role check include both Instructor and Teaching Assistant; and then further filter out the TeachingAssistant role next?

Why not do the following one check only:

@zqian The problems is that, A TA can be instructor in the course sections, but never a instructor in the main lecture course. LTI launch will send all the course role for this person that include other sections.

For Eg, here is the list of roles of TA in course, but if course has multiple sections and LTI will send all his roles in the sections a s well, that's the reason why I need to filter out by TA and instructor and figure out if TA or Instructor

```
[http://purl.imsglobal.org/vocab/lis/v2/institution/membership#Instructor,
http://purl.imsglobal.org/vocab/lis/v2/institution/membershipmembership/Instructor#TeachingAssistant,
http://purl.imsglobal.org/vocab/lis/v2/institution/membership#Instructor,
http://purl.imsglobal.org/vocab/lis/v2/institution/membershipmembership/Instructor#TeachingAssistant,
http://purl.imsglobal.org/vocab/lis/v2/institution/membership#Instructor,]
```

lsloan:
> > Misleading comment: Correct or remove it.
> 
> This PR did not introduce the change, Sam added that comment for whatever reason So i don't want to change it

If I'm not mistaken, you changed the code that follows the comment.  Your changes make the comment misleading.

jonespm:


```diff

Issues
======
+ Solved 2
- Added 1
           

Complexity increasing per file
==============================
- assets/src/containers/App.js  1
- dashboard/middleware/samesite.py  7
- dashboard/migrations/0016_adding_mysql_cache.py  1
- dashboard/ltiv1p3.py  8
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/ITS_Teaching_And_Learning/my-learning-analytics/pullRequest?prid=5829272&bid=19085657)

config/env_sample.json
```diff
@@ -85,19 +85,40 @@
...
-    "/*  LTI Course ID field.": "*/",
-    "LTI_CANVAS_COURSE_ID_FIELD": "custom_canvas_course_id",
-    "/*  LTI First Name.": "*/",
-    "LTI_FIRST_NAME": "lis_person_name_given",
-    "/*  LTI Last Name.": "*/",
-    "LTI_LAST_NAME": "lis_person_name_family",
+    "/*LTI 1.3 configuration":"*/",
+    "LTI_CONFIG": {
+        "/*canvas LMS domain nanme":"*/",
+        "https://canvas.test.instructure.com": [
```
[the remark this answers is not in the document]
lsloan: Depending on the point of view, that's wrong, too.

Do we want the actual hostname for UMich's Canvas or do we want to use fake names in this sample file?  If actual, `michedu` should be changed to `umich`.  Otherwise, it should be changed to something like `example` or `school_name_here`.

lsloan:
I'm doing a review now.  Will be done soon.

dashboard/ltiv1p3.py
```diff
@@ -0,0 +1,203 @@
...
+        logger.info(f"user {username} is not instructor in course {course_id}")
+        return False
+
+    ta = [teach_role for teach_role in teaching_roles if 'membership/Instructor#TeachingAssistant' in teach_role]
+    if len(ta) != 0:
+        logger.info(f'user {username} is not an instructor of course {course_id}')
+        return False
+    else:
+        logger.info(f'user {username} is an instructor in course {course_id}')
+        return True
```
lsloan, line 68: The logic doesn't seem right to me.  Doesn't Canvas allow users to be instructor and TA for a course at the same time?  If so, this function returns `False` for anyone who has both roles.

Since this function is named `check_if_instructor`, it should look for the explicit instructor role string in their list of roles and return a Boolean value for whether it was found.

dashboard/ltiv1p3.py
```diff
@@ -0,0 +1,203 @@
...
+        return user_roles
+
+    for role in course_roles:
+        split_ = role.split('#')
+        user_roles.append(split_[1])
+    logger.info(f'User role in course {user_roles}')
+    # this step eliminates duplicates of multiple same roles in the course Like TA can be instructor in multiple sections
+    user_roles = list(dict.fromkeys(user_roles))
+    # A TA can be instructor in other sections of the course but not in Lecture course so default to single role
+    user_roles = ['TeachingAssistant'] if not 'TeachingAssistant' else user_roles
```
lsloan, line 87: This looks very wrong.

Negating a static, non-empty string like `'TeachingAssistant'` **_always_** evaluates to `False`.  So the part of the statement following `else` will **_always_** be used.  This is equivalent to:

```
    user_roles = user_roles
```

dashboard/management/commands/createuser.py
```diff
@@ -1,5 +1,6 @@
 from django.core.management.base import BaseCommand
 from django.contrib.auth.models import User
+from django.conf import settings
 import random, string
 
 RANDOM_PASSWORD_DEFAULT_LENGTH = 8
```
lsloan, line 6: Delete this line.  With the new import, the old `RANDOM_PASSWORD_DEFAULT_LENGTH` is no longer used.

dashboard/urls.py
```diff
@@ -18,24 +18,27 @@
...
+from django.conf.urls import include
 from django.conf.urls.static import static
 from django.urls import path, re_path
 
 from dashboard.graphql.view import DashboardGraphQLView
 
 from django.views.decorators.cache import cache_page
 
-from . import views
+from . import views, ltiv1p3
```
lsloan, line 29: Again, I'm concerned about naming our module for a specific version of the standard, `ltiv1p3`.

requirements.txt
```diff
@@ -1,18 +1,18 @@
-gunicorn>=19.9.0,<19.9.99
```
lsloan, line 1: GH says there are conflicts in this file that need to be resolved.

dashboard/ltiv1p3.py
```diff
@@ -0,0 +1,203 @@
...
+
+
+def get_jwks(request):
+    try:
+        tool_conf = get_tool_conf()
+    except (OSError, Exception) as e:
+        error_message = {
+            'lti_error_message': f'Fetching JWKS failed due to {e}'
+        }
+        return JsonResponse(error_message)
```
lsloan, line 51:
```suggestion
        return JsonResponse(error_message, status=500)
```

dashboard/ltiv1p3.py
```diff
@@ -0,0 +1,203 @@
...
+        return True
+
+
+def role_check(roles):
+    course_roles = [role for role in roles if 'membership' in role]
+    logger.info(course_roles)
+    user_roles = []
+    if len(course_roles) == 0:
+        institute_roles = [role for role in roles if "institution" in role]
+        logger.info(f'user dont seems to have role in Course must be admin {institute_roles}')
```
lsloan, line 77:
```suggestion
        institute_roles = [role for role in roles if 'institution' in role]
        logger.info(f"User doesn't have course membership role, must be admin {institute_roles}")
```

dashboard/ltiv1p3.py
```diff
@@ -0,0 +1,203 @@
...
+        return user_roles
+
+    for role in course_roles:
+        split_ = role.split('#')
+        user_roles.append(split_[1])
+    logger.info(f'User role in course {user_roles}')
+    # this step eliminates duplicates of multiple same roles in the course Like TA can be instructor in multiple sections
+    user_roles = list(dict.fromkeys(user_roles))
```
lsloan, line 85: Even with the comment, this is confusing logic.

Rather than using a list for `user_roles` above, it would make more sense (and probably be more efficient) to use a set for it instead.  Instead of using `append` to the list, you would `add` to the set.  Sets automatically keep only one of each member.  If you try to add a duplicate to a set, it will be ignored without error.

dashboard/ltiv1p3.py
```diff
@@ -0,0 +1,203 @@
...
+    course_name = launch_data['https://purl.imsglobal.org/spec/lti/claim/context']['title']
+    roles = launch_data['https://purl.imsglobal.org/spec/lti/claim/roles']
+    username = custom_params['user_username']
+    course_id = custom_params['canvas_course_id']
+    email = launch_data['email']
+    first_name = launch_data['given_name']
+    last_name = launch_data['family_name']
+    user_img = launch_data['picture']
+
+    # Logging the user regardless course exits or not otherwise Django redirect to login page
```
lsloan, line 108: Confusing comment.
```suggestion
    # Add user to DB if not there; avoids Django redirection to login page
```

dashboard/ltiv1p3.py
```diff
@@ -0,0 +1,203 @@
...
+    try:
+        course_details = Course.objects.get(canvas_id=course_id)
+    except ObjectDoesNotExist:
+        if is_instructor or user_obj.is_staff:
+            canvas_long_id = canvas_id_to_incremented_id(course_id)
+            course_details = Course.objects.create(id=canvas_long_id, canvas_id=course_id, name=course_name)
+            CourseViewOption.objects.create(course_id=canvas_long_id)
+
+    if course_details is None:
+        logger.info(f'Course {course_id} dont have a cron run yet')
```
lsloan, line 133:
```suggestion
        logger.info(f'Course {course_id} data has not yet been loaded')
```

dashboard/ltiv1p3.py
```diff
@@ -0,0 +1,203 @@
...
+    elif course_details.term_id is not None:
+        logger.info(f'Course {course_id} is ready with Data')
```
lsloan, line 135:
```suggestion
        logger.info(f'Course {course_id} data is ready')
```

dashboard/ltiv1p3.py
```diff
@@ -0,0 +1,203 @@
...
+
+@csrf_exempt
+def login(request):
+    try:
+        tool_conf = get_tool_conf()
+    except (OSError, Exception) as e:
+        error_message = {
+            'lti_error_message': f'LTI Login failed due to {e}'
+        }
+        return JsonResponse(error_message)
```
lsloan, line 163:
```suggestion
        return JsonResponse(error_message, status=500)
```

dashboard/ltiv1p3.py
```diff
@@ -0,0 +1,203 @@
...
+
+    target_link_uri = request.POST.get('target_link_uri', request.GET.get('target_link_uri'))
+    if not target_link_uri:
+        error_message = {
+            'lti_error_message': 'LTI Login failed due to missing "target_link_uri" param'
+        }
+        return JsonResponse(error_message)
```
lsloan, line 170:
```suggestion
        return JsonResponse(error_message, status=500)
```

dashboard/ltiv1p3.py
```diff
@@ -0,0 +1,203 @@
...
+@require_POST
+@csrf_exempt
+def launch(request):
+    try:
+        tool_conf = get_tool_conf()
+    except (OSError, Exception) as e:
+        error_message = {
+            'lti_error_message': f'LTI Launch failed due to {e}'
+        }
+        return JsonResponse(error_message)
```
lsloan, line 185:
```suggestion
        return JsonResponse(error_message, status=500)
```

dashboard/ltiv1p3.py
```diff
@@ -0,0 +1,203 @@
...
+    cache_ttl = settings.DB_CACHE_CONFIGS['CACHE_TTL']
+    cache_lifetime = cache_ttl if cache_ttl else 7200
+    message_launch.set_public_key_caching(launch_data_storage, cache_lifetime=cache_lifetime)
+    try:
+       myla_globals = extracting_launch_variables_for_tool_use(request, message_launch)
+    except Exception as e:
+        error_message = {
+            'lti_error_message': f'LTI Launch failed due to {e}'
+        }
+        return JsonResponse(error_message)
```
lsloan, line 198:
```suggestion
        return JsonResponse(error_message, status=500)
```

lsloan:
> > Not sure I understand the logic here: why the first round of role check include both Instructor and Teaching Assistant; and then further filter out the TeachingAssistant role next?
> 
> Why not do the following one check only:
> 
> @zqian The problems is that, A TA can be instructor in the course sections, but never a instructor in the main lecture course. LTI launch will send all the course role for this person that include other sections.
> 
> For Eg, here is the list of roles of TA in course, but if course has multiple sections and LTI will send all his roles in the sections a s well, that's the reason why I need to filter out by TA and instructor and figure out if TA or Instructor
> 
> ```
> [http://purl.imsglobal.org/vocab/lis/v2/institution/membership#Instructor,
> http://purl.imsglobal.org/vocab/lis/v2/institution/membershipmembership/Instructor#TeachingAssistant,
> http://purl.imsglobal.org/vocab/lis/v2/institution/membership#Instructor,
> http://purl.imsglobal.org/vocab/lis/v2/institution/membershipmembership/Instructor#TeachingAssistant,
> http://purl.imsglobal.org/vocab/lis/v2/institution/membership#Instructor,]
> ```

I agree with @zqian that the logic looks wrong.  At the very least, if it's not wrong, the implementation is confusing and not straightforward.  Using the example role data you've given, you can construct a set of unique roles from it like this:

```python
role_strip_chars = '[]\n'
roles = set([e.strip(role_strip_chars) for e in role_data.split(',') \
  if e.strip(role_strip_chars)])
```

Then, to check whether the instructor role is present:

```python
role_instructor = 'http://purl.imsglobal.org/vocab/lis/v2/institution/membership#Instructor`
print(role_instructor in roles)  # prints True or False
```

pushyamig:
@lsloan I have resolved all the comment you had. I improved the roles checking logic and checking for instructor logic and I have also resolved the conflict in this PR

jonespm:


```diff

Issues
======
+ Solved 2
           

Complexity increasing per file
==============================
- dashboard/middleware/samesite.py  7
- dashboard/migrations/0016_adding_mysql_cache.py  1
- dashboard/lti_new.py  8
- assets/src/containers/App.js  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/ITS_Teaching_And_Learning/my-learning-analytics/pullRequest?prid=5829272&bid=19085657)

dashboard/settings.py
```diff
@@ -288,10 +288,10 @@
     },
 }
 RANDOM_PASSWORD_DEFAULT_LENGTH = ENV.get('RANDOM_PASSWORD_DEFAULT_LENGTH', 32)
-DB_CACHE_CONFIGS = ENV.get('DB_CACHE_CONFIGS', {"CACHE_TTL": 7200, "BACKEND": "django_mysql.cache.MySQLCache",
-                                                "LOCATION": "django_myla_cache",
-                                                "CACHE_KEY_PREFIX": "myla",
-                                                "CACHE_OPTIONS": {"COMPRESS_MIN_LENGTH": 5000, "COMPRESS_LEVEL": 6}
+DB_CACHE_CONFIGS = ENV.get('DB_CACHE_CONFIGS', {'CACHE_TTL': 7200, 'BACKEND': 'django_mysql.cache.MySQLCach',
```
lsloan, line 291: Typo in the module name, `MySQLCache`.  I think this would crash if run as it is.

```suggestion
DB_CACHE_CONFIGS = ENV.get('DB_CACHE_CONFIGS', {'CACHE_TTL': 7200, 'BACKEND': 'django_mysql.cache.MySQLCache',
```

I really don't like the popular Python style of using strings to refer to modules or other code symbols.    This typo demonstrates why.  The developer's IDE or the Python interpreter didn't catch this problem because the name of the module is in a string.

I think it'd be better if this could be done using code like `django_mysql.cache.MySQLCache.__name__` if a string is required, or just `django_mysql.cache.MySQLCache` if a reference to the module is acceptable.  Each of these may require importing the module first.

config/env_sample.json
```diff
@@ -89,17 +89,17 @@
     "STUDENT_DASHBOARD_LTI": false,
     "/*LTI 1.3 configuration":"*/",
     "LTI_CONFIG": {
-        "/*canvas LMS domain nanme":"*/",
+        "/*canvas LMS domain name":"*/",
         "https://canvas.test.instructure.com": [
```
lsloan, line 93: I think @zqian meant for this line, `"https://canvas.test.instructure.com"`, to be changed.
```suggestion
        "https://umich.test.instructure.com": [
```

dashboard/lti_new.py
```diff
@@ -16,19 +16,22 @@
 
 
 logger = logging.getLogger(__name__)
+INSTRUCTOR = 'Instructor'
+TA = 'TeachingAssistant'
+COURSE_MEMBERSHIP = 'membership'
```
lsloan, line 21: This is good.  I was thinking that these values should be in constants/variables.  They'd be helpful if the names of the roles ever change in the future.  I was hoping that maybe the "pylti1p3" module would have these defined already.

Ideally, the full role URIs would be used here.  It's better to be explicit.  We don't know whether Instructure will start sending some new role URI, like `http://instructure.com/role/membership#FUBAR-Instructor`, which we don't want, but would match the simple `Instructor` string this code is looking for.

dashboard/settings.py
jonespm: Yeah, that only gets you the name of the module though "MySQLCache". 

It looks like this is a "Django-ism" and [they have a pattern for this.](https://stackoverflow.com/a/33627620/3708872)

```
from django.utils.module_loading import import_string
my_module = import_string('django_mysql.cache.MySQLCache')
```

If it's successful my_module just has the name of the module, if not it throws an ImportError. So we *could* import this package and replace this string with `import_string('django_mysql.cache.MySQLCache')` for safety.

jonespm: And yeah, I was finishing up and reviewing some other changes today but I have this all checked out and ready to go for tomorrow morning.

pushyamig:
> I think @zqian meant for this line, "https://canvas.test.instructure.com", to be changed.

That needs to be like that. Otherwise LTI validation with fail. Initially when I was changed it to umich.instructure.com validation failed so that's how I learned that needs to stay like that.

Where are other canvas url doesn't need to be like that way. So I preferred umich.

pushyamig:
> I think it'd be better if this could be done using code like django_mysql.cache.MySQLCache.__name__ if a string is required, or just django_mysql.cache.MySQLCache if a reference to the module is acceptable. Each of these may require importing the module first.

I checked, I did not find something that emit full path to the class. I accepted your answer

pushyamig:
>This is good. I was thinking that these values should be in constants/variables. They'd be helpful if the names of the roles ever change in the future. I was hoping that maybe the "pylti1p3" module would have these defined already.

I don't see [PyLTI1.3](https://github.com/dmitry-viskov/pylti1.3/tree/master/pylti1p3) defined constants as such

> Ideally, the full role URIs would be used here. It's better to be explicit. We don't know whether Instructure will start sending some new role URI, like http://instructure.com/role/membership#FUBAR-Instructor, which we don't want, but would match the simple Instructor string this code is looking for.

Done

pushyamig:
> If it's successful my_module just has the name of the module, if not it throws an ImportError. So we could import this package and replace this string with import_string('django_mysql.cache.MySQLCache') for safety.

Done

dashboard/settings.py
lsloan: I forgot Python's `__name__` just gives the name of the module (or the "leaf node"), not the full hierarchy.

What I'm suggesting is the opposite of Django's `import_string()`.  Import a module so that it can be checked by the IDE AND will throw exceptions at runtime if it's not found.  When the module is imported, have a way to get its fully qualified name as a string to use in places where it's needed.

Or maybe an approach could be to import the module and reference it instead of putting its name in a string.  I'm not sure whether Django (or the DB module) supports that, but it would be a shame if it didn't.

dashboard/lti_new.py
```diff
@@ -0,0 +1,198 @@
...
+    if len(user_roles_in_course) == 0:
+        logger.info(f'User {username} do not have course membership role, must be admin {roles}')
+        return list()
+    short_role_str_list = set()
+    for role in user_roles_in_course:
+        short_role_str_list.add(role)
+    if INSTRUCTOR in short_role_str_list and TA in short_role_str_list:
+        logger.info(f'User {username} is a {TA} in the course')
+        short_role_str_list.remove(INSTRUCTOR)
+    return list(short_role_str_list)
```
lsloan, line 76: Since you've added the full role URIs and you no longer need to split the roles on "`#`", this can now be simplified to:

```suggestion
    user_membership_roles = set([role for role in roles if COURSE_MEMBERSHIP in role])
    if not user_membership_roles:
        logger.info(f'User {username} does not have course membership roles, must be admin {roles}')
    elif INSTRUCTOR in user_membership_roles and TA in user_membership_roles:
        logger.info(f'User {username} is a {TA} in the course')
        user_membership_roles.remove(INSTRUCTOR)
    return user_membership_roles
```

dashboard/lti_new.py
```diff
@@ -0,0 +1,198 @@
...
+from dashboard.common.db_util import canvas_id_to_incremented_id
+from django.core.exceptions import ObjectDoesNotExist
+from dashboard.models import Course, CourseViewOption
+from django.conf import settings
+
+
+logger = logging.getLogger(__name__)
+INSTRUCTOR = 'http://purl.imsglobal.org/vocab/lis/v2/membership#Instructor'
+TA = 'http://purl.imsglobal.org/vocab/lis/v2/membership/Instructor#TeachingAssistant'
+COURSE_MEMBERSHIP = 'membership'
```
lsloan, line 21: As the others are now explicit, make `COURSE_MEMBERSHIP` explicit, too.  We never know whether there will be some other "membership" URIs in the future, ones that we might not want to match.

```suggestion
COURSE_MEMBERSHIP = 'http://purl.imsglobal.org/vocab/lis/v2/membership'
```

config/env_sample.json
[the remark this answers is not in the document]
jonespm: I'd set all of these to be a generic and a comment that you need to fill this out for your institution. 

`https://canvas.instructure.com/`

dashboard/settings.py
lsloan: > If you look at the setting.py all modules are defined as strings, in INSTALLED_APPS, MIDDLEWARE, DATABASES.

Yeah, I hate that, too.  😆  I suppose they probably have a good reason for doing it this way.  🤞 

> I think this is just a convenience to avoid having to import every module in the settings. This really isn't any different from those.

Maybe this isn't the place to discuss, but why would importing the modules be an **_in_**-convenience?

> I think we just have to fix typos as they will come up in these cases and don't really see a need to break from the pre-existing patterns.

I begrudgingly agree that it's probably not the time to challenge established Django patterns. ¯\_(ツ)_/¯

config/env_sample.json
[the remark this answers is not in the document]
lsloan: `https://example.canvas.instructure.com/`, maybe?

dashboard/lti_new.py
```diff
@@ -0,0 +1,198 @@
+import logging, string, random
```
jonespm, line 1: I'd like to break this up into 

generic lti code
myla lti (that extends this generic code) 

But we can leave it like this for now. 

I'd probably rename this to something like myla_lti_1p3.py

dashboard/migrations/0016_adding_mysql_cache.py
```diff
@@ -0,0 +1,23 @@
+from django.db import migrations
```
jonespm, line 1: This migration file will need to be renamed up to 19
16 is used by #993
17 is used by #981
18 is going to be used by #995

dashboard/migrations/0016_adding_mysql_cache.py
```diff
@@ -0,0 +1,22 @@
...
+    ]
+
+    operations = [
+        migrations.RunSQL(
+            """
+            CREATE TABLE `django_myla_cache` (
+                cache_key varchar(255) CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci
+                                       NOT NULL PRIMARY KEY,
+                value longblob NOT NULL,
+                value_type char(1) CHARACTER SET latin1 COLLATE latin1_bin
```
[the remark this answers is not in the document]
jonespm: Yeah, for char(1) it probably won't matter and [this is what was recommended by the code](https://django-mysql.readthedocs.io/en/latest/cache.html) but should be `utf8mb4_unicode_ci` to be consistent.

It looks like it's fine if it just depends on the 0001 script, doesn't have to be on the 0015 script.

dashboard/lti_new.py
lsloan: I'm not enthusiastic about the version number in the module name.  However, I agree that it'd be good to separate generic and MyLA-specific code somewhat.  Do you mean different modules, or just the organization within this module?

Besides the pylit1p3 module, aren't there modules around that define important things like LTI role URIs?  Doesn't IMS Global make something like this available?

dashboard/settings.py
```diff
@@ -282,7 +288,23 @@
...
-
+RANDOM_PASSWORD_DEFAULT_LENGTH = ENV.get('RANDOM_PASSWORD_DEFAULT_LENGTH', 32)
+DB_CACHE_CONFIGS = ENV.get('DB_CACHE_CONFIGS',
+                           {'CACHE_TTL': 7200, 'BACKEND': import_string('django_mysql.cache.MySQLCache'),
+                            'LOCATION': 'django_myla_cache',
+                            'CACHE_KEY_PREFIX': 'myla',
+                            'CACHE_OPTIONS': {'COMPRESS_MIN_LENGTH': 5000, 'COMPRESS_LEVEL': 6}
+                            })
+
+CACHES = {
```
jonespm, line 299: It doesn't look like there's any easy way to disable this cache. I'm not sure if that's something we'll need to do.

When I setup Student Explorer [I had DummyCache as the default](https://github.com/tl-its-umich-edu/student_explorer/blob/2d5445a7b5ec0f7da8e634d7676c67a464ca4404/student_explorer/settings.py#L387). This might work better, but then someone might not configure it. 

This is probably fine though as a default.

dashboard/lti_new.py
jonespm: Yeah it could probably be called lti_advantage rather than have this specific version. 

AFIAK IMS doesn't support any python modules aside from the Caliper one. 

In a perfect situation there probably would be
* A separate project (or at least module) that has all these LTI constants
* A separate project that just has the LTI Advantage specific code for Django (like we had for 1.1)

The author of the library hasn't done this yet. I think we could split this out to only need to be included as INSTALLED_APPS [like the 1.1 code was](https://github.com/ccnmtl/django-lti-provider) but I haven't seen anyone that did that. The project author just had an [example with everything together.](https://github.com/dmitry-viskov/pylti1.3-django-example)

dashboard/lti_new.py
```diff
@@ -0,0 +1,198 @@
...
+    short_role = []
+    for role in roles:
+        short_role.append(role.split('#')[1])
+    return short_role
+
+
+def extracting_launch_variables_for_tool_use(request, message_launch):
+    launch_id = message_launch.get_launch_id()
+    launch_data = message_launch.get_launch_data()
+    custom_params = launch_data['https://purl.imsglobal.org/spec/lti/claim/custom']
```
jonespm, line 89: I think it would be useful here to have some logger.debug for custom_params and launch_data. I locally enabled this.

dashboard/lti_new.py
```diff
@@ -0,0 +1,198 @@
...
+    if not custom_params:
+        raise Exception(
+            f'Custom param LTI configuration [user_username=$User.username,canvas_course_id=$Canvas.course.id] not set'
```
jonespm, line 92: Could this message be a little clearer? I had to add this setup step and wasn't sure why it was failing.

Something like.

"You need have custom parameters configured on your LTI Launch. Please see the LTI installation guide on the Github Wiki for more information."

jonespm:


```diff

Issues
======
+ Solved 2
           

Complexity increasing per file
==============================
- dashboard/middleware/samesite.py  7
- dashboard/lti_new.py  8
- assets/src/containers/App.js  1
- dashboard/migrations/0016_adding_mysql_cache.py  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/ITS_Teaching_And_Learning/my-learning-analytics/pullRequest?prid=5829272&bid=19085657)

dashboard/urls.py
```diff
@@ -18,24 +18,27 @@
...
 import watchman.views
-
 urlpatterns = [
     path('', views.get_home_template, name = 'home'),
     path('status/', include('watchman.urls')),
     path('status/bare_status/', watchman.views.bare_status),
 
+
     path('admin/', admin.site.urls),
+    path('login/', lti_new.login, name="login"),
```
jonespm, line 39: I know it's a little late in the review but the previous LTI had these under /lti/*.

Would it be possible to move these in there like

```
/lti/login/
/lti/launch/
/lti/jwks/

```
Having them on the top level might conflict with something now or later, as they are semi-generic paths. I *feel* like the biggest change from this will be in the documentation on the wiki.

jonespm: And these should probably only be available if LTI is enabled. You could just check the settings for that. On localhost and dev we should still support local login.

pushyamig:
> And these should probably only be available if LTI is enabled. You could just check the settings for that. On localhost and dev we should still support local login.

I have changed this as suggested [`46264a0` (#982)](https://github.com/tl-its-umich-edu/my-learning-analytics/pull/982/commits/46264a07ce77cb4915e5b3b9b6e084837b5869b8)

dashboard/lti_new.py
```diff
@@ -0,0 +1,196 @@
...
+        logger.info(f'User {username} is a {TA} in the course')
+        user_membership_roles.remove(INSTRUCTOR)
+    return user_membership_roles
+
+
+def short_user_role_list(roles):
+    short_role = []
+    for role in roles:
+        short_role.append(role.split('#')[1])
+    return short_role
```
lsloan, line 80: I meant to suggest this earlier…

```suggestion
def short_user_role_list(roles):
    return [role.split('#')[1] for role in roles]
```

jonespm:


```diff

Issues
======
+ Solved 2
- Added 1
           

Complexity increasing per file
==============================
- dashboard/lti_new.py  8
- dashboard/migrations/0018_adding_mysql_cache.py  1
- assets/src/containers/App.js  1
- dashboard/middleware/samesite.py  7
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/ITS_Teaching_And_Learning/my-learning-analytics/pullRequest?prid=5829272&bid=19085657)

dashboard/lti_new.py
```diff
@@ -0,0 +1,196 @@
...
+    myla_globals = {
+        "username": username,
+        "is_superuser": user_obj.is_staff,
+        "login": settings.LOGIN_URL,
+        "logout": settings.LOGOUT_URL,
+        "user_image": user_img,
+        "primary_ui_color": settings.PRIMARY_UI_COLOR,
+        "help_url": settings.HELP_URL,
+        "google_analytics_id": settings.GA_ID,
+        "user_courses_info": [{"course_id": course_id, "course_name": course_name, "enrollment_type": short_roles_list}],
```
lsloan, line 141: `short_roles_list` is only used here, so remove it and use the function call here.

```suggestion
        'user_courses_info': [{
            'course_id': course_id,
            'course_name': course_name,
            'enrollment_type': short_user_role_list(user_roles)
        }],
```

dashboard/lti_new.py
```diff
@@ -0,0 +1,196 @@
...
+        user_obj = User.objects.get(username=username)
+    except User.DoesNotExist:
+        password = ''.join(random.sample(string.ascii_letters, settings.RANDOM_PASSWORD_DEFAULT_LENGTH))
+        user_obj = User.objects.create_user(username=username, email=email, password=password, first_name=first_name,
+                                            last_name=last_name)
+    user_obj.backend = 'django.contrib.auth.backends.ModelBackend'
+    django.contrib.auth.login(request, user_obj)
+    user_roles = course_user_roles(roles, username)
+    is_instructor = check_if_instructor(user_roles, username, course_id)
+    short_roles_list = short_user_role_list(user_roles)
```
lsloan, line 113: This goes along with my [other suggestion](https://github.com/tl-its-umich-edu/my-learning-analytics/pull/982/files#r463757435):

```suggestion
```

pushyamig:
@jonespm @lsloan I have resolved the changes requested, I am sorry if i missed anything. Lot of comment.

README.md
```diff
@@ -85,29 +85,12 @@ MyLA allows you to configure the primary color of the user interface (i.e., the
...
-Only basic LTI launches are supported at the moment (automatic account creation and redirection to the correct course). New courses are not added nor are course view options modified.
-
-The relative LTI launch url is `/lti/auth/` (ex: `https://example.com/lti/auth/`). *NOTE* The trailing slash after `/auth/` is required! You'll get an error about CSRF if you have the wrong URL! :)
-
-#### Canvas Notes
-* When you configure the tool in Canvas the "Privacy" option must be changed from Anonymous to Public to pass along the user information required.
-
-* You should use the [XML builder](https://www.edu-apps.org/build_xml.html) to generate and paste the XML that rather than manually adding it. This will allow options like "Course Navigation" extension.
+### LTI v1.3 Configuration
+* MyLA Supports LTI 1.3 and using [pylti1.3](https://github.com/dmitry-viskov/pylti1.3) library for LTI validation
```
lsloan, line 89: Is "validation" the right word?  Isn't it authentication, authentication, or handling?  Or maybe it doesn't need to be that specific, just remove the "library for LTI validation" part altogether.

dashboard/lti_new.py
```diff
@@ -0,0 +1,197 @@
...
+        }
+        return JsonResponse(error_message, status=500)
+    dummy_cache = DUMMY_CACHE in settings.DB_CACHE_CONFIGS['BACKEND']
+    launch_data_storage = DjangoCacheDataStorage(cache_name='default') if not dummy_cache else None
+    message_launch = DjangoMessageLaunch(request, tool_conf, launch_data_storage=launch_data_storage)
+    if not dummy_cache:
+        cache_ttl = settings.DB_CACHE_CONFIGS['CACHE_TTL']
+        cache_lifetime = cache_ttl if cache_ttl else 7200
+        # fetch platform's public key from cache instead of calling the API will speed up the launch process
+        message_launch.set_public_key_caching(launch_data_storage, cache_lifetime=cache_lifetime)
```
lsloan, line 185: The first two lines of this code block also appear in `login()`.  Maybe the repeated code can be reduced.

Try moving this code block to a new function that returns `[dummy_cache, launch_data_storage, cache_lifetime]`.  Both `login()` and `launch()` can call this new function and use the returned values that's needed in each place.  (`login()` would use only `launch_data_storage` and `launch()` would use that and `cache_lifetime`.  Each of them could use `dummy_cache`, too.)

dashboard/lti_new.py
```diff
@@ -0,0 +1,197 @@
...
+@require_POST
+@csrf_exempt
+def launch(request):
+    try:
+        tool_conf = get_tool_conf()
+    except (OSError, Exception) as e:
+        error_message = {
+            'lti_error_message': f'LTI Launch failed due to {e}'
+        }
+        return JsonResponse(error_message, status=500)
```
lsloan, line 177: This repeated code appears in both `login()` and `launch()` and can be refactored into a new function.

dashboard/lti_new.py
```diff
@@ -0,0 +1,197 @@
...
+        cache_lifetime = cache_ttl if cache_ttl else 7200
+        # fetch platform's public key from cache instead of calling the API will speed up the launch process
+        message_launch.set_public_key_caching(launch_data_storage, cache_lifetime=cache_lifetime)
+    try:
+       myla_globals = extracting_launch_variables_for_tool_use(request, message_launch)
+    except Exception as e:
+        error_message = {
+            'lti_error_message': f'LTI Launch failed due to {e}'
+        }
+        return JsonResponse(error_message, status=500)
```
lsloan, line 192: This section is included in the one repeated above, but it's used under different circumstances. (Not trying to use `get_tool_conf()` here.)   I think this error message and status code could be refactored into a new class based on `JsonResponse`.  It could be a subclass or composition.  The new class would let you simply call it as:

```python
return LtiError(lti_exception)
```

It should handle the message formatting and set the status code.

pushyamig:
> This repeated code appears in both login() and launch() and can be refactored into a new function.
@lsloan I was able to resolve this.

pushyamig:
> The new class would let you simply call it as:
return LtiError(lti_exception)

Done. I took out the LTI error as a separate class.

pushyamig:
@jonespm @lsloan I would like to merge this today, if you don't see lot of concerns with the code something minor I can resolve it as separate issue. please take a look at this in the morning.

dashboard/lti_new.py
```diff
@@ -24,12 +24,23 @@
 DUMMY_CACHE = 'DummyCache'
 
 
+class LTIError:
```
lsloan, line 27: When I suggested moving this error message to a new class, I had in mind that the new class would extend `Exception` and it could be invoked with a `raise e` call.  If it's working the way it's currently coded, then it's OK to leave it as is for now.

A new issue could be opened to handle this.

pushyamig:
@jonespm Let me know if you approve as well

jonespm:
Yeah, like I said at the meeting it looks like it's working. There are probably some things we could improve and things we might still need to fix but lets merge and test it.

Issue #768, opened 2019-10-23
User logging in throws error if course they were enrolled in has been deleted but cron was not run again


ssciolla:
Thank you for contributing to this project!

- Make sure to search the [issues](https://github.com/tl-its-umich-edu/my-learning-analytics/issues) for duplicates first!

## Expected behavior (A clear and concise description of what you expected to happen) : 

A user should be able to log in to MyLA without an error being thrown, regardless of their prior course affiliations or the state of the course list in the backend.

## Describe the bug (Tell us what happens instead of the expected behavior) :

@jonespm and I noticed when logging in to the test instance of MyLA via Shibboleth that an error will be thrown if 1) the user is part of a course that was recently deleted from the database and 2) the cron job has not yet been run with the updated list of courses.

@jennlove-um, I'm `@`ing you because I'd like you to be aware of this bug. Some recent maintenance may have caused this situation, which is good because now we know about it!

## Steps to Reproduce :
1. Add a course to an instance of MyLA and run the cron.
2. Remove that course from MyLA, without re-running the cron
3. Attempt to log in as a student in the deleted course

jennlove-um:
Testing passes: 
1. Add a course to MyLA and run cron
2. Access course as a student to verify it can be accessed and data is there
3. Delete course from MyLA and DO NOT run cron
4. Access course as a student.  No error is thrown.  Student sees the correct error message that course does not exist.

So even if instructors forget to remove the LTI in Canvas, the student will see a valid error message if the course is deleted in MyLA.

PR #984, opened 2020-07-10
Issue 768: check the course data frame length first before creating course list

zqian:


jonespm:
Fixes #768

dashboard/common/db_util.py
```diff
@@ -102,9 +102,13 @@ def get_user_courses_info(username):
             cursor.execute("select canvas_id, name from course where canvas_id in %s",[course_tuple])
             course_names = cursor.fetchall()
             df = pd.DataFrame(list(course_names))
-            df.columns = ["course_id", "course_name"]
-            course_info = df.to_dict(orient='records')
-            logger.info(f"User {username} is enrolled in these courses: {course_info}")
+
+            # create an empty dict
+            course_info = dict()
```
jonespm, line 107: It looks like course_info here is already set up as an empty list so these lines aren't needed. I tested without it and it works fine. 

The expected result from this is a list of objects
`[{column -> value}, … , {column -> value}]
`
So an empty list is expected.

PR #988, opened 2020-07-16
API to support course config settings (#985)

### LINKED ISSUE #985: API entry to support course config
Implement an API entry to support the UI changes described in #849.

Requirements:

* [x] Accept Boolean values for each of the available visualizations (currently 4) to indicate whether they are visible to students.
* [x] Accept Boolean value for whether the `show_grade_counts` feature is enabled.
* [ ] Allow requests only from instructor-like users.
* [x] If possible, accept partial requests.  That is, allow values for as few as one configuration option, as many as all of them, and any number in between.  That will enable the UI to save the values individually, as the user specifies them.

lsloan:
Resolves #985.

Accepts JSON payloads in the request body of the format:

```json
{
  "ap": { "enabled": true },
  "apv1": { "enabled": false },
  "gd": { "enabled": true, "show_grade_counts": true },
  "ra": { "enabled": true }
}
```

Where each of the root-level keys are optional, but at least one must be given.  The `enabled` key is required for any object given.  The `show_grade_counts` key for the `gd` object is optional.  If it is not given, that option won't be updated.

The "jsonschema" module is used for input validation.

Of the four items listed in the requirements for #985, I've not yet tested whether non-instructors are disallowed from using this new API entry.

I'm unsure under which conditions this API entry will be used, so I'm not sure whether the Django model updating code I used requires additional error handling.

I wouldn't be surprised if this needs some tweaks.  However, as I'm taking several days off from work next week, I wanted to try to get something working committed so that @chrisrrowland could potentially use it with the UI code he's developing.

### DIFF
--- a/dashboard/views.py
+++ b/dashboard/views.py
@@ -19,7 +19,7 @@ from dashboard.common import utils
 from django.core.exceptions import ObjectDoesNotExist
 from collections import namedtuple
 
-from dashboard.models import Course, CourseViewOption, Resource, UserDefaultSelection
+from dashboard.models import Course, CourseViewOption, Resource, UserDefaultSelection, User
 from dashboard.settings import RESOURCE_VALUES, RESOURCE_VALUES_MAP, RESOURCE_ACCESS_CONFIG
 from dashboard.settings import COURSES_ENABLED
 
@@ -127,6 +127,9 @@ def get_course_info(request, course_id=0):
     resp['total_weeks'] = total_weeks
     resp['course_view_options'] = CourseViewOption.objects.get(course=course).json(include_id=False)
     resp['resource_types'] = course_resource_list
+
+    course_users_list = User.objects.filter(course_id=course_id)
+    resp['course_user_exist'] = 1 if len(course_users_list) > 0 else 0
 
     return HttpResponse(json.dumps(resp, default=str))
 
--- a/dashboard/views.py
+++ b/dashboard/views.py
@@ -1,27 +1,29 @@
-from django.forms.models import model_to_dict
-from rules.contrib.views import permission_required, objectgetter
-
-import math, json, logging
+import json
+import logging
+from collections import namedtuple
 from datetime import timedelta
-from django.utils import timezone
+from json import JSONDecodeError
 
+import math
 import numpy as np
 import pandas as pd
 from django.conf import settings
 from django.contrib import auth
-from django.db import connection as conn
-from django.http import HttpResponse, HttpResponseForbidden
-from django.shortcuts import redirect, render
-from pinax.eventlog.models import log as eventlog
-from dashboard.event_logs_types.event_logs_types import EventLogTypes
-from dashboard.common.db_util import canvas_id_to_incremented_id
-from dashboard.common import utils
 from django.core.exceptions import ObjectDoesNotExist
-from collections import namedtuple
+from django.db import connection as conn
+from django.forms.models import model_to_dict
+from django.http import HttpResponse, HttpResponseForbidden, HttpResponseBadRequest, JsonResponse
+from django.shortcuts import redirect, render
+from django.utils import timezone
+from pinax.eventlog.models import log as eventlog
+from rules.contrib.views import permission_required, objectgetter
 
+from dashboard.common import utils
+from dashboard.common.db_util import canvas_id_to_incremented_id
+from dashboard.event_logs_types.event_logs_types import EventLogTypes
 from dashboard.models import Course, CourseViewOption, Resource, UserDefaultSelection, User
-from dashboard.settings import RESOURCE_VALUES, RESOURCE_VALUES_MAP, RESOURCE_ACCESS_CONFIG
 from dashboard.settings import COURSES_ENABLED
+from dashboard.settings import RESOURCE_VALUES, RESOURCE_VALUES_MAP, RESOURCE_ACCESS_CONFIG
 
 logger = logging.getLogger(__name__)
 # strings for construct resource download url
@@ -132,6 +134,77 @@ def get_course_info(request, course_id=0):
     resp['course_user_exist'] = 1 if len(course_users_list) > 0 else 0
 
     return HttpResponse(json.dumps(resp, default=str))
+
+
+@permission_required('dashboard.update_course_info',
+                     fn=objectgetter(Course, 'course_id', 'canvas_id'), raise_exception=True)
+def update_course_info(request, course_id=0):
+    """
+
+    :param request: HTTP `PUT` req.; body should contain the JSON body…
+    :param course_id: Integer Canvas course ID number, typically six digits or less.
+    :return: HttpResponse containing `{"default": "success"}` or `{"default": "fail"}`
+    """
+    logger.info(update_course_info.__name__)
+
+    if (request.method != 'PUT'):
+        return HttpResponseBadRequest(JsonResponse({'error': 'Invalid request method.'}))
+
+    course_id = canvas_id_to_incremented_id(course_id)
+    current_user = request.user.get_username()
+
+    bad_json_response = HttpResponseBadRequest(JsonResponse({'error': 'Request JSON malformed.'}))
+
+    try:
+        request_data = json.loads(request.body.decode('utf-8'))
+    except JSONDecodeError:
+        return bad_json_response
+
+    logger.info(request_data)
+
+    # validate root key names
+    valid_root_keys = ['course_view_options', 'show_grade_counts']
+    request_root_keys = request_data.keys()
+    invalid_root_keys = set(request_root_keys) - set(valid_root_keys)
+    if (invalid_root_keys):
+        return bad_json_response
+
+    # validate root key types
+    if (type(request_data.get('course_view_options', {})) is not dict) or \
+            (type(request_data.get('show_grade_counts', False)) is not bool):
+        return bad_json_response
+
+    # everything is valid, now OK to use request data
+
+    ############################
+    return JsonResponse({'fubar': 'tarfu'})
+    ############################
+
+    default_type = list(request_data.keys())[0]
+    default_type_value = request_data.get(default_type)
+    logger.info(
+        f'request to set default for type: {default_type} and default_type value: {default_type_value}')
+    # json for eventlog
+    data = {
+        'course_id': course_id,
+        'default_type': default_type,
+        'default_value': default_type_value
+    }
+    eventlog(request.user, EventLogTypes.EVENT_VIEW_SET_DEFAULT.value, extra=data)
+    key = 'default'
+    try:
+        obj, create_or_update_bool = UserDefaultSelection.objects. \
+            set_user_defaults(int(course_id), current_user, default_type, default_type_value)
+        logger.info(
+            f'''setting default returns with success with response {obj.__dict__} and entry created or Updated: {create_or_update_bool}
+                        for user {current_user} in course {course_id}''')
+        value = 'success'
+    except (ObjectDoesNotExist, Exception) as e:
+        logger.info(
+            f'updating default failed due to {e} for user {current_user} in course: {course_id}')
+        value = 'fail'
+    return HttpResponse(json.dumps({key: value}), content_type='application/json')
+
 
 
 # show percentage of users who read the resource within prior n weeks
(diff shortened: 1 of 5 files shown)


lsloan:
@ssciolla, I'd like your opinion on the one-off JSON validation that I've started to add in [dashboard/views.py](https://github.com/tl-its-umich-edu/my-learning-analytics/pull/988/files#diff-474394b3595bf5cedbf0bf5ee1080b66R165).  I've started by writing some code to check the root level keys and types are as they should be.  I'd finish by doing the same for the second level.  It's not the most elegant stuff, but I thought it might be overkill to use some general purpose JSON schema checking module, like "jsonschema" that we used in the course-inventory project.

On the other hand, I think this type of validation is what we should be doing for all other API entries that take input.  So, maybe we **_should_** use "jsonschema".  I could add it for this new entry I'm writing, then its usage could be expanded to other entries in the near future.

Does that seem reasonable?

ssciolla:
> @ssciolla, I'd like your opinion on the one-off JSON validation that I've started to add in [dashboard/views.py](https://github.com/tl-its-umich-edu/my-learning-analytics/pull/988/files#diff-474394b3595bf5cedbf0bf5ee1080b66R165)
> Does that seem reasonable?

We should be using some kind of validation method for input, so I appreciate that you are thinking about this. What you have (I haven't thoroughly reviewed yet, just skimmed) would probably **work** given short deadlines, a desire to get a working prototype out there. Improvements could be made iteratively once this is set up and @chrisrrowland has an endpoint to use.

In terms of best practices -- while I am still a fan of `jsonschema` -- in the realm of Django and a Web application, we should really be moving in the direction of GraphQL/Django REST Framework for managing API endpoints (yes, I know I said previously this issue would involve writing a "view"). I'm thinking that something closer to an ideal implementation would expand upon the work done by @andrew-gardener here: https://github.com/tl-its-umich-edu/my-learning-analytics/blob/master/dashboard/graphql/mutation.py

It doesn't look like the `CourseViewOption` model is implemented (https://github.com/tl-its-umich-edu/my-learning-analytics/blob/master/dashboard/graphql/objects.py), so it wouldn't be trivial to go down this route. But if significant work is going to be put into improving this, I'd do this before building a custom validation routine with `jsonschema`. Hope that answers your question.

lsloan:
Thanks for the feedback and ideas.

Overnight I thought about my question to you.  Using jsonschema would add a dependency to the project, although it seems to be a fairly stable one at v3.2.0.  It would also add JSON schema code to the program code, but it is a well-documented schema.  So those two potentially negative points actually work out to be pretty positive.

Another positive point is that using jsonschema's validation reduces the amount of potentially buggy code I may produce.

Chris and I are still discussing the schema in the comments of #985.

So this morning I took a few minutes and found that jsonschema does a great job with little code.  So I've changed that part of my code to use it instead.  I'll commit that soon and it'll be available here for perusal.

If it's overkill, don't worry, I can change it back to the single-purpose type of validation I committed yesterday.

lsloan:
@chrisrrowland and @ssciolla, this is ready for review.  Chris, you're welcome to use this code with the UI you're developing.  You can probably pull this into your branch to use it.  (Maybe pull isn't the right thing to do.  You or Sam may have a better idea how to handle it.)

lsloan:
> But if significant work is going to be put into improving this, I think I'd do this before building a custom validation routine with `jsonschema`. Hope that answers your question.

@ssciolla, it turns out that in terms of effort, it was easier and faster to use jsonschema for input validation than to write my own custom code for this.  I'll be interested to see what you may have to say about the code in review.

dashboard/views.py
```diff
@@ -134,6 +137,90 @@ def get_course_info(request, course_id=0):
...
+    """
+
+    :param request: HTTP `PUT` req.; body should contain the JSON body…
+    :param course_id: Integer Canvas course ID number, typically six digits or less.
+    :return: HttpResponse containing `{"default": "success"}` or `{"default": "fail"}`
+    """
+    logger.info(update_course_info.__name__)
+
+    if (request.method != 'PUT'):
+        return HttpResponseBadRequest(JsonResponse({'error': 'Invalid request method.'}))
```
lsloan, line 152: I noticed that using `JsonResponse` like this is better than the wordy `HttpResponse(json.dumps({key: value}),content_type='application/json')` technique used in other parts of this code.

The drawback to `JsonResponse` is that it always has a HTTP response code of `200`.  However, I found that by wrapping it in another kind of response, like `HttpResponseBadRequest`, it's easy to change that in a way that's easy to understand what it's doing.

Also, I think more of the API entry points should verify they were requested with the correct HTTP method.  I **_thought_** there was a way to include the method required as part of the path given in `urls.py`, but that hasn't been used in MyLA before.  Either I've misremembered or it was decided that it's not a good technique for MyLA.

dashboard/views.py
```diff
@@ -134,6 +137,90 @@ def get_course_info(request, course_id=0):
...
+                      'required': ['enabled'],
+                      'additionalProperties': False,
+                      'properties': {'enabled': {'type': 'boolean'},
+                                     'show_grade_counts': {'type': 'boolean'}}},
+                  'ra': {
+                      'type': 'object',
+                      'required': ['enabled'],
+                      'additionalProperties': False,
+                      'properties': {'enabled': {'type': 'boolean'}}}},
+              'minProperties': 1}
```
lsloan, line 189: I liked the way `jsonschema` helped out here.  I would liked to have skipped the repetition of `'additionalProperties': False` everywhere, though.  I asked around and there are supposedly ways to define it once near the beginning of the schema and have it apply to all objects, but I couldn't get it to work.  I think `jsonschema` doesn't support the version of the schema that allows that, but also, I couldn't get it to work with other online JSON Schema tools.

dashboard/views.py
```diff
@@ -134,6 +137,90 @@ def get_course_info(request, course_id=0):
...
+    except jsonschema.ValidationError:
+        return bad_json_response
+
+    # to translate short names returned by model back to original column names
+    view_column_names: dict = {
+        'ap': CourseViewOption.show_assignment_planning.field_name,
+        'apv1': CourseViewOption.show_assignment_planning_v1.field_name,
+        'gd': CourseViewOption.show_grade_distribution.field_name,
+        'ra': CourseViewOption.show_resources_accessed.field_name
+    }
```
lsloan, line 202: Demonstrating my avoidance of hard-coded strings.

dashboard/views.py
```diff
@@ -134,6 +137,90 @@ def get_course_info(request, course_id=0):
...
+                Course.objects.filter(pk=course_id).update(
+                    show_grade_counts=view_settings['show_grade_counts'])
+
+        CourseViewOption.objects.filter(pk=course_id).update(**view_data)
+    except (ObjectDoesNotExist, Exception) as e:
+        logger.info(
+            f'updating course visualization options failed due to {e} for user {current_user} in course: {course_id}')
+        success = False
+
+    return JsonResponse({'default': 'success' if success else 'fail'})
```
lsloan, line 221: I don't really like this `default` technique of indicating success status, but it seems to be the paradigm used in this part of the MyLA code. ¯\_(ツ)_/¯

dashboard/views.py
```diff
@@ -1,27 +1,30 @@
...
-from collections import namedtuple
+from rules.contrib.views import permission_required, objectgetter
 
+from dashboard.common import utils
+from dashboard.common.db_util import canvas_id_to_incremented_id
+from dashboard.event_logs_types.event_logs_types import EventLogTypes
 from dashboard.models import Course, CourseViewOption, Resource, UserDefaultSelection, User
-from dashboard.settings import RESOURCE_VALUES, RESOURCE_VALUES_MAP, RESOURCE_ACCESS_CONFIG
 from dashboard.settings import COURSES_ENABLED
+from dashboard.settings import RESOURCE_VALUES, RESOURCE_VALUES_MAP, RESOURCE_ACCESS_CONFIG
```
lsloan, line 27: IntelliJ IDEA has a nice feature to reorganize imports.  It seems to be very good.  It removes imports that were once used, but are no longer, and groups the imports by their origin, like base Python, framework, third-party modules, etc.

It's not necessary, but helpful.

dashboard/views.py
```diff
@@ -134,6 +137,90 @@ def get_course_info(request, course_id=0):
...
+    """
+    logger.info(update_course_info.__name__)
+
+    if (request.method != 'PUT'):
+        return HttpResponseBadRequest(JsonResponse({'error': 'Invalid request method.'}))
+
+    course_id = canvas_id_to_incremented_id(course_id)
+    current_user = request.user.get_username()
+
+    bad_json_response = HttpResponseBadRequest(JsonResponse({'error': 'Request JSON malformed.'}))
```
ssciolla, line 157: Something about this constructor chaining is leading to the `Content-Type` header being changed to `text/html; charset=utf-8` from `application/json`. It's not an issue with the successful response returned, so the `HttpResponseBadRequest` must be converting it. Maybe instead of using this class, you can just set the `status` in the constructor, i.e. `JsonResponse({'error': 'Request JSON malformed.'}, status=400)`.

Not a huge deal, but I thought I'd point it out. You can see the header with `response.__getitem__('Content-Type')`. If you do fix this, please remove the `HttpResponseBadRequest` import.

dashboard/views.py
```diff
@@ -1,27 +1,30 @@
...
-import math, json, logging
+import json
+import logging
+from collections import namedtuple
 from datetime import timedelta
-from django.utils import timezone
+from json import JSONDecodeError
 
+import jsonschema
+import math
```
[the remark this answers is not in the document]
lsloan: Probably.  Interesting that IntelliJ IDEA grouped it with third party modules.

dashboard/views.py
```diff
@@ -134,6 +137,90 @@ def get_course_info(request, course_id=0):
...
+    current_user = request.user.get_username()
+
+    bad_json_response = HttpResponseBadRequest(JsonResponse({'error': 'Request JSON malformed.'}))
+
+    try:
+        request_data: dict = json.loads(request.body.decode('utf-8'))
+    except JSONDecodeError:
+        return bad_json_response
+
+    schema = {'$schema': 'http://json-schema.org/draft-07/schema',
```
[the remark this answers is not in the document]
lsloan: Yes, I had given that a little thought as I was writing it.

I think that the schema is so specific to this API entry that it should stay here.  If the schema could be reduced in size (but not readability) and still work the same way, I'd feel that even more strongly.

If the schema were to be used by multiple API entries it would make sense to move it to a file.

So, I think I'll leave it here.  If you have any suggestions for improving its formatting or reducing it, please tell me, I'd love to know.

dashboard/views.py
```diff
@@ -134,6 +137,90 @@ def get_course_info(request, course_id=0):
...
+        for (view_key, view_settings) in request_data.items():
+            view_data[view_column_names[view_key]] = view_settings['enabled']
+            if (view_key == 'gd' and 'show_grade_counts' in view_settings.keys()):
+                Course.objects.filter(pk=course_id).update(
+                    show_grade_counts=view_settings['show_grade_counts'])
+
+        CourseViewOption.objects.filter(pk=course_id).update(**view_data)
+    except (ObjectDoesNotExist, Exception) as e:
+        logger.info(
+            f'updating course visualization options failed due to {e} for user {current_user} in course: {course_id}')
```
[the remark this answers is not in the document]
lsloan: I had to resist the urge to wrap the variables the way I'd like them, as `"{current_user}"` and `({course_id}}`.  I like to do that so I can see in the log messages whether these things are strings or numbers and where are the ends.  Especially because strings may have spaces at either end.

dashboard/views.py
lsloan: OK.  That's my bad.  I'd looked for a way to set the status code with `JsonResponse`, but I didn't find it.  I must have overlooked it in the documentation and examples.

dashboard/views.py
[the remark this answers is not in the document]
ssciolla: I'll leave it up to you. `views.py` is already gigantic, so anything to keep things compact is on my radar. I have formatting preferences in general, but they would make it longer, not shorter.

dashboard/views.py
ssciolla: I think its available because it's a subclass of `HttpResponse`.

dashboard/views.py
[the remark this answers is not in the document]
lsloan: I told the IDE to reorganize the imports again.  It put `math` in the right place this time.  I don't know why it was out of place before.  I didn't put it there intentionally.

dashboard/views.py
lsloan: The IDE import reorganization feature that I used again to address your `math` comment removed `HttpResponseBadRequest` for me.

dashboard/views.py
ssciolla: Do we need to specify a different status codes here for an error, like `50X`?

lsloan:
@chrisrrowland, this has been merged with a few changes based on code review comments.  You should probably pull from master to get the changes into your branch.

PR #991, opened 2020-07-22
Re-add git to Dockerfile as it's not found

### LINKED ISSUE #990: Git command not found after slim image switch

When starting up the instance there is an error about git not being available. This probably got removed when switching over to the slim image in #978.

```
Setting Git info variables
./start.sh: line 56: git: command not found
./start.sh: line 57: git: command not found
./start.sh: line 58: git: command not found
```

jonespm:
Fixes #990

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -37,7 +37,7 @@ WORKDIR /code
 COPY requirements.txt .
 RUN apt-get update && \
     apt-get install -y --no-install-recommends \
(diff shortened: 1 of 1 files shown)


jonespm:
Yeah, I think it's probably in the base python:3.8 and just not in the slim. I just noticed when I was looking at the logs but I also saw that this info was missing in the UI.

PR #977, opened 2020-06-22
issue_964 add alert for course when course data is not available

zqian:

PR #993, opened 2020-07-23
Instructor page to configure MyLA courses

### LINKED ISSUE #849: Instructor page to configure MyLA courses when added as LTI
For using MyLA as an LTI, the instructor needs a way to manage the configuration of MyLA for the course: 

**Business rules/functionality:**

- Instructors can add MyLA to a Canvas course as an LTI when they have the Teacher or Assistant role
- Adding MyLA defaults to enable all three visualizations (Resources Accessed, Assignment Planning, Grade Distribution)
- Instructors can toggle a setting to make each view visible or hidden for students in their course
- Instructors can select to show or hide a visualization from the Course View page (not a separate settings page).  The instructor view of this page is controlled by the user role.  Students cannot see the options to turn a view on or off.  

**Workflow for initial course set up in MyLA:**

1. Instructor selects to add MyLA to a course
2. System displays the MyLA Course View page for the course. A checkbox for each view indicates that the view is enabled.  
3. Instructor does not have to do anything else.  However, if they want to unselect a view, then they click the checkbox to uncheck it and that selection is saved.

**Workflow to make changes to MyLA course setting:**
1. Instructor accesses MyLA
2. Course view page displays.
3. Instructor checks or unchecks each view to make changes.

chrisrrowland:
Resolves #849 

Depends on #988


pushyamig:
The Enrollment info is not available in the DB until the course data loaded, incase of standalone tool it will throws 403 permission error initially once data loaded this info can be fetched

So in case of LTI launch, this info is obtained as part of launch in my WIP [PR](https://github.com/pushyamig/my-learning-analytics/blob/i962_lti_integration_new/assets/src/globals.js#L12)

Based on the LTI code this PR might need to adjusted.

chrisrrowland:
> The Enrollment info is not available in the DB until the course data loaded, incase of standalone tool it will throws 403 permission error initially once data loaded this info can be fetched
> 
> So in case of LTI launch, this info is obtained as part of launch in my WIP [PR](https://github.com/pushyamig/my-learning-analytics/blob/i962_lti_integration_new/assets/src/globals.js#L12)
> 
> Based on the LTI code this PR might need to adjusted.

Can  you explain what might need to be adjusted?  Its not clear to me how anyone would try to access a page for an enrollment that doesn't yet exist.

pushyamig:
> Can you explain what might need to be adjusted? Its not clear to me how anyone would try to access a page for an enrollment that doesn't yet exist.

Myla as a standalone tool, we manually added the course to DB and run the cron until then instructor/student can access Myla. So with standalone tool there is an API fetch that is happening `/[info](https://github.com/tl-its-umich-edu/my-learning-analytics/blob/master/dashboard/views.py#L63)` and with that we are also checking through rules, if the person accessing the Myla has [enrollment](https://github.com/tl-its-umich-edu/my-learning-analytics/blob/master/dashboard/rules.py#L25) in USER DB table and that is the time it will throw 403 permission error

With LTI instructor is actually enabling the myla in canvas and that will enable course addition in Myla Database and later based on the course entry we run the cron and data will be available next day. After the LTI launch (before data loaded ) instructor will get a pretty message saying data is not [loaded](https://github.com/tl-its-umich-edu/my-learning-analytics/pull/977). This is my understanding.

justin0022:
@chrisrrowland unfortunately I'm unable to do a detailed review, but I took a quick look at the changes and it looks reasonable to me

jonespm:


```diff

Issues
======
+ Solved 3
- Added 36
           

Complexity increasing per file
==============================
- assets/src/routes/routes.js  1
- assets/src/containers/IndexPage.js  1
- assets/src/components/SelectCard.js  6
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/ITS_Teaching_And_Learning/my-learning-analytics/pullRequest?prid=5922908&bid=19406302)

jonespm:


```diff

Issues
======
+ Solved 9
- Added 18
           

Complexity increasing per file
==============================
- assets/src/routes/routes.js  1
- assets/src/containers/IndexPage.js  1
- assets/src/components/SelectCard.js  6
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/ITS_Teaching_And_Learning/my-learning-analytics/pullRequest?prid=5922908&bid=19406302)

jonespm:


```diff

Issues
======
+ Solved 9
- Added 17
           

Complexity increasing per file
==============================
- assets/src/containers/IndexPage.js  1
- assets/src/routes/routes.js  1
- assets/src/components/SelectCard.js  6
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/ITS_Teaching_And_Learning/my-learning-analytics/pullRequest?prid=5922908&bid=19406302)

dashboard/common/db_util.py
```diff
@@ -131,4 +135,47 @@ def get_canvas_data_date():
...
+                    enrollment_type=course[2]
+                    current_grade = course[3]
+                    final_grade = course[4]
+                    enrollment = CourseEnrollment(name, course_id, None, enrollment_type, current_grade, final_grade)
+                    logger.info("APPENDING "+str(enrollment))
+                    course_enrollment.append(enrollment)
+    except Exception as err:
+        logger.info("Error in get_course_enrollment_info: "+str(err), exc_info = True)
+    return course_enrollment
+
```
ssciolla, line 181: I'd like to have a discussion with you about why all this is necessary when we already have an existing function, `get_user_courses_info` that could be expanded to include `enrollment_type`. In an ideal world, we'd change the database so there was an `enrollment` table and model, which would make our lives easier, but that can be a future effort. It feels like, while your intention was good, it's duplicating a lot of code and introducing things that may not be used.

dashboard/views.py
```diff
@@ -61,6 +61,7 @@ def get_home_template(request):
     return render(request, 'frontend/index.html')
 
 
+
```
ssciolla, line 64: Please remove this newline.

assets/src/components/SelectCard.js
```diff
@@ -1,10 +1,19 @@
...
 import CardContent from '@material-ui/core/CardContent'
 import CardMedia from '@material-ui/core/CardMedia'
 import Typography from '@material-ui/core/Typography'
+import { CardActions, IconButton, Snackbar, CircularProgress, Divider, Fab, Link as MUILink } from '@material-ui/core'
+import CheckBoxOutlineBlankIcon from '@material-ui/icons/CheckBoxOutlineBlank'
+import CheckBoxIcon from '@material-ui/icons/CheckBox'
+import CloseIcon from '@material-ui/icons/Close'
+import { Link as ReactLink } from 'react-router-dom'
+import { yellow, grey } from '@material-ui/core/colors'
+import SaveIcon from '@material-ui/icons/Save'
```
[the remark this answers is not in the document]
chrisrrowland: I believe they are still used, it's just that saving appears instantaneous so you don't see the effect.  If saving ever has a visible delay these will be noticeable.

dashboard/common/db_util.py
chrisrrowland: Was trying to make headway towards using type hinting and being new i didn't want to create any unexpected side effects.  I'm fine with doing it differently if there is an advantage.

assets/src/containers/CourseList.js
```diff
@@ -88,12 +87,11 @@ function CourseList (props) {
...
             : (
               <Grid container spacing={2}>
+                <Grid item xs={12} />
                 {
                   user.relatedCourses.map((course, key) => (
                     <Grid item xs={12} sm={6} lg={4} key={key}>
-                      <Link tabIndex={-1} style={{ textDecoration: 'none' }} to={`/courses/${course.course_id}`}>
-                        <SelectCard cardData={{ title: course.course_name }} />
-                      </Link>
+                      <SelectCard cardData={{ title: course.course_name, description: course.description, path: `/courses/${course.course_id}`, enrollmentType: JSON.parse(user.enrolledCourses).filter(c => c.course_id === course.course_id)[0].enrollment_type, isAdmin: props.user.admin }} />
```
ssciolla, line 94: Just a reminder that we talked about how `enrollmentType` isn't really necessary for this use of `SelectCard`, so maybe we can extract what's common to the `Course` and `CourseList` uses, or just use `Card` here with whatever linking is needed.

dashboard/common/db_util.py
chrisrrowland: Done

assets/src/containers/CourseList.js
chrisrrowland: [#1002 ](https://github.com/tl-its-umich-edu/my-learning-analytics/issues/1002)

dashboard/common/db_util.py
```diff
@@ -87,24 +88,24 @@ def get_default_user_course_id(user_id):
 
 def get_user_courses_info(username):
```
ssciolla, line 89: @pushyamig, can you take a look at this alternative approach I helped with? I know you are composing `myla_globals` differently in your code, putting `lti_`-specific items in there, but I guess I'm somewhat concerned that enrollment roles still need to be associated with a course, as I don't think we're disabling the ability for a user to navigate through the UI to different courses with MyLA enabled. How do you think we should combine this approach with what you are doing?

assets/src/components/SelectCard.js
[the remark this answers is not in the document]
ssciolla: Hm... that seems a little strange to me. Might be a little jarring if they only see it for a couple milliseconds and then wonder what it is. I don't know what options we have here. I understand it would be valuable if it did take a while (at that point, it functions as a `Spinner`), but how often does that really happen?

pushyamig:
>@pushyamig, can you take a look at this alternative approach I helped with? I know you are composing myla_globals differently in your code, putting lti_-specific items in there, but I guess I'm somewhat concerned that enrollment roles still need to be associated with a course, as I don't think we're disabling the ability for a user to navigate through the UI to different courses with MyLA enabled. How do you think we should combine this approach with what you are doing?

@ssciolla Below is the myla globals I am sending. I can add the enrollment info to the course info object. enrollment type is a list for me as it is possible that a person can have more than one role in the course Like TA. But for TA I am checking if he is instructor/TA and just sending him as TA. I don't think instructor in the course can have any other role like admin. But Felt it is safe to have it as a list

`{"username": "dududud", "is_superuser": false, "login": "/accounts/login/", "logout": "/accounts/logout", "user_image": "https://canvas.instructure.com/images/messages/avatar-50.png", "primary_ui_color": null, "help_url": "https://sites.google.com/umich.edu/my-learning-analytics-help/home", "google_analytics_id": "UA-124eeee-2", "user_courses_info": [{"course_id": 223333, "course_name": "course 102 WN 2020"}], "lti_launch_id": "lti1p3-launch-02d2d1a5-5677-4463-971c-bad2283faf68", "lti_role": ["Instructor"], "lti_is_course_data_loaded": true}`

Also another thing is the role is `Instructor` and not `Teacher`, either the UI should handle it in some way since DB has only enrollment info. I could pull enrollment info using canvas LTI custom param but that is addition of custom param [info](https://github.com/tl-its-umich-edu/my-learning-analytics/wiki/Configuration-for-LTI-1.3#configure-lti-settings). Here is canvas variable for [that](https://canvas.instructure.com/doc/api/file.tools_variable_substitutions.html) ( search as Canvas.membership.roles)

jonespm:


```diff

Issues
======
+ Solved 8
- Added 14
           

Complexity increasing per file
==============================
- assets/src/containers/Course.js  1
- assets/src/components/SelectCard.js  7
- assets/src/routes/routes.js  1
- assets/src/containers/IndexPage.js  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/ITS_Teaching_And_Learning/my-learning-analytics/pullRequest?prid=5922908&bid=19406302)

chrisrrowland:
> `{"username": "dududud", "is_superuser": false, "login": "/accounts/login/", "logout": "/accounts/logout", "user_image": "https://canvas.instructure.com/images/messages/avatar-50.png", "primary_ui_color": null, "help_url": "https://sites.google.com/umich.edu/my-learning-analytics-help/home", "google_analytics_id": "UA-124eeee-2", "user_courses_info": [{"course_id": 223333, "course_name": "course 102 WN 2020"}], "lti_launch_id": "lti1p3-launch-02d2d1a5-5677-4463-971c-bad2283faf68", "lti_role": ["Instructor"], "lti_is_course_data_loaded": true}`

Like @ssciolla said, I think the role(s) would need to be associated with the user's course, rather than the user.  The payload would include information for all of their courses, not just the one that the LTI was initiated from?
I'd also think passing all of the roles might be useful, if we ever wanted to make a distinguished between instructor/ta/etc in the future, though that could always be done later, too.

dashboard/common/db_util.py
ssciolla: Can we continue this conversation in the thread? It's nice to be able to collapse this later once it's resolved.

You are right regarding multiple roles for a course, I'll look into that; I think we should resolve it here. But yes, I do think we need to attach the role(s) at the course level, not globally for that user.

assets/src/routes/routes.js
```diff
@@ -1,41 +1,45 @@
...
 
-const routes = (courseId, activeViews) => {
+const routes = (courseId, activeViews, activeOnly=true) => {
   const allViews = [
     {
       path: `/courses/${courseId}/resources`,
       title: 'Resources Accessed',
       icon: Grade,
       description: 'See what resources you and your peers are viewing.',
       image: '/static/images/file_access_trends_icon.png',
```
lsloan, line 10: Not part of this PR, but I just wanted to say it'll be nice to reorganize the backend to include models for the various visualizations that includes things like title, description, icon, image, etc.  Visualizations could be treated as plug-in modules.

assets/src/components/SelectCard.js
```diff
@@ -26,36 +31,177 @@ const styles = theme => ({
...
+    }).catch(e => {
+      setResponseMessage('Error saving setting')
+      console.log('Save Error ' + e)
+      setSnackbarOpen(true)
+    }).finally(
+      setSaving(false)
+    )
+  }
+
+  var saveAsync = function (isEnabled) {
```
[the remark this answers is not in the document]
chrisrrowland: my standard linter in vs code doesn't complain. Codacy might

assets/src/containers/SideDrawer.js
```diff
@@ -30,19 +30,25 @@ function SideDrawer (props) {
...
+    enrollmentType,
+    isAdmin
   } = props
 
   const [selectedIndex, setSelectedIndex] = useState(false)
 
+  const isTeacherOrAdmin = () => {
+    return isAdmin || enrollmentType === 'TeacherEnrollment'
+  }
+
```
ssciolla, line 43: Can we use the utility function you've defined here?

assets/src/util/roles.js
```diff
@@ -0,0 +1,24 @@
...
+/**
+  * Array of string values that are functionally equivalent in representing 'instructor' role
+  */
+const instructorRoles = [
+  'TeacherEnrollment', // student_dashboard.user enrollment_type
+  'Instructor' // LTI role
+]
+
+/**
+ * Return true if the provided criteria represent the role of an a
```
ssciolla, line 14: Is this comment correct?

dashboard/common/db_util.py
```diff
@@ -87,26 +88,24 @@ def get_default_user_course_id(user_id):
...
-
-            if df.size > 0:
-                df.columns = ["course_id", "course_name"]
-                course_info = df.to_dict(orient='records')
-                logger.info(f"User {username} is enrolled in these courses: {course_info}")
+                course_info.append({
+                    'course_id': int(incremented_id_to_canvas_id(course[0])),
+                    'course_name': course[1],
+                    'enrollment_type': course[2]
+                })
```
ssciolla, line 107: So I still think it might be worth revisiting the PR I opened against your `instructor-admin` branch. The way it is now, you will get multiple separate objects/dictionaries for each enrollment in a course, meaning the course name and ID will be repeated in two separate objects if you are both a student and instructor. The PR I opened basically just nests the data so `enrollment_type` becomes `enrollment_types`.

assets/src/containers/Course.js
```diff
@@ -65,6 +68,8 @@ function Course (props) {
                         {...props}
                         courseInfo={courseInfo}
                         courseId={courseId}
+                        enrollmentType={enrollmentType}
+                        isAdmin={user.admin}
```
ssciolla, line 72: Question: do we need to know whether they are specifically an instructor or admin within `IndexPage` or `SideDrawer`? Could we just consolidate into an `instructorOrAdmin` or `teacherOrAdmin` boolean prop? This way, you may be able to just call your `isTeacherOrAdmin` utility function here in `Course.js` and pass the result where necessary.

dashboard/lti_new.py
```diff
@@ -13,7 +13,7 @@
 from pylti1p3.contrib.django import DjangoOIDCLogin, DjangoMessageLaunch, DjangoCacheDataStorage
 from pylti1p3.tool_config import ToolConfDict
 
-from dashboard.models import Course, CourseViewOption
+from dashboard.models import Course, CourseViewOption, User as mylaUser
```
ssciolla, line 16: Can we alias instead as `MylaUser`, so we can tell it's a class?

assets/src/containers/Course.js
```diff
@@ -18,6 +18,7 @@ function Course (props) {
   const [loaded, error, courseInfo] = useCourseInfo(courseId)
   const [sideDrawerState, setSideDrawerState] = useState(false)
 
+  const enrollmentType = user.relatedCourses.length > 0 ? user.relatedCourses.filter(c => c.course_id === props.courseId)[0].enrollment_type : undefined
```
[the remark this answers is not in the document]
ssciolla: Also, whether or not you merge the PR I opened against your branch, there is still the possibility of multiple `enrollmentTypes`, i.e. `instructor` and `student`; we'll need to handle that somehow.

ssciolla:
When I view the course as an admin, disable a view, and then visit that view, it doesn't let me see the data. Is that the desired behavior? I think we need instructors/admins to be able to still see the visualizations, so they can assess whether it should be enabled or not.

assets/src/components/CourseListCard/CourseListCard.js
```diff
@@ -0,0 +1,71 @@
...
+      </Link>
+    </Card>
+  )
+}
+
+CourseListCard.propTypes = {
+  path: PropTypes.string.isRequired,
+  courseName: PropTypes.string.isRequired,
+  description: PropTypes.string
+}
```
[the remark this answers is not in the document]
chrisrrowland: I just came across it on the react [page](https://reactjs.org/docs/typechecking-with-proptypes.html)

assets/src/containers/App.js
```diff
@@ -22,7 +22,7 @@ function App (props) {
       <Route path='/' exact render={props => <CourseList {...props} user={user} />} />
       <Route path='/courses' exact render={props => <CourseList {...props} user={user} />} />
       <Route exact path='/lti/launch'><Redirect to={ltiCourseId} /></Route>
-      {courseId ? <Course user={user} courseId={courseId} {...props} /> : null}
+      {courseId ? <Course user={user} courseId={parseInt(courseId)} {...props} /> : null}
```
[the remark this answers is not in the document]
chrisrrowland: You give me too much credit.  I had to google parseInt vs Number.  Sounds like parseInt can "work" when the string isn't really a good number parseInt("123four") returns 123, where Number("123four") will return NaN

Sounds like Number is better suited here, even though it _should_ always work the same with expected input.

assets/src/components/CourseListCard/CourseListCard.lazy.js
```diff
@@ -0,0 +1,11 @@
+import React, { lazy, Suspense } from 'react';
```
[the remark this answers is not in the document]
chrisrrowland: I used [generate-react-cli](https://www.npmjs.com/package/generate-react-cli) to create the component.  Presumably it does it with some good patterns.  I haven't tried to use the lazy loading version but I did want to keep it there as a reminder to try.

assets/src/containers/SideDrawer.js
chrisrrowland: Yeah I missed that one, thanks.

assets/src/util/roles.js
chrisrrowland: Is any comment, really?

But in this case no.

assets/src/containers/Course.js
chrisrrowland: Currently there is no need within those contexts do distinguish which role is granting them the functionality so your suggestion would be fine.  It seems like there could be functionality down the road that does distinguish between the two, but who knows.  I think it's fine either way now, but I don't feel particularly motivated to change it.  If it seems important I can go ahead and do it.

dashboard/lti_new.py
chrisrrowland: Is this just about case?  MylaUser vs mylaUser?

chrisrrowland:
> When I view the course as an admin, disable a view, and then visit that view, it doesn't let me see the data. Is that the desired behavior? I think we need instructors/admins to be able to still see the visualizations, so they can assess whether it should be enabled or not.

I think we need @jennlove-um to comment on this.  It's working as I intended it to, but I see where you're coming from and that sounds like it could be useful.  Might need some more detail about how this should work if we need to visually distinguish that the instructor/admin is viewing a disabled view.

dashboard/lti_new.py
ssciolla: Haha, um, yes, it is about case. The Python standard (which as you'll often hear is a recommendation, not a rule) is to use `CapWords` for classes (https://www.python.org/dev/peps/pep-0008/#class-names). I was/would be momentarily confused if a class didn't follow this naming scheme. However, if you feel strongly about wanting it to be `camelCase`, I will not fight you on it.

assets/src/containers/Course.js
ssciolla: Seems like it could simplify the code, and the number of places you call your utility function, but it's not a high priority. One potential argument in favor of doing it this way would be because really it should be `enrollmentTypes` (see other comments), so maybe that's another level of input complexity that you'd avoid by moving the processing up in the application.

assets/src/components/CourseListCard/CourseListCard.lazy.js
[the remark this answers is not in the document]
ssciolla: So you're not using this at all? Must have missed that. I guess I'd prefer to not have unused code in the codebase.

The semicolon thing is just to keep things consistent with the rest of the frontend code, follow `standard`.

jennlove-um:
I agree that would be useful and reasonable for instructors or admins to see the views without turning them on for students, but it's new functionality.  As admins, what we've done in the past is use myla-test to view data in a course before turning it on in prod.  In prod, we kind of had this ability by taking advantage of the lag between turning MyLA on for a course in prod and the links showing up on pages for students..

Unless that is an easy change, I'd say we open an enhancement request for that new functionality.  It's not the highest priority for this release.  Instructors could turn the view on, look at it, then turn it right back off with little risk of students seeing the data in the time it takes them to view the visualization and decide whether to use it.

ssciolla:
> Unless that is an easy change, I'd say we open an enhancement request for that new functionality. It's not the highest priority for this release. Instructors could turn the view on, look at it, then turn it right back off with little risk of students seeing the data in the time it takes them to view the visualization and decide whether to use it.

@chrisrrowland, @jennlove-um has spoken! Feel free to disregard my concern for now.

assets/src/routes/routes.js
ssciolla: @lsloan, if this is something you want to happen, I'd suggest creating an issue; no one will remember the suggestion if it only lives here. I'm resolving this for now.

assets/src/components/CourseListCard/CourseListCard.js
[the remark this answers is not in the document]
ssciolla: Cool! Thanks for sharing.

dashboard/common/db_util.py
chrisrrowland: I merged the PR

dashboard/lti_new.py
```diff
@@ -118,10 +118,15 @@ def extracting_launch_variables_for_tool_use(request, message_launch):
     roles = launch_data['https://purl.imsglobal.org/spec/lti/claim/roles']
     username = custom_params['user_username']
     course_id = custom_params['canvas_course_id']
+    canvas_course_long_id = canvas_id_to_incremented_id(course_id)
+    canvas_user_id = custom_params['canvas_user_id']
+    canvas_user_long_id = canvas_id_to_incremented_id(canvas_user_id)
     email = launch_data['email']
     first_name = launch_data['given_name']
     last_name = launch_data['family_name']
+    user_name = launch_data['name']
```
ssciolla, line 127: Nit-pick: variable name could be clearer here, it's super close to `username`, which already exists.

dashboard/common/utils.py
```diff
@@ -66,7 +67,7 @@ def get_myla_globals(current_user):
     logout_url = ""
     google_analytics_id = ""
 
-    is_superuser = current_user.is_superuser
+    is_superuser = current_user.is_staff
```
ssciolla, line 70: Nit-pick: what inspired this change? Kinda weird to me that we're still calling it `is_superuser`.

pushyamig:
> Nit-pick: what inspired this change? Kinda weird to me that we're still calling it is_superuser.
Probably something from the past. is_staff is the status that is actually give admin access. So this was bug that got fixed here. Did not update all the JS variables. We can do it as a separate issue if that is required

assets/src/components/SelectCard.js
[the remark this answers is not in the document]
ssciolla: Huh, maybe I was mis-remembering. Certainly you get results if you google `prefer-const`, but my linter doesn't complain either. Let's not worry about it. I may research later.

dashboard/lti_new.py
jonespm: Yeah I'd really just import this as the class it is `User` or minimally change it to `MylaUser`. It's just a case issue. 

We're actually branded as `MyLA` so technically should be `MyLAUser` but that is a nitpick.

pushyamig:
>Yeah I'd really just import this as the class it is User or minimally change it to MylaUser. It's just a case issue.

We're actually branded as MyLA so technically should be MyLAUser but that is a nitpick.

@jonespm I resolved it here https://github.com/pushyamig/my-learning-analytics/blob/i1009_fix_becourse_user_course_link/dashboard/lti_new.py#L17

PR #1008, opened 2020-08-10
Add links to info about views on the course view page

### LINKED ISSUE #1000: Add links to info about views on the course view page
On the course view tiles, add a link to more information about the view.  This will help instructors learn more about a view when they are considering whether to turn it on or not.

On each visualization tile for a course, add an information icon that links to the online help for that page view.

chrisrrowland:
Closes #1000 

This is dependent on changes in PR #993 

View specific URLs are configured in settings.py and can be overridden via env.json.

There are currently default U of M pages linked for Resources Accessed, Assignment Planning (v1), and Grade Distribution.

Assignment Planning (v2) has no link.


assets/src/routes/routes.js
```diff
@@ -1,41 +1,50 @@
 import { Grade } from '@material-ui/icons'
+import { viewHelpURLs } from '../globals'
```
chrisrrowland, line 2: This wasn't done anywhere else, so I'm not sure if it is considered bad or if its OK.

jonespm:


```diff

Issues
======
+ Solved 4
- Added 32
           

Complexity increasing per file
==============================
- assets/src/components/CourseListCard/CourseListCard.lazy.js  1
- dashboard/lti_new.py  1
- assets/src/util/roles.js  2
- assets/src/components/CourseListCard/CourseListCard.js  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/ITS_Teaching_And_Learning/my-learning-analytics/pullRequest?prid=6002792&bid=19679514)

dashboard/lti_new.py
```diff
@@ -169,6 +169,12 @@ def extracting_launch_variables_for_tool_use(request, message_launch):
         'user_image': user_img,
         'primary_ui_color': settings.PRIMARY_UI_COLOR,
         'help_url': settings.HELP_URL,
+        'view_help_urls': { 
+            'ra':settings.URL_VIEW_RESOURCES_ACCESSED,
+            'apv1':settings.URL_VIEW_ASSIGNMENT_PLANNING_V1,
+            'ap':settings.URL_VIEW_ASSIGNMENT_PLANNING_V2,
+            'gd':settings.URL_VIEW_GRADE_DISTRIBUTION 
+        }  ,
```
pushyamig, line 177: This file change may not be needed if this code get merged, https://github.com/tl-its-umich-edu/my-learning-analytics/pull/1010

dashboard/common/utils.py
```diff
@@ -89,6 +89,13 @@ def get_myla_globals(current_user):
         "logout": logout_url,
         "primary_ui_color": primary_ui_color,
         "google_analytics_id": google_analytics_id,
-        "help_url": help_url
+        "help_url": help_url,
+        "view_help_urls": { 
+            'ra':settings.URL_VIEW_RESOURCES_ACCESSED,
+            'apv1':settings.URL_VIEW_ASSIGNMENT_PLANNING_V1,
+            'ap':settings.URL_VIEW_ASSIGNMENT_PLANNING_V2,
```
pushyamig, line 96: Can we just keep the name as assignment_planning instead of v2? The idea is if Assignment redesign is the way to go in future we will remove all the `VIEW_ASSIGNMENT_PLANNING_V1`. SO the naming convention doesn't needs to be attached with versioning

dashboard/settings.py
```diff
@@ -45,6 +45,11 @@
 LOGOUT_URL = '/accounts/logout'
 HELP_URL = ENV.get("HELP_URL", "https://sites.google.com/umich.edu/my-learning-analytics-help/home")
 
+URL_VIEW_RESOURCES_ACCESSED =       ENV.get("URL_VIEW_RESOURCES_ACCESSED", "https://its.umich.edu/academics-research/teaching-learning/my-learning-analytics/support/resources-accessed")
+URL_VIEW_ASSIGNMENT_PLANNING_V1 =   ENV.get("URL_VIEW_ASSIGNMENT_PLANNING_V1", "https://its.umich.edu/academics-research/teaching-learning/my-learning-analytics/support/assignment-planning")
+URL_VIEW_ASSIGNMENT_PLANNING_V2 =   ENV.get("URL_VIEW_ASSIGNMENT_PLANNING_V2", None)
```
pushyamig, line 50: rename to `URL_VIEW_ASSIGNMENT_PLANNING`

dashboard/common/db_util.py
```diff
@@ -136,4 +137,4 @@ def get_canvas_data_date():
                 return date
     except Exception:
         logger.info("Value could not be found from metadata", exc_info = True)
-    return datetime.min
\ No newline at end of file
+    return datetime.min
```
pushyamig, line 140: I don't see any code changes associated with this file can the changes be undone?

assets/src/globals.js
```diff
@@ -22,6 +22,8 @@ const user = Object.freeze({
   LTIIsCourseDataLoaded: mylaGlobals.lti_is_course_data_loaded
 })
 
+const viewHelpURLs = Object.freeze(mylaGlobals.view_help_urls)
```
pushyamig, line 25: I don't feel Object.freeze is strictly needed since these are static's and will have less effect of changing, and if somebody will change.

assets/src/routes/routes.js
pushyamig: IMO This should be ok since you are just pulling what is needed rather bring all the stuff

ssciolla:
FYI, @pushyamig, I'm working with @chrisrrowland on trying to clean up the commit history here.

jonespm:


```diff

Issues
======
+ Solved 2
- Added 7
           
```


See the complete overview on [Codacy](https://app.codacy.com/manual/ITS_Teaching_And_Learning/my-learning-analytics/pullRequest?prid=6002792&bid=19679514)

dashboard/settings.py
```diff
@@ -45,6 +45,11 @@
 LOGOUT_URL = '/accounts/logout'
 HELP_URL = ENV.get("HELP_URL", "https://sites.google.com/umich.edu/my-learning-analytics-help/home")
 
+URL_VIEW_RESOURCES_ACCESSED =       ENV.get("URL_VIEW_RESOURCES_ACCESSED", "https://its.umich.edu/academics-research/teaching-learning/my-learning-analytics/support/resources-accessed")
+URL_VIEW_ASSIGNMENT_PLANNING_V1 =   ENV.get("URL_VIEW_ASSIGNMENT_PLANNING_V1", "https://its.umich.edu/academics-research/teaching-learning/my-learning-analytics/support/assignment-planning")
+URL_VIEW_ASSIGNMENT_PLANNING =   ENV.get("URL_VIEW_ASSIGNMENT_PLANNING", None)
+URL_VIEW_GRADE_DISTRIBUTION =       ENV.get("URL_VIEW_GRADE_DISTRIBUTION", "https://its.umich.edu/academics-research/teaching-learning/my-learning-analytics/support/grade-distribution")
```
ssciolla, line 51: @chrisrrowland, it'd be nice to build these URLs off of a base, maybe `INFO_URL`. I'm assuming that the `HELP_URL` needs to be different. If they are the same (@jennlove-um), then we should try to consolidate, and we'd need to think about updating the `AvatarModal` code to use what you're doing here. 

Also, I'd delete the spaces after the assignment operators (e.g. `URL_VIEW_RESOURCES_ACCESSED = ENV...`)

assets/src/components/SelectCard.js
```diff
@@ -117,23 +136,54 @@ const SelectCard = props => {
...
     }
   }
 
+  function getHelpLink (cardData) {
+    if (cardData.helpUrl) {
+      return (
+        <Typography gutterBottom variant='h5' component='h4' className={classes.title}>
+          <Tooltip title={'About ' + cardData.title}>
+            <MUILink className={classes.infoLink} href={cardData.helpUrl} target='_blank' rel='noopener noreferrer'>
+              <InfoIcon />
```
ssciolla, line 157: Nit-pick: we could make this icon a little bigger to fill the space, be more noticeable.

assets/src/components/SelectCard.js
```diff
@@ -117,23 +136,54 @@ const SelectCard = props => {
...
+      )
     } else {
       return null
     }
   }
 
+  function getHelpLink (cardData) {
+    if (cardData.helpUrl) {
+      return (
+        <Typography gutterBottom variant='h5' component='h4' className={classes.title}>
```
ssciolla, line 154: Nit-pick: Is `Typography` necessary here?

assets/src/components/SelectCard.js
```diff
@@ -117,23 +136,54 @@ const SelectCard = props => {
...
-        <Typography gutterBottom variant='h5' component='h4' className={classes.title}>
-          {cardData.title}
-        </Typography>
-        <Typography component='p' className={classes.description}>
-          {cardData.description}
-        </Typography>
+        <Grid container className={classes.titleArea}>
+          <Grid item xs={11}>
+            <Link tabIndex={-1} to={cardData.path} className={classes.viewLink}>
+              <Typography gutterBottom variant='h5' component='h4' className={classes.title}>
```
ssciolla, line 173: Nit-pick: We have a variant mapping in `globals.js` for Web accessibility purposes. I think the `h5` variant looks good, but I think we actually want it to show up as an `h2` (or maybe `h1`). I think `h5` maps to `h1`, so maybe just change `component` here to `h2`. This is mostly to support screen reading, I think; we want to avoid skipping header levels.

assets/src/components/SelectCard.js
```diff
@@ -117,23 +136,54 @@ const SelectCard = props => {
...
       <CardContent className={classes.content}>
-        <Typography gutterBottom variant='h5' component='h4' className={classes.title}>
-          {cardData.title}
-        </Typography>
-        <Typography component='p' className={classes.description}>
-          {cardData.description}
-        </Typography>
+        <Grid container className={classes.titleArea}>
+          <Grid item xs={11}>
+            <Link tabIndex={-1} to={cardData.path} className={classes.viewLink}>
```
ssciolla, line 172: Can the `Link` be wrapped around the `Grid container`, instead of around each individual `Grid item`?

assets/src/components/SelectCard.js
```diff
@@ -1,11 +1,13 @@
 /* global fetch */
 import React, { useState } from 'react'
 import { withStyles } from '@material-ui/core/styles'
-import { Card, CardActionArea, CardActions, CardContent, CardMedia, CircularProgress, Divider, Fab, IconButton, Snackbar, Typography } from '@material-ui/core'
+import { Card, CardActionArea, CardActions, CardContent, CardMedia, CircularProgress, Divider, Fab, IconButton, Link as MUILink, Snackbar, Tooltip, Typography } from '@material-ui/core'
 import CheckBoxOutlineBlankIcon from '@material-ui/icons/CheckBoxOutlineBlank'
 import CheckBoxIcon from '@material-ui/icons/CheckBox'
 import CloseIcon from '@material-ui/icons/Close'
 import { Link } from 'react-router-dom'
+import InfoIcon from '@material-ui/icons/Info'
```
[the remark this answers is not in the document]
chrisrrowland: The material-ui documentation says individual imports is 'safer' whereas one line yield the 'best developer experience' however you had to take care to [minimize your bundle size](https://material-ui.com/components/icons/#usage) if you do it that way.  I'm not sure if that is something we're already doing appropriately?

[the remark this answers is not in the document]
ssciolla: Good point, didn't think about that. I wasn't necessarily saying that it should be all on one line. At the least, the icon imports could be one after another.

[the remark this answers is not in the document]
ssciolla: Seems like we're probably safe to use "tree-shaking", per the ES6 modules and `webpack` versions limitations listed here: https://material-ui.com/guides/minimizing-bundle-size/#when-and-how-to-use-tree-shaking

@pushyamig, can you confirm?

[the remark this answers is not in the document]
ssciolla: It's different for icons I guess. We have this babel import (`babel-plugin-named-asset-import`), not sure if that does it.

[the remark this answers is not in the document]
ssciolla: I'll admit I'm ignorant about both `webpack` and `babel`, so thanks for raising this issue.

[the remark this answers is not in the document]
chrisrrowland: I am too, I didn't want to make a change that might have a negative impact on the  bundles size.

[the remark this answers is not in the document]
pushyamig: We are on Webpack 4 i see the tree-shaking is supported that particular version. For prod webpack build it does minimize the bundle size. bundle can be further compressed as I am doing it [here](https://github.com/pushyamig/my-learning-analytics/blob/i987_optimize_webpack/package.json#L137) and [here](https://github.com/pushyamig/my-learning-analytics/blob/i987_optimize_webpack/webpack.config.js#L38) 

but we can even break down the bundle size using webpack code-[splitting](https://webpack.js.org/guides/code-splitting/) instead of one single js file we can have multiple as canvas does 

Screen Shot 2020-08-14 at 12 10 43 PM

Whatever chris is doing is fine, we might need to look at code-splitting as a future issue

dashboard/settings.py
chrisrrowland: I removed the extra indentation.

However I like the flexibility of having the full url rather than sharing a prefix.  Odds are a common prefix would work, but this way allows admins to point at their documentation no matter where it is, rather than requiring it to all be co-located.

assets/src/components/SelectCard.js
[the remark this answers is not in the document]
pushyamig: > It's different for icons I guess. We have this babel import (babel-plugin-named-asset-import), not sure if that does it.

Probably it is doing it when making it as big bundle, since in the `webpack.config.js`. 
```
rules: [
      {
        test: /\.(js|mjs|jsx|ts|tsx)$/,
        loader: 'babel-loader'
```

I really don't know details of it. when making the bundle it just add a simple import in the js file instead of multiple reference based on the how many file it reference. I am learning this as well

assets/src/routes/routes.js
pushyamig: This should be ok.

assets/src/components/SelectCard.js
chrisrrowland: There may be a better way to do it, but the `Grid container` has both the link to the view and the link to help, so I had to come up with a way to wrap the individual areas that should link to different places.  This was what I came up with.

I imagine there would be a way to have a relatively positioned overlay over the info icon with a higher z-index and the help link, perhaps, but this is what I came up with.

dashboard/common/utils.py
```diff
@@ -89,6 +89,13 @@ def get_myla_globals(current_user):
         "logout": logout_url,
         "primary_ui_color": primary_ui_color,
         "google_analytics_id": google_analytics_id,
-        "help_url": help_url
+        "help_url": help_url,
```
pushyamig, line 92: @chrisrrowland can you make sure the `help_url` is `target = _blank` as well, since it was not before. I am not sure if you are already addressing it here but that's not a requirement but would be nice to have it

dashboard/settings.py
ssciolla: I think having documentation that isn't co-located might be a problem, but certainly this does provide more flexibility. Not a huge deal. I'm more concerned about whether the `HELP_URL` default and `AvatarModal` components needs to be updated.

dashboard/common/utils.py
chrisrrowland: I'm not even sure where help_url is used :-o

assets/src/components/SelectCard.js
ssciolla: Mmm, I see. Didn't think about how this would affect that other item. What you have looks good for now and works!

dashboard/common/utils.py
pushyamig: It AvatarModal component,
Screen Shot 2020-08-14 at 12 39 37 PM

assets/src/components/SelectCard.js
chrisrrowland: So it has been using h4 all along, and changing that makes the text larger.. is there some functionality here that makes the screen reader treat it as a h2 but the browser render it as a h4?  Or are you saying we should just make it a h2?

dashboard/common/utils.py
chrisrrowland: Thanks, I'll include that as well.

assets/src/components/SelectCard.js
ssciolla: We want the style to be `h5` (I think) and the actual HTML element to be `h2`. Based on what I'm seeing in the [`Typography` API](https://material-ui.com/api/typography/), you can change this by modifying `component` to be `h2`. You're saying that affected the style? My understanding is `variant` controls the style, but I could be wrong.

assets/src/globals.js
chrisrrowland: I hadn't used this before and was following the example  of mylaglobals.  Is there a downside to using it? I can remove it for sure.

dashboard/settings.py
chrisrrowland: I have updated those per @pushyamig to open in a new tab, is [this](https://github.com/tl-its-umich-edu/my-learning-analytics/pull/1008/commits/7d8997d8f73dcf5e96541535e63e96a01bc9d362) what you meant?

assets/src/components/SelectCard.js
chrisrrowland: I can set it to 'large' size, not sure if there is more flexibility than 3 sizes.
Is this better? @ssciolla @jennlove-um 
image

chrisrrowland: Vs
image

dashboard/settings.py
ssciolla: No, this is a separate problem. Essentially, you're improving on (but also replicating) a different help URL set up that I referred you to when you started working on this issue: https://github.com/tl-its-umich-edu/my-learning-analytics/blob/f9b613a96a2ec4c84870eb2a920fce8d416a7c85/assets/src/components/AvatarModal.js#L96-L105

I checked with @jennlove-um, and the links from the `AvatarModal` and the `SelectCard`s should be the same. Now, you could configure them to be the same, but I really think these should be populated in the same way. Am I making more sense now?

pushyamig: Yes, it should be a target blank since it is external link. Thanks for doing the @chrisrrowland

ssciolla: You can keep the `HELP_URL` settings variable (in fact you have to, since there should be a help landing page linked to somewhere), but I'd rather see this logic shown above replaced with some way of pulling these other globals you've set up in MyLA globals.

assets/src/components/SelectCard.js
chrisrrowland: It is allowing the icon to align with the title text.  Would there be a better way to do it?

assets/src/components/SelectCard.js
chrisrrowland: Ok I changed the h4 to h2.

dashboard/settings.py
chrisrrowland: Hmm, well i see why you wanted a base url in that case.   I did not understand that these urls were supposed to be related in any way to those urls.  Is that the case?  Are those URLs supposed to be replaced by the new ones?  That was definitely not clear to be based on the issue description, but maybe it assumed I had more insight into the issue than I did.

@jennlove-um how is this supposed to work?  Are the info links on the select cards supposed to link to the same context aware help links for each of the views, or are these independent links?

jonespm:


```diff

Issues
======
+ Solved 2
- Added 6
           
```


See the complete overview on [Codacy](https://app.codacy.com/manual/ITS_Teaching_And_Learning/my-learning-analytics/pullRequest?prid=6002792&bid=19679514)

assets/src/components/SelectCard.js
jennlove-um: I like the larger size better.

dashboard/settings.py
jennlove-um: The info links on the select cards should be context aware like each visualization is.  When an instructor wants to learn more about one of the views, it makes sense to take them directly to the help page for that view.

dashboard/settings.py
```diff
@@ -45,6 +45,11 @@
 LOGOUT_URL = '/accounts/logout'
 HELP_URL = ENV.get("HELP_URL", "https://sites.google.com/umich.edu/my-learning-analytics-help/home")
 
+URL_VIEW_RESOURCES_ACCESSED = ENV.get("URL_VIEW_RESOURCES_ACCESSED", "https://its.umich.edu/academics-research/teaching-learning/my-learning-analytics/support/resources-accessed")
```
jonespm, line 48: I don't think we should have defaults in the code for umich resources, these should also be documented in the env_sample file: https://github.com/tl-its-umich-edu/my-learning-analytics/blob/master/config/env_sample.json

We can suggest these as help documentation that may want to be customized.

assets/src/components/SelectCard.js
chrisrrowland: Switched to the larger icon

dashboard/settings.py
chrisrrowland: @ssciolla I refactored to be more consistent with how help urls are used

dashboard/settings.py
chrisrrowland: Talked to @jennlove-um and going to leave this as is and treat handling lack of help in a future issue.

lsloan: These should be suffix URL fragments that are added to a base help URL.

assets/src/components/AvatarModal.js
```diff
@@ -94,14 +95,23 @@ function AvatarModal (props) {
...
+      } else if (url.includes('assignmentsv1')) {
+        target = viewHelpURLs.apv1
+      } else if (url.includes('assignments')) {
+        target = viewHelpURLs.ap
+      } else if (url.includes('resources')) {
+        target = viewHelpURLs.ra
+      } else {
+        target = viewHelpURLs.home
+      }
+      return target || viewHelpURLs.home
```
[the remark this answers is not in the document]
chrisrrowland: All of the help urls can be independently configured so it's possible that any one of them is undefined.
Currently they all have (U of M) defaults except the new assignment planning.

So out of the box viewHelpURLs.ap is undefined and in that case it will use the 'default' help home page.

The default help home page could also be overridden to be undefined, which would just not work well.  But someone would have to do that purposefully.. we're planning to address removing u of m defaults and handling lack of configured help in the future.

assets/src/components/AvatarModal.js
```diff
@@ -94,14 +95,23 @@ function AvatarModal (props) {
...
+        target = viewHelpURLs.gd
+      } else if (url.includes('assignmentsv1')) {
+        target = viewHelpURLs.apv1
+      } else if (url.includes('assignments')) {
+        target = viewHelpURLs.ap
+      } else if (url.includes('resources')) {
+        target = viewHelpURLs.ra
+      } else {
+        target = viewHelpURLs.home
+      }
```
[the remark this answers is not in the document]
chrisrrowland: I agree, but that was going to require more  significant change and I wanted to get it done asap.  Considering the rate at which new views are added, this is not likely to cause many problems in the meantime.

dashboard/settings.py
chrisrrowland: I didn't see any user benefit in requiring urls to be relatively located so i removed that to permit flexibility.
The previous design required the help pages to all have specific urls based off a prefix.  Which I suppose you are advocating.  I thought this was inflexible.  There are benefits to both I guess.

PR #1014, opened 2020-08-12
Instructor preview

### LINKED ISSUE #1013: Allow instructors to see a preview of the views before enabling them for students
Instructors should be able to see a preview of each visualization for their course before enabling the view for students.  This will allow them to make informed decisions about whether to turn a view on for students.

On the course views page, allow instructors to view a visualization even when it is disabled for students. When they are on the individual view, provide a banner or other visual marker that indicates this view is not enabled for students.  Text: "Preview Mode: This view is currently disabled for students."

chrisrrowland:
Resolves #1013 

image


jonespm:


```diff

Issues
======
+ Solved 10
- Added 20
           

Complexity increasing per file
==============================
- assets/src/components/PreviewBanner/PreviewBanner.js  2
- assets/src/containers/AssignmentPlanningV2.js  1
- assets/src/containers/AssignmentPlanning.js  1
- assets/src/containers/ResourcesAccessed.js  1
- assets/src/components/PreviewBanner/PreviewBanner.lazy.js  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/ITS_Teaching_And_Learning/my-learning-analytics/pullRequest?prid=6011441&bid=19706744)

chrisrrowland:
Cleaned up the commit.

assets/src/containers/AssignmentPlanning.js
```diff
@@ -133,82 +135,85 @@ function AssignmentPlanning (props) {
...
-                  : <Spinner />
-              }
-            </div>
-            <div className={classes.section}>
-              <Grid container>
-                <Grid item xs={12} md={10}>
-                  <Typography variant='h6' gutterBottom>Assignments Due by Date</Typography>
+    <>
+      <PreviewBanner isDisabled={disabled} />
+      <div className={classes.root}>
```
ssciolla, line 140: Looks like you needed to create this external `<>` so you could put the `Banner` on the same level as the `root` `div`. I'm not sure (but could be convinced) about this decision. I think it might look cleaner if the `Banner` was on the main `Paper` with the rest of the content -- the width would then be the same as the other `Banner` indicating no data. We could even consider putting this banner below the View title.

Less important side note: Another advantage of this would be that you'd technically be editing many fewer lines of code. If it's really necessary, that's fine, but it does end up obscuring history, if you want to know when/why a particular part of the component was changed in the past.

assets/src/components/PreviewBanner/PreviewBanner.js
```diff
@@ -0,0 +1,21 @@
...
+
+const PreviewBanner = props => {
+  const { isDisabled } = props
+
+  if (isDisabled) {
+    return (<AlertBanner>Preview Mode: This view is currently disabled for students.</AlertBanner>)
+  } else {
+    return null
+  }
+}
```
ssciolla, line 13: I don't really see a problem with this, but I'm not sure it's really worth another component now that I look at it.

Instead of this, you could just have this line in each view:

```
{disabled ? <AlertBanner>Preview Mode: This view is currently disabled for students.</AlertBanner> : undefined}
```

assets/src/components/PreviewBanner/PreviewBanner.lazy.js
```diff
@@ -0,0 +1,11 @@
+import React, { lazy, Suspense } from 'react';
+
+const LazyPreviewBanner = lazy(() => import('./PreviewBanner'));
+
+const PreviewBanner = props => (
+  <Suspense fallback={null}>
+    <LazyPreviewBanner {...props} />
+  </Suspense>
+);
```
ssciolla, line 9: You're generating these through some command, correct? Are you using it here? It doesn't look like it to me. Would this lazy approach make more sense for components with state? I need to be educated on how these work, when they make sense. Creating more directories and files for this (especially if we're not going to use it) seems like potentially too much.

chrisrrowland: Yeah I am generating them with a command.  I was operating under the assumption that the tool was implementing best practices.  I have tried to use the lazy version of the component but it looks like it requires some additional configuration to make it work.  I had left the previous one in b/c I thought it might make sense to take a pass at a later date to use the lazy version (as I assumed it was preferred for performance)  I didn't think there was really any downside to having a file that wasn't used but I can delete it.

assets/src/components/PreviewBanner/PreviewBanner.js
chrisrrowland: Sure.  Is it better to have that exact thing 4 times or have a component that can be changed in one place?

assets/src/containers/AssignmentPlanning.js
chrisrrowland: I liked the idea of the banner effectively existing outside of the view so it would be a true preview. I hadn't considered the blame impact. I care more about the structure of the code than making it easier to go through history, but I don't feel strongly about either aspect in this case, though I'd tend to leave it how it is currently..

ssciolla: Okay, I think you convinced me. Might me worth checking with @jennlove-um on the appearance, but I'm fine with this.

assets/src/components/PreviewBanner/PreviewBanner.js
ssciolla: I'm ambivalent and I guess it doesn't matter that much.

assets/src/components/PreviewBanner/PreviewBanner.lazy.js
ssciolla: I guess I need to learn more about these. I generally prefer to keep the codebase compact and only including code you're actually going to run. That might be a narrow-minded approach though.

chrisrrowland: As do I.  I don't think it would be difficult to come up with it again if it was gone, however if it was a practice to use eventually I think the folder structure makes sense.  I'll go ahead and delete it for now.  Not like lazy loading something this small would be likely to be impactful anyways, I imagine.

assets/src/components/PreviewBanner/PreviewBanner.js
chrisrrowland: I removed the component and the folders and the lazy stuff.

ssciolla: Yeah, the repetition kinda bugs me, but it also didn't feel like the `PreviewBanner` was doing enough to warrant a whole new component. Hm... Anyways, what you currently have is fine for now -- maybe in the future we can determine all the things these views have a common and extract those somehow.

PR #1011, opened 2020-08-11
#826 - load MiVideo usage data from UDP

lsloan:
Resolves #826.

I scrapped the approach I had been taking and started fresh.

Rather than trying to combine UDP and Kaltura API data using a new technique of my own, I decided to re-use the existing UDP-only approach and adapt it to work with MiVideo events.  The efficiency I thought my technique would bring probably wouldn't save as much resources as I expected.

The strange thing about this when I test it is that I see many messages logged to the console that say rows with various resource names and IDs are being updated, but when the process is done and I check my DB contents, I don't see any of that.

~Also, I've been testing with my personal test course in production Canvas (https://umich.instructure.com/courses/374350).  The course has several test students and they've all viewed the MiVideo content I published.  I can see Caliper events from their activity in production UDP.  However, those users are never added to the `user` table of my DB.  The cron code isn't doing what I expect.  I've probably misunderstood some of the subtleties of the process.~

UDW delays were a problem when it came to testing.  I'd added new students to my course to test with and it seemed to take more than 24 hours for those students to appear in the UDW course roster.

## Caveat

I'm having trouble testing the MyLA UI for this change.  This code gets the MiVideo data into the DB, but I've not actually seen it in the UI, either via LTI or SAML launch.  I'm still working on verifying that the resources are shown for courses, but I'd appreciate tips to get that working.

PR #1018, opened 2020-08-13
Course_id still in resource table after being dropped by conversion

### LINKED ISSUE #1017: Course_id still in resource table after being dropped by conversion

@chrisrrowland noticed this error when running on an empty database

`student_dashboard  | [2020-08-13 08:22:17 -0400] [ERROR] [models.py:268] Couldn't fetch Resource list in Course 17700000000335188 due to: Cannot resolve keyword 'course_id' into field. Choices are: id, name, resource_id, resource_type, resourceaccess`

This wasn't an issue in my database but resource for some reason had the course_id column still. It looks like pandas `to_sql` was adding this back in. We'll need to fix this so it's not created anymore and the query works.

jonespm:
Fixes #1017

Co-authored-by: chrisrrowland <chris.robert.rowland@gmail.com>

### DIFF
--- a/dashboard/cron.py
+++ b/dashboard/cron.py
@@ -363,13 +363,8 @@ class DashboardCronJob(CronJobBase):
                          f'{resource_access_df.dtypes}')
 
             # Make resource data from resource_access data
-            resource_df = resource_access_df.copy(deep=True)
-            resource_df = resource_df.drop(
-                columns=['user_id', 'access_time', 'course_id'])
-            resource_df = resource_df.drop_duplicates('resource_id')
-
-            # set resource_id as index for pangres.upsert
-            resource_df: pd.DataFrame = resource_df.set_index('resource_id')
+            resource_df = resource_access_df.filter(["resource_id", "resource_type", "name"])
+            resource_df = resource_df.drop_duplicates(["resource_id"])
 
             logger.debug(f'resource_df:\n'
                          f'{resource_df}\n'
--- a/dashboard/models.py
+++ b/dashboard/models.py
@@ -263,7 +263,7 @@ class ResourceQuerySet(models.QuerySet):
         :return:
         """
         try:
-            return list(self.values_list('resource_type', flat=True).distinct().filter(course_id=course_id))
+            return list(self.values_list('resource_type', flat=True).distinct().filter(resourceaccess__course_id=course_id))
         except(self.model.DoesNotExist, Exception) as e:
             logger.error(f"Couldn't fetch Resource list in Course {course_id} due to: {e}")
             return None


jonespm:
Yeah, the cron used to be doing a copy then a drop.

I was reading that if I do a filter that action does a copy so we don't need these two actions and it's clearer what columns are being kept. I was going to modify this to drop the additional course_id column but now it only keeps the columns it needs.

jonespm:
This was broken by #1011, will need to fix some conflicts

dashboard/cron.py
```diff
@@ -303,11 +303,8 @@ def update_with_bq_access(self):
             logger.debug(resource_access_df)
 
             # Because we're pulling all the data down into one query we need to manipulate it a little bit
-            # Make a copy of the access dataframe
-            resource_df = resource_access_df.copy(deep=True)
-            # Drop out the columns user and access time from resource data frame
-            resource_df = resource_df.drop(["user_id", "access_time"], axis=1)
-            # Drop out the duplicates
+            # This creates a new resource df of the resource_access with specific columns
+            resource_df = resource_access_df.filter(["resource_id", "resource_type", "name"], axis=1)
```
lsloan, line 307: Instead of `axis=1`, use `columns=` with the list of column names.  It makes the code easier to read and maintain.  No one would need to remember or look up that axis 1 refers to columns.
```suggestion
            resource_df = resource_access_df.filter(columns=["resource_id", "resource_type", "name"])
```

dashboard/models.py
```diff
@@ -263,7 +263,7 @@ def get_course_resource_type(self, course_id):
         :return:
         """
         try:
-            return list(self.values_list('resource_type', flat=True).distinct().filter(course_id=course_id))
+            return list(self.values_list('resource_type', flat=True).distinct().filter(resourceaccess__course_id=course_id))
```
lsloan, line 266: This is a good addition.

I've been thinking that we may need to be a little more sophisticated with identifying resource types.  MiVideo resources have at least two types, video and audio.  At some point may want to be specific about this.  I can't think of many good reasons, but knowing the specific type would allow us to show a more appropriate icon for the resource in the UI views.

lsloan:
> This was broken by #1011, will need to fix some conflicts

I wouldn't say "broken" per se.  I just address this issue a different way in that PR.

I hope there's no hard feelings about that.  I would have liked to resolve the conflict in #1011 if I had more time.  Zhen told me that because an approval to my PR had already been given and we're trying to meet a tight deadline, I should merge it.  I wanted to keep the PR open longer for more possible feedback, too.

dashboard/cron.py
```diff
@@ -303,11 +303,8 @@ def update_with_bq_access(self):
...
             resource_df = resource_df.drop_duplicates(["resource_id"])
```
lsloan, line 308: I feel bad about leaving you in the lurch with a conflict.  I was being encouraged to merge my PR ASAP.

Here's what the conflict involves:

```
<<<<<<< issue_1017 -- Incoming Change
            # Because we're pulling all the data down into one query we need to manipulate it a little bit
            # This creates a new resource df of the resource_access with specific columns
            resource_df = resource_access_df.filter(["resource_id", "resource_type", "name"], axis=1)
            resource_df = resource_df.drop_duplicates(["resource_id"])
=======
            # Make resource data from resource_access data
            resource_df = resource_access_df.copy(deep=True)
            resource_df = resource_df.drop(
                columns=['user_id', 'access_time', 'course_id'])
            resource_df = resource_df.drop_duplicates('resource_id')

            # set resource_id as index for pangres.upsert
            resource_df: pd.DataFrame = resource_df.set_index('resource_id')
>>>>>>> master -- Current Change
```

Where the upper part is your code and the lower one contains my change.  I think it can be resolved like this:

```suggestion
            # Make resource data from resource_access data
            resource_df = resource_access_df.filter(
                columns=['resource_id', 'resource_type', 'name'])
            resource_df = resource_df.drop_duplicates('resource_id')
```

It's mostly a matter of choosing your changes instead of mine.

lsloan: I forgot to say:  The `resource_df.set_index()` part of the conflict threw me off.  At first I thought you had dropped that from the code because it wasn't necessary.  However, I see that you still have it, but I had reformatted it in my code.  It was different enough to confuse the conflict detection in the diff.  (Also the conflict editor was WAY off on line numbers.  It said these changes were around line 366, when they're actually around 305.)

dashboard/cron.py
```diff
@@ -303,11 +303,8 @@ def update_with_bq_access(self):
             logger.debug(resource_access_df)
 
             # Because we're pulling all the data down into one query we need to manipulate it a little bit
-            # Make a copy of the access dataframe
-            resource_df = resource_access_df.copy(deep=True)
```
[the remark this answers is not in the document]
jonespm: I don't think the deep copy was necessary. But I was less experienced with Pandas when this was committed over a year ago. 

It looks like either dropping the unneeded or filtering just specified needed columns works fine.

dashboard/cron.py
jonespm: I'm not [seeing columns](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.filter.html) as a kwarg for filter. I do see `axis='columns'` which is clearer. Is this what you might have meant?

jonespm:
Yeah, I didn't mean *broken* but what's another good word for conflicted WRT code? I don't think saying the _code is conflicted_ sounds good either. ;) That sounds a little too emotional.

dashboard/cron.py
jonespm: It also looks like the default for axis here is 'columns' for DataFrame so it's probably not necessary.

lsloan: I admit I didn't look up the docs for `filter()`, but `drop()` supports the `columns=` argument, so I assumed `filter()` would, too.

If the default axis is columns, then it makes sense to drop it.  If you try `columns=` with with `filter()` and it works, then I'd use that.  When it comes to working with pandas, I remember @ssciolla's comment to me in a past PR that it's better to not be so clever that it becomes complicated for future maintainers.  So I avoided chaining too many methods together and I think using named arguments makes them easier to understand.

Sorry for soapboxing. 😆

dashboard/models.py
jonespm: Yeah it's pretty generic right now, determined by just a few times. This was desired by the project team to just have a few top level types like **files, videos, reading.** 

I do think we could add a feature (maybe I even did where we could have a multi-select that would be like

```
Videos -> MiVideo
          Lecture Capture
          Zoom
```

Or something like that but not worth future proofing around this for now since this could relatively easily be added.

dashboard/cron.py
jonespm: Yeah this code didn't have much in it, resolving this shouldn't take long, probably even less time than replying to these comments! ;)

And yeah, all of this could use a good refactoring to make it easier to manage, though we've been trying to avoid too many ETL changes pending future backend changes (like UDP).

jonespm: Okay, just resolving it like this.

dashboard/cron.py
jonespm: Yeah, this is throwing an error

`TypeError: filter() got an unexpected keyword argument 'columns'`
Without axis looks like it works. I also hit #1025 when testing this and assigned that to you, should be an easy fix. I just worked around it with the ignore parameter but thinking about it, it seems like it would be worth printing a warning in the logs.

PR #1024, opened 2020-08-14
#1006 When adding a new course, need to update resource_access from the beginning of the term

### LINKED ISSUE #1006: When adding a new course, need to update resource_access from the beginning of the term
If a course is added part way through the semester (even a few days/week later) it's events will only populate from the time it's added.

To fix this I believe we'll have to make the job 

- [x] Check if there are any courses newly added since the last run

- [x] If there are, run the update from the start of the term for the new courses

- [x] Add a new column to the courses table to indicate when the data was last refreshed for that course

The way Bigquery works it doesn't matter if you run it for select courses or all courses, it's the same cost currently, so we'll just have to remove all of the events since the start of this new courses term and update them.

I believe checking if there are new courses might be useful for other reasons (like if we wanted the cron to run more frequently). For instance we *could* run the cron hourly but only if there's new courses, otherwise we'd just do the daily run.

_Originally posted by @jonespm in https://github.com/tl-its-umich-edu/my-learning-analytics/issues/973#issuecomment-669401541_

jonespm:
Fixes #1006


jonespm:
I think I need to change the is_any_course_new to return the earlier start date on the new course.

Otherwise it will run from the beginning of all the courses.

So this is why this is draft.

jonespm:


```diff

Issues
======
- Added 3
           
```


See the complete overview on [Codacy](https://app.codacy.com/manual/ITS_Teaching_And_Learning/my-learning-analytics/pullRequest?prid=6026291&bid=19751148)

.vscode/launch.json
```diff
@@ -26,7 +26,8 @@
                 }
             ],
             "port": 3001,
-            "host": "localhost"
+            "host": "localhost",
+            "justMyCode": false
```
jonespm, line 30: This change is unrelated but helped debugging this as it let me look into some libraries for issues and is a minor change.

jonespm:


```diff

Issues
======
- Added 3
           

Complexity decreasing per file
==============================
+ dashboard/cron.py  -1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/ITS_Teaching_And_Learning/my-learning-analytics/pullRequest?prid=6026291&bid=19751148)

jonespm:


```diff

Issues
======
- Added 6
           

Complexity decreasing per file
==============================
+ dashboard/cron.py  -1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/ITS_Teaching_And_Learning/my-learning-analytics/pullRequest?prid=6026291&bid=19751148)

jonespm:


```diff

Issues
======
- Added 5
           

Complexity decreasing per file
==============================
+ dashboard/cron.py  -1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/ITS_Teaching_And_Learning/my-learning-analytics/pullRequest?prid=6026291&bid=19751148)

dashboard/common/utils.py
```diff
@@ -1,28 +1,14 @@
...
-from dashboard.models import Course
-from typing import Union
+from dashboard.common import db_util
+from dashboard.models import Course, ResourceAccess
 
 
 logger = logging.getLogger(__name__)
 
-
-def find_earliest_start_datetime_of_courses():
```
jonespm, line 13: I moved this into models.py

pushyamig:
I will review it now

dashboard/models.py
```diff
@@ -171,8 +203,9 @@ class Course(models.Model):
     GRADING_CHOICES = [('Percent', 'Percent'), ('Point', 'Point'), ]
     show_grade_type = models.CharField(verbose_name="Show Grade Type", max_length=255,
                                          choices=GRADING_CHOICES, default='Percent')
+    last_cron_run = models.DateTimeField(verbose_name="Last time cron ran for this course", null=True, blank=True)
```
ssciolla, line 206: Couple thoughts I had here, but I'm not sure how important they are.

1) I don't love `last_cron_run` as a variable name. `cron` as a term is related to a specific technology, which aren't even using anymore (though we are still using `DjangoCronJob`). I think I'd prefer something like `data_last_updated_at`. That could still potentially be useful in a world where we don't ETL (though maybe not).

2) In its current implementation, I believe this is just going to set all of the `Course.last_cron_run` values to the same value (which is repetitive). If I expected this to be the same in perpetuity, I'd say this needs to be normalized more. `CronRun` or `DataUpdate` could be a table, and we could FK to the `course`, allowing `null`s as an indicator of "this course (and child data) wasn't updated as part of any cron run yet". If we moved to a workflow where individual courses could be updated alone or as a group (suggested recently by @chrisrrowland), then **maybe** it makes sense to do it this way. Still, if you're ever going to run multiple at the same time and not distinguish between when one course was updated and when the next was, a new table and FK may make sense.

My two cents. Feel free to disagree. To a fault, I tend to prefer database normalization, though I know it can take more time and does not always seem worth it. In fact, I think I elected not to do something similar in `placement-exams` recently.

dashboard/cron.py
```diff
@@ -656,6 +648,9 @@ def do(self):
         if courses_added_during_cron:
             logger.warning(f'During the run, users added {len(courses_added_during_cron)} course(s): {courses_added_during_cron}')
             logger.warning(f'No data was pulled for these courses.')
+            # Set all of the courses to have been updated now (this is the same set update_course runs on)
```
ssciolla, line 651: I don't think this should be indented?

dashboard/cron.py
```diff
@@ -656,6 +648,9 @@ def do(self):
...
+
+        Course.objects.get_supported_courses().update_all_last_cron_run()
```
ssciolla, line 653: I don't think this is what you want. I think you want `Course.objects.filter(self.valid_locked_course_ids).update_all_last_cron_run()`, or something similar. You don't want to update courses added during the cron.

Aside from the course set you are updated, I'm a little confused by the chaining of `get_supported_courses().update_all_last_cron_run()`. Doesn't `get_supported_courses` return a list of `int`s and `update_all_last_cron_run` act on a `QuerySet`?

Sorry if I'm missing something here. I'm not that familiar with defining custom `QuerySet` objects.

ssciolla (commented):
Some initial comments, based on reading the code. I will do actual testing this afternoon.

dashboard/models.py
```diff
@@ -150,14 +151,45 @@ def get_supported_courses(self):
...
+            logger.info(f"No course listed. Returning None as the earliest_start_datetime_of_course. ")
+        return earliest_start
 
+    def update_all_last_cron_run(self):
+        """ Updates the last cron run to now for all courses in QuerySet
+        """
+        # Finally update the time on all courses updated
+        self.update(last_cron_run=datetime.datetime.now())
+
+    def last_cron_run(self) -> Optional[datetime.datetime]:
```
pushyamig, line 177: This function can be renamed to `get_last_cron_run`

dashboard/models.py
```diff
@@ -150,14 +151,45 @@ def get_supported_courses(self):
...
+        """ Returns the datetime of the last cron run of all courses
+            This checks for any courses where the last_cron_run value is null.
+
+        :return: datetime. Either the earliest or None
+        :rtype: None
+        """
+        new_courses = self.filter(last_cron_run__isnull=True)
```
pushyamig, line 184: This variable can be renamed `new_courses_from_last_cron_run`. It is bit long but it make sense  to me

dashboard/models.py
```diff
@@ -185,14 +218,14 @@ def course_date_range(self):
             start = self.term.date_start
         else:
             logger.warning("No date_start value was found for course or term; setting to current date and time")
-            start = datetime.now(pytz.UTC)
+            start = datetime.datetime.now(pytz.UTC)
```
pushyamig, line 221: NitPick: I feel `datetime.datatime` can have a better import like `from datetime import datetime`. This is something going over all the file. But next time somebody touch this should fix it

.vscode/launch.json
pushyamig: So changing this let you step into underline library code like pandas etc? I though you can do that even now with step into over that line of code

pushyamig:
This code is working.
I ran the cron before the running this code locally, when I added a course and ran cron I can see the data with resources access, then I added another course( from same/different term) and ran the cron but still i did not see the data with subsequent courses. 

With this PR I did not see that issue. I don't want to understand why was that an issue with current master on local as it might me less concerning and nobody else did not complaint.

I feel this can be merged

pushyamig (approved):

dashboard/cron.py
ssciolla: Hm, looks like it's still a `QuerySet`, it just evaluates to a list of `int`s instead of a list of `Course` objects, I guess.

dashboard/models.py
```diff
@@ -150,14 +151,45 @@ def get_supported_courses(self):
...
+            logger.info(f"Earliest start datetime for all courses: {earliest_start.isoformat()} found in course {earliest_course.canvas_id}")
+        else:
+            logger.info(f"No course listed. Returning None as the earliest_start_datetime_of_course. ")
+        return earliest_start
 
+    def update_all_last_cron_run(self):
+        """ Updates the last cron run to now for all courses in QuerySet
+        """
+        # Finally update the time on all courses updated
+        self.update(last_cron_run=datetime.datetime.now())
```
ssciolla, line 175: This is throwing a warning. I think we should use `datetime.datetime.now(tz=pytz.UTC)` or `datetime.datetime.now(tz=timezone.utc)` (with the latter, you need `from django.utils import timezone`).

```
RuntimeWarning: DateTimeField Course.last_cron_run received a naive datetime (2020-09-10 16:11:00.695774) while time zone support is active.
```

dashboard/models.py
ssciolla: I tend to agree, but it's not a huge deal. I prefer to import as `from datetime import datetime, timedelta`.

dashboard/cron.py
```diff
@@ -246,18 +248,9 @@ def update_with_bq_access(self):
...
-        try:
-            latest_resource = ResourceAccess.objects.latest('access_time')
-            latest_resource_time = latest_resource.access_time
-        except ResourceAccess.DoesNotExist:
-            latest_resource_time = course_start_time
-
-        logger.info(f"Earliest Start: {course_start_time} Latest resource: {latest_resource_time}")
+        last_cron_run = Course.objects.last_cron_run()
 
+        status += deleteAllRecordInTable("resource_access", f"WHERE access_time > %s", [last_cron_run,])
```
ssciolla, line 253: Feels like there is a reason for this, but having trouble articulating it specifically. Can you explain? If `last_cron_run` uses a course start date, it'd reset `resource_access` essentially. If it's the last time the cron ran, it'd wipe out records after that, but would there be any?

dashboard/models.py
ssciolla: I think maybe 1) is worth considering, but given time constraints and possible refactoring in the future, I'm fine with  ignoring 2) for now.

dashboard/models.py
ssciolla: Should we be using a timestamp from the beginning of the cron run? Feels like this way we might lose some data if the events occur during the run?

dashboard/cron.py
ssciolla: I'm still concerned about the use of `Course.objects.get_supported_courses` here. If a course is added during the cron, you're going to give it a `last_cron_run` value, even though it wasn't updated, right?

ssciolla (changes requested):
More comments. I think I've identified a couple of bugs that need to be addressed, but please set me right if they aren't really issues.

jonespm:


```diff

Issues
======
- Added 5
           

Complexity increasing per file
==============================
- dashboard/migrations/0019_course_last_cron_run.py  1
         

Complexity decreasing per file
==============================
+ dashboard/cron.py  -1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/ITS_Teaching_And_Learning/my-learning-analytics/pullRequest?prid=6026291&bid=19751148)

.vscode/launch.json
jonespm: Yeah, the idea is whenever you're on a line that goes into a library (like Pandas) you can Step Into and it will go into that code. It's skipped over it for me without this unless I actually downloaded the library and added it to my workspace. I think I was mostly using it for the upsert library in this case.

dashboard/cron.py
```diff
@@ -66,18 +67,20 @@ def util_function(data_warehouse_course_id, sql_string, mysql_table, table_ident
 
 
 # execute database query
-def executeDbQuery(query):
+def executeDbQuery(query: str, params: List=None) -> ResultProxy:
```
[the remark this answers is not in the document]
jonespm: I didn't even really notice it wasn't. I generally don't PEP-8 everything as I'm going through it, I was just adding type hinting here with the new parameter. I could update these though.

jonespm:


```diff

Issues
======
- Added 5
           

Complexity increasing per file
==============================
- dashboard/migrations/0019_course_last_cron_run.py  1
         

Complexity decreasing per file
==============================
+ dashboard/cron.py  -1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/ITS_Teaching_And_Learning/my-learning-analytics/pullRequest?prid=6026291&bid=19751148)

dashboard/cron.py
jonespm: Yeah, fixed, I think I was thinking of putting it in here for some reason but it didn't make sense in the end, just didn't fix the comment.

dashboard/models.py
jonespm: I can update this, I feel like there was some reason I wasn't but I don't remember what it was.

dashboard/models.py
jonespm: Usually in Python when you're thinking of adding a "get" prefix, it really should be a [@property](https://docs.python.org/3/library/functions.html#property). So I can change these to use that decorator.

dashboard/models.py
jonespm: Okay fixed

dashboard/cron.py
jonespm: Yeah, I think I was back and forth between putting it into this block but just wanted something working. I'll look at this and change it to your suggestion.

dashboard/models.py
jonespm: Yeah, good point it should probably lock down the time at the beginning. If there's a little duplicate next run it's not a problem but if it's too late then it is. I'll capture the time at the start of the run.

dashboard/models.py
jonespm: It's not working as I expected, I'm just going to use your suggestion to speed it up.

dashboard/models.py
jonespm: 1) Okay, that sounds like a good suggestion. Easier to make it now than change it. I think I'm going to drop the `_at` and just call it `data_last_updated`. 
2) Currently this is the case. The only reason that this column was added it that some on the dev calls say they wanted a field in the database to see when the course was updated. We update all courses together at the same time (except for new courses that haven't yet been run)

Because it would just be a 1:1 table and all of these will have a value, I opted for a column rather than a FK to another table. 

I wrote this method so it could take a QuerySet and operate on a limited set of values rather than all of them. I'm planning on fixing another of your suggestions to only update the locked courses, so it's possible a course added during the run won't get updated. There might also be cases we want to disable courses or if all of the views are disabled don't run the ETL on that course. In that case those won't get updated either. But we don't do that yet.

dashboard/cron.py
jonespm: Yeah my idea is it last_cron_run (now renamed to data_last_updated) would return the start date of the new courses. This is either going to be the date set on the course, the start of the term or "now" if all those fail. (That logic is in course_date_range)

So there's a chance it will wipe it all out. However if we have multiple terms (say Winter 2019 and Winter 2020) and the new course is just Winter 2020, it will just wipe out data up to Winter 2020 and repopulate that. This is really just protecting from a full database wipeout for really old courses. Maybe that isn't as necessary as I'm thinking.

dashboard/models.py
ssciolla: I'm fine with whatever you want to do here. But it wouldn't be 1:1, it'd 1:M from `DataUpdate` to `Course`. Most of the courses would share the same timestamp, so to avoid repeating the data in the database, you just store it once on `DataUpdate`.

jonespm:


```diff

Issues
======
- Added 4
           

Complexity increasing per file
==============================
- dashboard/migrations/0019_course_last_cron_run.py  1
         

Complexity decreasing per file
==============================
+ dashboard/cron.py  -1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/ITS_Teaching_And_Learning/my-learning-analytics/pullRequest?prid=6026291&bid=19751148)

dashboard/models.py
jonespm: That really becomes a 1:M with the CronJobLog then we wouldn't need a new table.

Maybe that's most appropriate, but I'd also like to remove this library since we really aren't using it and it hasn't been updated since 2018. I think this might be worth fixing if we ever wrote our own more simplified cronjob, but am reluctant to add a FK to that table.

dashboard/cron.py
ssciolla: Oh, I was thinking that the `get_last_cron_run` was acting on all courses, not just the new ones. This is starting to make more sense. So is this mainly to reduce duplicates when adding new data? `pangres` upsert would handle them, but you're making things easier?

dashboard/cron.py
jonespm: Yeah, it returns a QuerySet which seems convenient for chaining other calls. 

I could write it like you have it though if it was in the if statement.

dashboard/cron.py
jonespm: Yeah, this is really just to set the *best* date for the BigQuery date. Pangres is fine for updating resources but we don't have any easy way to upsert into the resources_access, so it has to delete up to a date and re-insert.

ssciolla: Oh, gotcha, thought the upsert was happening here, too. I'm all caught up. This should be fine.

dashboard/cron.py
jonespm: I guess it doesn't need to be, but if there aren't any then it won't do anything. So I guess same thing.

jonespm: Yeah, I see what you mean, it should just update the valid_locked ones, so it's fine where it is.

jonespm:
Okay I think I got *all* of the issues mentioned here! And it's still running!

dashboard/admin.py
```diff
@@ -62,7 +62,7 @@ class CourseAdmin(admin.ModelAdmin):
     form = CourseForm
     list_display = ('canvas_id', 'name', 'term', 'show_grade_counts', 'course_link', '_courseviewoption', 'show_grade_type')
     list_select_related = True
-    readonly_fields = ('term',)
+    readonly_fields = ('term','data_last_updated',)
```
ssciolla, line 65:
```suggestion
    readonly_fields = ('term', 'data_last_updated',)
```

dashboard/models.py
```diff
@@ -171,8 +205,9 @@ class Course(models.Model):
     GRADING_CHOICES = [('Percent', 'Percent'), ('Point', 'Point'), ]
     show_grade_type = models.CharField(verbose_name="Show Grade Type", max_length=255,
                                          choices=GRADING_CHOICES, default='Percent')
+    data_last_updated = models.DateTimeField(verbose_name="Last time data last updated for this course", null=True, blank=True)
```
ssciolla, line 208:
```suggestion
    data_last_updated = models.DateTimeField(verbose_name="Time data for this course was last updated", null=True, blank=True)
```

ssciolla: Would have to re-generate the migration for that. Not a huge deal, but it was weird to me "last" was repeated.

dashboard/models.py
```diff
@@ -150,13 +151,46 @@ def get_supported_courses(self):
...
+        else:
+            logger.info(f"No course listed. Returning None as the earliest_start_datetime_of_course. ")
+        return earliest_start
+
+    def update_all_data_last_updated(self, updated_date=None):
+        """ Updates the last cron run to now for all courses in QuerySet
+        """
+        # Finally update the time on all courses updated
+        if updated_date is None:
+            updated_date = datetime.now(pytz.UTC)
```
ssciolla, line 176: Nit-pick: What's the reason for making `updated_date` optional? I don't think you're doing this now anywhere, and even if you wanted to, you could just pass `datetime.now(tz=pytz.UTC)` to this.

You could add a type hint here for `updated_date` and the return type (`None`).

ssciolla: Now that I'm looking at this more, would it be so bad to remove this entirely and just use `SomeQuerySet.update(data_last_updated=some_datetime)`?

dashboard/cron.py
```diff
@@ -246,18 +249,9 @@ def update_with_bq_access(self):
...
-        try:
-            latest_resource = ResourceAccess.objects.latest('access_time')
-            latest_resource_time = latest_resource.access_time
-        except ResourceAccess.DoesNotExist:
-            latest_resource_time = course_start_time
-
-        logger.info(f"Earliest Start: {course_start_time} Latest resource: {latest_resource_time}")
+        data_last_updated = Course.objects.get_data_last_updated()
 
+        status += delete_all_records_in_table("resource_access", f"WHERE access_time > %s", [data_last_updated,])
```
ssciolla, line 254: So this works if `data_last_updated` is `null`, I guess?

dashboard/models.py
```diff
@@ -150,13 +151,46 @@ def get_supported_courses(self):
...
+        """
+        sorted_courses = sorted(self.all(), key=lambda course: course.course_date_range.start)
 
-    def get_supported_courses(self):
-        return self.get_queryset().get_supported_courses()
+        earliest_start = None
+        if len(sorted_courses) > 0:
+            earliest_course = sorted_courses[0]
+            earliest_start = earliest_course.course_date_range.start
+            logger.info(f"Earliest start datetime for all courses: {earliest_start.isoformat()} found in course {earliest_course.canvas_id}")
```
ssciolla, line 166:
```suggestion
            logger.info(f"Earliest start datetime for CourseQuerySet: {earliest_start.isoformat()} found in course {earliest_course.canvas_id}")
```

dashboard/models.py
```diff
@@ -150,13 +151,46 @@ def get_supported_courses(self):
...
+        else:
+            logger.info(f"No course listed. Returning None as the earliest_start_datetime_of_course. ")
```
ssciolla, line 168:
```suggestion
            logger.info(f"No courses in CourseQuerySet; returning None as the earliest_start_datetime")
```

dashboard/cron.py
ssciolla: Doesn't look fixed?

dashboard/cron.py
ssciolla: I guess `null` is impossible, since `update_with_bq_access` won't be invoked if `self.locked_valid_course_ids` is empty.

dashboard/cron.py
```diff
@@ -246,18 +249,9 @@ def update_with_bq_access(self):
...
-
-        # the earliest latest date of all resources_accessed
-        try:
-            latest_resource = ResourceAccess.objects.latest('access_time')
-            latest_resource_time = latest_resource.access_time
-        except ResourceAccess.DoesNotExist:
-            latest_resource_time = course_start_time
-
-        logger.info(f"Earliest Start: {course_start_time} Latest resource: {latest_resource_time}")
+        data_last_updated = Course.objects.get_data_last_updated()
```
ssciolla, line 252: Does this need to be limited to the locked course list also?

ssciolla (commented):
My apologies; this is looking pretty good. Just a couple more questions and some textual change suggestions.

dashboard/models.py
jonespm: Yeah, I can fix this in the migration too.

dashboard/cron.py
jonespm: Yeah, that check blocks the case of no courses and no date getting into here.

dashboard/models.py
```diff
@@ -150,13 +151,46 @@ def get_supported_courses(self):
             logger.info("Courses did not exist", exc_info = True)
         return []
```
jonespm, line 152: This isn't part of the change but I think this needs to return an empty QuerySet. Should probably add a typehint here. 

It sounds like the best way to do it is
Course.objects.none()

jonespm:


```diff

Issues
======
- Added 2
           

Complexity increasing per file
==============================
- dashboard/migrations/0019_course_data_last_updated.py  1
         

Complexity decreasing per file
==============================
+ dashboard/cron.py  -1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/ITS_Teaching_And_Learning/my-learning-analytics/pullRequest?prid=6026291&bid=19751148)

dashboard/models.py
jonespm: I created #1090 to improve this. I don't want to fix this because I'd like to remove django-cron and fixing this correctly would involve a FK to that table. If we had a simpler table to manage the crons (or even just re-used this table locally without that package) it would be easier.

ssciolla: Yeah, I'm in favor of removing `django-cron`. Doesn't seem like it would be too much work to just re-write this as our own class. As I've already indicated, no worries about dealing with this now.

dashboard/cron.py
jonespm: Yeah, good catch.

dashboard/models.py
jonespm: Since I'm only calling it one it seems fine. I was reading about how explicitly named methods can be better than more complex ones on an object, but this one is pretty straightforward.

jonespm:


```diff

Issues
======
- Added 2
           

Complexity increasing per file
==============================
- dashboard/migrations/0019_course_data_last_updated.py  1
         

Complexity decreasing per file
==============================
+ dashboard/cron.py  -1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/ITS_Teaching_And_Learning/my-learning-analytics/pullRequest?prid=6026291&bid=19751148)

ssciolla (approved):
I think this looks good! Let's merge so we can get more eyeballs on it in QA.

PR #1026, opened 2020-08-18
Enable mypy type hinting for VSCode

### LINKED ISSUE #1022: Enable mypy type hinting for VScode

As we're starting to use type hints, we should minimally enable the mypy type hinting checker for VSCode. Otherwise any errors might not be noticed.

jonespm:
Fixes #1022

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,8 +1,9 @@
 {
-    "python.formatting.provider": "autopep8",
-    "python.formatting.autopep8Args": ["--ignore=E501", "--max-line-length=120"],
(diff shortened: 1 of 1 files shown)


jonespm:
Without this no problems for mypy (or other problems) seem to appear in VSCode. This does not enable pylint or flake8 since they will pop up a lot of new problems when enabled.

jonespm:
If we merge this VSCode will start reporting some problems but there's an issue where it will only [report problems in the files opened](https://github.com/microsoft/vscode/issues/13953). This is good but also things we should fix.

Issue #1027, opened 2020-08-19
Fix existing mypy issues in dashboard


ssciolla:
Thank you for contributing to this project!

- Make sure to search the [issues](https://github.com/tl-its-umich-edu/my-learning-analytics/issues) for duplicates first!

## Expected behavior (A clear and concise description of what you expected to happen) : 

We should try to resolve known type issues flagged by mypy.

## Describe the bug (Tell us what happens instead of the expected behavior) :

Much of the application is not type-hinted (which is okay for now), but there are a few areas that are, and running `mypy` flags a couple issues. These should be fixed. Also, `mypy` is using Python 3.7 automatically, so we should add a `mypy.ini` to specify 3.8, so we can use features like `TypedDict`.

pushyamig:
I ran `mypy dashboard` and I am seeing it is reporting issue and since most of the files in Myla are not type hinted this is ok. Most of the error are of this nature
`Skipping analyzing 'django.contrib': found module but no type hints or library stubs`

How do I test if mypy is using python 3.8 instead of python 3.7?

jonespm:
In VSCode the version of Python it's using is on the bottom toolbar.
![image](https://user-images.githubusercontent.com/27447/91065550-9f94bb80-e5fe-11ea-843b-ad2d47347df8.png)

If you don't have 3.8 installed locally you probably could do that with [Homebrew](https://opensource.com/article/19/5/python-3-default-mac).

ssciolla:
@pushyamig, I think `TypedDict` is only available in 3.8, so that may be an indication of whether it's working (there are errors related to the issues I tried to fix with that).

jonespm:
I think the best thing for the future would be if we had a CI script that would do the check rather than relying on the localhost. This could be done in Jenkins or Travis or locally in a Docker container.

pushyamig:
I have 3.8 installed and that's why my VENV running, So I  guess I did not see any issue with TypeDict as mentioned. So I wan't sure why the specific versioning importance. I guess if somebody are still running 3.7 they might see an error with type hinting.

pushyamig:
Test Passes

PR #1028, opened 2020-08-19
Fix current mypy issues in dashboard (#1027)

ssciolla:
This PR adds a `mypy.ini` file specifying the Python 3.8 version and fixes mypy errors currently flagged by running `mypy` on the command line. An error flagged in a migration file was ignored. This PR aims to resolve issue #1027.

### DIFF
--- a/dashboard/admin.py
+++ b/dashboard/admin.py
@@ -1,3 +1,5 @@
+from typing import Tuple
+
 from django import forms
 from django.conf import settings
 from django.contrib import admin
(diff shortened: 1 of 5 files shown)


jonespm:
I fixed a few of these on #1024 on the code I was using and this will cause a major conflict on that as I moved stuff out of db_utils into models, but this is fine.

ssciolla:
@jonespm, only conflict I see is the addition of `Any` in `cron.py`, which should be easy to resolve?

jonespm:
Yeah, I moved many methods method out of db_utils because I'm seeing how doing them in the models and making better use of the ORM is simplifying some of this. ;)

Maybe it won't be too bad and I was looking at a different method. I'll rebase now.

PR #1032, opened 2020-08-20
Need to add an index after filter for pangres

jonespm:


jonespm:
I'm closing this since we'll just fix it in #1033

PR #1033, opened 2020-08-20
Warn about old queries, but support them (#1025)

### LINKED ISSUE #1025: ETL incompatible with old queries
Existing queries should continue to work after #826 change.

This exception is thrown now when using existing queries without the user_login_name column.
```
  File "/usr/local/lib/python3.8/site-packages/pandas/core/indexes/base.py", line 5018, in drop
    raise KeyError(f"{labels[mask]} not found in axis")
KeyError: "['user_login_name'] not found in axis"
```

It looks like we *could* just add errors="ignore" to this statement to keep it going. I'm thinking it might be better to do a try/except instead here to log a message to the user that they may want to look at updating to use the new query format (if necessary).

```
                resource_access_df = resource_access_df.drop(
                    columns='user_login_name', errors="ignore")
```

lsloan:
Fixes #1025.

**_NOTE:_** I recommend you tell GitHub to ignore whitespace when you view the diffs of this PR.  I changed just a few lines of code, but also indented a block of code to be inside a conditional.  GitHub's diff shows this as a big change unless whitespace is ignored.  (I wish Python used brackets!)

This change skips over converting `user_login_name` to `user_id` during the ETL's event query processing if the name column isn't available.

### DIFF
--- a/dashboard/cron.py
+++ b/dashboard/cron.py
@@ -306,49 +306,53 @@ class DashboardCronJob(CronJobBase):
                          f'{resource_access_df}\n'
                          f'{resource_access_df.dtypes}')
 
-            # process data which contains user login names, not IDs
-            if len(resource_access_df[
-                       resource_access_df['user_id'] == -1]) > 0:
-                login_names = ','.join(
-                    map(repr, resource_access_df['user_login_name']
-                        .drop_duplicates().dropna().values))
-
-                logger.debug(f'login_names:\n{login_names}')
-
-                # get user ID as string because pd.merge will convert int64 to
-                # scientific notation; converting SN to int64 causes
-                # Obi-Wan problems (off by one)
-                user_id_df = pd.read_sql(
-                    'select sis_name as user_login_name,'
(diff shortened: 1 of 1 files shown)


dashboard/cron.py
```diff
@@ -365,6 +369,8 @@ def update_with_bq_access(self):
             # Make resource data from resource_access data
             resource_df = resource_access_df.filter(["resource_id", "resource_type", "name"])
             resource_df = resource_df.drop_duplicates(["resource_id"])
+            # pangres.upsert() requires DataFrame to have index
+            resource_df = resource_df.set_index('resource_id')
```
jonespm, line 373: I just added this back in https://github.com/tl-its-umich-edu/my-learning-analytics/pull/1032

Realized I must have dropped it when I resolved some conflicts

jonespm: I think we can just merge this one and close the other.

lsloan: Yeah, it wouldn't run for me without that index.

PR #1035, opened 2020-08-20
Fixes #1034 - Enable pylint to catch errors and fatal issues.

### LINKED ISSUE #1034: Enable pylint in VSCode at least to catch errors

I was noticing that we don't have PyLint enabled at all to even catch errors. This has been making it harder to debug in VSCode. #1023 will turn on support for PEP-8 stuff later and build on this.

jonespm:
Currently disabled for any lower messages like warnings as there are many in the code. Maybe we want these on but really need to see the errors.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -11,5 +11,6 @@
     "python.linting.mypyEnabled": true,
     "[javascript]": {
         "editor.tabSize": 2
-    }
+    },
+    "editor.formatOnSaveMode": "modifications"
 }
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,7 +1,12 @@
 {
-    "python.linting.pylintArgs": ["--load-plugins=pylint_django"],
     "python.linting.enabled": true,
-    "python.linting.pylintEnabled": false,
+    "python.linting.pylintEnabled": true,
+    "python.linting.pylintArgs": [
+        "--disable=C0111", // missing docstring
+        "--load-plugins=pylint_django",
+        "--disable=all",
+        "--enable=F,E" // Only enable errors for now, will turn on warnings in a future fix
+    ],
     "python.linting.flake8Enabled": false,
     "python.linting.mypyEnabled": true,
     "[javascript]": {
--- a/requirements.txt
+++ b/requirements.txt
@@ -50,3 +50,6 @@ django-csp==3.6
 
 jwcrypto==0.8
 pycryptodome==3.9.8
+# These are for local development
+pylint==2.5.3
+pylint-django==2.3.0


.vscode/settings.json
```diff
@@ -1,7 +1,12 @@
 {
-    "python.linting.pylintArgs": ["--load-plugins=pylint_django"],
     "python.linting.enabled": true,
-    "python.linting.pylintEnabled": false,
+    "python.linting.pylintEnabled": true,
+    "python.linting.pylintArgs": [
+        "--disable=C0111", // missing docstring
```
lsloan, line 5: Does VSCode support comments in its config file JSON?

lsloan:
I'd like to review this, but I do not use VSCode.  It'd be great if this were done in a more general way, to support linting via CLI or other IDEs.

.vscode/settings.json
jonespm: It actually does.

https://code.visualstudio.com/docs/languages/json#_json-with-comments

Normal JSON this would be a problem in.

I believe it's more [json5](https://json5.org/) style.

jonespm:
I think the *ideal* would be.
* We decide what linter(s) we want to support
* We specify the linting rules we want in their appropriate config files. I think for pylint this is `.pylintrc`. mypy looks like `.mypy.ini`. 
* Then we have a script that can run it on the command line and configuration that will also run it for the different IDE's.

I feel like this is just a "quick fix" as VSCode wasn't reporting any obvious errors and it made it much harder to refactor. 

I could see if a .pylintrc file would work instead of these args. That's not going to happen today though.

requirements.txt
```diff
@@ -49,3 +49,7 @@ PyLTI1p3==1.7.0
 django-webpack-loader==0.6.0
 jsonschema==3.2.0
 django-csp==3.6
+
+# These are for local development
+pylint==2.5.3
+pylint-django==2.3.0
```
ssciolla, line 55: What about `mypy`?

But more importantly, shouldn't we specify dev dependencies somewhere else? Keeping images for deployment small -- where these packages are mostly irrelevant -- seems like the ideal. I'm not sure what the best practice is these days, though I found this older post: https://stackoverflow.com/questions/17803829/how-to-customize-a-requirements-txt-for-multiple-environments

Most of the options would also require changing `Dockerfile`. We could postpone adding these dependencies to a later date, just change `.vscode`.

ssciolla (commented):
Do I have to do anything special to get `pylint` linting errors/warnings to show up in VSCode? It doesn't seem to be working for me when I just checkout your branch.

requirements.txt
jonespm: Did you restart VSCode? Or at least close/re-open the workspace editor. I don't think it re-reads this the first time. 

Yeah, I wasn't sure what to do with these dependencies either. I figured that someone developing on localhost would be pip install'ing the requirements so their IDE doesn't complain about missing dependencies but some aren't needed in prod. 

It's not big but it doesn't look hard to do from that StackOverflow article. I don't think we currently have anything else that's for local dev only and they aren't that big so I didn't think about that.

jonespm: I also think it's possible that VSCode will prompt to install these if it needs them, so this might not be needed either.

pushyamig:
i Tried to open the VScode after this setting, and it did not ask for installing the pylint-django automatically, I guess we need to install it through pip

jonespm:
I like @ssciolla about the chained requirements but I'm not going to jump on fixing this for this release. Even though without this configured VSCode doesn't report any problems at all which has made coding a lot more annoying. (And I generally have been pulling this patch in locally just to get work done)

This is also not a code change, it's just affecting local dev (aside from the 2 requirements that take up a little space on the image)

ssciolla:
> I like @ssciolla about the chained requirements but I'm not going to jump on fixing this for this release.

Okay, @jonespm, so just to be clear, do you want to 1) just forget about this for the current release; 2) update this removing the package additions; or 3) merge it as is?

jonespm:
I already have this bumped off to the next release, we don't have to worry about it this week. I'll try #2 but haven't gotten to it yet.

requirements.txt
jonespm: I created this idea as a new issue. I don't think it's a big deal that these are included as they are very small dependencies, a few MB total. I think if we cared about image size there's a lot more cleanup ideas, including one I issue I just created that cuts down almost 200MB.

PR #1004, opened 2020-08-04
I952 package update to latest 

pushyamig:
This PR aims to resolve #952. This PR has dependency on this [PR](https://github.com/tl-its-umich-edu/my-learning-analytics/pull/997) related to Graphene-django update

As part of LTI 1.3 update Django version is updated to 3. But this PR is aiming to update all the latest version pip package.

Certain packages don't have an update since couple of years. I feel either we have to revisit then for alternatives

Pinax-eventlog library had an latest release of 4.0.1 and update to that break for Mysql as migration file is incorrect for mysql more on that [here](https://github.com/pinax/pinax-eventlog/issues/32).

PR #1045, opened 2020-08-24
#1004 403 error message wording change

### LINKED ISSUE #1044: Change the wording for the 403 error message
When an instructor adds MyLA to a course, and the cron has not yet run, students get an error message when they try to access MyLA.  This is because MyLA does not have the students listed as members of the course until after the cron runs.  

The current error messages says "You do not have access to course 123456."

Change the wording to "You do not have access to course 123456 or data for the course is still being loaded.  Check back in 24 hours."

For a future release, we can work on a solution that displays the same icon that shows up for teachers when data has not yet been loaded.

pushyamig:
Fixes #1044


pushyamig:
I am merging this now, comments are welcome

PR #1058, opened 2020-08-26
Fixes #1055 - env_sample comment breaks LTI support

### LINKED ISSUE #1055: `env_sample.json` comment breaks LTI support
## Problem

The following comment in `env_sample.json` breaks the LTI support:

```json
        "/*canvas LMS domain name, fill this out for your institution.":"*/",
```

This can be seen in the file at https://github.com/tl-its-umich-edu/my-learning-analytics/blob/9e0816d/config/env_sample.json#L97.

If that comment is left in the config file (Why wouldn't one leave it there?  None of the other comments cause trouble.), when attempting to launch the application, the following message appears in the browser:

```json
{"lti_error": "Launch failed due to Invalid tool conf format. Allowed types of elements: list or dict"}
```

## Solution

Do one of the following:

* Remove the comment
* Move the comment to a place that will not interfere with the required structure
* ~Reformat the comment to comply with the required structure~  
  This approach is very problematic.  If the beginning of the comment becomes the key to an array of a single object, and the end of the comment is a key-value pair in that object, then the PyLTI1p3 module giver errors about various keys are missing (e.g., `auth_login_url`, etc.).  That was produced using the following change:

  ```diff
  -        "/*canvas LMS domain name, fill this out for your institution.":"*/",
  +        "/*canvas LMS domain name, fill this out for your institution.":[{"...":"*/"}],
  ```

As noted in #1056, adding support for Hjson and converting `env.json` to that format would resolve this issue as well.  However, if that issue is not approved or would take longer to implement, then this issue would be a quick fix.

jonespm:


lsloan:
Merge at will!

pushyamig:
TEsted Locally, Looks good

Issue #1062, opened 2020-08-27
Error logged in cron when running course process on warehouse_course_dict


jonespm:
We saw this issue in cron this morning when it was running.

I'm not sure exactly what caused it but this line (568) doesn't have any try/except around it.

https://github.com/tl-its-umich-edu/my-learning-analytics/blob/2a9bab8580fe452575edbc3388289e7dbdfa2237/dashboard/cron.py#L568

My best guess is that there was a course in the local database that wasn't yet in the UDW, so when it tried to look it up in the map this failed. This should be a little more resilient to potential errors. It's a little hard to read this output but the problem was an IndexError on 

`warehouse_course_dict: Dict[str, Any] = warehouse_courses_data.loc[warehouse_courses_data[\'id\'] == course.id].iloc[0].to_dict`

This also was only logged an INFO message because it was put in the cronlog and Jenkins didn't catch it. I created a new issue for that. 

```
[2020-08-27 07:26:45 -0400] [DEBUG] [utils.py:125] (0.013) DELETE FROM `django_myla_cache` WHERE cache_key = 'myla:1:DashboardCronJob'; args=('myla:1:DashboardCronJob',)
[...]

PR #1064, opened 2020-08-27
Lock list of course IDs driving other cron methods after verification (#1062)

ssciolla:
This PR creates a new instance variable on `DashboardCronJob` that is used for storing the course IDs verified by the initial `verify_course_ids` method. This variable is then used in lieu of `Course.objects.get_supported_courses`, because this list is static (while users could add new courses during the cron run, which would be returned by `get_supported_courses` or other `QuerySet` at other points of the run). The PR aims to resolve issue #1062.

### DIFF
--- a/dashboard/cron.py
+++ b/dashboard/cron.py
@@ -611,9 +611,6 @@ class DashboardCronJob(CronJobBase):
 
         # Lock in valid course IDs that data will be pulled for.
         course_df: pd.DataFrame = course_verification.course_data
-        # It not be a best practice to set instance variables in a method, but seemed less messy than overriding the __init__
-        # from CronJobBase. See https://stackoverflow.com/questions/19284857/instance-attribute-attribute-name-defined-outside-init
-        # I'll probably look into this more later. - Sam
         self.valid_locked_course_ids: List[int] = course_df['id'].to_list()
 
         # continue cron tasks
--- a/dashboard/cron.py
+++ b/dashboard/cron.py
@@ -143,7 +143,7 @@ class DashboardCronJob(CronJobBase):
             courses_data = pd.concat(course_dfs).reset_index()
         else:
             logger.info("No course records were found in the database.")
-            courses_data = pd.DataFrame()
+            courses_data = pd.DataFrame(columns=["id", "canvas_id", "enrollment_term_id", "name", "start_at", "conclude_at"])
 
         CourseVerification = namedtuple("CourseVerification", ["invalid_course_ids", "course_data"])
         return CourseVerification(invalid_course_id_list, courses_data)
@@ -615,10 +615,8 @@ class DashboardCronJob(CronJobBase):
             return (status,)
 
         # Lock in valid course IDs that data will be pulled for.
-        if not course_verification.course_data.empty:
-            self.valid_locked_course_ids = course_verification.course_data['id'].to_list()
-        else:
-            self.valid_locked_course_ids = []
+        self.valid_locked_course_ids = course_verification.course_data['id'].to_list()
+        logger.info(f'Valid locked course IDs: {self.valid_locked_course_ids}')
 
         # continue cron tasks
 
--- a/dashboard/cron.py
+++ b/dashboard/cron.py
@@ -156,8 +156,7 @@ class DashboardCronJob(CronJobBase):
         status += deleteAllRecordInTable("user")
 
         # loop through multiple course ids
-        for data_warehouse_course_id in Course.objects.get_supported_courses():
-
+        for data_warehouse_course_id in self.valid_locked_course_ids:
 
             # select all student registered for the course
             user_sql=f"""with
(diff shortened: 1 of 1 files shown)


dashboard/cron.py
```diff
@@ -610,12 +609,16 @@ def do(self):
             logger.info("************ total status=" + status + "/n/n")
             return (status,)
 
+        # Lock in valid course IDs that data will be pulled for.
+        course_df: pd.DataFrame = course_verification.course_data
+        self.valid_locked_course_ids: List[int] = course_df['id'].to_list()
```
ssciolla, line 614: It may not be a best practice to initialize instance variables outside a constructor, but it seemed less messy than overriding the `__init__` from `CronJobBase` See https://stackoverflow.com/questions/19284857/instance-attribute-attribute-name-defined-outside-init I'll be looking into this more.

dashboard/cron.py
```diff
@@ -644,6 +647,10 @@ def do(self):
             logger.info("** informational")
             status += self.update_unizin_metadata()
 
+        courses_added_during_cron: List[int] = list(set(Course.objects.get_supported_courses()) - set(self.valid_locked_course_ids))
+        logger.warning(f'During the run, users added {len(courses_added_during_cron)} course(s): {courses_added_during_cron}')
+        logger.warning(f'No data was pulled for these courses.')
```
ssciolla, line 652: I suppose it's possible the deleting of a course during the cron could cause problems. Is this a case we need to handle?

jonespm: It was decided that we don't need to handle this.

dashboard/cron.py
jonespm: I believe you just can define an __init__ here and call 
`super().__init__()`

You could also put it in the do() method as that will always be called first.

ssciolla: I think you need to pass through the constructor parameters to the `super().__init()` call. Not sure what they are, I'd have to look into the code, since it's not documented. Even if I did figure it out, I'd be concerned that might easily be broken by updates to that package (though there haven't been many lately).

jonespm: Yeah I didn't see this one having any parameters other than self. 
https://github.com/Tivix/django-cron/blob/7fa5c2d2835520d8c920f3c103937ab292d9e545/django_cron/__init__.py#L52

ssciolla: Hm, nice, I'll try that. Thanks for doing the work for me!

jonespm: No problem, just working my crunch time developer support shift! :)

ssciolla: Done! Thought about initializing as `None`, but it throws `mypy` error(s) and I thought it was kinda much to check everywhere whether it was a list. `mypy` seems to be happy with this. See https://stackoverflow.com/questions/60925137/using-mypy-with-with-lazy-initialization-of-instance-attributes

ssciolla:
When I was testing, I added these lines after self.valid_locked_course_ids is set (line 619), ensuring that I import `time`.

```
        logger.info(f'Course.objects.get_supported_courses(): {Course.objects.get_supported_courses()}')
        logger.info(f'valid_locked_course_ids: {self.valid_locked_course_ids}')
        logger.info('sleeping...')
        time.sleep(60)
        logger.info(f'Course.objects.get_supported_courses(): {Course.objects.get_supported_courses()}')
        logger.info(f'valid_locked_course_ids: {self.valid_locked_course_ids}')
```

Then while the sleep is occurring, you can go into the Django admin and add a new course.

zqian (approved):
The patch works for me. Tried the suggested "Sleep time" for testing.

jonespm (approved):
Everything looks good now!

ssciolla:
@jonespm and @zqian, I was double-checking my work, and I realized I was not handling the case where the DataFrame is empty (i.e. there are no courses). Could one of you look at my latest commit (https://github.com/tl-its-umich-edu/my-learning-analytics/pull/1064/commits/1af95aaba5778604de10f2973421ade99e15f196) before I merge?

dashboard/cron.py
```diff
@@ -615,7 +615,10 @@ def do(self):
             return (status,)
 
         # Lock in valid course IDs that data will be pulled for.
-        self.valid_locked_course_ids = course_verification.course_data['id'].to_list()
+        if not course_verification.course_data.empty:
```
jonespm, line 618: This would also be the same as
course_verification.course_data.get("id", pd.Series(dtype=pd.Int64Dtype)).to_list()

ssciolla: Shorter, but probably less readable?

jonespm: This is mentioned as an [anti-pattern:](https://docs.quantifiedcode.com/python-anti-patterns/correctness/not_using_get_to_return_a_default_value_from_a_dictionary.html)

"Although there is nothing wrong with this code, it is verbose and inefficient because it queries the dictionary twice." 

I think it's only less readable because we have to pass a pd.Series (and that in turn has a warning without a dtype on an empty Series) but that also makes it more explicit too.

So it's fine either way. the performance isn't a big deal

ssciolla: You're extrapolating a bit on that anti-pattern, but you are probably correct. I will make the change.

jonespm: It might be better also if `verify_course_ids` on line 146 returns this as a DataFrame

`pd.DataFrame(columns=["id", "canvas_id", "enrollment_term_id", "name", "start_at", "conclude_at"])`

Then you don't have to do either of these!

ssciolla: Hm, I like that. Possibly more verbose, but I like that the `DataFrame` maintains the same structure.

jonespm (approved):

Issue #1053, opened 2020-08-25
The logout link in the profile menu should be removed for the LTI


jennlove-um:
The logout link does not make sense for the LTI.  It doesn't necessarily break anything if a user logs out but it looks wrong.  They can easily access MyLA again by launching the LTI.

## Steps to Reproduce :
1. Launch MyLA
2.Under the profile menu, click Logout
3.User is logged out & Django login screen displays

## Screenshots/Video (If possible, add media to help explain your problem) :
![image](https://user-images.githubusercontent.com/39634416/91227967-6ee37d80-e6f5-11ea-8395-ed6605a60be8.png)

jennlove-um:
This may be related to #1060

ssciolla:
@zqian, why was this issue closed? I think we still need to disable the "Log out" option for the Avatar dropdown, right?

jennlove-um:
This is a duplicate of #1079

PR #1073, opened 2020-09-03
Only enable Django backend login methods if explicitly loaded or not …

### LINKED ISSUE #1060: Only enable django backend login methods if explicitly loaded or not  other auth.

`DjangoModelBackend` seems to be loaded if either LTI or SAML are not loaded, not either. 

It should really only be enabled at the end of the settings of there is no other `AUTHENTICATION_BACKENDS` available OR if explicitly enabled via a config (which can be useful for a test server).

jonespm:
…other auth.

Fixes #1060

* This adds a new config property `ENABLE_BACKEND_LOGIN` (default false) that will force the login to be available. This can be useful for testing servers
* There's a slight reorganization of the settings with the introduction of the variable ENABLE_BACKEND_LOGIN. We can more efficiently take care of this
* Some of the functionality had to be taken care of in urls.py. I protected the Django admin login with staff_member_required to keep it open and moved the login patterns behind the setting.


dashboard/settings.py
```diff
@@ -417,17 +418,13 @@
...
-    LOGIN_REDIRECT_URL = '/'
-    LOGOUT_REDIRECT_URL='/'
 
-# Give an opportunity to disable LTI
-if ENV.get('STUDENT_DASHBOARD_LTI', False):
-    if not 'django.contrib.auth.backends.ModelBackend' in AUTHENTICATION_BACKENDS:
-        AUTHENTICATION_BACKENDS += ('django.contrib.auth.backends.ModelBackend',)
+    ENABLE_BACKEND_LOGIN = False
 
+# Give an opportunity to disable LTI
```
ssciolla, line 424: This comment doesn't make sense here anymore?

dashboard/settings.py
```diff
@@ -496,6 +493,15 @@ def show_debug_toolbar(request):
 SESSION_COOKIE_SAMESITE = ENV.get("SESSION_COOKIE_SAMESITE", None)
 CSRF_COOKIE_SAMESITE = ENV.get("CSRF_COOKIE_SAMESITE", None)
 
+# Give a chance to enable the backend login via a setting
+ENABLE_BACKEND_LOGIN = ENV.get("ENABLE_BACKEND_LOGIN", ENABLE_BACKEND_LOGIN)
+
```
ssciolla, line 498: This looks like it would work, but I wonder if instead of re-assigning throughout, you could just handle it all here, doing something like this:
```
ENABLE_BACKEND_LOGIN = True
# would have to create the STUDENT_DASHBOARD_SAML variable
if STUDENT_DASHBOARD_SAML or STUDENT_DASHBOARD_LTI:
   ENABLE_BACKEND_LOGIN = False
# Allow for ENABLE_BACKEND_LOGIN override
 ENABLE_BACKEND_LOGIN = ENV.get("ENABLE_BACKEND_LOGIN", ENABLE_BACKEND_LOGIN)
```

ssciolla (commented):
Couple initial comments. Doing some testing now.

dashboard/settings.py
jonespm: Yeah I was trying to think of the best way to do this and re-wrote it a few times. This is needed by the urls.py.

The only thing we'd need to remember is if another auth method got enabled we'd have to add it there. But that would be the same as in the other case too. I can change this.

ssciolla: It may not be a huge difference, but I like that the logic for `ENABLE_BACKEND_LOGIN` is all condensed, you don't have to go searching for variable references.

jonespm: How about 
`ENABLE_BACKEND_LOGIN = False if STUDENT_DASHBOARD_SAML or STUDENT_DASHBOARD_LTI else True `

jonespm: STUDENT_DASHBOARD_SAML wasn't set and I noticed the default was True. I'm going to change this to False and have this as a setting.

ssciolla: One liner works. For some reason I haven't come around to using the ternary approach in Python as much yet, but use it all the time in JS.

ssciolla:
Thanks for the change, I'll get to testing this with LTI next week. I think we do need to think about issue #1053 with this. Something might be broken in the avatar dropdown with this change? Need to take a look at `myla_globals` and the FE code.

dashboard/settings.py
```diff
@@ -496,6 +486,17 @@ def show_debug_toolbar(request):
 SESSION_COOKIE_SAMESITE = ENV.get("SESSION_COOKIE_SAMESITE", None)
 CSRF_COOKIE_SAMESITE = ENV.get("CSRF_COOKIE_SAMESITE", None)
 
+ENABLE_BACKEND_LOGIN = False if STUDENT_DASHBOARD_SAML or STUDENT_DASHBOARD_LTI else True 
+
+# Allow for ENABLE_BACKEND_LOGIN override
+ENABLE_BACKEND_LOGIN = ENV.get("ENABLE_BACKEND_LOGIN", ENABLE_BACKEND_LOGIN)
```
pushyamig, line 492: I felt `ENABLE_BACKEND_LOGIN` on L489 can be written as `CHECK_TO_ENABLE_BACKEND_LOGIN` since that is actually checking if there is need to enable it or not and actually the action of enabling it is happening on L492

dashboard/settings.py
```diff
@@ -417,16 +415,8 @@
...
     }
-else:
-    AUTHENTICATION_BACKENDS += ('django.contrib.auth.backends.ModelBackend',)
-    LOGIN_REDIRECT_URL = '/'
-    LOGOUT_REDIRECT_URL='/'
-
-# Give an opportunity to disable LTI
-if ENV.get('STUDENT_DASHBOARD_LTI', False):
-    if not 'django.contrib.auth.backends.ModelBackend' in AUTHENTICATION_BACKENDS:
-        AUTHENTICATION_BACKENDS += ('django.contrib.auth.backends.ModelBackend',)
```
pushyamig, line 428: I guess this was always there from before since LTIV1p1, some how the logic here adding it exclusively. I did not give any serious though on this when upgrading to v1p3. Thanks for taking care of this

config/env_sample.json
```diff
@@ -128,8 +128,11 @@
         "CACHE_KEY_PREFIX": "myla",
         "CACHE_OPTIONS": {"COMPRESS_MIN_LENGTH": 5000, "COMPRESS_LEVEL": 6}
     },
+    "/*  Set this to true to enable the internal (Django) Login methods.": "*/",
+    "/*  This is enabled by default if no external auth (Currently LTI/SAML) is set.": "*/",
+    "ENABLE_BACKEND_LOGIN": false,
```
pushyamig, line 133: I feel this needs to be added to the LTI wiki page.

dashboard/urls.py
```diff
@@ -100,6 +97,14 @@
         path('lti/jwks/', lti_new.get_jwks, name="get_jwks"),
     )
 
+if settings.ENABLE_BACKEND_LOGIN:
+    from django.contrib.auth import views as auth_views
+    # Login patterns for testing, SAML should be installed in prod
+    urlpatterns += (
+        path('accounts/login/', auth_views.LoginView.as_view(), name='login'),
+        path('accounts/logout/', auth_views.LogoutView.as_view(), name='logout'),
```
pushyamig, line 105: This is a nice way to reduce the access to the login/logout URL. 

In `settings.py` I see as below 
```
# Add any settings you need to be available to templates in this array
SETTINGS_EXPORT = ['LOGIN_URL','LOGOUT_URL','DEBUG', 'GA_ID', 'RESOURCE_VALUES']
```

do we need to have a check if when either saml/lti check enabled and then add LOGIN_URL and LOGOUT_URL?

dashboard/settings.py
jonespm: Yeah, that's just figuring out the default, I can rename that initial variable.

dashboard/urls.py
jonespm: The SETTINGS_EXPORT is for this package https://pypi.org/project/django-settings-export/

So those settings are available in the template. I think we always want some URL sent to be able to login logout. 

We may want the to disable the login/logout URL in the UI in some cases (like if LTI is enabled?) but that seems like a separate issue.

config/env_sample.json
jonespm: Okay I updated this

jonespm:
Any additional comments or approval? Would like to merge this today to get #1076 merged too.

ssciolla:
I still want to run this again, I'll get back to you soon.

jonespm:
@ssciolla Sam mentioned there was an issue launching LTI with the backend disabled. I'm looking at this now.

ssciolla (approved):
Okay, now LTI is working for me. I say let's merge this so we can test SAML and move on to other stuff. Thanks for the changes!

pushyamig (approved):

dashboard/urls.py
jonespm: It looks like #784 moved many of these out of the template to the view so I think you're right we don't need this package anymore. I'll create a new issue.

jonespm: Created as #1084

dashboard/urls.py
```diff
@@ -29,12 +30,15 @@
 from . import views
 
 import watchman.views
+
+# Disable the Django admin login page
+admin.site.login = staff_member_required(admin.site.login, login_url=settings.LOGIN_URL)
+
 urlpatterns = [
     path('', views.get_home_template, name = 'home'),
```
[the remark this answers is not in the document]
jonespm: Yeah, I'd changed LOGIN_URL in the LTI piece to be /lti/login but that doesn't really work either.

I think we kind of just need a 404 page setup if someone tries to access it directly. We should probably have custom error pages for all of these? 

https://docs.djangoproject.com/en/3.1/topics/http/views/#customizing-error-views

jonespm:


```diff

Issues
======
- Added 1
           
```


See the complete overview on [Codacy](https://app.codacy.com/manual/ITS_Teaching_And_Learning/my-learning-analytics/pullRequest?prid=6127818&bid=20044124)

config/env_sample.json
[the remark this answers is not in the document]
jonespm: Yeah the purpose of this setting now really just seems to be to enable the URLs. I can update the description here.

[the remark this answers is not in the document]
jonespm: I'm going to have to test with SAML still, and I think this might need some more work especially if we end up writing a custom Backend for LTI.

PR #1078, opened 2020-09-08
issue-1070: operate on dataframe copy to prevent Pandas SettingWithCopyWarning warning

zqian:
Without the fix, I can repeat the problem by going to MyLA courses and open Assignment Planning view.

The SettingWithCopyWarning is a type of Pandas warning message, and [there are ways to fix the warning](https://www.dataquest.io/blog/settingwithcopywarning/). This fix is to get a copy of dataframe object first before further operations.

### DIFF
--- a/dashboard/views.py
+++ b/dashboard/views.py
@@ -574,26 +574,25 @@ def assignments(request, course_id=0):
         df['avg_score']= df.apply(no_show_avg_score_for_ungraded_assignments, axis=1)
     df['avg_score']=df['avg_score'].fillna('Not available')
 
-    df3 = df[df['towards_final_grade'] > 0.0]
     # operate on dataframe copy to prevent Pandas "SettingWithCopyWarning" warning
-    df3 = df3.copy()
-    df3[['score']] = df3[['score']].astype(float)
-    df3['graded'] = df3['graded'].fillna(False)
-    df3[['score']] = df3[['score']].astype(float)
-    df3['percent_gotten'] = df3.apply(lambda x: user_percent(x), axis=1)
-    df3.sort_values(by=['graded', 'due_date_mod'], ascending=[False, True], inplace=True)
-    df3.reset_index(inplace=True)
-    df3.drop(columns=['index'], inplace=True)
+    df_progressbar = df.loc[df['towards_final_grade'] > 0.0].copy()
+    df_progressbar[['score']] = df_progressbar[['score']].astype(float)
+    df_progressbar['graded'] = df_progressbar['graded'].fillna(False)
+    df_progressbar[['score']] = df_progressbar[['score']].astype(float)
+    df_progressbar['percent_gotten'] = df_progressbar.apply(lambda x: user_percent(x), axis=1)
+    df_progressbar.sort_values(by=['graded', 'due_date_mod'], ascending=[False, True], inplace=True)
+    df_progressbar.reset_index(inplace=True)
+    df_progressbar.drop(columns=['index'], inplace=True)
 
     assignment_data = {}
-    assignment_data['progress'] = json.loads(df3.to_json(orient='records'))
+    assignment_data['progress'] = json.loads(df_progressbar.to_json(orient='records'))
 
     # Group the data according the assignment prep view
-    df2 = df[df['towards_final_grade'] >= percent_selection]
-    df2.reset_index(inplace=True)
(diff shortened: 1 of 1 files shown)


dashboard/views.py
```diff
@@ -575,6 +575,8 @@ def assignments(request, course_id=0):
     df['avg_score']=df['avg_score'].fillna('Not available')
 
     df3 = df[df['towards_final_grade'] > 0.0]
+    # operate on dataframe copy to prevent Pandas "SettingWithCopyWarning" warning
+    df3 = df3.copy()
```
jonespm, line 579: This looks fine to me. It might be better to improve line 577 to be

`df3 = df.loc[df['towards_final_grade'] > 0.0].copy()`

Also what is df3? I think Sam mentioned this when we were reviewing this. It might be wordier but something more like `df_positive_grades` or something might be a better name than df3. df2 looks like `df_greater_that_percent_selection`. Perhaps this could be better refactored.

pushyamig: Yes, I do agree on this and it can be named little better. The `df3` is for progress bar at the top of the assignment

pushyamig: on line591 which is below  
`df2 = df[df['towards_final_grade'] >= percent_selection]`
can also be written as 
df2 = df.loc[df['towards_final_grade'] > percent_selection].copy()

zqian: Thanks for the suggestions. I have renamed the dataframe names from df2 and df3; also added .copy() to more dataframe objects as suggested.

dashboard/views.py
```diff
@@ -574,24 +574,25 @@ def assignments(request, course_id=0):
...
-    assignment_data['progress'] = json.loads(df3.to_json(orient='records'))
+    assignment_data['progress'] = json.loads(df_progressbar.to_json(orient='records'))
 
     # Group the data according the assignment prep view
-    df2 = df[df['towards_final_grade'] >= percent_selection]
-    df2.reset_index(inplace=True)
-    df2.drop(columns=['index'], inplace=True)
-    logger.debug('The Dataframe for the assignment planning %s ' % df2)
-    grouped = df2.groupby(['week', 'due_dates'])
+    df_greaterthan_percent = df.loc[df['towards_final_grade'] >= percent_selection].copy()
```
pushyamig, line 591: You can say this a `df_plan` the Json has 2 parts one is data for progress and other is for assignment planningtable view

dashboard/views.py
```diff
@@ -643,8 +644,8 @@ def get_course_assignments(course_id):
     assignments_in_course[['points_possible','group_points']]=assignments_in_course[['points_possible','group_points']].fillna(0)
     assignments_in_course[['points_possible', 'group_points','weight']] = assignments_in_course[['points_possible', 'group_points','weight']].astype(float)
     consider_weight=is_weight_considered(course_id)
-    df2 = assignments_in_course[['weight','group_points','grp_id']].drop_duplicates()
-    hidden_assignments = are_weighted_assignments_hidden(course_id, df2)
+    df_assignment = assignments_in_course[['weight','group_points','grp_id']].drop_duplicates().copy()
```
pushyamig, line 647: is this `df_assignment` or `df_greaterthan_percent`?

dashboard/views.py
zqian: renamed the dataframe to df_plan

dashboard/views.py
zqian: This is in another function get_course_assignments, just another instance of "df2" but totally unrelated to the previous example.

pushyamig:
I still see the warning in the logs
```
student_dashboard  | /usr/local/lib/python3.8/site-packages/pandas/core/frame.py:3990: SettingWithCopyWarning:
student_dashboard  | A value is trying to be set on a copy of a slice from a DataFrame
student_dashboard  |
student_dashboard  | See the caveats in the documentation: https://pandas.pydata.org/pandas-docs/stable/user_guide/indexing.html#returning-a-view-versus-a-copy
student_dashboard  |   return super().drop(
```

zqian:
I cannot reproduce this remaining warning consistently.

Let's stay with the current fixes for this release, and I have created #1086 to track down the remaining issue.

jonespm (approved):

pushyamig (approved):

PR #1081, opened 2020-09-10
Issue 1079 remove login logout lti

### LINKED ISSUE #1079: Disable logout URL when LTI is configured
If you're launching via LTI we should disable the logout link in the UI as it really won't do anything. Maybe should just pass something to the myla globals whether or not to show the logout? This was kind of mentioned on this ticket.

However it's possible to ENABLE_BACKEND_LOGIN and login that way. In that case we should still have the logout link. So this might not be completely straightforward to fix.

pushyamig:
Fixes #1079

This issues removes the logout URL from the UI doesn't send in the Login URL when LTI. But this should be available with SAML/local 

This PR is based off of #1079 1073

### DIFF
--- a/assets/src/containers/App.js
+++ b/assets/src/containers/App.js
@@ -11,7 +11,7 @@ function App (props) {
 
   if (!user.isLoggedIn) {
     if (user.loginURL === '') {
-      return (<WarningBanner>You should launch this tool from your Canvas Course</WarningBanner>)
+      return (<WarningBanner>This tool needs to be launched from a Canvas course.</WarningBanner>)
     }
     return (window.location.href = user.loginURL)
   }
--- a/assets/src/containers/App.js
+++ b/assets/src/containers/App.js
@@ -4,11 +4,15 @@ import { matchPath } from 'react-router'
 import GoogleAnalyticsTracking from '../components/GoogleAnalyticsTracking'
 import CourseList from './CourseList'
 import Course from './Course'
+import WarningBanner from '../components/WarningBanner'
 
 function App (props) {
   const { location, user, gaId } = props
 
   if (!user.isLoggedIn) {
+    if (user.loginURL === '') {
+      return (<WarningBanner>You should launch this tool from your Canvas Course</WarningBanner>)
+    }
     return (window.location.href = user.loginURL)
   }
   const coursePageMatch = matchPath(location.pathname, '/courses/:courseId/')
--- a/dashboard/common/utils.py
+++ b/dashboard/common/utils.py
@@ -86,10 +86,10 @@ def get_myla_globals(request):
         username = current_user.get_username()
         user_courses_info = get_user_courses_info(username, course_id)
 
-    if not settings.STUDENT_DASHBOARD_LTI:
+    if settings.SHOW_LOGOUT_LINK:
         login_url = settings.LOGIN_URL
-    if not settings.STUDENT_DASHBOARD_LTI:
         logout_url = settings.LOGOUT_URL
+
     if settings.GA_ID:
         google_analytics_id = settings.GA_ID
     primary_ui_color = settings.PRIMARY_UI_COLOR
--- a/dashboard/common/utils.py
+++ b/dashboard/common/utils.py
@@ -86,9 +86,9 @@ def get_myla_globals(request):
         username = current_user.get_username()
         user_courses_info = get_user_courses_info(username, course_id)
 
-    if settings.LOGIN_URL:
+    if not settings.STUDENT_DASHBOARD_LTI:
         login_url = settings.LOGIN_URL
-    if settings.LOGOUT_URL:
+    if not settings.STUDENT_DASHBOARD_LTI:
         logout_url = settings.LOGOUT_URL
     if settings.GA_ID:
         google_analytics_id = settings.GA_ID
--- a/assets/src/components/AvatarModal.js
+++ b/assets/src/components/AvatarModal.js
@@ -156,14 +156,20 @@ function AvatarModal (props) {
                 ? SwitchCourses()
                 : null
             }
-            <Link style={{ textDecoration: 'none' }} href={user.logoutURL}>
-              <ListItem button>
-                <ListItemIcon>
-                  <LogoutIcon />
-                </ListItemIcon>
-                <ListItemText inset primary='Logout' className={classes.text} />
-              </ListItem>
-            </Link>
+            {
+              user.logoutURL !== ''
+                ? (
+                  <Link style={{ textDecoration: 'none' }} href={user.logoutURL}>
+                    <ListItem button>
+                      <ListItemIcon>
+                        <LogoutIcon />
+                      </ListItemIcon>
+                      <ListItemText inset primary='Logout' className={classes.text} />
+                    </ListItem>
+                  </Link>
+                )
+                : null
+            }
           </List>
         </Grid>
       </Grid>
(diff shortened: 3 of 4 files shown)


jonespm:


```diff

Issues
======
- Added 1
           
```


See the complete overview on [Codacy](https://app.codacy.com/manual/ITS_Teaching_And_Learning/my-learning-analytics/pullRequest?prid=6160227&bid=20123248)

jonespm:
I'll review this after it's rebased.

pushyamig:
OK, I rebased. Please check again!

dashboard/common/utils.py
```diff
@@ -86,9 +86,9 @@ def get_myla_globals(request):
         username = current_user.get_username()
         user_courses_info = get_user_courses_info(username, course_id)
 
-    if settings.LOGIN_URL:
+    if not settings.STUDENT_DASHBOARD_LTI:
         login_url = settings.LOGIN_URL
-    if settings.LOGOUT_URL:
+    if not settings.STUDENT_DASHBOARD_LTI:
         logout_url = settings.LOGOUT_URL
```
ssciolla, line 92: For some reason, when I try to access `localhost:5001`, it's doing that looping thing which I've pointed out before: it keeps trying and failing to load. This isn't happening on `master` (you just get a 404 when redirected to `accounts/login`). I can't explain this yet. `myla_globals['login_url']` must be in use somewhere and the logic you changed is breaking something?

ssciolla (commented):
This isn't working for me yet. See the comment about going straight to the MyLA domain. This happens whether `ENABLE_BACKEND_LOGIN` is `true` or `false`, I believe.

dashboard/common/utils.py
ssciolla: Just a note to say that this is successfully hiding the "Logout" button, but the behavior around visiting the domain and using the default backend seems to have been disrupted somehow.

jonespm:


```diff

Complexity increasing per file
==============================
- assets/src/containers/App.js  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/ITS_Teaching_And_Learning/my-learning-analytics/pullRequest?prid=6160227&bid=20123248)

assets/src/containers/App.js
```diff
@@ -4,11 +4,15 @@ import { matchPath } from 'react-router'
...
 import CourseList from './CourseList'
 import Course from './Course'
+import WarningBanner from '../components/WarningBanner'
 
 function App (props) {
   const { location, user, gaId } = props
 
   if (!user.isLoggedIn) {
+    if (user.loginURL === '') {
+      return (<WarningBanner>You should launch this tool from your Canvas Course</WarningBanner>)
```
ssciolla, line 14:
```suggestion
      return (<WarningBanner>This tool needs to be launched from a Canvas course.</WarningBanner>)
```

ssciolla (approved):
I think this is okay. I'd rather use `null` over the empty string as the default, but I won't insist on that change right now.

It's a little confusing when you use the default login while also using LTI; the "Logout" button is gone from the Avatar drop-down, even when you aren't in the Canvas iframe. Maybe that's alright since it mimics what the user would see? In production, we wouldn't want these on at the same time yet. I think there's probably a way to put in `mylaGlobals` where this is getting launched from (determined by the referrer?) and use that to decide whether to show "Logout". That might make things more consistent, which might be desirable if you wanted to enable a prod backdoor. It'd be nice to hear @jonespm's feedback on this, too.

pushyamig:
> I think there's probably a way to put in mylaGlobals where this is getting launched from (determined by the referrer?) and use that to decide whether to show "Logout". That might make things more consistent

Even we put the logout but it won't do much since the `/accounts/logout` is disabled in Backend. I am not how feasible option based on the referrer you enable the backed and display option. Not for the PR :)

pushyamig:
> It's a little confusing when you use the default login while also using LTI; the "Logout" button is gone from the Avatar drop-down, even when you aren't in the Canvas iframe. Maybe that's alright since it mimics what the user would see?

Yes, When authentication is done in LTI and not through Django form/SAML, then it make sense not to show the logout button.

ssciolla:
> > I think there's probably a way to put in mylaGlobals where this is getting launched from (determined by the referrer?) and use that to decide whether to show "Logout". That might make things more consistent
> 
> Even we put the logout but it won't do much since the `/accounts/logout` is disabled in Backend. I am not how feasible option based on the referrer you enable the backed and display option. Not for the PR :)

If `ENABLE_BACKEND_LOGIN` is `true`, then `/accounts/logout` is still active, isn't it?

pushyamig:
> If ENABLE_BACKEND_LOGIN is true, then /accounts/logout is still active, isn't it?

Yes, here is the logic for it
https://github.com/tl-its-umich-edu/my-learning-analytics/blob/master/dashboard/urls.py#L100

ssciolla:
That's my point You **can** log out if you log in using the `ENABLE_BACKEND_LOGIN` while LTI is also enabled, but it's not visible in the drop-down, because you're always disabling it when LTI is eanbled.

> > If ENABLE_BACKEND_LOGIN is true, then /accounts/logout is still active, isn't it?
> 
> Yes, here is the logic for it
> https://github.com/tl-its-umich-edu/my-learning-analytics/blob/master/dashboard/urls.py#L100

That's my point. You **can** log out if you log in using the backend triggered by `ENABLE_BACKEND_LOGIN` while LTI is also enabled, but it's not visible in the drop-down, because the change you added always hides it when LTI is enabled.

pushyamig:
> That's my point. You can log out if you log in using the backend triggered by ENABLE_BACKEND_LOGIN while LTI is also enabled, but it's not visible in the drop-down, because the change you added always hides it when LTI is enabled.

With LTI, the option is set to `ENABLE_BACKEND_LOGIN = False` and the account URL are disabled. SO you cannot login/logout by typing in URL

pushyamig:
```
CHECK_ENABLE_BACKEND_LOGIN = False if STUDENT_DASHBOARD_SAML or STUDENT_DASHBOARD_LTI else True 

# Allow for ENABLE_BACKEND_LOGIN override
ENABLE_BACKEND_LOGIN = ENV.get("ENABLE_BACKEND_LOGIN", CHECK_ENABLE_BACKEND_LOGIN)
```

ssciolla:
> With LTI, the option is set to `ENABLE_BACKEND_LOGIN = False` and the account URL are disabled. SO you cannot login/logout by typing in URL

That's wrong; you can force it to enable both LTI and the default backend by using the override, specifying `"ENABLE_BACKEND_LOGIN": true` in `env.json`.

pushyamig:
> That's wrong; you can force it to enable both LTI and the default backend by using the override, specifying "ENABLE_BACKEND_LOGIN": true in env.json.

we are enabling Modelbackend with LTI
I felt we are going to set it `ENABLE_BACKEND_LOGIN` = false for LTI and my assumption is we are going to disable the account URL's as well @jonespm  please share your thoughts!

jonespm:


```diff

Complexity increasing per file
==============================
- assets/src/containers/App.js  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/ITS_Teaching_And_Learning/my-learning-analytics/pullRequest?prid=6160227&bid=20123248)

pushyamig:
@ssciolla upon discussion with @jonespm 
for LTI `ENABLE_BACKEND_LOGIN` will be `False` and `account/login` and `account/logout` will be disabled in backend and not show in the UI.

In case of SAML and Django default forms those will be enabled. 

I feel this PR is serving the purpose as intended

jonespm (approved):

Issue #1082, opened 2020-09-10
Add alt text to the data is loading image


jennlove-um:
Add the following alt-text to the image that displays before the course data is loaded: "Mouse running on wheel with text course data being processed. Try back in 24 hours."

chrisrrowland:
Using text w/ quotes
'Mouse running on wheel with text "course data being processed Try back in 24 hours"'

PR #1091, opened 2020-09-14
Add alt text for mouse on wheel loading image (#1082)

chrisrrowland:
Set alt text via title on CardMedia element.

I tested using the [ChromeVox](https://chrome.google.com/webstore/detail/chromevox-classic-extensi/kgejglhpjiefppelpmljglcjbhoiplfn?hl=en) chrome extension.

By default I (Windows) had to press Shift+Alt+A Shift+Alt+A to toggle ChromeVox on and off.  Shift+Alt is the "Chrome Vox Key" and is different on different platforms. On MacOS its Control+Command, per the documentation.

Navigating to the image (I clicked on it) triggers the text to be read aloud.

### DIFF
--- a/assets/src/containers/Course.js
+++ b/assets/src/containers/Course.js
@@ -67,7 +67,11 @@ function Course (props) {
             {courseInfo.course_data_loaded === 0
               ? (
                 <Card className={classes.card}>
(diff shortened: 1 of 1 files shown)


zqian (approved):
Verified with ChromeVox and mouseover for alt text.

PR #1100, opened 2020-10-02
Elimiate an unnecessary warning message for admins

### LINKED ISSUE #1061: remove redundant log warning message for MyLA admin
There are redundant warning messages in the MyLA log file for admin users:

[2020-08-27 08:05:00 -0400] [WARNING] [db_util.py:113] Couldn't find user <admin user id> in enrollment info. The user could be an admin, or enrollment data has not been populated yet.

Admins never have problem for accessing MyLA tool. Those messages can be removed.

chrisrrowland:
Closes #1061 

This eliminates a warning message for admin users.

I believe this situation can only occur in standalone.

### DIFF
--- a/dashboard/common/db_util.py
+++ b/dashboard/common/db_util.py
@@ -94,13 +94,15 @@ class CourseEnrollment(TypedDict):
     enrollment_types: List[str]
 
 def is_superuser(user_name: str) -> bool:
-    logger.info(is_superuser.__name__+f' \'{user_name}\'')
+    logger.debug(is_superuser.__name__+f' \'{user_name}\'')
 
     user = DjangoUser.objects.filter(username=user_name)
     if user.count() == 0:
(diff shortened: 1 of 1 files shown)


dashboard/common/db_util.py
```diff
@@ -92,6 +93,14 @@ class CourseEnrollment(TypedDict):
     course_name: str
     enrollment_types: List[str]
 
+def is_superuser(user_name: str) -> bool:
+    logger.info(is_superuser.__name__+f' \'{user_name}\'')
```
zqian, line 97: If we want to log this, maybe it is better to also log the result of is_superuser call. Like after line 103

I think the log level could be debug instead of info.

zqian (changes requested):

zqian (approved):

PR #1101, opened 2020-10-06
Added `getjwk` management command (issue #1065)

### LINKED ISSUE #1065: add JWK management tool for LTI config
The next version of the LTI specification will _require_ tools provide their public key via JWKS URL.  This info was provided by Bracken Mosbacker (IMS Global), Eric Preston (Blackboard), and Viktor Haag (D2L) in a conversation on IMS Global's Slack workspace (https://imsglobal.slack.com/messages/C04GUTJTN/p1598555022006400).

This conversation started as I (@lsloan) asked the IMS Global community about whether PEM or JWK public key is preferred.  I asked because I was trying to update our documentation (wiki) to include the option of using the generated JWK instead of the PEM.

Make a Django management tool to give JWK JSON, when necessary.

See also: #1103 (update LTI config docs)

## Testing

See the `getjwk` section of the LTI deployment documentation at https://github.com/tl-its-umich-edu/my-learning-analytics/wiki/Deploy:-Configuration-for-LTI-1.3-Deployment#using-mylas-django-management-tool

lsloan:
Resolves #1065.

## Review procedures
<details>
  <summary>Details…</summary>
  
  * I **_humbly_** request comments on this PR be nested under a single review per reviewer. It helps me when the comments are organized this way, especially if the comments request changes in the code.
  * If I've requested a review from you but you are definitely unable to review this PR, please leave a comment saying so (please state why you're unable to review, that would be appreciated), tell the other reviewers that the review will be left to them, and remove your name from the list of reviewers.
</details>

## About
Move JWK generation out of the custom `createkeys` management command.

The JWK rarely (practically never) needs to be generated for storing anywhere, because the `/lti/jwks/` service uses PyLTI1p3 to generate it on demand.

The new `getjwk` management command is for convenience only, to use in those extremely rare cases the JWK is needed.  Instead of generating the JWK in this code, it calls the code in PyLTI1p3 to do the work.

This new command is mentioned in the updated LTI config. doc. in the MyLA wiki.

## Testing

See the `getjwk` section of the LTI deployment documentation at https://github.com/tl-its-umich-edu/my-learning-analytics/wiki/Deploy:-Configuration-for-LTI-1.3-Deployment#using-mylas-django-management-tool


jonespm:


```diff

Complexity increasing per file
==============================
- dashboard/management/commands/getjwk.py  2
- dashboard/keyfileutils.py  2
         

Complexity decreasing per file
==============================
+ dashboard/management/commands/createkeys.py  -1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/ITS_Teaching_And_Learning/my-learning-analytics/pullRequest?prid=6313191&bid=20526237)

lsloan:
@jonespm, @ssciolla, @zqian: I opened this little PR a couple of weeks ago.  I'd like to get it settled so that I don't need to revisit it later when I've moved on to other tasks.

dashboard/management/commands/createkeys.py
```diff
@@ -67,17 +40,14 @@ def handle(self, *args, **options: dict):
...
+            keyFileUtils.KeyFileType.PRIVATE,
+            privateKey.decode('utf-8'))
+        self.stdout.write(result)
 
-        self._writeKeyFile(self.KeyFileType.JWK,
-                           json.dumps(public_jwk))
+        result = keyFileUtils.writeKeyFile(
+            keyFileUtils.KeyFileType.PUBLIC,
+            publicKey.decode('utf-8'))
+        self.stdout.write(result)
```
ssciolla, line 53: Nit-pick: This is probably okay, but I don't love the reassignment to `result`. I think I'd go with separate variables.

dashboard/management/commands/getjwk.py
```diff
@@ -0,0 +1,56 @@
...
+        group.add_argument(
+            f'--{self.dumpOption}',
+            action='store_true',
+            default=False,
+            required=False,
+            help='Dump the JWK JSON to stdout instead of writing it to a file',
+            dest=self.dumpOption)
+
+    def handle(self, *args, **options: dict):
+        jwkResponse: JsonResponse = lti_new.get_jwks(None)
```
ssciolla, line 38: Should we change `lti_new.get_jwks` so it doesn't require you to pass an unnecessary parameter?

dashboard/management/commands/getjwk.py
```diff
@@ -0,0 +1,56 @@
...
+        jwkObject = json.loads(jwkResponse.content.decode('utf8'))
+        jwkJson = json.dumps(jwkObject['keys'][0])
+
+        if options.get(self.dumpOption) is True:
```
ssciolla, line 42: Is the `is True` necessary?

dashboard/management/commands/getjwk.py
```diff
@@ -0,0 +1,56 @@
...
+            self.stdout.write(jwkJson)
+        else:
+            configDir: str = os.path.dirname(
+                os.getenv('ENV_FILE', '/secrets/env.json'))
```
ssciolla, line 46: There's a similar pattern in `settings.py`, you could probably make this a variable there and then import it through `from django.conf import settings`.

dashboard/management/commands/createkeys.py
```diff
@@ -1,63 +1,36 @@
-import json
 import os
 from datetime import datetime
-from enum import Enum
 
 from Crypto.PublicKey import RSA
 from Crypto.PublicKey.RSA import RsaKey
 from django.core.management.base import BaseCommand, CommandParser
-from jwcrypto.jwk import JWK
```
ssciolla, line 9: Can we remove this dependency from `requirements.txt` now?

ssciolla (commented):
Refactor looks okay to me. I added a couple comments. I still need to actually test this (which I'll probably do on Monday). Am I right in my understanding that this new `getjwk` command is very unlikely to be used in practice and is even less likely to be part of our production workflows?

jonespm:
I think all of these variables should be in snake_case which is the prevailing style of this and most Python/Django projects. But I guess we can revisit and fix this when/if we implement flake8 checking on this project. 

Otherwise I don't see anything that @ssciolla hasn't mentioned

ssciolla (approved):
I just tested the Django management command `getjwk`, and I didn't see any issues. If that `jwcrypto` dependency isn't needed (or is being used as a sub-dependency), please remove it from `requirements.txt.` However, I'm approving; all other notes I'm leaving up to your discretion.

The documentation page from the Wiki was helpful in re-clarifying the purpose of this new command/how all these keys work, and I think it will be helpful to have this all sorted out in the future. Sorry for taking a while to get back to you; I know it's nice to get timely feedback on work. I guess I felt (and still feel somewhat) that this isn't a top priority, given what the team wants to accomplish before the end of the year.

dashboard/keyfileutils.py
```diff
@@ -0,0 +1,31 @@
+from enum import Enum
+
+
+class KeyFileUtils:
+    timestampFormat = '%Y%m%d%H%M%S'
```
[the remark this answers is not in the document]
lsloan: Would you expect the code in this file to become part of `dashboard/common/utils.py` or should this file be moved to `dashboard/common/keyfileutils.py`?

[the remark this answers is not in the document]
ssciolla: I was thinking the latter.

dashboard/management/commands/createkeys.py
lsloan: Interesting!  Could you explain why you prefer that?  Is it a common practice or your personal preference?

What do you think of not assigning the results to variables, and wrap the `writeKeyFile()` calls in `self.stdout.write()` directly.  That is…

```python
        self.stdout.write(keyFileUtils.writeKeyFile(
            keyFileUtils.KeyFileType.PRIVATE,
            privateKey.decode('utf-8')))
```

dashboard/management/commands/getjwk.py
lsloan: Yes, I could do that.  I considered it, but it wasn't strictly necessary in the scope of implementing this feature.  Since you're thinking along the same lines and the argument of `lti_new.get_jwks()` isn't even used, I think it's worth doing.

dashboard/management/commands/createkeys.py
ssciolla: Probably personal preference, but something about it seems error-prone to me. In a limited case like this, there is low risk, but in a more complicated situation, I could see a developer losing track of what the current value is. For me, whatever small gains in terms of efficiency, memory usage, or succinctness when using a single variable are outweighed by the increased clarity when using multiple variables.

dashboard/management/commands/getjwk.py
lsloan: Although I can use `*_` to let Django pass a Request object without error, yet not require an argument, this doesn't seem ideal.  This tool uses `lti_new.get_jwks()` to avoid repeating the call to `pylti1p3` methods.  However, since this method returns JSON in a HTTP response, the tool must decode that to use it.

It would probably be better to refactor `lti_new.get_jwks()` into two methods.  One for Django to call that returns a JSON HTTP response and one to get the payload for that response.

ssciolla: I'm not really following, but if this is going to require a lot of tinkering with `lti_new`, I would vote to defer this, so you can get the people who worked on that module originally involved in whatever changes you're thinking about.

lsloan: I've updated `lti_new` in the past, so I think it's not a problem.

lsloan: I've updated `lti_new` in the past, so I think it's not a problem.

lsloan:
> …I felt (and still feel somewhat) that this isn't a top priority, given what the team wants to accomplish before the end of the year.

## TL;DR

I disagree. 😆 However, I sincerely appreciate your code review comments and testing.

## ~Details~ Rant

I feel it was a mistake to provide LTI support in MyLA without clear documentation and without implementing the configuration-by-URL feature.  I think providing good documentation and support should be a top priority, especially if it is to replace misleading docs.

While implementing better docs and config-by-URL, I learned that the instructions for generating JWKS was totally unnecessary.  I wanted to remove it completely, but other developers on the team objected.  Moving the code to a Django management tool is my way of compromising.

Regardless of whether this issue is a top priority, it was my assignment and I tried to implement it diligently.  Our process requires us to review each other's code, so I hope others will be conscientious enough to give reviews in a timely manner or leave a message saying they can't do it (for whatever reason).  As a group, we've had conversations about how much time is wasted when developers have to refocus after we've been interrupted or asked to switch to working on something else.  I think finishing an issue, opening a pull request for it, and getting code reviews on it only after more than a month has passed is along the same lines.  I'm already working on other issues, but I need to go back and refocus to respond to the comments.

Perhaps I'm foolish and I should have just accepted the approval and merged it. But I want to be respectful of my colleagues' opinions, learn from them, and maybe improve my code in the process.

OK.  Rant over.  Sorry.

dashboard/management/commands/createkeys.py
lsloan: I might be able to remove the dependency for `pycryptodome`, too.

ssciolla:
> I feel it was a mistake to provide LTI support in MyLA without clear documentation and without implementing the configuration-by-URL feature. I think providing good documentation and support should be a top priority, especially if it is to replace misleading docs.
> 
> While implementing better docs and config-by-URL, I learned that the instructions for generating JWKS was totally unnecessary. I wanted to remove it completely, but other developers on the team objected. Moving the code to a Django management tool is my way of compromising.

I'm not sure if I was a party to all of these discussions, since I don't remember the compromise or your voting to remove it entirely. I agree regarding the need for clear documentation, and it does make sense to separate this out if it's not really needed at this point. I have more to say about prioritizing, but I'll be quiet for now. Feel free to DM if you want to hear MY rant.

dashboard/management/commands/getjwk.py
lsloan: Maybe not. 🤔

jonespm:
I feel like the most effective way to get reviews might be if we had a weekly scheduled session where we all got together for the sole purpose of reviewing PR's. Then everyone could explain the issue, we could either resolve and merge it there or someone specific could be assigned to look at it afterward. Otherwise nobody feels like they need to take responsibility for it.

I'm also generally more of the philosophy ["Commit Often, Perfect Later" ](https://sethrobertson.github.io/GitBestPractices/#commit). Just get in quick reviews so nothing obvious is missed or can be improved and just fix anything up afterward. 

I also personally found this hard to review and read since the coding style mixes Javascript (Node.js) and Python.

PR #1104, opened 2020-10-08
Goalinput 941

### LINKED ISSUE #941: Are goal input changes stored too quickly?

Entering goals in the Assignment Planning (Redesign) view should be as simple and seamless a process as possible, and the number of GraphQL calls should be minimized.

Upon entering a goal in the view, each new character entered or deleted seems to trigger a new call storing the setting (separate calls for each step of the in-progress goal are visible in the Network tab of Developer Tools). While this doesn't seem to cause any problems technically speaking, it may be jarring to users who expect only one Snackbar notification after a new goal has been entered (not multiple, beginning after the first character change). Also, it might help performance a little to make only one GraphQL after a goal has been fully entered.

## Steps to Reproduce :
1. Enter a goal in any of the input fields in the Assignment Planning Redesign view.
2. Observe the Snackbar activation as soon as one character is entered.
3. Examine the Network tab of Developer Tools to see the separate calls for each character change.

## Additional context (Add any other context about the problem here) :

@jennlove-um and @justin0022, do either of you think this is a significant problem? I find it to be noticeable, but I can't speak for everyone. In terms of a solution, I think the ideal would be to have some kind of listener waiting for the user to stop entering data or to click away. I vaguely remember there being something like this already in the works or implemented, but I'm not seeing any behavior like this now.

chrisrrowland:
Closes #941 

The debouncing in useSetUserSettingGQL.js was not working correctly.  It was queuing up a series of queries and executing them all after a delay related to the size of the queue and blocking the UI.

I removed this and put debouncing in AssignmentGoalInput so that it can visually accept changes and then apply them appropriately debounced.


justin0022:
Hi @chrisrrowland, just took a look at the code, and it looks good to me. Thanks for looking into this - I'm still not certain why the debouncing wasn't working correctly the way it was previously set up.

justin0022 (approved):

chrisrrowland:
> Hi @chrisrrowland, just took a look at the code, and it looks good to me. Thanks for looking into this - I'm still not certain why the debouncing wasn't working correctly the way it was previously set up.

Not 100% sure either.  Still a beginner with react!

Issue #1056, opened 2020-08-26
add Hjson support for `env.json`


lsloan:
Adding Hjson support for the `env.json` configuration file would make maintaining and reading that file easier.  Comments would no longer need to be done in the hackish `"/* comment":"*/"` style.  It's that style that caused the bug in #1055, so this feature would also fix that bug.  Besides supporting comments, Hjson is a more relaxed syntax than JSON, which may be helpful to people setting up this application who are less familiar with the syntax.

Implementation would probably involve using a Hjson module to load the file, which converts the contents to JSON.  The resulting JSON would then be handled the usual way, with Python's `json.loads()`.

jonespm:
Thanks for creating this issue, I thought we'd already had an issue for this but it must be in another project. The 2 best alternatives I've seen are json5 and hjson. I've been using hjson for more projects lately as it's a simple swap. Previous json works and new hjson files add in new syntax to make it easier to read.

jonespm:
Cool, looks like you can just use hjson command to convert too! https://hjson.github.io/users.html Might try this out.

PR #1105, opened 2020-10-09
Implement LTI config by URL (#1054)

### LINKED ISSUE #1054: add support for LTI config by URL
Besides "manual entry" and "paste JSON" methods, LTI configuration can be done via "enter URL".  That method takes a "JSON URL", from which presumably the JSON used in the "paste JSON" method can be retrieved from a service.

Add a new entry to the MyLA API that can be used to generate the JSON.  Give the URL of that API entry to the LMS configuration to let it load the JSON itself.

Bonus points for:

* Including the template for the config JSON in the `env.json` file
* Include code to auto-generate the various key files needed

The goal is to greatly reduce the number of steps required to configure the tool while keeping it safe and secure.

### Test main feature
1. Go to Canvas' LTI Developer Key creation page:
    * https://umich.instructure.com/accounts/1/developer_keys#lti_key_modal_opened
1. In the "Method" menu, select the "Enter URL" item.
1. Copy the following URL and paste it into the "JSON URL" field:
    * https://nnnnnnnnnnnn.ngrok.io/lti/config/
    * Replace "nnnnnnnnnnnn" with the ID number of your ngrok server.
1. Copy the following URL and paste it into the "Redirect URIs" field:
    * https://nnnnnnnnnnnn.ngrok.io/lti/launch/
    * Replace "nnnnnnnnnnnn" with the ID number of your ngrok server.
1. (Optional, recommended) Enter meaningful values into the "Key Name", "Owner Email", and "Notes" fields.
1. Click the "Save" button.

### Test `LTI_CONFIG_TEMPLATE_PATH` property in `env.json`
1. Copy `assets/config/lti_config_template.json` from the project directory to the `/secrets` directory on the Docker machine (i.e., `/secrets/lti_config_template.json`).
1. Make some changes to `/secrets/lti_config_template.json`. (E.g., change the `title` or `description`, remove some values from `scopes`, etc.)
1. Change or add the `LTI_CONFIG_TEMPLATE_PATH` property to the `env.json` that will be used by the Docker machine.  It should look something like:
    * `"LTI_CONFIG_TEMPLATE_PATH": "/secrets/lti_config_template.json",`
1. Stop and restart the Docker machine to get MyLA running with the just-edited `env.json`.
1. In a web browser, request the config.  If using ngrok, the URL will be something like:
    * https://nnnnnnnnnnnn.ngrok.io/lti/config/
    * Replace "nnnnnnnnnnnn" with the ID number of your ngrok server.
1. Verify that the configuration JSON received corresponds to that in the file you edited.
1. With the Docker machine running, rename the file `/secrets/lti_config_template.json` to something else.  (E.g., `/secrets/lti_config_template-FUBAR.json`.)
1. Again, in a web browser, request the config.  If using ngrok, the URL will be something like:
    * https://nnnnnnnnnnnn.ngrok.io/lti/config/
    * Replace "nnnnnnnnnnnn" with the ID number of your ngrok server.
1. Verify that you receive error messages similar to the following:
    * In the web browser: `{"lti_error": "Error reading LTI template file \"/secrets/lti_config_template.json\": ([Errno 2] No such file or directory: '/secrets/lti_config_template.json')"}` (This should also have an HTTP status code of 500.  The status may not be easy to find in all browsers.  It may be useful to use `curl`, like `curl -i https://nnnnnnnnnnnn.ngrok.io/lti/config/`.)
    * In the Docker console output: `student_dashboard  | [2020-10-23 16:20:27 -0400] [ERROR] [lti_new.py:39] LTI error: Error reading LTI template file "/secrets/lti_config_template.json": ([Errno 2] No such file or directory: '/secrets/lti_config_template.json')`

lsloan:
Resolves #1054.

I'm working on wiki docs for this new feature.

## Changes based on code review
* [x] Move the LTI config template out of `config/env_sample.json` and into a new file, `assets/config/lti_config_template.json`.  See: https://github.com/tl-its-umich-edu/my-learning-analytics/pull/1105#discussion_r502669168
* [x] Change `"target_link_uri": "{base_url}{url_login}"` in template from `url_login` to `url_launch`.  See: https://github.com/tl-its-umich-edu/my-learning-analytics/pull/1105#discussion_r505511939

## Testing
More testing details in the issue, #1054. 
1. Go to Canvas' LTI Developer Key creation page:
    * https://umich.instructure.com/accounts/1/developer_keys#lti_key_modal_opened
1. In the "Method" menu, select the "Enter URL" item.
1. Copy the following URL and paste it into the "JSON URL" field:
    * https://nnnnnnnnnnnn.ngrok.io/lti/config/
    * Replace "nnnnnnnnnnnn" with the ID number of your ngrok server.
1. Copy the following URL and paste it into the "Redirect URIs" field:
    * https://nnnnnnnnnnnn.ngrok.io/lti/launch/
    * Replace "nnnnnnnnnnnn" with the ID number of your ngrok server.
1. (Optional, recommended) Enter meaningful values into the "Key Name", "Owner Email", and "Notes" fields.
1. Click the "Save" button.

### DIFF
--- a/config/env_sample.json
+++ b/config/env_sample.json
@@ -117,53 +117,6 @@
             }
         ]
     },
-    "LTI_CONFIG_TEMPLATE": {
-        "title": "MyLA autoconfig-{timestamp}",
-        "scopes": [
-            "https://purl.imsglobal.org/spec/lti-ags/scope/lineitem",
-            "https://purl.imsglobal.org/spec/lti-ags/scope/lineitem.readonly",
-            "https://purl.imsglobal.org/spec/lti-ags/scope/result.readonly",
-            "https://purl.imsglobal.org/spec/lti-ags/scope/score",
-            "https://purl.imsglobal.org/spec/lti-nrps/scope/contextmembership.readonly",
-            "https://canvas.instructure.com/lti/public_jwk/scope/update",
-            "https://canvas.instructure.com/lti/account_lookup/scope/show",
-            "https://canvas.instructure.com/lti/data_services/scope/create",
-            "https://canvas.instructure.com/lti/feature_flags/scope/show",
-            "https://canvas.instructure.com/lti/data_services/scope/list_event_types",
-            "https://canvas.instructure.com/lti/data_services/scope/destroy",
-            "https://canvas.instructure.com/lti/data_services/scope/list",
-            "https://canvas.instructure.com/lti/data_services/scope/update",
-            "https://canvas.instructure.com/lti/data_services/scope/show"
-        ],
-        "extensions": [
-            {
-                "domain": "{host}",
-                "platform": "canvas.instructure.com",
-                "settings": {
-                    "platform": "canvas.instructure.com",
-                    "placements": [
-                        {
-                            "default": "disabled",
-                            "placement": "course_navigation",
-                            "message_type": "LtiResourceLinkRequest",
-                            "target_link_uri": "{base_url}{url_launch}"
-                        }
-                    ]
-                },
-                "privacy_level": "public"
-            }
-        ],
-        "public_jwk": {},
-        "description": "MyLA automatically generated configuration",
-        "custom_fields": {
-            "user_username": "$User.username",
-            "canvas_user_id": "$Canvas.user.id",
-            "canvas_course_id": "$Canvas.course.id"
-        },
-        "public_jwk_url": "{base_url}{url_get_jwks}",
-        "target_link_uri": "{base_url}{url_launch}",
-        "oidc_initiation_url": "{base_url}{url_login}"
-    },
     "/*Database caching using mysql cache for ltiv1p3":"*/",
     "DB_CACHE_CONFIGS": {
         "/*cache timeout":"*/",
--- a/config/env_sample.json
+++ b/config/env_sample.json
@@ -117,6 +117,8 @@
             }
         ]
     },
+    "/*Full path in container to LTI config template.  If unset or null, defaults to location of 'config/lti_config_template.json' (usually '/code/assets/config/lti_config_template.json' in container)":"*/",
+    "LTI_CONFIG_TEMPLATE_PATH": null,
     "/*Database caching using mysql cache for ltiv1p3":"*/",
     "DB_CACHE_CONFIGS": {
         "/*cache timeout":"*/",
--- a/config/env_sample.json
+++ b/config/env_sample.json
@@ -146,7 +146,7 @@
                             "default": "disabled",
                             "placement": "course_navigation",
                             "message_type": "LtiResourceLinkRequest",
-                            "target_link_uri": "{base_url}{url_login}"
+                            "target_link_uri": "{base_url}{url_launch}"
                         }
                     ]
                 },
--- a/config/env_sample.json
+++ b/config/env_sample.json
@@ -117,6 +117,53 @@
             }
         ]
     },
+    "LTI_CONFIG_TEMPLATE": {
+        "title": "MyLA autoconfig-{timestamp}",
+        "scopes": [
+            "https://purl.imsglobal.org/spec/lti-ags/scope/lineitem",
+            "https://purl.imsglobal.org/spec/lti-ags/scope/lineitem.readonly",
+            "https://purl.imsglobal.org/spec/lti-ags/scope/result.readonly",
+            "https://purl.imsglobal.org/spec/lti-ags/scope/score",
+            "https://purl.imsglobal.org/spec/lti-nrps/scope/contextmembership.readonly",
+            "https://canvas.instructure.com/lti/public_jwk/scope/update",
+            "https://canvas.instructure.com/lti/account_lookup/scope/show",
+            "https://canvas.instructure.com/lti/data_services/scope/create",
+            "https://canvas.instructure.com/lti/feature_flags/scope/show",
+            "https://canvas.instructure.com/lti/data_services/scope/list_event_types",
+            "https://canvas.instructure.com/lti/data_services/scope/destroy",
+            "https://canvas.instructure.com/lti/data_services/scope/list",
+            "https://canvas.instructure.com/lti/data_services/scope/update",
+            "https://canvas.instructure.com/lti/data_services/scope/show"
+        ],
+        "extensions": [
+            {
+                "domain": "{host}",
+                "platform": "canvas.instructure.com",
+                "settings": {
+                    "platform": "canvas.instructure.com",
+                    "placements": [
+                        {
+                            "default": "disabled",
+                            "placement": "course_navigation",
+                            "message_type": "LtiResourceLinkRequest",
+                            "target_link_uri": "{base_url}{url_login}"
+                        }
+                    ]
+                },
+                "privacy_level": "public"
+            }
+        ],
+        "public_jwk": {},
+        "description": "MyLA automatically generated configuration",
+        "custom_fields": {
+            "user_username": "$User.username",
+            "canvas_user_id": "$Canvas.user.id",
+            "canvas_course_id": "$Canvas.course.id"
+        },
+        "public_jwk_url": "{base_url}{url_get_jwks}",
+        "target_link_uri": "{base_url}{url_launch}",
+        "oidc_initiation_url": "{base_url}{url_login}"
+    },
     "/*Database caching using mysql cache for ltiv1p3":"*/",
     "DB_CACHE_CONFIGS": {
         "/*cache timeout":"*/",
--- a/dashboard/lti_new.py
+++ b/dashboard/lti_new.py
@@ -69,7 +69,7 @@ def check_if_success_getting_tool_config(tool_config):
         logger.info('Success in fetching LTI tool configuration')
         return True
     else:
-        logger.error(tool_config)
+        logger.error(f'Invalid LTI configuration: "{tool_config}"')
         return False
 
 
@@ -84,7 +84,8 @@ def generate_config_json(request: HttpRequest) -> \
         Union[HttpResponse, LTIError]:
     config = get_tool_conf()
     if not check_if_success_getting_tool_config(config):
-        return LTIError(config).response_json()
+        return LTIError(f'Invalid LTI configuration: "{config}"')\
+            .response_json()
 
     parameters = {
         'timestamp': datetime.now().isoformat(),
--- a/dashboard/lti_new.py
+++ b/dashboard/lti_new.py
@@ -90,11 +90,10 @@ def get_jwks(_):
 
 
 def generate_config_json(request: HttpRequest) -> \
-        Union[HttpResponse, LTIError]:
+        Union[HttpResponse, JsonResponse]:
     config = get_tool_conf()
-    if not check_if_success_getting_tool_config(config):
-        return LTIError(f'Invalid LTI configuration: "{config}"')\
-            .response_json()
+    if not is_config_valid(config):
+        return lti_error(f'Invalid LTI configuration: "{config}"')
 
     parameters = {
         'timestamp': datetime.now().isoformat(),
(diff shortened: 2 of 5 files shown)


config/env_sample.json
```diff
@@ -117,6 +117,53 @@
             }
         ]
     },
+    "LTI_CONFIG_TEMPLATE": {
```
jonespm, line 120: It doesn't look like this is something anyone would have to ever modify and everything is replaced in the code.

If it's just going to be a static json template we should probably just have it out in it's own file, part of the app separate from this json and config.

lsloan: I wholeheartedly agree!

I [asked on Slack for opinions](https://app.slack.com/client/T69CPPGNP/GNSJ3RMHN/thread/GNSJ3RMHN-1602174463.002200) about where the JSON template should go.

I wanted to support HJSON for the template file to allow comments in it.  But when I thought about the places in the project where the file could go, the `secrets` directory came to mind.  Since `env.json` already existed there, I thought it might as well go there.  It doesn't support HJSON (see: #1056), unfortunately, but it seems like a good place for it.

I also thought that there's a chance the format of the LTI config JSON could change in the future.  Either because of the way LTI is implemented or some of the Canvas-related values change.  So putting the LTI config template in the main app config makes it easier for others to update it as necessary.

Suggest a place for the template file to go and I'll put it there.

lsloan:
FYI: My biggest regret with this feature is that I couldn't make the string formatting work in a subclass of [`JSONEncoder`](https://docs.python.org/3.8/library/json.html?highlight=jsonencoder#json.JSONEncoder).  It's fairly easy to add support for encoding classes that are currently unsupported.  Trying to change how it encodes strings, a fundamental part of JSON, was elusive.

I decided to give in and go with this pre-encoding step instead.

I'd still welcome any ideas of how to make a JSONEncoder subclass solution work, though!

Also, it just occurred to me, maybe the parameter values (other than timestamp) should be filled into the template when it's being *decoded* from the config file instead.  Maybe `JSONDecoder` stands a better chance of handling this easily, although not as flexibly as I'd like (loss of timestamp, etc.).

config/env_sample.json
jonespm: I see, I haven't been looking at slack (outside of prod support) this week to focus on data studio. I see what you were talking about though. To me it feel like it should go in the dashboard folder as it's part of the app. Assets is where the current static file are stored (when collect static is run).

I'd probably put it in like `assets/config` or `assets/json` ? 

Then you should be able to just read it with `static("config/lti_config_template.json") `

pushyamig: Yeah, I agree that the JSON template right now only covers the properties that we us in MyLA and we might want to add based on the needed. For Example `"default": "disabled",` this was added later as we did not want tool to be installed by default

pushyamig: Having this as a configuration option should be a way to go `config/lti_config_template.json` so that in case of changes we can just change the configuration. But there might be certain options we might associate with the code

dashboard/lti_new.py
```diff
@@ -75,6 +79,48 @@ def get_jwks(request):
...
+    elif isinstance(args, MutableSequence):
+        return [recursive_format(value, **kwargs)
+                for value in args]
+    else:
+        return args.format(**kwargs)
+
+
+def generate_config_json(request: HttpRequest) -> Union[
+    JsonResponse, LTIError]:
+    config = get_tool_conf()
```
pushyamig, line 105: What is the need for doing this? This is something PyLTI1.3 library after grabbing the LTI_CONFIG values try to do the LTI login/launch. when we configure the `/lti/config/` Myla simply return the JSON.

config/env_sample.json
```diff
@@ -117,6 +117,53 @@
...
+        "title": "MyLA autoconfig-{timestamp}",
```
pushyamig, line 121: If we don't specify the `KeyName` in the LTI dev key, will this value here replaced their?

config/env_sample.json
pushyamig: Their might be certain LTI configuration options we might not have used might use in future having it configuration option is good and having it as separate file(`lti_config_template.json`) is even better so that it doesn't clutter the env.json as that file purpose is purely for properties that differ by env

config/env_sample.json
pushyamig: I got confused with KeyName and title.

dashboard/lti_new.py
```diff
@@ -75,6 +79,48 @@ def get_jwks(request):
...
+        return LTIError(config).response_json()
+
+    params = {
+        'timestamp': datetime.now().isoformat(),
+        'host': request.get_host(),
+        'base_url': urllib.parse.urlunsplit(
+            [request.scheme, request.get_host()] + 3 * ['']),
+        'url_login': reverse('login'),
+        'url_launch': reverse('launch'),
+        'url_get_jwks': reverse('get_jwks'),
```
pushyamig, line 116: I like the way you were autofilling this rather me changing it manual in config

config/env_sample.json
lsloan: @pushyamig, I'm confused.

Which do you mean?

* [ ] Move LTI config template to `config/lti_config_template.json`
* [ ] Keep LTI config in `config/env_sample.json` (and therefore, the deployed `/secrets/env.json`)
* [ ] something else

lsloan: Thanks for clarifying!  I'll move it to `config/lti_config_template.json`.

dashboard/lti_new.py
lsloan: Thanks.

dashboard/lti_new.py
lsloan: Because the LTI should be enabled and configured properly before an LMS requests the config and tries to use it.

That seems important to me.

config/env_sample.json
lsloan: As I move this to another file, I had initially thought I would take the opportunity to add HJSON support to this.

However, as I think about it more, I have an opportunity to simplify this code.  Supporting HJSON for commenting and formatting probably won't be important for this file.  The comments would need to be parsed out before it could be sent to the requesting LMS anyway.  That would just add another step to the current process of replacing all the parameters in the string values and serializing everything back to JSON.

Instead, my thinking is to just use the new JSON file contents as one big string, apply formatting to it with the defined parameters, and return that to the LMS.  It wouldn't need the `recursive_format()` function at all.  That's a little disappointing, but it's probably worthwhile to reduce the complexity of the code.

dashboard/lti_new.py
lsloan: @pushyamig (or @jonespm), can you confirm whether checking MyLA's LTI config before generating the config for the LMS is OK, or does it cause a problem?

pushyamig: I tested it locally and it did create configuration and I pointed those config in my local instance and it worked as expected. I did that yesterday and not with updated changes. Should work even with recent changes as it is not something drastically different

pushyamig: @lsloan I don't see any harm done, it just instantiate the tool config. I personally would have not done it. Since this is purely communication between Myla and Canvas to get the LTI keys. Even let's say if we did not put any LTI_CONFIGS then it will throw errors. I don't know if Canvas notifies with the response we are sending or defaults to its own error message. I am fine either way

config/env_sample.json
lsloan: > Then you should be able to just read it with `static("config/lti_config_template.json") `

There were a few implementations of `static()` in Django, but none of them gave me the path to the file or its contents, as I was expecting.

In the documentation for Django's `staticfiles` app (https://docs.djangoproject.com/en/3.1/ref/contrib/staticfiles/#finders-module), I found that `finders.find()` would give me the full path to the file, so I use that like this:

```python
open(finders.find('config/lti_config_template.json')).read()
```

dashboard/lti_new.py
lsloan: To test what happens when MyLA's LTI configuration is invalid, I altered the code a little to simulate that.

## Results

### Canvas

Canvas showed the message

> Request failed with status code 422

MyLA's response actually has a status code of 500.  I believe 422 is the response from Canvas' internal processing to its UI.  (422 is "Unprocessable Entity", probably the error given when Canvas didn't find the expected content in the LTI config JSON.)

That message is good enough to tell the user something was wrong with MyLA.

### MyLA

MyLA logged an error message.  I'm making some changes to that log message now, to be a little more specific.

> [2020-10-16 14:58:16 -0400] [ERROR] [lti_new.py:72] Invalid LTI configuration: "0"

That message should help MyLA administrators that something is wrong.

The content of the response also includes the error.

```json
{"lti_error": "Invalid LTI configuration: \"0\""}
```

## Conclusion

It's helpful to check the LTI configuration, even if it's not being used in that block of code.

lsloan:
@jonespm and @pushyamig, please review this again and tell me if you find problems or give it an approval so I can merge it.

I've fixed a mistake in the `target_link_uri` property of the template and moved the template to a new file, which also simplified the logic.

dashboard/lti_new.py
pushyamig: Thanks for the detailed text. I guess Canvas will show the error response based on it's interpretation and it is fine that way. +1 with log message

dashboard/lti_new.py
```diff
@@ -64,17 +69,40 @@ def check_if_success_getting_tool_config(tool_config):
...
+        'base_url': urllib.parse.urlunsplit(
+            [request.scheme, request.get_host()] + 3 * ['']),
+        'login_url_suffix': reverse('login'),
+        'launch_url_suffix': reverse('launch'),
+        'jwks_url_suffix': reverse('get_jwks'),
+    }
+
+    return HttpResponse(
+        open(finders.find(
+            'config/lti_config_template.json')).read() % parameters,
```
pushyamig, line 102: I though we want to upload the `lti_config_template.json` as secret  as in `env.json`. so that we can change it on fly if we have to. Because all the file that are place in `config` are copied to a local `mylasecrets` and in docker-compose `${HOME}/mylasecrets:/secrets`. 
@jonespm Thoughts???

dashboard/lti_new.py
```diff
@@ -64,17 +69,40 @@ def check_if_success_getting_tool_config(tool_config):
...
+        content_type='application/json')
```
pushyamig, line 103: `open(finders.find(` this statement seems to directly opening the file and making an assumption the we will find that file there and exception handling/logging are not in place? 
Just before rendering as HTTPResponse we need to check if file exists and params are rightly applied and log accordingly stating the issue

lsloan: Yes, it does use the file without checking for errors.

If you think it's necessary to check for errors reading the file, I can add that.  I didn't think it was absolutely necessary because the file is being read from the project folders, as suggested by @jonespm.  That is, we have more control over the file than if it were in `/secrets`.

dashboard/lti_new.py
lsloan: @pushyamig, if that's what you meant, I didn't understand your comments.

In my comment at (https://github.com/tl-its-umich-edu/my-learning-analytics/pull/1105#discussion_r505579885), I asked for clarification.  You chose the "Move LTI config template to assets/config/lti_config_template.json" option and gave the comment a "thumbs up" reaction.

I also repeated the option in the comment at (https://github.com/tl-its-umich-edu/my-learning-analytics/pull/1105#discussion_r505583005) so there would not be any misunderstanding.

Putting the template into a secret is overkill.  It doesn't contain anything that would be considered "secret" and it's unlikely to be changed.

pushyamig: @lsloan Sorry for confusing comment. But Please keep the file as in the config directory and it's fine. I do agree we don't change the template file that often and if we do have to change something in emergency we can do it in canvas directly. This template is only used during initial creating time of LTI Key.

pushyamig (approved):

lsloan:
I appreciate the approval, @pushyamig.

I've been thinking about this issue this morning.

Although I don't think putting the template inside the env.json or a file in /secrets is the right thing to do, I think it wouldn't hurt if we provide some kind of "escape hatch" for admins setting this up, in case they need to change the template.  (To address a LTI spec change, etc.)

So, I think I'll add an option to env.json that will normally be null, but if it's not and it contains a pathname to an valid alternate template file, it will be used instead.  That will require a little more error handling around the use of the file, but I think it will be worth it.

Issue #986, opened 2020-07-14
Remove bootstrap references from code


jonespm:
Thank you for contributing to this project!

- Make sure to search the [issues](https://github.com/tl-its-umich-edu/my-learning-analytics/issues) for duplicates first!

## Describe the bug (Tell us what happens instead of the expected behavior) :

We still have one reference to bootstrap (in the rangeslider) but I don't think we're using it at all anymore and it semi-conflicts with the use of Materia-UI so we should just clean it all up.

jennlove-um:
It looks like this was resolved with #1115.  I'm closing this issue.

Issue #648, opened 2019-08-08
Create sourcemaps for front-end development


justin0022:
Currently webpack isn't generating [sourcemaps](https://webpack.js.org/configuration/devtool/) so when in Chrome Devtools, it's difficult to debug since the code is transpiled.

pushyamig:
In `webpack.config.js` if I do this `devtool: 'inline-source-map'` it creates the resource map but the trick is we need to know if this the build is in dev or prod mode. As it is suggested not to do it for prod mode

justin0022:
https://webpack.js.org/configuration/devtool/#production I think there's an option to disable in prod. I'll play around with the options and try to increase build speed.

jennlove-um:
MyLA is currently using sourcemaps, so I'm closing this issue.

PR #1115, opened 2020-10-26
I951 node 14 update

### LINKED ISSUE #951: Update Node base image to 14

Node 14 was released and it's a new even image. We should update this for the fall release. Additionally we should update any node modules if needed.

This was mentioned on #950

pushyamig:
Fixes #951 #648 #986 
1. The main goal was to upgrade to Node version 14 and clean up package.json
	- since Node 12 has entered into long term support. Node team tends to keep the even version long term support and odd version more like experimental. So that will be reason why I did not update it to 15. Multi-stage build is using Debian now
2. Added sourcemap support of easy debugging in the Dev environment
3. Cleaned up boostrap references. Resources rc-slider has some references of boostrap.css of it but it was specific to it's implementation and nothing with the boostrap library in package.json. Removing the boostrap.css did not had any impact on range slider functionality
4. I have upgrades babel version to 7 and cleaned up some reference to Babel 6. I felt it was necessary to do it.
5. I have also removed certain redundant libraries put in package.json. I tried to do a build and run and I don't see any bad effects due to removal. 
6. There still some Warning with deprecation of packages mostly with Jest and webpack. I tried to upgrade the Jest related package but did not had much effect until the package dependencies get updated. 
	- I Did not update webpack to 5 since the django-webpack-loader doesn't quite support that yet. That library is not well supported and saw this new library supporting [webpack](https://github.com/shonin/django-manifest-loader) splitchunks. Some research is needed here
	- I upgraded django-webpack-loader as I missed during django 3 upgrade
7. React 17 is out last month and I felt this can be taken as separate issue and similary upgrading material UI/d3 etc. These need a little more attention so did not feel like doing it here

Try to do a docker prune and run the build and also test the JEST 
`docker system prune -a --force --volumes`


pushyamig:
@justin0022 your input are valuable with `package.json` update since you are the person who set it up. Let me know if you need help with running MyLA locally
@chrisrrowland you have been working on Frontend mostly try to run locally making I did not do really crazy :)
@jonespm you are the docker expert so please review it from that standpoint.

jonespm (approved):
I think the Docker changes look good. I reviewed the rest and didn't notice anything but not an expert on it.

package.json
```diff
@@ -82,60 +91,29 @@
     ]
   },
   "devDependencies": {
-    "@babel/core": "~7.3.4",
+    "@babel/core": "~7.12.1",
+    "@babel/preset-env": "~7.12.1",
+    "@babel/preset-react": "~7.12.1",
+    "@babel/cli": "~7.12.1",
+    "@babel/register": "~7.12.1",
```
ssciolla, line 98: Looks like `babel`'s setup change quite a bit? Any documentation or articles you followed to update these?

package.json
```diff
@@ -82,60 +91,29 @@
...
+    "js-cookie": "~2.2.1",
     "mini-css-extract-plugin": "0.5.0",
     "node-sass": "~4.14.1",
-    "optimize-css-assets-webpack-plugin": "5.0.1",
-    "pnp-webpack-plugin": "1.4.1",
-    "popper.js": "~1.14.7",
-    "postcss-flexbugs-fixes": "4.1.0",
-    "postcss-loader": "3.0.0",
-    "postcss-preset-env": "6.6.0",
-    "postcss-safe-parser": "4.0.1",
```
ssciolla, line 119: For packages like these, how did you determine that they weren't being used? I'm not asking because I'm suspicious, I'm just trying to learn : )

ssciolla (approved):
I built this, ran the `jest` test suite, and navigated around the site without any console warnings or errors. From my ignorant perspective, this all looks good to me! I left some comments and questions, mostly asking for details about how you went about updating these.

ssciolla:
Forgot to ask: would this be a good time to add `package-lock.json` to the repo? 
https://docs.npmjs.com/cli/v6/configuring-npm/package-lock-json

jonespm:
Yeah, I hadn't done much with npm before doing a few tutorials on React/Vue and understand the purpose of the package-lock now. Makes sense to me to include this now and use NPM for managing these front-end dependencies rather than doing it manually.

pushyamig:
> Forgot to ask: would this be a good time to add package-lock.json to the repo?
docs.npmjs.com/cli/v6/configuring-npm/package-lock-json

I though about this but some how the docker staged build generate package-lock.json during the `npm install` but during the next step in the build the file is not carried on so when i went and looked in the webpack watcher container `docker exec -it webpack_watcher /bin/sh` i did not find the file. So it was not copy/paste the file from docker build. I had to go in the container and do a `npm install`. I was wondered why is that happening?

ssciolla:
> I though about this but some how the docker staged build generate package-lock.json during the `npm install` but during the next step in the build the file is not carried on so when i went and looked in the webpack watcher container `docker exec -it webpack_watcher /bin/sh` i did not find the file. So it was not copy/paste the file from docker build. I had to go in the container and do a `npm install`. I was wondered why is that happening?

Not sure about what happens to `package-lock.json` in these containers, but you could probably generate a `package-lock.json` just by doing a `npm install` outside of Docker on the command-line. I think I've done that in the past. It shouldn't be any different from what is generated internally, I don't think...

package.json
pushyamig: I just looked up the purpose of a package in `npmjs.com` and tried to look up the reference in the code and googled about the packages, ran the build and tested. I just went by rationale is build failed for missing package, runtime error, ran through couple of browsers and If none happened then I removed it. 

As you mentioned we might need to see as we all start using new version of `package.json` and find error due to this.

pushyamig:
> Not sure about what happens to package-lock.json in these containers, but you could probably generate a package-lock.json just by doing a npm install outside of Docker on the command-line. I think I've done that in the past. It shouldn't be any different from what is generated internally, I don't think...

I wanted to avoid that I guess I can do it out side as well. I will commit this file with this PR. But I did not investigate further as i felt i did not want to spend too much as that will side track the issue. I will open an issue to figure this out

package.json
pushyamig: Babel 7 has some notable changes like they packages started with `@` instead of `babel-core`. I saw a mix of both Babel 6 and 7. This is good to start with https://babeljs.io/docs/en/v7-migration. I read little more articles. Just followed build, run, test and delete approach.

package.json
```diff
@@ -15,11 +18,17 @@
     "lodash.debounce": "^4.0.8",
     "lodash.isequal": "^4.5.0",
     "rc-slider": "^8.7.1",
-    "react-helmet": "^6.1.0"
+    "react-helmet": "^6.1.0",
+    "react": "^16.10.2",
+    "react-app-polyfill": "^2.0.0",
+    "react-dom": "^16.10.2",
+    "react-ga": "~2.5.7",
+    "react-router-dom": "^5.0.0"
```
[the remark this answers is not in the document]
pushyamig: To me these are core packages used in FE and not just for Dev packages when used purely for building purpose that's what I understood when I read about so I changed if they have any impact but did not so I kept it here

pushyamig:
Fixed the Codacy alerts and commited in package-lock.json file.

package.json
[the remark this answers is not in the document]
ssciolla: Yeah, I agree, it makes more sense to me that they'd be core `dependencies`.

package.json
ssciolla: Thanks for the info!

chrisrrowland:
I noticed this warning.  Does this happen for apple folks as well?  I'll have to check if it was happening before to see if it's new.

Didn't break anything as far as I have seen so far, just noting it here.

```
Step 10/25 : RUN npm prune --production &&     find /usr/src/app/node_modules -type d -name "examples" -print0 | xargs -0 rm -rf
 ---> Running in e1121c2507a4
npm WARN enoent ENOENT: no such file or directory, open '/package.json'
npm WARN . No description
npm WARN . No repository field.
npm WARN . No README data
npm WARN . No license field.

up to date in 0.474s
```

pushyamig:
> I noticed this warning. Does this happen for apple folks as well? I'll have to check if it was happening before to see if it's new.
Didn't break anything as far as I have seen so far, just noting it here.
Step 10/25 : RUN npm prune --production &&     find /usr/src/app/node_modules -type d -name "examples" -print0 | xargs -0 rm -rf
 ---> Running in e1121c2507a4
npm WARN enoent ENOENT: no such file or directory, open '/package.json'
npm WARN . No description
npm WARN . No repository field.
npm WARN . No README data
npm WARN . No license field.
up to date in 0.47

I don't think I introduced this change, it was happening before and by looking at the  warning before I tried to fill in all the details about package.json and that was not the cause of the issue. This is intermediate step with multi-stage docker build I don't know if this is related to a node modules missing package.json. I can see when I login to the webpack_watcher container package.json exists.

We can tackle this as a separate issue

chrisrrowland:
> I don't think I introduced this change,
> 
> We can tackle this as a separate issue

Yeah I went back and confirmed this was happening before as well. I just hadn't noticed.

chrisrrowland (commented):
I poked around the site.  Didn't notice anything not working, and no errors popping up.

chrisrrowland (approved):
I poked around the site.  Didn't notice anything not working, and no errors popping up.

Issue #1113, opened 2020-10-22
Make the Assignment Planning view easier to scroll and see


jennlove-um:
### Describe your problem or feature you'd like added
UX testing and feedback from students indicated that they want to be able to see the progress bar, goal, and assignments at the same time.  They also had difficulty scrolling through the two parts of the screen.  We need to simplify the view to address this.

### Describe the solution you'd like
Make the top section of the screen with the progress bar and goal smaller.  Make the top bar and the goal field sticky so that they stay in view when scrolling through assignments. For desktop view, the goal and progress bar could be placed next to one another to save space.

pushyamig:
I started to work more are less we need to make both Progress bar and table header sticky i.e table header needs to stack with Progress bar. I felt Goal and progress bar will be in single line so I changed it as 

![Screen Shot 2020-11-17 at 9 53 37 AM](https://user-images.githubusercontent.com/8579775/99405525-df480800-28ba-11eb-9456-460bc85d0ed4.png)

Table Header needs to be sticky for the table content and sticky to the progress bar. I am running in to this issue where both header and progress are stacks up confusingly as you scroll

![Screen Shot 2020-11-17 at 9 59 35 AM](https://user-images.githubusercontent.com/8579775/99406393-d441a780-28bb-11eb-91d7-3120d8de41c6.png)

pushyamig:
I did not see the feasibility of making the Top portion ( goal + progress ) stack on top of Table header. It's like stacking 2 things and since the table introduces the scroll this make double scroll avoidable with standalone tool and 3 scroll bar with LTI. 

Having a triple scroll would be very annoying but something to keep in mind for next time when we introduce a scroll.

chrisrrowland:
The goal text now overlaps the Grade Progress text
![image](https://user-images.githubusercontent.com/17617998/102415228-d20b5f80-3fc5-11eb-86a6-90226ac8d6fa.png)
![image](https://user-images.githubusercontent.com/17617998/102415243-d6d01380-3fc5-11eb-8af9-ee7124d83e95.png)

ssciolla:
Thanks, @chrisrrowland. This is something we knew about and I showed to @jennlove-um before proceeding with the new layout. I'm going to work on a solution for this (I had one but it was nixed), but this is an edge case (most people won't be entering goals that low).

jennlove-um:
[...]

PR #1122, opened 2020-11-18
Assignment Planning V2: Assignment Name, Assignment Group, Graded Only filters.

### LINKED ISSUE #1114: Add filters to the Assignment Planning list of assignments
Student feedback from UX testing indicated that they would like a way to sort or filter the list of assignments to make it more manageable.

Add the following filters to the view: 
- [x] allow students to filter by assignment status (graded, not yet graded, unsubmitted)
- [x] filter by text search of assignments names (students enter text in a field to only include items with that text in the name)
- [x] filter by assignment group (note: this filter and associated column should be named "Type" to match Canvas vernacular)

Filters do not need to persist when students leave the view.
The action of selecting a filter needs to be saved as an event to the MyLA event log to track usage for research purposes.

chrisrrowland:
Closes #1114 

Added filters above the assignment planning v2 table.

Assignment name search logic simply looks for the search string to be present in the assignment name, case insensitive.

Graded only assignment checkbox toggles graded only on and off.  Checked will show only graded assignments,  Unchecked will show all assignments.  A graded column is added that shows a checkbox or nothing if the assignment is graded.

Additionally, assignment group filter is added but hidden.  Just need to remove the hidden attribute and uncomment the  filter chain line that filters by assignmentGroup. Waiting on UI requirements.


justin0022:
Starting this review now

justin0022:
Screen Shot 2020-11-20 at 11 06 14 AM
I'm not sure yet if this issue is caused by this PR, but looks like the green line alignment is off

pushyamig:
I will review this now

pushyamig:
you might want to have this branch up to date with master

zqian:
The branch has been updated with master. I have tested the PR locally:

1. assignment group filter is shown by default. 
2. all filters (assignment name, assignment group, and graded) work as desired.
3. I did not see the off-aligned green grade line problem, as Justin reported. 

image

I would say this PR is good for merge.

pushyamig:
Screen Shot 2020-11-24 at 10 39 41 AM

I feel the newly introduced section for filtering color is not sync with the color layout for the site

pushyamig:
> The action of selecting a filter needs to be saved as an event to the MyLA event log to track usage for research purposes.

According to the requirement the user action with filters needs to be saved `eventlog_log` table and currently that is not happening

pushyamig:
Currently, the current goal/ max goal lines are not being set for user other than admin. I don't feel this PR broke it but some where during the changes for assignment planning this is broken @jennlove-um  can you test and confirm from master branch. That needs to be fixed

pushyamig:
This is my personal view but my comment can be ignored, the filtering bar is taking up the space on top of table header and it look very odd and the whole table looks bulky

pushyamig:
The search word here needs to be part ( partials test / casing as handled) as assignment name, So it user mistypes it will be recognized. It is fair since that's what it is asked. But might need to think if misspelled word needs to be recognized. That will definitely out of scope might need a external library integration, these are just my thoughts

pushyamig:
To me the Filter section doesn't tell me what I am supposed to do, May be some text saying filter some where is needed

assets/src/components/AssignmentTable.js
```diff
@@ -120,170 +144,272 @@ function AssignmentTable (props) {
...
+  }
+
+  const [assignmentGroupFilterArray, setAssignmentGroupFilterArray] = useState([])
+  function handleGroupNameArrayFilterChange (e) {
+    setAssignmentGroupFilterArray(e.target.value)
+  }
+
+  function handleGradedOnlyFilterChange (e) {
+    setGradedOnly(!gradedOnly)
+  }
```
pushyamig, line 178: Nit Pick: The functions like above in various places can be rewritten as below. this is consistence with the pattern used in the code

```
const handleGradedOnlyFilterChange = e => {
setGradeOnly(!gradedOnly)
}
```

assets/src/components/AssignmentTable.js
```diff
@@ -120,170 +144,272 @@ function AssignmentTable (props) {
...
+
+  const ITEM_HEIGHT = 48
+  const ITEM_PADDING_TOP = 8
+  const MenuProps = {
+    PaperProps: {
+      style: {
+        maxHeight: ITEM_HEIGHT * 4.5 + ITEM_PADDING_TOP,
+        width: 250
+      }
+    }
```
pushyamig, line 168: I am asking this for my knowledge ... I see that this is used in the `Select` element, and `MenuProps` is an object and how exactly these are applied to the element

assets/src/components/AssignmentTable.js
```diff
@@ -120,170 +144,272 @@ function AssignmentTable (props) {
...
-                  <TableCell
-                    className={classes.tableCell + ' ' + classes.tableHeadCell}
-                    key={key}
-                  >
-                    {heading}
-                  </TableCell>
-                ))
+      <div>
+        <Grid container className={classes.filterArea}>
+          {/* Hidden until UI requirements are hashed out */}
```
pushyamig, line 191: how exactly are these hidden?

assets/src/components/d3/createResourceAccessChart.js
```diff
@@ -175,9 +175,9 @@ function createResourceAccessChart ({ data, weekRange, gradeSelection, resourceT
     .attr('height', availHeight)
     .attr('aria-hidden', 'true')
     .attr('transform', `translate(${margin.left}, ${margin.top})`)
-    .on('wheel.zoom', scroll)
+    .on('wheel.zoom', scroll, { passive: true })
```
pushyamig, line 178: Usually we try to keep changes separate between issues this is touching Resource accessed. It is fine for now as this is doesn't have a huge effect and it is just UI improvement

pushyamig:
This PR mostly reflect Filtering table and it is functional, but there are certain comment I raised are purely related to UI @jennlove-um  can weigh in thoughts and Event logging is not happening for filtering action can be taken up as separate issue

pushyamig (changes requested):
I left comments after my review

assets/src/components/AssignmentTable.js
```diff
@@ -120,170 +144,272 @@ function AssignmentTable (props) {
...
+          <Grid item xs={12} sm={5}>
+            <FormControl className={classes.formControl}>
+              <InputLabel>Assignment Group</InputLabel>
```
pushyamig, line 194: This Form is outside of the table, and we might be run into visibility issue as screen size varies. Since stacking order of element might not work perfectly. On a regular size laptop should work ok. for #1113 I am trying to minimize the top screen real estate I may need to consider this layout as well

chrisrrowland:
> Screen Shot 2020-11-24 at 10 39 41 AM
> 
> I feel the newly introduced section for filtering color is not sync with the color layout for the site

What are you looking to see here?  I made the filter region darker than the header so it would be clear they were different.

assets/src/components/AssignmentTable.js
chrisrrowland: Fixed

assets/src/components/AssignmentTable.js
chrisrrowland: It is how properties are set on the Select element in the  [material UI multi select](https://material-ui.com/components/selects/#multiple-select) documentation.

assets/src/components/AssignmentTable.js
chrisrrowland: Forgot to delete that comment.  Used to have a hidden attribute when we were talking about not supporting that filter.

assets/src/components/d3/createResourceAccessChart.js
chrisrrowland: Yeah this was completely unrelated, I was just eliminating a warning message.  I will try to keep these to separate issues in the future.

assets/src/components/AssignmentTable.js
chrisrrowland: It is outside the table (the table doesn't support filtering..) but it is [responsive ](https://material-ui.com/components/grid/)so the filter itself should respond ok to various screen sizes.

chrisrrowland:
> To me the Filter section doesn't tell me what I am supposed to do, May be some text saying filter some where is needed

Changed the text, is this better?
image

justin0022:
> Screen Shot 2020-11-20 at 11 06 14 AM
> 
> I'm not sure yet if this issue is caused by this PR, but looks like the green line alignment is off

I realized that the green lines are like that because the assignments are worth 0% of the final grade.

justin0022:
I was experiencing some significant lag/performance issues, but after running it in production mode (by modifying the `docker-compose.yml` to `command: sh -c "tar zxvf node_modules/all_symlinks.tgz && npm run prod"`, I'm getting much better performance, went from unusable to completely useable.

assets/src/components/AssignmentTable.js
```diff
@@ -120,170 +144,271 @@ function AssignmentTable (props) {
     }
   }, [currentWeekRow.current])
 
+  const [assignmentGroupNames, setAssignmentGroupNames] = useState([])
+  useEffect(() => {
+    const allGroupNames = assignments.map(ag => ag.assignmentGroup.name)
+    var uniqueGroupNames = [...new Set(allGroupNames)]
+    uniqueGroupNames.sort()
+    setAssignmentGroupNames(uniqueGroupNames)
```
justin0022, line 152: Could you simplify this to: 

```js
    const allGroupNames = assignments.map(ag => ag.assignmentGroup.name)
    const uniqueGroupNames = [...new Set(allGroupNames)].sort()
    setAssignmentGroupNames(uniqueGroupNames)
```

assets/src/components/AssignmentTable.js
```diff
@@ -120,170 +144,271 @@ function AssignmentTable (props) {
     }
   }, [currentWeekRow.current])
 
+  const [assignmentGroupNames, setAssignmentGroupNames] = useState([])
```
justin0022, line 147: I see why you declared the state here, but I think the usual convention in React is to declare all state variables near the top of the component. Could you move this line to be with the other `useState` calls above?

assets/src/components/AssignmentTable.js
```diff
@@ -120,170 +144,271 @@ function AssignmentTable (props) {
...
+  useEffect(() => {
+    const allGroupNames = assignments.map(ag => ag.assignmentGroup.name)
+    var uniqueGroupNames = [...new Set(allGroupNames)]
+    uniqueGroupNames.sort()
+    setAssignmentGroupNames(uniqueGroupNames)
+  }, [assignments])
+
+  const [assignmentFilter, setAssignmentFilter] = useState('')
```
justin0022, line 155: Same for this line, please put where the other `useState` calls are above.

assets/src/components/AssignmentTable.js
```diff
@@ -120,170 +144,271 @@ function AssignmentTable (props) {
...
+  const handleNameFilterChange = e => {
+    setAssignmentFilter(e.target.value)
+  }
```
justin0022, line 158: I think this line could be removed, and put inline on line 216:
`onChange={e => setAssignmentFilter(e.target.value)}`

assets/src/components/AssignmentTable.js
```diff
@@ -120,170 +144,271 @@ function AssignmentTable (props) {
...
+  const MenuProps = {
+    PaperProps: {
+      style: {
+        maxHeight: ITEM_HEIGHT * 4.5 + ITEM_PADDING_TOP,
+        width: 250
+      }
+    }
+  }
+
+  const [assignmentGroupFilterArray, setAssignmentGroupFilterArray] = useState([])
```
justin0022, line 171: Same for this line, please put where the other `useState` calls are above.

assets/src/components/AssignmentTable.js
```diff
@@ -120,170 +144,271 @@ function AssignmentTable (props) {
...
+  const handleGroupNameArrayFilterChange = e => {
+    setAssignmentGroupFilterArray(e.target.value)
+  }
```
justin0022, line 174: I think this could be removed, and put inline on line 198: 
`onChange={e => setAssignmentGroupFilterArray(e.target.value)}`

assets/src/components/AssignmentTable.js
```diff
@@ -120,170 +144,271 @@ function AssignmentTable (props) {
...
+
+  const handleGradedOnlyFilterChange = e => {
+    setGradedOnly(!gradedOnly)
+  }
```
justin0022, line 178: I think this line could also be removed and placed inline on line 225:
`onChange={() => setGradedOnly(!gradedOnly)}`

assets/src/components/AssignmentTable.js
```diff
@@ -120,170 +144,271 @@ function AssignmentTable (props) {
...
+
+  const ITEM_HEIGHT = 48
+  const ITEM_PADDING_TOP = 8
+  const MenuProps = {
+    PaperProps: {
+      style: {
+        maxHeight: ITEM_HEIGHT * 4.5 + ITEM_PADDING_TOP,
+        width: 250
+      }
+    }
```
justin0022, line 168: I'm wondering if this could be defined in another file and imported - is there any other component that could potentially use `menuProps`?

justin0022 (changes requested):
Some minor refactor suggestions, but otherwise looks good and works as expected!

pushyamig:
Since the Logging of the filtering is taken as separate issue, I think this issue can be merged.

assets/src/components/AssignmentTable.js
chrisrrowland: Currently I only see the Material Select being used in one other relevant place (ResourcesAccessed) but with no style properties defined.
I think it could make sense to put this somewhere common if we decide that we want this common appearance across more components but I feel like it might be early to know that is going to happen?

pushyamig:
I see one formatting error `my-learning-analytics/assets/src/components/ProgressBarV2.js:51:75: There should be no space before '}'`. can you fix this?

pushyamig (changes requested):

chrisrrowland:
> @chrisrrowland We can ignore codacy alerts since that is not configured with Standard. I don't see these are raised as errors with Standards

Ok, i was just trying to eliminate some of the red flags it was reporting.  I was just starting down the road of trying to switch from using standardjs to [ESLint with "standard js style"](https://travishorn.com/setting-up-eslint-on-vs-code-with-javascript-standard-style-aa3ab75bcc1c) in my environment, but  maybe that's the wrong thing to do.
After my last commit to fix some of these complaints codacy complained that they were still present, even though they were literally the only changes i made.  I'm not sure how it works.. when I browsed to the code within codacy the style issues it was complaining about weren't even there any more

pushyamig (approved):

Issue #1125, opened 2020-11-24
Assignment-Redesign: Current grade and max Possible grade line is not shown for students 


pushyamig:
Thank you for contributing to this project!

- Make sure to search the [issues](https://github.com/tl-its-umich-edu/my-learning-analytics/issues) for duplicates first!

### Expected behavior (A description of what you expected to happen) : 
The current/Max possible grades line is show for the default admin view but it is not visible after you become student. This was working before some where recent changes to the assignment planning v2 broke this. 

### Steps to Reproduce :
1. Become as a student user and current/max grade line is missing
2. 
3.

pushyamig:
This problem seems to happen for certain courses only and not for ever course. Needs to figure out what is the root cause for this.

pushyamig:
Test plan: 
1. Choose a course where some of the assignments points_possible =0 
2. you can also set any of the assignments `points_possible `column to 0 in `assignment` table to trigger this case

jennlove-um:
Testing passes: Current and max possible grades display correctly when a course has an assignment with 0 points.

PR #1126, opened 2020-11-30
#1125 fixes missing current and Max grade line for AssignV2

pushyamig:
#1125 

The missing of the current and Max grades was happening with certain assignment doesn't had points meaning points = 0 or null (no point at all). I can say sometime instructor creates these assignments as a reading assignments or for attendance purposes. The progress bar is not considering those assignments in those cases


chrisrrowland (approved):
Confirmed the missing current and max grade lines before fix are present w/ fix.

justin0022 (approved):

PR #1127, opened 2020-12-01
Fixes #1108 - Problem starting up when not setting GUNICORN_RELOAD

### LINKED ISSUE #1108: Problem starting up when not setting GUNICORN_RELOAD

If you don't set the environment variable GUNICORN_RELOAD it doesn't look like it's starting up. This is related to a change on #998. 

Because "${GUNICORN_RELOAD}" is quoted now it's causing a problem for the command line. I think that we should reconfigure these GUNICORN options to just allow you to pass whatever you want. Also it should probably reload when in development (or PTVSD) mode. 

### Steps to Reproduce :
1. Don't set GUNICORN_RELOAD
2. Try to startup
3. Error of "gunicorn: error: unrecognized arguments: "

jonespm:
Really just removed the quotes from gunicorn. With the exec it was causing it to expand to single quotes when not set and this is always either set to a specific variable or unset so safe.

Also setting it to reload now if debugging is on.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -27,8 +27,7 @@
             ],
             "port": 3001,
             "host": "localhost",
-            "justMyCode": false,
-            "subProcess": true
+            "justMyCode": false
         }
     ]
 }
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -27,7 +27,8 @@
             ],
             "port": 3001,
             "host": "localhost",
-            "justMyCode": false
+            "justMyCode": false,
+            "subProcess": true
         }
     ]
 }
--- a/start.sh
+++ b/start.sh
@@ -24,7 +24,7 @@ fi
 if [ "${GUNICORN_RELOAD}" ]; then
     GUNICORN_RELOAD="--reload"
 else
-    GUNICORN_RELOAD=""
+    GUNICORN_RELOAD=
 fi
 
 DOMAIN_JQ='.ALLOWED_HOSTS | . - ["127.0.0.1", "localhost", ".ngrok.io"] | if . | length == 0 then "localhost" else .[0] end'
@@ -81,11 +81,12 @@ if [ "${IS_CRON_POD:-"false",,}" == "false" ]; then
         echo "Starting Gunicorn for PTVSD debugging"
         # Workers need to be set to 1 for PTVSD
         GUNICORN_WORKERS=1
+        GUNICORN_RELOAD="--reload"
     fi
     exec gunicorn dashboard.wsgi:application \
         --bind 0.0.0.0:${GUNICORN_PORT} \
         --workers="${GUNICORN_WORKERS}" \
-        "${GUNICORN_RELOAD}"
+        ${GUNICORN_RELOAD}
  
 else
     if [ -z "${CRONTAB_SCHEDULE}" ]; then
--- a/start.sh
+++ b/start.sh
@@ -82,10 +82,12 @@ if [ "${IS_CRON_POD:-"false",,}" == "false" ]; then
         # Workers need to be set to 1 for PTVSD
         GUNICORN_WORKERS=1
         GUNICORN_RELOAD="--reload"
+        GUNICORN_TIMEOUT=0
     fi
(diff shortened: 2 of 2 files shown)


.vscode/launch.json
```diff
@@ -27,7 +27,8 @@
             ],
             "port": 3001,
             "host": "localhost",
-            "justMyCode": false
+            "justMyCode": false,
+            "subProcess": true
```
pushyamig, line 31: What is the purpose of the `subProccess`?

jonespm: I saw [this issue initially](https://github.com/microsoft/ptvsd/issues/1890#issuecomment-548176248) about Debugger exiting and this being a possible fix for it. It probably isn't necessary but it also doesn't seem to hurt.

jonespm: I can remove it as I don't think it makes a difference. I also noticed [ptvsd is deprecated](https://github.com/microsoft/ptvsd) as of 7 months ago and [debugpy](https://github.com/microsoft/debugpy/) replaces it. Will create a new issue for that.

start.sh
```diff
@@ -81,11 +81,14 @@ if [ "${IS_CRON_POD:-"false",,}" == "false" ]; then
         echo "Starting Gunicorn for PTVSD debugging"
         # Workers need to be set to 1 for PTVSD
         GUNICORN_WORKERS=1
+        GUNICORN_RELOAD="--reload"
+        GUNICORN_TIMEOUT=0
```
pushyamig, line 85: I can see `"PTVSD_ENABLE": true,` stopping on breakpoint is happening and do not  disconnect after 30 sec.

start.sh
```diff
@@ -81,11 +81,14 @@ if [ "${IS_CRON_POD:-"false",,}" == "false" ]; then
         echo "Starting Gunicorn for PTVSD debugging"
         # Workers need to be set to 1 for PTVSD
         GUNICORN_WORKERS=1
+        GUNICORN_RELOAD="--reload"
```
pushyamig, line 84: The reload is happening when `"PTVSD_ENABLE": true,` but not every-time the python code changes. 

```
student_dashboard  | [2020-12-02 12:24:42 -0500] [32] [INFO] Worker reloading: /code/dashboard/views.py modified
student_dashboard  | [2020-12-02 12:24:42 -0500] [32] [INFO] Worker exiting (pid: 32)
student_dashboard  | [2020-12-02 12:24:43 -0500] [44] [INFO] Booting worker with pid: 44

```

jonespm: Are you saying there's some problem with Gunicorn reloading changes? Or that this is working as expected?

pushyamig: I can see gunicorn server is starting up with out any issue when extra setting are removed from the `.env` and `docker-compose`.

But I don't see reload is happening as frequently I need when python code change. Infact I only saw one time. I don't know this is something specific to my machine

pushyamig: Also I have noticed when running with "PTVSD_ENABLE": true ( i.e one worker ) things are really slow. the website initial load happens and when I reload things are very slow. I tested this from standalone tool.

jonespm: Yeah, it will be a lot slower. We won't be running in production on this mode though. But this should have been about the same performance as before as runserver is single threaded. 

Can't use ptvsd in a multithreaded environment. Not sure if debugpy changes that.

These things you're mentioning related to code changes or performance aren't related to this change though. Can open a new issue and look at that more if you wanted but this really should get in otherwise it will cause issues with production instances.

pushyamig: Agreed. But I feel Python code changes are not picked is concerning for me when developing. Might be that it is purely my issue if other are not seeing it. I haven't being working on python code lately so i did not notice this earlier

pushyamig: I am fine with slowness when running with ptvsd enabled. I will enable it when need to use the breakpoint feature

pushyamig (approved):

jonespm:
I will test it more. I have another PR switching from ptvsd to debugpy. That might have some performance enhancements. If necessary we can switch back to django runserver, that isn't a big change anymore.

PR #1130, opened 2020-12-01
Fixes #667 - In cron access denied to bigquery should be an error.

### LINKED ISSUE #667: In cron, access denied to big query should be an error

If you cannot connect to big query it prints up a message

`[2019-08-16 16:14:20 -0400] [1] [INFO] 403 POST https://www.googleapis.com/bigquery/v2/projects/xxx-xxx-xxx/jobs: Access Denied: Project xxx-xxx-xxx: The user xxx-xxx-xxx@xxx-xxx-xxx.iam.gserviceaccount.com does not have bigquery.jobs.create permission in project xxx-xxx-xxx.`

Instead of an [INFO] This really should probably be an [ERROR] message which would fail the job

### LINKED ISSUE #746: If big query query is incorrect it just prints an INFO and continues, should be ERROR.

On this line when bigquery runs if there's a syntax error with the query (which is defined in the settings) it just seems to print an error message and continue processing. 
https://github.com/tl-its-umich-edu/my-learning-analytics/blob/578abaac4efa47f881d9d444b893d1eb59c10b50/dashboard/cron.py#L270

This will cause the files view to be empty and the process should be stopped here. I think we should move the files update (on line 515) to be before any other updates, then it would catch this syntax error first.

Since we can easily change this now in the settings it's possible this might be changed and an error a little harder to notice, though the view would be empty and there would be no data.

jonespm:
Fixes #746 - Prints error and fails is there is an error in query

Also updates bigquery library to latest version with pandas support

### DIFF
--- a/dashboard/cron.py
+++ b/dashboard/cron.py
@@ -296,7 +296,7 @@ class DashboardCronJob(CronJobBase):
                 total_bytes_billed += bq_job.total_bytes_billed
             # If there is an exception here print an error message and return
             except Exception as e:
-                err_msg = f"There was an exception running the bigquery job of {e.message}"
+                err_msg = f"There was an exception running the BigQuery job of {e.message}"
                 # Print this as an error so it will trigger an error in the cron
                 logger.error(err_msg)
                 return status + err_msg
--- a/dashboard/cron.py
+++ b/dashboard/cron.py
@@ -289,10 +289,17 @@ class DashboardCronJob(CronJobBase):
             job_config.query_parameters = query_params
 
             # Location must match that of the dataset(s) referenced in the query.
-            bq_query = bigquery_client.query(final_bq_query, location='US', job_config=job_config)
-
-            resource_access_df: pd.DataFrame = bq_query.to_dataframe()
-            total_bytes_billed += bq_query.total_bytes_billed
+            try:
+                bq_job = bigquery_client.query(final_bq_query, location='US', job_config=job_config)
+                # This is the call that could result in an exception
+                resource_access_df: pd.DataFrame = bq_job.result().to_dataframe()
+                total_bytes_billed += bq_job.total_bytes_billed
+            # If there is an exception here print an error message and return
+            except Exception as e:
+                err_msg = f"There was an exception running the bigquery job of {e.message}"
+                # Print this as an error so it will trigger an error in the cron
+                logger.error(err_msg)
+                return status + err_msg
 
             resource_access_row_count = len(resource_access_df)
             if resource_access_row_count == 0:
--- a/dashboard/cron.py
+++ b/dashboard/cron.py
@@ -289,17 +289,10 @@ class DashboardCronJob(CronJobBase):
             job_config.query_parameters = query_params
 
             # Location must match that of the dataset(s) referenced in the query.
-            try:
-                bq_job = bigquery_client.query(final_bq_query, location='US', job_config=job_config)
-                # This is the call that could result in an exception
-                resource_access_df: pd.DataFrame = bq_job.result().to_dataframe()
-                total_bytes_billed += bq_job.total_bytes_billed
-            # If there is an exception here print an error message and return
-            except Exception as e:
-                err_msg = f"There was an exception running the BigQuery job of {e.message}"
-                # Print this as an error so it will trigger an error in the cron
-                logger.error(err_msg)
-                return status + err_msg
+            bq_job = bigquery_client.query(final_bq_query, location='US', job_config=job_config)
+            # This is the call that could result in an exception
+            resource_access_df: pd.DataFrame = bq_job.result().to_dataframe()
+            total_bytes_billed += bq_job.total_bytes_billed
 
             resource_access_row_count = len(resource_access_df)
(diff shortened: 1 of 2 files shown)


jonespm:
TODO: If this fails, then the dates shouldn't be updated.

zqian:
I have added an extra ',' into the BQ sql to make it invalid. I did get an error in the log file as: 

"[2020-12-04 15:13:13 -0500] [ERROR] [cron.py:301] There was an exception running the bigquery job of Syntax error: Expected end of input but got "," at [1:34]

(job ID: 9a7b982e-f6b9-4fe8-869e-2e2b11d13577)"

I guess you can either address the date update problem with this PR or open another issue.

dashboard/cron.py
```diff
@@ -289,10 +289,17 @@ def update_with_bq_access(self):
...
-            resource_access_df: pd.DataFrame = bq_query.to_dataframe()
-            total_bytes_billed += bq_query.total_bytes_billed
+            try:
+                bq_job = bigquery_client.query(final_bq_query, location='US', job_config=job_config)
+                # This is the call that could result in an exception
+                resource_access_df: pd.DataFrame = bq_job.result().to_dataframe()
+                total_bytes_billed += bq_job.total_bytes_billed
+            # If there is an exception here print an error message and return
+            except Exception as e:
+                err_msg = f"There was an exception running the bigquery job of {e.message}"
```
zqian, line 299: bigquery -> BigQuery instead

requirements.txt
```diff
@@ -39,7 +39,7 @@ pysaml2==5.1.0
 SQLAlchemy==1.3.18
 psycopg2==2.8.5
 mysqlclient==1.4.6
-google-cloud-bigquery==1.26.0
+google-cloud-bigquery[pandas]==2.4.0
```
zqian, line 42: why not use [google-cloud-bigquery](https://pypi.org/project/google-cloud-bigquery/) directly? Newest is on 2.5.0

jonespm: `pandas` is a now an '[extra](https://stackoverflow.com/a/52475030/3708872)' for google-cloud-bigquery package. If you don't include it like this there is no support for Pandas, which we're using in the cron.

jonespm: It's a way of creating a package now and not requiring unnecessary packages that you may not need support for. Kind of cool really.

jonespm:
@zqian I think it should address both issues. I also noticed that this already had a try/except in the main loop to catch errors like this so what I'm doing here really isn't necessary. What we *want* to do is have some way of stopping the date getting updated when there is an exception, so this PR may need to be reworked.

jonespm:
I'm going to convert this back to draft as it needs some more work next week.

dashboard/cron.py
```diff
@@ -289,10 +289,10 @@ def update_with_bq_access(self):
             job_config.query_parameters = query_params
 
             # Location must match that of the dataset(s) referenced in the query.
-            bq_query = bigquery_client.query(final_bq_query, location='US', job_config=job_config)
-
-            resource_access_df: pd.DataFrame = bq_query.to_dataframe()
-            total_bytes_billed += bq_query.total_bytes_billed
+            bq_job = bigquery_client.query(final_bq_query, location='US', job_config=job_config)
+            # This is the call that could result in an exception
+            resource_access_df: pd.DataFrame = bq_job.result().to_dataframe()
```
jonespm, line 294: This was a slight change because of the difference in the library

pushyamig:
I will review this now

dashboard/cron.py
```diff
@@ -289,10 +289,10 @@ def update_with_bq_access(self):
             job_config.query_parameters = query_params
 
             # Location must match that of the dataset(s) referenced in the query.
-            bq_query = bigquery_client.query(final_bq_query, location='US', job_config=job_config)
-
-            resource_access_df: pd.DataFrame = bq_query.to_dataframe()
-            total_bytes_billed += bq_query.total_bytes_billed
+            bq_job = bigquery_client.query(final_bq_query, location='US', job_config=job_config)
```
pushyamig, line 292: Probably this is step the rate limit error happens. just an observation probably since the error is raised to the outer function where this method is called from. 

Optional: can we wrap this line `try-except` and raises exception exclusivity not sure what a best practice here since it will work either way

pushyamig:
This change looks functional, But I could not test with actual rate limiting scenario nor could anybody unless we reach the limit in reality. This is working with a simulated situation of manually raising the error. Probably QA of this going to be harder

pushyamig (approved):

dashboard/cron.py
jonespm: Yeah, I originally had it localized with a try/except but I didn't know what specific action to take that the outer one wouldn't just handle. It basically just logged a message and Raised the exception. 

It has to either get an Exception out of this method or return a tuple that indicates status. Seems more pythonic to just let it do the Exception.

jonespm:
Yeah, I didn't test with the rate limit, I just tested that the query didn't work and that the user didn't have access (the credentials were wrong).

PR #1141, opened 2020-12-04
#877 Event_logging addition for assignment v2

### LINKED ISSUE #877: [Redesign assignment] Event logging
Send the system-generated scores when saving user defaults to Database, this is to support Event logging

pushyamig:
Fixes #877

Currently the object is emitted as this 
```
{"goalGrade": "85", "goalGradePrev": 0, "currentGrade": 84.5, "maxPossible": 90, "assignments": [{"assignmentId": "17700000001139432", "goalGradeSetByUser": true, "goalGrade": 4.3, "goalGradePrev": 43}, {"assignmentId": "17700000001056893", "goalGradeSetByUser": true, "goalGrade": 69.6, "goalGradePrev": 68.6}, {"assignmentId": "17700000001056906", "goalGradeSetByUser": true, "goalGrade": 69, "goalGradePrev": 69.7}]}
```
https://github.com/tl-its-umich-edu/my-learning-analytics/issues/877#issuecomment-731243921


pushyamig:
The actual description of what needs to be done is here

https://github.com/tl-its-umich-edu/my-learning-analytics/issues/877#issuecomment-731243921

chrisrrowland:
When I set an assignment grade, the previous course goal grade is getting logged as 0

```
{
   "course_id":17700000000400707,
   "default_type":"assignment",
   "default_value":{
      "goalGrade":"86",
      "goalGradePrev":0,
      "currentGrade":91.6,
      "maxPossible":94.6,
      "assignments":[
	...
      ]
   }
}
```

chrisrrowland:
I reloaded the page and set a single assignment goal 3 => 13 but an assignment goal i previously set is still being logged

```
{
   "course_id":17700000000400707,
   "default_type":"assignment",
   "default_value":{
      "goalGrade":"86",
      "goalGradePrev":0,
      "currentGrade":91.6,
      "maxPossible":94.6,
      "assignments":[
         {
            "assignmentId":"17700000001139234",
            "goalGradeSetByUser":true,
            "goalGrade":0,
            "goalGradePrev":1
         },
         {
            "assignmentId":"17700000001139362",
            "goalGradeSetByUser":true,
            "goalGrade":13,
            "goalGradePrev":3
         },
	...
      ]
   }
}
```

assets/src/containers/AssignmentPlanningV2.js
```diff
@@ -74,6 +74,7 @@ function AssignmentPlanningV2 (props) {
 
   const [assignments, setAssignments] = useState([])
   const [goalGrade, setGoalGrade] = useState('')
+  const [goalGradePrev, setGoalGradePrev] = useState('')
```
pushyamig, line 77: > When I set an assignment grade, the previous course goal grade is getting logged as 0

I did set an empty string but internally when useState is called it is setting to 0. or probably internally goalGrade is zero so it is like that

`const preGrade = usePrevious(goalGrade)`

Is that confusing? should we change it to empty string ''

chrisrrowland (changes requested):
It seems like the individual event logs contain historical changes..

When I make a new single assignment grade change the log seems to contain the most historical setting for any assignment goal that has been set.

I am assuming the log should only contain the single thing that chaged.

Also the overall goal is being reported as "goalGradePrev":0 whenever an assignment goal changes.

assets/src/containers/AssignmentPlanningV2.js
chrisrrowland: I think ideally the goalGradePrev property wouldn't be present unless there was a change... what do you think?

pushyamig:
> I reloaded the page and set a single assignment goal 3 => 13 but an assignment goal i previously set is still being logged

It will since those are the assignment goals changes the user made for that assignment so it will log it. This is simply capturing user default setting and storing it `eventlog_log` table.

If we have to change to just log what changed then i feel we discussed to have new design pattern for capture event changes behavior for UI

pushyamig:
@jennlove-um can you provide your input on this https://github.com/tl-its-umich-edu/my-learning-analytics/pull/1141#issuecomment-742645772

jennlove-um:
Based on the time left for this release, we may have to accept this for now and explain how the logging works in the documentation.  However, it does seem like it will make it harder to analyze the data if the same change looks like it happened over and over again on different dates in the event log when it was actually only done once by the student. (If I'm understanding the issue correctly.)

assets/src/containers/AssignmentPlanningV2.js
pushyamig: Ok, I am only sending the `gradePrev` value when it is there

chrisrrowland: I just updated my course goal from grade  from 86 to 84 and the previous  grade (86) was not in the log

```
{
   "course_id":17700000000400707,
   "default_type":"assignment",
   "default_value":{
      "goalGrade":"84",
      "currentGrade":91.6,
      "maxPossible":94.6,
      "assignments":[
```

pushyamig: It is at the very end of the object? can you check?

chrisrrowland:
Set a course goal grade.
Set an assignment goal grade.

goalGradePrev is present in the event log.

Seems that if you set the course goal during your session it will get logged with later events.

chrisrrowland (changes requested):
Set a course goal grade.
Set an assignment goal grade.

goalGradePrev is present in the event log.

Seems that if you set the course goal during your session it will get logged with later events.

chrisrrowland (approved):
@jennlove-um Has approved the way the logging is working, so i have no issues

andrew-gardener:
Sorry to bring this up after its been closed for a while but was just wondering if this could be simplified slightly (front end wise) by not storing the previous grade in the event log? This would be due to the fact that that value can be retrieved from a previous event log entry.

aka when analyzing the event log you would probably get all events for a student to build their history of activity anyways and that would inherently provide the previous value entered over time.

Not sure if there is a specific analysis method in mind though.

pushyamig:
@jennlove-um based on the Andrews [comment](https://github.com/tl-its-umich-edu/my-learning-analytics/pull/1141#issuecomment-782323034) what are your thoughts?

jennlove-um:
This was the functionality requested by the research team for their analysis of the use data. After this term, when they've had a chance to work with the data, we will have a better sense of whether the previous grade is necessary or could be derived.  Since they haven't worked with the event data yet for this view, we don't have that feedback about their analysis yet.

PR #1142, opened 2020-12-08
Requirements updates for Assignment Planning V2 filters.

### LINKED ISSUE #1114: Add filters to the Assignment Planning list of assignments

### LINKED ISSUE #1139: Assignment filtering is selected and assignment goal change is not updated in UI

When ever user selects a assignment filter option and changes a assignment goal grade the UI is not updating with change. But this is not an issue when no filter option is selected
### Steps to Reproduce :
1. Set a Course goal grade
2. select a assignment filter
3. Change assignment goal grade and UI is not updating, I can almost say the UI update is not happening every-time. But I can see user setting saved alert is trigged for it sometime

chrisrrowland:
Closes #1114 
Fixes #1139

Replace graded only filter with assignment status filter
Add clear filters button
Status icons for all 3 states, color scheme to match progress bar scheme
Column names and filter text adjusted

### DIFF
--- a/assets/src/components/AssignmentTable.js
+++ b/assets/src/components/AssignmentTable.js
@@ -180,7 +180,7 @@ function AssignmentTable (props) {
         <Grid container className={classes.filterArea}>
           <Grid item xs={12} sm={5}>
             <FormControl className={classes.formControl}>
-              <InputLabel>Filter by Assignment Group</InputLabel>
+              <InputLabel>Filter by Type</InputLabel>
               <Select
                 labelId='assignment-group-checkbox-label'
                 id='assignment-group-mutiple-checkbox'
@@ -203,8 +203,8 @@ function AssignmentTable (props) {
           </Grid>
           <Grid item xs={12} sm={5}>
             <FormControl className={classes.formControl}>
-              <InputLabel>Filter by Assignment Name</InputLabel>
-              <Input value={assignmentFilter} placeholder='Filter by assignment name...' onChange={e => setAssignmentFilter(e.target.value)} />
+              <InputLabel>Filter by Name</InputLabel>
+              <Input value={assignmentFilter} placeholder='Filter by name...' onChange={e => setAssignmentFilter(e.target.value)} />
             </FormControl>
           </Grid>
           <Grid item xs={12} sm={1}>
@@ -230,8 +230,8 @@ function AssignmentTable (props) {
                   [
                     'Week',
                     'Due',
-                    'Assignment Group',
-                    'Assignment Name',
+                    'Type',
+                    'Name',
                     'Percent of Final Grade',
                     'Score / Out of',
                     'Graded',
--- a/assets/src/components/AssignmentTable.js
+++ b/assets/src/components/AssignmentTable.js
@@ -28,6 +28,7 @@ import ConditionalWrapper from './ConditionalWrapper'
 import StyledTextField from './StyledTextField'
 import { calculateWeekOffset } from '../util/date'
 import { roundToXDecimals, getDecimalPlaceOfFloat } from '../util/math'
+import { assignmentStatus } from '../util/assignment'
 
 const styles = theme => ({
   root: {
@@ -104,13 +105,7 @@ function AssignmentTable (props) {
     handleInputBlur
   } = props
 
-  const assignmentStatus = {
-    GRADED: 'Graded',
-    SUBMITTED: 'Not Yet Graded',
-    UNSUBMITTED: 'Unsubmitted'
-  }
-
-  const assignmentStatusNames = [assignmentStatus.GRADED, assignmentStatus.SUBMITTED, assignmentStatus.UNSUBMITTED]
+  const assignmentStatusNames = Object.values(assignmentStatus)
   const [filteredAssignments, setFilteredAssignments] = useState(assignments)
 
   const [popoverEl, setPopoverEl] = useState({ popoverId: null, anchorEl: null })
--- a/assets/src/components/AssignmentTable.js
+++ b/assets/src/components/AssignmentTable.js
@@ -103,15 +103,26 @@ function AssignmentTable (props) {
     handleInputBlur
   } = props
 
+  const assignmentStatus = {
+    GRADED: 'Graded',
+    SUBMITTED: 'Not Yet Graded',
+    UNSUBMITTED: 'Unsubmitted'
+  }
+
+
   const [popoverEl, setPopoverEl] = useState({ popoverId: null, anchorEl: null })
 
   const [gradedOnly, setGradedOnly] = useState(false)
 
   const [assignmentGroupNames, setAssignmentGroupNames] = useState([])
 
+  const [assignmentStatusNames, setAssignmentStatusNames] = useState([assignmentStatus.GRADED, assignmentStatus.SUBMITTED, assignmentStatus.UNSUBMITTED])
+
   const [assignmentFilter, setAssignmentFilter] = useState('')
 
   const [assignmentGroupFilterArray, setAssignmentGroupFilterArray] = useState([])
+
+  const [assignmentStatusFilterArray, setAssignmentStatusFilterArray] = useState([])
 
   const tableRef = useRef(null)
   const currentWeekRow = useRef(null)
@@ -219,18 +230,27 @@ function AssignmentTable (props) {
             </FormControl>
           </Grid>
           <Grid item xs={12} sm={1}>
-            <FormControlLabel
-              className={classes.formControl}
-              control={
-                <Checkbox
-                  checked={gradedOnly}
-                  onChange={() => setGradedOnly(!gradedOnly)}
-                  name='checkedB'
-                  color='primary'
-                />
-              }
-              label='Graded'
-            />
+            <FormControl className={classes.formControl}>
+                <InputLabel>Filter by Status</InputLabel>
+                <Select
+                  labelId='assignment-status-checkbox-label'
+                  id='assignment-status-mutiple-checkbox'
+                  multiple
+                  value={assignmentStatusFilterArray}
+                  onChange={e => setAssignmentStatusFilterArray(e.target.value)}
+                  input={<Input />}
+                  renderValue={(selected) => selected.join(', ')}
+                  MenuProps={MenuProps}
+                  width='250px'
+                >
+                  {assignmentStatusNames.map((name) => (
+                    <MenuItem key={name} value={name}>
+                      <Checkbox checked={assignmentStatusFilterArray.indexOf(name) > -1} />
+                      <ListItemText primary={name} />
+                    </MenuItem>
+                  ))}
+                </Select>
+              </FormControl>
           </Grid>
         </Grid>
         <TableContainer className={classes.container}>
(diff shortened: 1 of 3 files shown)


chrisrrowland:
Fixes #1139

assets/src/util/assignment.js
```diff
@@ -8,39 +8,60 @@ const clearGoals = assignments => assignments
...
-      ...assignments[key],
-      inputFocus,
-      inputBlur: !inputFocus
-    },
-    ...assignments.slice(key + 1)
-  ]
+const setAssigmentGoalInputState = (assignmentId, assignments, inputFocus) => {
+  const assignment = assignments.filter(a => a.id === assignmentId)
+  if (assignment.length !== 1) {
+    console.log('Error finding unique assignment id')
```
chrisrrowland, line 14: This is one of those "should never happen" things.  To better handle that would need to change the way the function is used.  Same with the other 2 of this ilk.

pushyamig:
I will review this today

assets/src/components/AssignmentTable.js
```diff
@@ -92,16 +104,27 @@ function AssignmentTable (props) {
     handleInputBlur
   } = props
 
-  const [popoverEl, setPopoverEl] = useState({ popoverId: null, anchorEl: null })
+  const assignmentStatus = {
+    GRADED: 'Graded',
+    SUBMITTED: 'Not Yet Graded',
+    UNSUBMITTED: 'Unsubmitted'
+  }
```
pushyamig, line 111: This Object can be separated out and can be put in a `utils/assignments.js` and be reused in `AssignmentTable.js` and in the `AssignmentPlanningV2.js`

assets/src/components/AssignmentTable.js
```diff
@@ -92,16 +104,27 @@ function AssignmentTable (props) {
...
+
+  const assignmentStatusNames = [assignmentStatus.GRADED, assignmentStatus.SUBMITTED, assignmentStatus.UNSUBMITTED]
```
pushyamig, line 113: This can be simplified as `Object.values(assignmentStatus)`, this will render as a list
Screen Shot 2020-12-09 at 12 02 51 PM

assets/src/components/AssignmentTable.js
```diff
@@ -156,6 +179,19 @@ function AssignmentTable (props) {
...
+  useEffect(() => {
+    setFiltersAreClear(assignmentNameFilter.trim().length === 0 && assignmentStatusFilterArray.length === 0 && assignmentGroupFilterArray.length === 0)
+  }, [assignmentNameFilter, assignmentStatusFilterArray, assignmentGroupFilterArray])
+
+  useEffect(() => {
+    setFilteredAssignments(
+      assignments
+        .filter(assignment => assignmentNameFilter.trim().length === 0 || assignment.name.toUpperCase().includes(assignmentNameFilter.toUpperCase()))
+        .filter(assignment => assignmentStatusFilterArray.length === 0 || (assignmentStatusFilterArray.indexOf(assignmentStatus.GRADED) >= 0 && assignment.graded) || (assignmentStatusFilterArray.indexOf(assignmentStatus.SUBMITTED) >= 0 && assignment.submitted && !assignment.graded) || (assignmentStatusFilterArray.indexOf(assignmentStatus.UNSUBMITTED) >= 0 && !(assignment.graded || assignment.submitted)))
+        .filter(assignment => assignmentGroupFilterArray.length === 0 || assignmentGroupFilterArray.indexOf(assignment.assignmentGroup.name) >= 0)
```
pushyamig, line 191: These 3 filter blocks inner logic can be separated as `const` name reflecting what it is doing and simply applied

This is just a pseudo code
```
const filterByAssignmentName = () => {
return assignment => assignmentNameFilter.trim().length === 0 || assignment.name.toUpperCase().includes(assignmentNameFilter.toUpperCase())
}
```

I think you can also separate the functions to utils folder, if you prefer as I see that the `AssignmentTable.js` is becoming very huge

assets/src/components/AssignmentTable.js
```diff
@@ -92,16 +104,27 @@ function AssignmentTable (props) {
...
 
-  const [popoverEl, setPopoverEl] = useState({ popoverId: null, anchorEl: null })
+  const assignmentStatus = {
+    GRADED: 'Graded',
+    SUBMITTED: 'Not Yet Graded',
+    UNSUBMITTED: 'Unsubmitted'
+  }
+
+  const assignmentStatusNames = [assignmentStatus.GRADED, assignmentStatus.SUBMITTED, assignmentStatus.UNSUBMITTED]
+  const [filteredAssignments, setFilteredAssignments] = useState(assignments)
```
pushyamig, line 114: I see that all the some of functions/const is using `filteredAssignments` instead of `assignments`. Can you explain what exactly the idea behind it. Since I can see `assignment` and `filteredAssignments` use approperiately)

pushyamig (changes requested):
UI not updating when assignment goal changes is not happening anymore.

assets/src/util/assignment.js
pushyamig: I don't think if we want to put console.log in JS code since it might appear browser for user, this case is a rare case.
But I think this can be handles like this https://stackoverflow.com/questions/41040266/remove-console-logs-with-webpack-uglify/41041580#41041580.

pushyamig: but since you are trying to find an assignment id from a list if you find 2 can you check if both are same and remove one. can that be handled like that?

pushyamig: @ssciolla @jonespm what are your thoughts on having console.log in JS code?

assets/src/components/AssignmentTable.js
```diff
@@ -382,13 +437,15 @@ function AssignmentTable (props) {
                           </div>
                         </TableCell>
                         <TableCell className={classes.veryNarrowCell}>
-                          {a.graded ? <DoneIcon /> : <div />}
+                          <Tooltip title={a.graded ? assignmentStatus.GRADED : a.submitted ? assignmentStatus.SUBMITTED : assignmentStatus.UNSUBMITTED}>
+                            {a.graded ? <GradedIcon className={classes.graded} /> : a.submitted ? <SubmittedIcon className={classes.ungraded} /> : <UnsubmittedIcon className={classes.unsubmitted} />}
+                          </Tooltip>
```
pushyamig, line 442: Where in the UI this little block is used?

assets/src/util/assignment.js
ssciolla: Hm, yeah, these cases can be tricky. In terms of logging in production, I've seen some people do it. Not sure if it's necessary here. What happens in the UI if this occurs? In this case, `console.error` would be more appropriate.

pushyamig:
One weird behavior I am seeing (but no necessary needs a fix) is when I change an assignment goal 4.5 ( with out using the up/down arrow ) when I delete 5 it also deletes `.`. It's like 2 deletes in one

assets/src/util/assignment.js
chrisrrowland: In the UI it would simply not apply the filter b/c its in a bad/unexpected state.. trying to find an assignment with the provided id (that it got from the client side - so it should exist) and it can't or it finds it more than once.  Neither of those situations should happen.  console.error would make more sense.. i left it in as someway that someone would be able to provide some feedback if they observed strange behavior, but it would obviously rely on that being manually provided to someone supporting the issue..

assets/src/components/AssignmentTable.js
chrisrrowland: #1139  was because the keys being used by the functions passed along to AssignmentTable were using indexes into the array of assignments as a 'key' for performing updates.  When the list was filtered inplace the keys started pointing to the wrong array index.  To fix this I switched to using the assignment id as the key.  

However there are several helper functions in AssignmentTable.js that explicitly reference consecutive array indices and I wanted to keep them functionality unchanged so I had used the filtered copy of the array.

Not sure if that really explains it well or not.

assets/src/components/AssignmentTable.js
chrisrrowland: Tooltip for the icon displayed in the Status column

chrisrrowland:
> One weird behavior I am seeing (but no necessary needs a fix) is when I change an assignment goal 4.5 ( with out using the up/down arrow ) when I delete 5 it also deletes `.`. It's like 2 deletes in one

Looks like it is because roundToXDecimals is being applied on every onChange event.

If I ever get back to that debounced input it should stop it from happening immediately (it would still happen if you paused after deleting the 5)

assets/src/components/AssignmentTable.js
chrisrrowland: Refactored the filter logic into individual functions

assets/src/components/AssignmentTable.js
pushyamig: Can you please add this in the comment? It make other people who step into the code easier if they have little bit of context

assets/src/util/assignment.js
```diff
@@ -232,7 +232,14 @@ const createUserSettings = (courseId, viewName, setting) => {
   return mutation
 }
 
+const assignmentStatus = {
+  GRADED: 'Graded',
+  SUBMITTED: 'Not Yet Graded',
+  UNSUBMITTED: 'Unsubmitted'
+}
```
pushyamig, line 239: Can you apply the change to the `AssignmentPlanningV2.js` file as well? since this code can be reused their. Should be simple Change!

chrisrrowland: Yeah, thanks I forgot that.

assets/src/util/assignment.js
```diff
@@ -232,7 +232,14 @@ const createUserSettings = (courseId, viewName, setting) => {
   return mutation
 }
 
+const assignmentStatus = {
+  GRADED: 'Graded',
```
pushyamig, line 236: Since this a an object, I feel it follow the camel casing standard. it should not be All caps

pushyamig:
I will rerun locally one more time

assets/src/util/assignment.js
chrisrrowland: I don't care either way but should they not be all caps?
https://google.github.io/styleguide/jsguide.html#features-objects-enums

pushyamig: Ok, Agreed. It's confusing to tell if this is an object or Enum. so in this scenario this a enum. and make sense to keep it like that

pushyamig:
These changes looks Good!

pushyamig (approved):

assets/src/util/assignment.js
chrisrrowland: I mean it's both right?  

Lets convert MyLA to Typescript real quick

ssciolla: We should .... but maybe not right now.

chrisrrowland: We can squeeze that in before Friday no probelmo 🤣

PR #1143, opened 2020-12-08
Updating docker ignore to ignore SQL and a few more file extensions.

### LINKED ISSUE #1134: Update the existing .dockerignore file to this repository to ignore additional unnecessary files for the build

When building with Docker, things like temporary files and SQL dumps shouldn't be included in the image. We aren't pushing these images anywhere from local host (which is good) but still they increase the size of the images locally and carry a risk if they were published

Just need to update the .dockerignore file: https://docs.docker.com/engine/reference/builder/#dockerignore-file

We already have one but it's missing some file extensions that get into the image.

jonespm:
Fixes #1134

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -3,6 +3,14 @@ saml
 bq_cred.json
 .idea/
 .DS_Store
+*.sql
+*.md
+*.sample
+*.tgz
+*.txt
+# However keep this file
+!requirements.txt
+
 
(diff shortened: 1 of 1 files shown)


jonespm:
Are there other files locally? These are the only ones I seem to have that aren't in this list. This should make the builds more secure and smaller for those that leave temp files in their local directory.

pushyamig (approved):

PR #1119, opened 2020-11-17
1113 Reducing the goal and progress bar real estate above the table

pushyamig:
#1113 
This PR is trying to reduce the assignment V2 top real estate so that progress bar is in visible table when changing the table and not have a sticky effect that table header has. More on the problem description is [here](https://github.com/tl-its-umich-edu/my-learning-analytics/issues/1113#issuecomment-740961211)

This change is on dev instance https://dev-myla.tl.it.umich.edu/. Please note that MyLA will be running as LTI tool so the screen width reduces in that case

PR #1145, opened 2020-12-08
Making top navigation fixed

### LINKED ISSUE #1137: Make the top navigation "fixed"

The top navigation bar should remain on the page as the page is scrolled down. Users may want to interact with this at any time and we may want to add more content in here.

This may be similar to #1113 and maybe have some fixed position.

jonespm:
Fixes #1137
Screen_Recording_2020-12-08_at_4 09 10_PM

### DIFF
--- a/assets/src/containers/DashboardAppBar.js
+++ b/assets/src/containers/DashboardAppBar.js
@@ -94,6 +94,7 @@ function DashboardAppBar (props) {
           </Popover>
         </Toolbar>
       </AppBar>
+      <Toolbar />
     </div>
   )
 }
--- a/assets/src/containers/DashboardAppBar.js
+++ b/assets/src/containers/DashboardAppBar.js
@@ -51,7 +51,7 @@ function DashboardAppBar (props) {
 
   return (
     <div>
-      <AppBar className={classes.root} position='static'>
+      <AppBar className={classes.root} position='fixed'>
         <Toolbar>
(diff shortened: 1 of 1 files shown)


jonespm:
@pushyamig Not sure if you were doing anything with this in #1119.

assets/src/containers/DashboardAppBar.js
```diff
@@ -51,7 +51,7 @@ function DashboardAppBar (props) {
 
   return (
     <div>
-      <AppBar className={classes.root} position='static'>
+      <AppBar className={classes.root} position='fixed'>
```
pushyamig, line 54: This change is chopping off the Title of all the views like I don't see `Resource access` title at all

jonespm:
Yeah, hmm, yeah I see that, that's not so good. ;)

assets/src/containers/DashboardAppBar.js
pushyamig: May be try adding a margin as suggested here https://stackoverflow.com/a/43795590/724238

jonespm:
Looks like [adding an empty Toolbar will bump down the content. ](https://github.com/mui-org/material-ui/issues/16844#issuecomment-517208151)

Looking better for me now, will update the screenshot

pushyamig (approved):
Looks Good!

PR #1148, opened 2020-12-09
Assignment Planning 2 assignments not sorted by due date

### LINKED ISSUE #1146: Assignments are not sorted by due date
Assignments are sorted by week, but within the week do not appear to be sorted by due date.

I'd assume they should be sorted by Week > Due Date > Group (*This is new) > Name 

image

chrisrrowland:
Fixes #1146 

Assignments are no sorted by multiple fields in decreasing precedence.

Due Date > Assignment Group Name > Assignment Name

They are being sorted by the literal due date (localDate field) however the UI currently shows the due date truncated to the  Month/Day so some assignments may appear in a confusing order with respect to alphabetical sorting of their group name/name.

For example, here the due date for Quiz 2 is earlier in the day than the due date for A&S 4, but that information is not currently reflected in the UI.  One might expect A&S to appear before Quizzes in an alphabetical sort.

image

### DIFF
--- a/assets/src/util/assignment.js
+++ b/assets/src/util/assignment.js
@@ -11,7 +11,7 @@ const clearGoals = assignments => assignments
 const setAssigmentGoalInputState = (assignmentId, assignments, inputFocus) => {
   const assignment = assignments.filter(a => a.id === assignmentId)
   if (assignment.length !== 1) {
-    console.log('Error finding unique assignment id')
+    console.error('Error finding unique assignment id')
     return assignments
   } else {
     const key = assignments.indexOf(assignment[0])
@@ -30,7 +30,7 @@ const setAssigmentGoalInputState = (assignmentId, assignments, inputFocus) => {
 const setAssignmentGoalGrade = (assignmentId, assignments, goalGrade) => {
   const assignment = assignments.filter(a => a.id === assignmentId)
   if (assignment.length !== 1) {
-    console.log('Error finding unique assignment id')
+    console.error('Error finding unique assignment id')
     return assignments
   } else {
     const key = assignments.indexOf(assignment[0])
@@ -49,7 +49,7 @@ const setAssignmentGoalGrade = (assignmentId, assignments, goalGrade) => {
 const setAssignmentGoalLockState = (assignmentId, assignments, checkboxState) => {
   const assignment = assignments.filter(a => a.id === assignmentId)
   if (assignment.length !== 1) {
-    console.log('Error finding unique assignment id')
+    console.error('Error finding unique assignment id')
     return assignments
   } else {
     const key = assignments.indexOf(assignment[0])
@@ -252,7 +252,14 @@ const createUserSettings = (courseId, viewName, setting) => {
   return mutation
 }
 
+const assignmentStatus = {
+  GRADED: 'Graded',
+  SUBMITTED: 'Not Yet Graded',
+  UNSUBMITTED: 'Unsubmitted'
+}
+
 export {
+  assignmentStatus,
   calculateAssignmentGoalsFromCourseGoal,
   calculateWeight,
   calculateCurrentGrade,
--- a/assets/src/util/assignment.js
+++ b/assets/src/util/assignment.js
@@ -167,22 +167,18 @@ const sumAssignmentGoalGrade = assignments => sum(
 )
 
 // assignmentGroup.name can be null
-const sortByAssignmentGroupName = (assignments) => {
-  return assignments.sort((a, b) => {
-    return (a.assignmentGroup.name === null && b.assignmentGroup.name === null) ? 0
-      : a.assignmentGroup.name ? a.assignmentGroup.name.localeCompare(b.assignmentGroup.name)
-        : b.assignmentGroup.name.localeCompare(a.assignmentGroup.name)
-  })
-}
+const sortByAssignmentGroupName = assignments => assignments.sort((a, b) => {
+  return (a.assignmentGroup.name === null && b.assignmentGroup.name === null) ? 0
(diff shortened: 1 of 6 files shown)


assets/src/util/assignment.js
```diff
@@ -145,14 +166,34 @@ const sumAssignmentGoalGrade = assignments => sum(
   assignments.map(a => a.goalGrade || 0)
 )
 
-const sortAssignmentsByWeek = assignments => {
-  const assignmentsWithDueDates = assignments
+// assignmentGroup.name can be null
+const sortByAssignmentGroupName = (assignments) => {
+  return assignments.sort((a, b) => {
+    return (a.assignmentGroup.name === null && b.assignmentGroup.name === null) ? 0
+      : a.assignmentGroup.name ? a.assignmentGroup.name.localeCompare(b.assignmentGroup.name)
```
justin0022, line 173: Are you trying to sort alphabetically here? 

Could this possibly be simplified to: 

```js
const sortByAssignmentGroupName = assignments => assignments.sort((a, b) => {
  if (a.assignmentGroup.name === null || b.assignmentGroup.name === null) return 0
  if (a.assignmentGroup.name < b.assignmentGroup.name) return -1
  if (a.assignmentGroup.name > b.assignmentGroup.name) return 1
  else return 0
})
```

assets/src/util/assignment.js
```diff
@@ -145,14 +166,34 @@ const sumAssignmentGoalGrade = assignments => sum(
...
+}
+
+// name can be null
+const sortByAssignmentName = (assignments) => {
+  return assignments.sort((a, b) => {
+    return (a.name === null && b.name === null) ? 0
+      : a.name ? a.name.localeCompare(b.name)
+        : b.name.localeCompare(a.name)
+  })
+}
```
justin0022, line 185: I think this can be simplified to:

```js
const sortByAssignmentName = assignments => assignments.sort((a, b) => {
  if (a.name < b.name) return -1
  if (a.name > b.name) return 1
  else return 0
})
```

assets/src/util/assignment.js
```diff
@@ -8,39 +8,60 @@ const clearGoals = assignments => assignments
...
-      goalGrade: goalGrade === '' ? '' : roundToXDecimals(Number(goalGrade), 3),
-      goalGradeSetByUser: !!goalGrade
-    },
-    ...assignments.slice(key + 1)
-  ]
+const setAssignmentGoalGrade = (assignmentId, assignments, goalGrade) => {
+  const assignment = assignments.filter(a => a.id === assignmentId)
+  if (assignment.length !== 1) {
+    console.log('Error finding unique assignment id')
+    return assignments
```
justin0022, line 34: Have you encountered this case happening, where the assignment is not found with the `assignmentId` supplied?

justin0022 (changes requested):

assets/src/util/assignment.js
chrisrrowland: Would that not fail to correctly sort elements reliably when one of the assignmentGroup.name values is null?

assets/src/util/assignment.js
chrisrrowland: No, frankly should be impossible

chrisrrowland: The assignmentId being passed in came from a list of assignmentIds for the assignments that the client was already displaying, so there is no reason this should ever happen.  But you know what they say  😛

pushyamig:
@chrisrrowland can you fix the conflicts here. Looks like you branch is not up to date with master

assets/src/util/assignment.js
pushyamig: typically assignment's will always be in a group, if that is not the case you are seeing then atleast we can reduce one return statement like this `=> assignments.sort((a, b) => {` and keep the rest as if and else logic same

assets/src/util/assignment.js
pushyamig: Comments same as above

pushyamig (approved):
Looks Good Functionally!

assets/src/util/assignment.js
justin0022: Ah okay, thanks for confirming!

assets/src/util/assignment.js
justin0022: > Would that not fail to correctly sort elements reliably when one of the assignmentGroup.name values is null?

I think I need to test that case to double check, but I think if either one or both of the `assignmentGroup.names` is null, it won't sort that pair.

chrisrrowland: Yeah that is what I meant by unreliably.  They will retain the order they were presented in, so if they were swapped as input into different sorts they would come out in a different order.

chrisrrowland: What I meant was that if the 2 inputs were in a different order on two different sorts, if it just returned 0 if either assignmentGroup,name was null, they 2 inputs would be in a different order after the sort was applied so 2 arrays with identical contents in a different order could be sorted with different outcomes.

assets/src/util/assignment.js
chrisrrowland: Comment same as above

assets/src/util/assignment.js
chrisrrowland: > typically assignment's will always be in a group, if that is not the case you are seeing then atleast we can reduce one return statement like this `=> assignments.sort((a, b) => {` and keep the rest as if and else logic same

assignmentgroup cannot be null but  assignmentgroup.name can be.

pushyamig:
I approved this before and good to go

chrisrrowland:
> I approved this before and good to go
Github still wants @justin0022 to approve the changes

Issue #878, opened 2020-02-05
[Redesign assignmnent] Make goal box error orange, not red


justin0022:
Need to set this in the theme object.

jennlove-um:
Testing passes: The new color displays when the goal score does not match the math for the assignment scores.

PR #1154, opened 2020-12-10
Introduce warning color to theme palette, use in StyledTextField (#878)

ssciolla:
This PR adds a `warning` color to the `defaultPalette` for use in the Assignment Redesign (specifically for the `StyledTextField`). Because `warning` is a standard color in Material UI (see [Palette](https://material-ui.com/customization/palette/)), it can be used with the `color` prop or `withStyles(theme => ({})...`. I conferred with @jennlove-um on the hex value choices; the `dark` variant is used for the text to ensure it meets W3C contrast requirements but fits with the `main` variant used in the outline and bottom line. The PR aims to resolve issue #878.

Note: the `info` color in the palette is also a shade of orange. We will be separately assessing whether it makes sense to modify that value for consistency/clarity.

### DIFF
--- a/assets/src/defaultPalette.js
+++ b/assets/src/defaultPalette.js
@@ -26,8 +26,10 @@ const defaultPalette = {
   info: {
     main: '#FFB74D'
   },
+  /*  The 'warning' color is used with text or interface elements to alert the user something may be wrong or
+      need their attention. */
   warning: {
-    main: '#FF6E00'
+    main: '#D84315'
   }
 }
 
--- a/assets/src/defaultPalette.js
+++ b/assets/src/defaultPalette.js
@@ -22,11 +22,12 @@ const defaultPalette = {
   link: {
     main: '#0000EE'
   },
-  /*  The 'info' color is used to confirm a submitted or stored by system state .
-    For now, this should be used directly through theme.palette.info.main instead of the color prop
-    (i.e. color='info'), which isn't fully supported. */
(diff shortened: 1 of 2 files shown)


assets/src/defaultPalette.js
```diff
@@ -22,11 +22,14 @@ const defaultPalette = {
   link: {
     main: '#0000EE'
   },
-  /*  The 'info' color is used to confirm a submitted or stored by system state .
-    For now, this should be used directly through theme.palette.info.main instead of the color prop
-    (i.e. color='info'), which isn't fully supported. */
+  /*  The 'info' color is used to confirm a submitted or stored by system state. */
```
ssciolla, line 25: @zqian, because `info` is a standard Material UI color name, I believe you can use it with the `color` prop in Material UI components.

chrisrrowland (approved):
This all looks technically sound to me and the right way to approach this.

That said the error theme and warning theme's are almost indistinguishable so it might be worth revisiting this topic when there is more time to focus on that specifically.  Warning is darker and kind of looks more ominous than error does, to me anyways.

Warning Theme
image

Error Theme
image

(I tested the above by just changing warning to error in theme.palette.warning.* in StyledTextField.js)

ssciolla:
Thanks for the review, @chrisrrowland. I agree we need to revisit the palette, but I hope having it wired it up this way is an improvement. Going to merge now.

PR #1157, opened 2020-12-10
Scroll to current week improvement

chrisrrowland:
Trying here to use a precise scrollTo to the _previous week_ instead of an approximate offset to the current week.

There is a still a hard coded offset but it is the fixed height of the table header row, which  should allow for precise scrolling, unless the style changes.

Need to do some testing with edge cases like current week = first week of assignments.


assets/src/components/AssignmentTable.js
```diff
@@ -164,21 +169,29 @@ function AssignmentTable (props) {
...
     setAssignmentGroupNames(uniqueGroupNames)
   }, [assignments])
 
+  useEffect(() => {
+    setWeeksPresentInCollection([...new Set(assignments.map(a => a.week))].sort((a, b) => { return a - b }))
+    const indexOfCurrentWeek = weeksPresent.indexOf(currentWeek)
+    if (indexOfCurrentWeek > 0) {
+      setPreviousWeek(weeksPresent[indexOfCurrentWeek - 1])
+    }
+  }, [assignments])
```
pushyamig, line 193: when user set the course goal from previous, in that case the scroll is not happening at all. I tried to remove it and tried various option but still scroll don't work. Some how when course goal is set ( from before ) assignment is [] when this effect takes place

chrisrrowland:
Getting consistency required less variance in the row heights.  I'm not sure why.
I narrowed the assignment group name column slightly and widened the assignment name column.
Assignment name is truncated at 50 characters and has a tooltip with the whole name.  I didn't think hard about 50 just chose a round number so this could be adjusted.
Also set a minWidth for the Score column because the  "/max" portion was wrapping on narrow screens which was ugly anyways.

assets/src/components/AssignmentTable.js
justin0022: @chrisrrowland could you explain why you're setting `weeksPresent.current` here? I'm actually not exactly sure what this block of code is doing in general - could you explain it at a high level? Thanks!

chrisrrowland: I believe @pushyamig added some of that, but the underlying concept was that not all calendar weeks necessarily have any assignments, so it may go 
Week 1
Week 2
Week 4
Week 5
etc

And when working off of the previous week, we just wanted it to be the previous week with assignments, so if it was currently Week 5 we'd scroll relative to Week 4, and if it were Week 4 we'd scroll relative to week 2.

justin0022: Ah, I see, thanks for clarifying.

pushyamig: For some reason when user never sets a course goal ( with filtered assignment in place) the `AssignmentTable` lacks an exact render so the assignment weeks value is available in between the render and not on the final render. So I have store all the week list as a local variable `weeksPresent`.

pushyamig (approved):

PR #1158, opened 2020-12-11
issue_1120 update the cron.py to not filter out user with null sis_user_id

zqian:


jonespm:
It looks like this was originally fixed as a bug in #324 because it was causing issues. Looking at that ticket the problem was more about duplicate users being returned but the duplicate users all had null sis_id's.

I think if we removed this constraint we'd still have to have something remove the duplicate users which could be based by user_id instead, either in the SQL or in the data frame, before writing it to the database.

I don't see any constraints or anything based on this column otherwise.

jonespm:
The specific course before that was causing the problem was 252307. Should make sure this still works with this change, otherwise we'd have to figure out some way to still make it work. I believe this course had a duplicate user with an SIS ID and without an SIS ID in the same class.

zqian:
Right. I found there are pairs of user entries where the one with empty sis_id has pattern of "xxxx@umich.edu" in the sis_name column, while there is another corresponding user entry with just "xxxx" in the sis_name and valid sis_id. 

I will try to rewrite the sql to filter out those false/duplicated entries.

jonespm:
I believe we may be able to replace this with a [ROW_NUMBER window function.](https://docs.aws.amazon.com/redshift/latest/dg/r_WF_ROW_NUMBER.html) This may need a little more testing though. 

The idea behind this is it will assign a row number to each user_id that is the same, while sorting by sis_user_id asc. This will put the null values lower in the rows and row_number 1 will be the first one either with an SIS_ID or null. I feel like even if this isn't it it's possibly close.

Saw this post here that included this toohttps://www.sisense.com/blog/4-ways-to-join-only-the-first-row-in-sql/

```
user_info as (select p.unique_name,p.sis_user_id, u.name, u.id as user_id, u.global_canvas_id
from (SELECT ROW_NUMBER() OVER (PARTITION BY user_id order by sis_user_id asc) AS row_number, * FROM pseudonym_dim) as p join user_dim u on u.id = p.user_id WHERE row_number = 1)
```

zqian:
Thanks, @jonespm. 

When executing the current MyLA "get user" script, I see duplicates in query results due to enrollments to multiple sections included in the same course. The query can certainly be rewritten, and below is my proposal to get unique users per course. I am writing Python scripts to compare this query result with what we have in MyLA prod database.

I would try out your suggestion, too. Let's discuss it after the break. 

with
                 course_fact as (select enrollment_id, current_score, final_score from course_score_fact
                                where course_id='{course_id}'),
                 enroll_data as (select DISTINCT
                                 e.user_id,
                                 e.type as type,
                                 c.current_score as current_grade,
                                 c.final_score as final_grade
                                 from enrollment_dim e
                                  left
                                 join course_fact c on e.id = c.enrollment_id
                                 where e.course_id = '{course_id}'
                                 and e.type in ('StudentEnrollment', 'TaEnrollment', 'TeacherEnrollment')
                                 and e.workflow_state= 'active'
                                 order by user_id, type, current_score, final_score),
                 user_enroll as (
                                 select p.unique_name as sis_name,
                                 p.sis_user_id as sis_id,
                                 u.name as name,
                                 u.global_canvas_id as user_id,
                                 '{course_id}' as course_id,
                                 en.current_grade as current_grade,
                                 en.final_grade as final_grade,
                                 en.type as enrollment_type
                                from enroll_data en, pseudonym_dim p,user_dim u
                                where p.workflow_state ='active'
                                and en.user_id = p.user_id
                                and u.id = p.user_id
                                 ),
                 final as (select user_id, name, sis_id, sis_name, course_id, current_grade, final_grade, enrollment_type  from user_enroll)
                 select * from final
                 order by final.user_id desc

zqian:
SQL updated according to suggestion. I have tested the "add_user" cron results for a couple of courses, and compared with prod MyLA db instance. All look good. Ready for a final review, @jonespm

jonespm (approved):

PR #1163, opened 2020-12-11
#1151 adding aria-label to index page buttons

### LINKED ISSUE #1151: Add aria-label="same value as the title of the 'i' link" to IndexPage buttons
Based on the accessibility review done in #907, we need to add the aria-label="same value as the title of the 'i' link" to IndexPage buttons.

pushyamig:
Fixes #1151 
I can see the when user tab through the Indexpage button I can see screen reader says what ever on the title of [routes.js](https://github.com/tl-its-umich-edu/my-learning-analytics/blob/4cabdbf34e3743894dba859a6e8b3e092c55e17d/assets/src/routes/routes.js)

### DIFF
--- a/assets/src/components/SelectCard.js
+++ b/assets/src/components/SelectCard.js
@@ -195,7 +195,7 @@ const SelectCard = props => {
   return (
     <>
       <Card className={classes.card} elevation={2}>
-        <CardActionArea>
+        <CardActionArea aria-label={'About ' + cardData.title}>
           {getLinkContents(cardData)}
         </CardActionArea>
         {
--- a/assets/src/components/SelectCard.js
+++ b/assets/src/components/SelectCard.js
@@ -195,7 +195,7 @@ const SelectCard = props => {
   return (
     <>
       <Card className={classes.card} elevation={2}>
-        <CardActionArea aria-label={'About ' + cardData.title}>
+        <CardActionArea aria-label={cardData.title}>
           {getLinkContents(cardData)}
         </CardActionArea>
         {


ssciolla (commented):
Okay, I may be reading the issue wrong, but I still don't think this is what was being asked for. The issue mentions the "i" link, which I believe is the i link. I believe the title referred to is the `title` referenced in the `getHelpLink` function here: https://github.com/tl-its-umich-edu/my-learning-analytics/blob/6998236f47b2857c4e68b44156e03e42a1f747eb/assets/src/components/SelectCard.js#L154-L166

In other words, I think we need to add an `aria-label` saying "About {name of view}" to the "i" icon links. VoiceOver already announces the name of the view when you just have the select card selected, but when you select the "i" link button, it doesn't say anything useful about it being a link to the help docs. I hope that is clear. Please reach out to me if you have more questions.

pushyamig:
> Okay, I may be reading the issue wrong, but I still don't think this is what was being asked for. The issue mentions the "i" link, which I believe is the "i" icon linking to docs specific to each view. I believe the title referred to is the title referenced in the getHelpLink function here:

@ssciolla Your understanding of the issue is correct and it was indeed the change to "i" link button. So according to @gsilver after consulting with Brandon he doesn't see this issue with "i" link needs to fixed. Thoughts???

What I fixed is something different which I don't know if an issue or nor. Should I close the PR?

ssciolla:
> What I fixed is something different which I don't know if an issue or nor. Should I close the PR?

Yeah, I don't think think what you've done is necessary, though probably wouldn't hurt anything. I still kinda wonder if the change that was requested would be helpful, but if they're saying it's not necessary, then we can focus on other things. You can close this PR.

PR #1164, opened 2020-12-14
Ap2 feedback

chrisrrowland:
Change 'Not Yet Graded' to 'Pending Grade'
Change "X" icon on clear filter button to "Clear Filters" text
Add an alert banner to say "No assignments match your filter selections" when no assignments match the selected filters.


pushyamig (approved):
Looks Good!

PR #1170, opened 2020-12-16
When a new course is added with a start date in the future, nothing i…

### LINKED ISSUE #1138: When a new course is added with a start date in the future, nothing is run and a day is missed

When a course (or courses) are added with a future start date (after current date) it should really not consider needing to populate this as a new course and skip it. 

If it doesn't it, when this future course is added, the system will do an resources accessed update taking this date as the the update date. This results in nothing getting updated for that date but the record of courses being updated still being made.

### Steps to Reproduce :
1. Add a course with a future start date (after today)
2. Run the cron
3. It should skip this course, but it's using this date in the future as the update date.

jonespm:
…s run and a day is missed.

* Changes the logic to return the minimum of either the earlier last run
  or the start time.
* Changes latest to earliest in the Django. Previously this didn't
  matter since all of these times were the same.

Fixes #1138

### DIFF
--- a/dashboard/models.py
+++ b/dashboard/models.py
@@ -187,7 +187,7 @@ class CourseQuerySet(models.QuerySet):
             earliest_start_date_of_new_courses = new_courses.earliest_start_datetime()
 
         # Return the lower value of existing_earliest and new_earliest, otherwise return None
-        return min((val for val in [earliest_data_last_updated, earliest_start_date_of_new_courses] if val is not None), default=None)
+        return min(filter(None, (earliest_data_last_updated, earliest_start_date_of_new_courses)), default=None)
 
 
 class Course(models.Model):
--- a/dashboard/models.py
+++ b/dashboard/models.py
@@ -182,12 +182,12 @@ class CourseQuerySet(models.QuerySet):
 
         earliest_data_last_updated = existing_courses.earliest("data_last_updated").data_last_updated
         # If there are new courses (courses with no last run) return the earliest time of all
-        earliest_start_date_of_new_courses = None
+        earliest_start_datetime = None
         if len(new_courses) > 0:
-            earliest_start_date_of_new_courses = new_courses.earliest_start_datetime()
+            earliest_start_datetime = new_courses.earliest_start_datetime()
 
         # Return the lower value of existing_earliest and new_earliest, otherwise return None
-        return min(filter(None, (earliest_data_last_updated, earliest_start_date_of_new_courses)), default=None)
+        return min(filter(None, (earliest_data_last_updated, earliest_start_datetime)), default=None)
 
 
 class Course(models.Model):
--- a/dashboard/models.py
+++ b/dashboard/models.py
@@ -172,17 +172,22 @@ class CourseQuerySet(models.QuerySet):
         """ Returns the datetime of the last cron run of all courses
             This checks for any courses where the data_last_updated value is null.
 
+            If there are new courses it will return the lower value of the last run date or the earliest start date.
+
         :return: datetime. Either the earliest or None
-        :rtype: None
+        :rtype: None, datetime
         """
         new_courses = self.filter(data_last_updated__isnull=True)
+        existing_courses = self.filter(data_last_updated__isnull=False)
 
+        existing_earliest = existing_courses.earliest("data_last_updated").data_last_updated
         # If there are new courses (courses with no last run) return the earliest time of all
+        new_earliest = None
         if len(new_courses) > 0:
-            return new_courses.earliest_start_datetime()
-        # Otherwise return the latest cron run
-        else:
-            return self.all().latest("data_last_updated").data_last_updated
+            new_earliest = new_courses.earliest_start_datetime()
+
+        # Return the lower value of existing_earliest and new_earliest, otherwise return None
+        return min((val for val in [existing_earliest, new_earliest] if val is not None), default=None)
 
 
 class Course(models.Model):
--- a/dashboard/models.py
+++ b/dashboard/models.py
@@ -168,7 +168,7 @@ class CourseQuerySet(models.QuerySet):
             logger.info(f"No courses in CourseQuerySet; returning None as the earliest_start_datetime")
         return earliest_start
 
-    def get_data_last_updated(self) -> Optional[datetime]:
+    def get_data_earliest_date(self) -> Optional[datetime]:
         """ Returns the datetime of the last cron run of all courses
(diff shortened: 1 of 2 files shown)


dashboard/models.py
```diff
@@ -172,17 +172,22 @@ def get_data_last_updated(self) -> Optional[datetime]:
         """ Returns the datetime of the last cron run of all courses
```
ssciolla, line 172: This method name may not be appropriate anymore, since you're using `earliest_start_datetime` inside of it (which has nothing to do with getting the `data_last_updated` value. I guess I'd argue there should be some higher-level method (`determine_resource_data_filter`?) that uses the results of `earliest_start_datetime` and `get_data_last_updated` to figure out what date should be used to filter.

Either that, or rename this method?

dashboard/models.py
```diff
@@ -172,17 +172,22 @@ def get_data_last_updated(self) -> Optional[datetime]:
...
+        new_earliest = None
         if len(new_courses) > 0:
-            return new_courses.earliest_start_datetime()
-        # Otherwise return the latest cron run
-        else:
-            return self.all().latest("data_last_updated").data_last_updated
+            new_earliest = new_courses.earliest_start_datetime()
+
+        # Return the lower value of existing_earliest and new_earliest, otherwise return None
+        return min((val for val in [existing_earliest, new_earliest] if val is not None), default=None)
```
ssciolla, line 190: I'm finding this difficult to reason about so I'm writing some basic pseudocode describing the behavior I think we should be aiming for (not necessarily just in these methods, but in general in determining the cutoff for the cron's queries).

```
if not new_courses:
   return data_last_updated
else:
   if not data_last_updated (all courses are new) or new_courses.earliest_start_datetime < data_last_updated:
      return new_courses.earliest_start_datetime
   else:
      return data_last_updated
```

Aren't we kind of missing the cases where we would still want to use the `data_last_updated` (when all courses are old or all the new courses are more recent than when the data was last updated)? Still seems like we'd want to avoid fetching data we already have. I'm also not sure we need to know the `existing_earliest`?

ssciolla (commented):
I started looking/thinking about this, and I'm leaving some feedback. I could be completely off-base here, so please clarify if I'm forgetting some important details. I'm happy to talk with you about the PR directly.

dashboard/models.py
ssciolla: Okay, wait, I was misreading what was going on with `existing_earliest`. Let me re-examine.

ssciolla: Yeah, what you have is basically what I wrote in pseudocode, maybe even more concise. I think `existing_earliest` could be better named, since the naming of `new_earliest` made me keep thinking it was using `earliest_start_datetime`. I think I'd prefer just `data_last_updated`.

dashboard/models.py
jonespm: When I was fixing this I noticed that using latest instead of earliest was incorrect since we'd need to go back to the earliest updated course. But since all the dates were the same it *really* didn't matter at the moment. 

I can change this method name here to be consistent with this.

dashboard/models.py
lsloan, line 190: Clever way of getting the minimum of the values when they may be `None`.  I've tried to think of a simpler way to do this, but this seems to be the best.

Could `existing_earliest` ever be `None`?  I guess it probably could if the data had never been updated.

lsloan (approved):
This is a clever solution.

dashboard/models.py
jonespm: Thanks for thinking about it. I did basically the same thing and tested it in a few scenarios by manipulating the database. I think the test plan on the issue should be pretty good and is what is expected by this pseudocode. 

I basically do the same thing by returning the min value of the two rather than doing some other date checks here. Really I just want. I can look at the variable names and try to improve them.

Really what I was thinking of is that last line

# New courses means that the data has never been updated
# Data Last Updated is a field added once courses have been processed at least once
`return min(earliest_data_last_updated, earliest_start_date_of_new_courses)
`
However `earliest_data_last_updated` would be `None` on the first run. `earliest_start_date_of_new_courses` would be `None` if there are no new courses. Both would be `None` if there are no courses at all and cron is run. 

If one of them is `None` this caused a problem so had to protect this with a for loop. And if **both** of them are `None`, still should just return `None` since cron is protecting this case. This could have been a lot more if statements as your pseudocode has, I just felt like I'd miss a check.

jonespm: Thanks, yeah originally I had a bunch of if statements (as Sam's suggested) and was like there has to be something better and it's possible a value slips thought the cracks.

I almost had it but when the List was None it gave another error, luckily [Python 3.4 added a default](`earliest_start_date_of_new_courses` would be None if there are no new courses.) to return if it's None, otherwise it has to be in a try/except. So this seemed perfect. 

Yeah, both or either have a possibility of being None.

lsloan: Yeah, looking back at your conversation with Sam, I see the evolution your code went through.  I avoided looking at the earlier comments when I was giving my review so I could look at it without preconceived opinions.

lsloan: FWIW, the only improvements I could think of for the logic are:

## Remove `is not None`
```suggestion
        return min((v for v in [existing_earliest, new_earliest] if v), default=None)
```
This is slightly less easier to read, though.

## Use `filter()`
```suggestion
        return min(filter(None, (existing_earliest, new_earliest)), default=None)
```
I think `filter()` is pretty self-explanatory.

dashboard/models.py
jonespm: Renamed this method to `get_data_earliest_date`

dashboard/models.py
jonespm: Renamed these methods to match the example here

jonespm: I guess with the default "is not none" isn't needed. 

I read about using [filter verses a for loop here. ](https://stackoverflow.com/a/53212104/3708872)

It sounds biggest difference is a simple filter like this will filter out zeroes whereas the for loop doesn't. You'd need a more complex function or lambda. In this case I don't **think** we'd have to worry about this that problem since it's doing it based on datetime. Otherwise they probably have the same result and mostly based on syntax preference. It doesn't matter much to me either way.

jonespm: Seems fine going with filter to me, I'm okay with whatever "Commit suggestion" is being pressed here.

jonespm: I committed the filter suggestion.

PR #1171, opened 2020-12-16
Fix config error in JWK code (Issue #1153)

### LINKED ISSUE #1153: config error in JWK code
With LTI enabled, in the url `/lti/jwks` it gives the error:

```
NameError at /lti/jwks/
name 'check_if_success_getting_tool_config' is not defined
```

We should probably either remove this route or fix this method so this it exists.

https://github.com/tl-its-umich-edu/my-learning-analytics/blob/b66eaa21027c0de38a1df0c9d3a0a7786e0d600a/dashboard/lti_new.py#L105

lsloan:
Fixes #1153.

Replace an old reference that had crept back in while resolving conflicts that arose while working on other issues.

### DIFF
--- a/dashboard/lti_new.py
+++ b/dashboard/lti_new.py
@@ -102,7 +102,7 @@ def is_config_valid(config: ToolConfDict):
 
 def generate_jwks() -> Dict[str, list]:
     config = get_tool_conf()
-    if not check_if_success_getting_tool_config(config):
+    if not is_config_valid(config):
         raise LTIException(config)
     return config.get_jwks()
 


jonespm (approved):
Looks good, I didn't notice that it was just this method getting renamed and this one missed.

lsloan:
Yeah, it's the same logic, just a simpler function name.  At the time I changed it, I was thinking of refactoring it all.  However, this validator-like function has some logging statements in it that I thought might be desirable.  I didn't feel like trying to determine whether they should appear somewhere else.

Issue #1067, opened 2020-08-28
Hide object level data on the confirmation page in admin delete page in Django


jonespm:
## Expected behavior (A clear and concise description of what you expected to happen) : 
When an administrator deletes a MyLA course in Django, they get a confirmation page.  That page should only have the summary of the data being deleted.  It should not list all of the individual objects since that list can get long and does not provide additional value in determining whether to delete the course.

## Describe the bug (Tell us what happens instead of the expected behavior) :
Sometimes deleting  course that has a lot of resources and resources_accessed  information was taking take a long time and generated a 504. While the 504 no longer occurs, the list of resources is still long and unnecessary.

## Steps to Reproduce :
1. Add some courses and run the cron, the courses need to have a decent amount of resources and resource_access data pulled in.
2. Select to delete the course
3. The confirmation page displays with summary and list of objects. The object list is unnecessary.

## Additional context (Add any other context about the problem here) :

Ideally we can just have it display the summary with just the details or skip the confirmation page.

I found this article describing a little about that: https://stackoverflow.com/questions/28365331/how-to-skip-delete-confirmation-page-in-django-admin-for-specific-model

I don't think we need the details.

ssciolla:
https://docs.djangoproject.com/en/3.1/ref/contrib/admin/actions/#actions-that-provide-intermediate-pages

ssciolla:
@jonespm, I was able to override the admin templates so they don't show all those details; see https://github.com/tl-its-umich-edu/my-learning-analytics/commit/93f64e7a272dfe9264d941293c9dcdec600e775a The pages are still loading pretty slowly, which I guess makes sense since the data is still behind the template. I could modify the views to try to change how it loads things, but feels tricky. Maybe some degree of slowness is unavoidable if we're talking about deleting 500 or more Resource Access records?

ssciolla:
Okay, now I'm not sure what to do. After talking with @jennlove-um and doing some testing, this only seems to be an issue on the `test` instance (and by association, Canvas test). Production works fine, and renders hundreds of related objects pretty quickly.

The template overrides I created did not fix the problem on the test instance when I deployed them. I'm not sure what the resource limits or settings we have for the test MyLA deployment and DB, but this feels like a Canvas test bottleneck to me. Even loading the course list in the admin is pretty slow.

jonespm:
[...]

PR #1172, opened 2020-12-16
Add modified admin delete confirmation templates for course (#1067)

ssciolla:
This PR overrides portions of the `delete_confirmation.html` and `delete_selected_confirmation.html` templates for course objects so as to hide the extensive related-object information included by default. Summaries of what related objects will be deleted are still included. The PR aims to resolve issue #1067.


ssciolla:
I couldn't find any ways to further limit the code copied verbatim here, since the most-specific block containing the things I needed to change is `content`. Let me know if you have any further ideas. Adding these template overrides seems somewhat risky for maintenance purposes, since any small change to the template language or the admin templates themselves could break this. I'm not sure how else to mitigate that.

ssciolla:
@jonespm, any thoughts about this? I found a Django feature request about this problem (https://code.djangoproject.com/ticket/10919), and there were some suggestions about modifying the template context using method overrides, but that kind of thing might also be difficult to maintain and may also not solve the data marshaling problem (i.e. the data is still being queried for, which is probably the root cause of long-loading times). This PR "does the thing", but I think I'm still uncertain about whether the maintenance cost is worth the benefit.

jonespm (approved):
I think this looks like a good fix!

You're extending the template and just overriding the content. These files don't change much. Maybe in the future Django could break this into better organized blocks but this should be good and only require minor updates when we pick up major new releases of Django.

jonespm:
I also tested this out locally and I think the summary looks better, Django should have this as an option by default. Amazed that issue was still open after 11 years.

ssciolla:
Thanks, @jonespm! I fixed a spacing issue I saw, and I tweaked the language on the modified portion of the selected confirmation page for clarity. I'll merge shortly.

PR #1173, opened 2020-12-17
#877 refactoring event logging

### LINKED ISSUE #877: [Redesign assignment] Event logging

pushyamig:
Fixes #877, Refactor of Event Logging, Now only sending what changed as part of user action instead of log of changes

### DIFF
--- a/assets/src/components/AssignmentTable.js
+++ b/assets/src/components/AssignmentTable.js
@@ -137,6 +137,7 @@ function AssignmentTable (props) {
   const [previousWeek, setPreviousWeek] = useState()
 
   const weeksPresent = useRef([])
+  const scrollOnce = useRef(false)
 
   const tableRef = useRef(null)
   const previousWeekRow = useRef(null)
@@ -178,12 +179,13 @@ function AssignmentTable (props) {
 
   // this effect scrolls to current week of assignments if it exists
   useEffect(() => {
-    if (previousWeekRow.current) {
+    if (!scrollOnce.current && previousWeekRow.current) {
       const tableHeaderOffset = 35 // And the universe said 'Let the offset be 35'
       tableRef.current.scrollTo({
         top: previousWeekRow.current.offsetTop - tableHeaderOffset,
         behavior: 'smooth'
       })
+      scrollOnce.current = true
     }
   }, [previousWeekRow.current, filteredAssignments])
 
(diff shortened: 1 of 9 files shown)


assets/src/containers/AssignmentPlanningV2.js
```diff
@@ -122,21 +124,31 @@ function AssignmentPlanningV2 (props) {
     settingChanged
   })
 
-  const handleAssignmentGoalGrade = (key, assignmentGoalGrade) => {
+  const handleAssignmentGoalGrade = (key, assignmentGoalGrade, prevGoalGrade) => {
+    const v = {
+      assignmentId: key,
+      assignGoalGrade: assignmentGoalGrade,
+      assignPrevGoalGrade: roundToXDecimals(prevGoalGrade, 1)
```
chrisrrowland, line 131: Why do you round the previous goal grade here?

pushyamig: otherwise it is logging it as `41.786` where are in the UI it is shown as 42

pushyamig: I just felt I will log what is shown to the user. The Calculation the way it is done it is holding up to 3 digit decimal places and user setting also stored like that.

pushyamig: So event logs are captured separately from user setting, I felt I can log what they see

chrisrrowland: I gotcha, I thought the goals were rounded before being saved, rather than before being displayed.

assets/src/components/AssignmentTable.js
```diff
@@ -137,6 +137,7 @@ function AssignmentTable (props) {
   const [previousWeek, setPreviousWeek] = useState()
 
   const weeksPresent = useRef([])
+  const scrollOnce = useRef(false)
```
chrisrrowland, line 140: Was thinking about this and it may make more sense to call it shouldScrollToCurrentWeek or something so its more clear what the use is rather than the function.

That way if we decided some action should warrant scrolling more than once per page load we can turn it back on and it makes more sense than setting scrollOnce to false again.

chrisrrowland (approved):
I've tested several scenarios and I think this looks like a nice improvement on the information that is captured by isolating individual events into each log, which I presume will be easier to consume for research purposes.

I've perused the code on a limited basis and I don't see have any concerns, though I don't have enough time to fully dig into all of the changes.

This seems like an important thing to get in for some of the other issues to utilize.

PR #1175, opened 2020-12-17
Update apv2 image and description

### LINKED ISSUE #1121: Create a different icon and description for each assignment view

Each assignment view (the v1 and the current) both have the same icons and description if enabled. We should add some additional text/image to differentiate them so users know they're two different tools. Also in the dropdown menu they should have a different name.

Create a new image and add a new description. Not sure if we should just leave it as Assignment Planning (V1) and Assignment Planning as the two names or have something more creative?

- [x] Provide a new icon for the new assignment planning
- [ ] Rename old assignment (Assignment Planning V1)
- [x] Provide a better description for the new assignment planning

Ideally we'll remove Assignment Planning (V1) but we're not ready for this yet.

chrisrrowland:
Fixes #1121 

Added new image for assignment planning redesign and updated text.


assets/src/routes/routes.js
```diff
@@ -27,8 +27,8 @@ const routes = (courseId, activeViews, activeOnly = true) => {
       path: `/courses/${courseId}/assignments`,
       title: 'Assignment Planning',
       icon: Grade,
-      description: 'See what assignments have the greatest impact on your grade.',
-      image: '/static/images/assignments_planning_icon.png',
+      description: 'See what assignments have the greatest impact on your grade and set grade goals.',
+      image: '/static/images/assignment_planning_icon.png',
```
ssciolla, line 31: Nit-pick: the naming could be clearer here. The single character difference between the other is easy to miss. We could say something like `assignment_planning_goals.png` to avoid the `v1`, `v2, `old`, `new` language.

ssciolla (approved):
Looks good to me! I'll leave it up to you whether you want to tweak the file name(s).

assets/src/routes/routes.js
chrisrrowland: I figured the original one was going away so it wouldn't be confusing for long :-p

PR #1176, opened 2020-12-17
Adding adding survey button to top menu

### LINKED ISSUE #1107: Add pop-up to gather student feedback when using MyLA
For research purposes, add a pop-up window to MyLA to collect feedback from students about their experience using MyLA. This feature needs to be configurable so that it can be turned on and off for a MyLA instance.  It does not have to be configurable at the course level.

TBD: The details of when to pop up the question, what question(s) to use, and how often to show the question are still being determined. 

We could consider using Qualtrics surveys and using embed codes.

jonespm:
* Adding constance for dynamic config
* Conditinally showing survey link based on config

Fixes #1107

### DIFF
--- a/assets/src/components/SurveyModal.js
+++ b/assets/src/components/SurveyModal.js
@@ -56,19 +56,13 @@ export default function SurveyModal(props) {
   const [modalStyle] = React.useState(getModalStyle);
   const [open, setOpen] = React.useState(false);
 
-  const handleOpen = () => {
-    setOpen(true);
-  };
-
-  const handleClose = () => {
-    setOpen(false);
-  };
+  const toggleOpen = () => setOpen(!open)
 
   const body = (
     <div style={modalStyle} className={classes.paper}>
       <DialogTitle disableTypography className={classes.dialogTitle}>
         <h4 id="survey-modal-title">Take the My Learning Analytics Survey</h4>
-        <IconButton onClick={handleClose}>
+        <IconButton onClick={toggleOpen}>
             <CloseIcon />
         </IconButton>
       </DialogTitle>
@@ -83,10 +77,10 @@ export default function SurveyModal(props) {
 
   return (
     <div>
-      <Button variant="contained" color="secondary" onClick={handleOpen}>Take Survey</Button>
+      <Button variant="contained" color="secondary" onClick={toggleOpen}>Take Survey</Button>
       <Modal
         open={open}
-        onClose={handleClose}
+        onClose={toggleOpen}
         aria-labelledby="survey-modal-title"
         aria-describedby="survey-modal-description"
       >
--- a/assets/src/components/SurveyModal.js
+++ b/assets/src/components/SurveyModal.js
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, { useState } from 'react'
 import { makeStyles } from '@material-ui/core/styles'
 import Button from '@material-ui/core/Button'
 import DialogTitle from '@material-ui/core/DialogTitle'
@@ -53,8 +53,8 @@ const useStyles = makeStyles((theme) => ({
 export default function SurveyModal (props) {
   const classes = useStyles()
   // getModalStyle is not a pure function, we roll the style only on the first render
-  const [modalStyle] = React.useState(getModalStyle)
-  const [open, setOpen] = React.useState(false)
+  const [modalStyle] = useState(getModalStyle)
+  const [open, setOpen] = useState(false)
 
   const toggleOpen = () => setOpen(!open)
 
--- a/assets/src/components/SurveyModal.js
+++ b/assets/src/components/SurveyModal.js
@@ -6,18 +6,12 @@ import IconButton from '@material-ui/core/IconButton'
 import Modal from '@material-ui/core/Modal'
 import CloseIcon from '@material-ui/icons/Close'
 
-const top = 10
-const left = 50
-
-function getModalStyle () {
-  return {
-    top: `${top}%`,
-    left: `${left}%`,
-    transform: `translate(-${left}%)`
-  }
-}
-
 const useStyles = makeStyles((theme) => ({
+  modal: {
+    top: '10%',
+    left: '50%',
+    transform: 'translate(-50%)'
+  },
   paper: {
     position: 'absolute',
     width: 400,
@@ -52,14 +46,12 @@ const useStyles = makeStyles((theme) => ({
 
 export default function SurveyModal (props) {
   const classes = useStyles()
-  // getModalStyle is not a pure function, we roll the style only on the first render
-  const [modalStyle] = useState(getModalStyle)
   const [open, setOpen] = useState(false)
 
   const toggleOpen = () => setOpen(!open)
 
   const body = (
-    <div style={modalStyle} className={classes.paper}>
+    <div className={`${classes.paper} ${classes.modal}`}>
       <DialogTitle disableTypography className={classes.dialogTitle}>
         <h4 id='survey-modal-title'>Take the My Learning Analytics Survey</h4>
         <IconButton onClick={toggleOpen}>
--- a/assets/src/components/SurveyModal.js
+++ b/assets/src/components/SurveyModal.js
@@ -67,11 +67,11 @@ export default function SurveyModal (props) {
         </IconButton>
       </DialogTitle>
 
-      <p id='survey-modal-description'>
+      <div id='survey-modal-description'>
         <div className={classes.iframeContainer}>
           <iframe className={classes.iframe} src={props.surveyURL} height='600px' width='400px' />
         </div>
-      </p>
+      </div>
     </div>
   )
 
--- a/assets/src/components/SurveyModal.js
+++ b/assets/src/components/SurveyModal.js
@@ -1,0 +1,97 @@
+import React from 'react';
+import { makeStyles } from '@material-ui/core/styles';
+import Button from '@material-ui/core/Button';
+import DialogTitle from '@material-ui/core/DialogTitle';
+import IconButton from '@material-ui/core/IconButton';
+import Modal from '@material-ui/core/Modal';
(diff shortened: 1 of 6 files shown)


assets/src/components/SurveyModal.js
```diff
@@ -0,0 +1,97 @@
...
+      </p>
+    </div>
+  );
+
+  return (
+    <div>
+      <Button variant="contained" color="secondary" onClick={handleOpen}>Take Survey</Button>
+      <Modal
+        open={open}
+        onClose={handleClose}
```
pushyamig, line 89: `handleClose` and `handleOpen` function can be inlined as this `onClose={() => setOpen(true)}` and `onClose={() => setOpen(false)}`. This is still WIP just noticed it

assets/src/components/SurveyModal.js
```diff
@@ -0,0 +1,97 @@
...
+    width: '100%',
+    height: '100%'
+  }
+}));
+
+export default function SurveyModal(props) {
+  const classes = useStyles();
+  // getModalStyle is not a pure function, we roll the style only on the first render
+  const [modalStyle] = React.useState(getModalStyle);
+  const [open, setOpen] = React.useState(false);
```
pushyamig, line 57: Minor: You can import the `useState` and can simply set it as `const [open, setOpen] = useState(false);

assets/src/components/SurveyModal.js
jonespm: Thanks, I based this example from the code on https://material-ui.com/components/modal/

Maybe they expected more to possibly be done with the open and close methods. I can update this.

jonespm:
TODO (From meeting)
* Make survey button a little more noticable (Just change the text color? 
* Explore using parameters to capture more about the survey taker, course, etc. https://www.qualtrics.com/support/survey-platform/survey-module/survey-flow/standard-elements/passing-information-through-query-strings/

assets/src/components/SurveyModal.js
```diff
@@ -0,0 +1,97 @@
...
+    right: 0,
+    width: '100%',
+    height: '100%'
+  }
+}));
+
+export default function SurveyModal(props) {
+  const classes = useStyles();
+  // getModalStyle is not a pure function, we roll the style only on the first render
+  const [modalStyle] = React.useState(getModalStyle);
```
ssciolla, line 56: So you need to run this on first render because of the iframe and absolute/relative positioning, correct? I'm still not super comfortable with this pattern, so I'm making sure I understand.

Is there anything stopping you from just using `const modalStyle = getModalStyle()`? That worked for me. If we can avoid the state variable (it only gets set once), that seems ideal.

assets/src/components/SurveyModal.js
```diff
@@ -0,0 +1,97 @@
...
+  const body = (
+    <div style={modalStyle} className={classes.paper}>
+      <DialogTitle disableTypography className={classes.dialogTitle}>
+        <h4 id="survey-modal-title">Take the My Learning Analytics Survey</h4>
+        <IconButton onClick={handleClose}>
+            <CloseIcon />
+        </IconButton>
+      </DialogTitle>
+
+      <p id="survey-modal-description">
```
ssciolla, line 76: The `p` tag here is throwing a warning because of the `div` descendant. I was able to change it to `div` without any issues.

assets/src/components/SurveyModal.js
```diff
@@ -0,0 +1,97 @@
...
+          <iframe className={classes.iframe} src={props.surveyURL} height="600px" width="400px"></iframe>
+        </div>
+      </p>
+    </div>
+  );
+
+  return (
+    <div>
+      <Button variant="contained" color="secondary" onClick={handleOpen}>Take Survey</Button>
+      <Modal
```
ssciolla, line 87: I'm guessing you have a reason (the iframe?) for using `Modal` specifically and not one of the more specific components, like `Dialog`? Just curious as to some of these implementation details. I tried replacing `Modal` with `Dialog` and thing started breaking.

ssciolla (commented):
Nice work, @jonespm, here are some comments. Some are more important than others (since they may help performance or resolve warnings), many are just style or syntax related improvements. Others are questions because some patterns are new to me. Let me know when you want me to give it a final review.

assets/src/components/SurveyModal.js
ssciolla: Actually, I investigated further, and I'm not seeing (but could still easily be missing something) why we can't just create a new class to use with `useStyles`. I added the following object to the style array, and changed that opening body `div` tag to look like: `<div className={`${classes.paper} ${classes.modal}`}>`. That seems to not break anything?

`  modal: {
    top: '10%',
    left: '50%',
    transform: 'translate(-50%)'
  }

assets/src/components/SurveyModal.js
jonespm: Yeah, this works, thanks.

dashboard/settings.py
```diff
@@ -484,6 +486,12 @@
         'django.contrib.auth.backends.ModelBackend',
     )
 
+CONSTANCE_BACKEND = 'constance.backends.database.DatabaseBackend'
+
+CONSTANCE_CONFIG = {
+    'SURVEY_URL': ('', 'URL To qualtrics survey', str),
+}
```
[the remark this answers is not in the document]
jonespm: Yeah, I wanted to get this in first then we could figure out the other configs. I could have made adding CONSTANCE a separate issue but it's really simple and not really worthwhile without adding *something*

assets/src/containers/DashboardAppBar.js
```diff
@@ -41,6 +43,7 @@ function DashboardAppBar (props) {
   const {
     classes,
     onMenuBarClick,
+    onSurveyButtonClick,
```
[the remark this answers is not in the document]
jonespm: Yeah, must have been leftover from some earlier idea

assets/src/components/SurveyModal.js
```diff
@@ -0,0 +1,97 @@
+import React from 'react';
+import { makeStyles } from '@material-ui/core/styles';
+import Button from '@material-ui/core/Button';
+import DialogTitle from '@material-ui/core/DialogTitle';
+import IconButton from '@material-ui/core/IconButton';
+import Modal from '@material-ui/core/Modal';
+import CloseIcon from '@material-ui/icons/Close';
```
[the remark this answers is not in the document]
jonespm: No I've never run this. We might want to look into setting it up as a Github action check if we want to block commits that break standard. Looks like this one? https://github.com/marketplace/actions/standardjs-lint-checks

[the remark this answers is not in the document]
ssciolla: Sure, I can provide tips on setting it up for VSCode, or you can run via `node`.

GitHub action seems like a good idea.

dashboard/settings.py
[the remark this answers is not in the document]
jonespm: I created #1200 to capture this request for a future release.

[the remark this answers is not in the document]
ssciolla: Cool, to be clear, I'm not objecting to the shoehorning of `constance` here, I just wanted to make sure we document what could be done in the future with this. It could be related to some of the other configuration issues we have open.

assets/src/components/SurveyModal.js
jonespm: I don't think I tried it with Dialog, I don't 100% remember why. I feel like I might have read the description and looked at the examples of Dialog and it didn't seem to fit what this was doing.

assets/src/components/SurveyModal.js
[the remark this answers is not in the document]
jonespm: Great, well I got standard working locally and fixed all of these. If we ever get unit tests I'd include this in actions but we could at probably still create an issue to set it up to run the frontend tests and this check.

assets/src/components/SurveyModal.js
jonespm: Sure, I'll update this to be consistent, though I personally **slightly** prefer the more explicit syntax of this here.

assets/src/containers/DashboardAppBar.js
```diff
@@ -68,6 +71,11 @@ function DashboardAppBar (props) {
             {courseName}
           </Link>
           <div className={classes.grow} />
+          {
+            surveyURL  
+            ? <SurveyModal surveyURL={surveyURL}/>
+            : null
```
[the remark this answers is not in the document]
jonespm: Okay, thanks! Updated.

assets/src/components/SurveyModal.js
jonespm: I updated, but where did you see this warning?

ssciolla: Web Console

assets/src/components/SurveyModal.js
jonespm: Sure I think that's fine. I think I had it like this because I was playing around a lot with the style positioning here and the variables to get it positioned well for mobile and desktop. But just hardcoding it would work.

Thanks for the review!

assets/src/components/SurveyModal.js
jonespm: Oh okay, I didn't look there. This is fixed though. I don't see any issues other than the CSP one about the iframe. We might have to add this https://umich.qualtrics.com to the config for production.

jonespm:
Thanks for the reviews, I've incorporated all the comments. Much of this was new to me too. I still need to look into the Qualtrics issue and pick a better font color to make the button more visible.

pushyamig:
This Looks Good. I am not sure why this is still a Draft but if there are no more changes this is good to go

jonespm:
Sorry, I clicked ready for review. Wasn't sure why it was in draft anymore either.

jonespm:
There were new issues #1215 #1216 and #1217 created for follow-up features.

pushyamig (approved):

PR #1178, opened 2020-12-17
After recent change, if DEBUGPY_ENABLE not set the server still uses …

### LINKED ISSUE #1177: After recent change, if DEBUGPY_ENABLE not set the server still uses debug mode

If DEBUGPY_ENABLE is not set in the config then system starts up in debug mode. It should start up in production mode.

Relates to #1128, it seems like the check for whether or not the variable is set isn't working correctly.

### Steps to Reproduce :
1. Don't define `DEBUGPY_ENABLE` anywhere at all
2. Start up MyLA
3. You should see **"Starting Gunicorn for production"** in the logs
4. Then in the environment set `"DEBUGPY_ENABLE": false,` restart and you should see in the logs the same message **"Starting Gunicorn for production"**
5. Finally set `"DEBUGPY_ENABLE": true,` restart you should see **"Starting Gunicorn for DEBUGPY debugging"** in the logs

jonespm:
…debug mode

Fixes #1177

### DIFF
--- a/start.sh
+++ b/start.sh
@@ -47,7 +47,6 @@ else
     DOMAIN=$(echo "${ENV_JSON}" | jq -r -c "${DOMAIN_JQ} | values")
 fi
 
-echo "$DEBUGPY_ENABLE"
 echo "Waiting for DB"
 while ! nc -z "${MYSQL_HOST}" "${MYSQL_PORT}"; do   
   sleep 1 # wait 1 second before check again
@@ -74,8 +73,8 @@ fi
 
 # If these values aren't set or they're set to false
 # This syntax substitutes False if null or unset
-if [ "${IS_CRON_POD:-"false",,}" == "false" ]; then
-    if [ "${DEBUGPY_ENABLE:-"false",,}" == "false" ]; then
+if [ "${IS_CRON_POD:-"false"}" == "false" ]; then
+    if [ "${DEBUGPY_ENABLE:-"false"}" == "false" ]; then
         echo "Starting Gunicorn for production"
     else
         echo "Starting Gunicorn for DEBUGPY debugging"


start.sh
```diff
@@ -74,8 +73,8 @@ fi
 
 # If these values aren't set or they're set to false
 # This syntax substitutes False if null or unset
-if [ "${IS_CRON_POD:-"false",,}" == "false" ]; then
-    if [ "${DEBUGPY_ENABLE:-"false",,}" == "false" ]; then
```
pushyamig, line 78: What are those extra commas mean previously?

pushyamig (approved):

start.sh
jonespm: I'm really not sure. I added them back last summer on 6f77ce7b93198e6aa53bd6112e9445a8c8dad8bd but I was trying to research them and couldn't figure out why I'd have done that.

If it's not set it sets the value to `"false",,` which doesn't compare. 

I [looked at this chart and this thread](https://stackoverflow.com/a/16753536/3708872) and it seems like either just having `-"false"` alone would do what I want which is what this code was like before this change.

So either I introduced a bug or there was some other reason for this that I'm can't find yet.

jonespm: I tested this this time with all 3 cases and put that in the issue for a test plan.
Testing set to false (is false)
Testing set to true (is true)
Testing not set at all (is false)

There is a problem if it's set to anything other than false/true (like no). It the comparison fails and it runs the cron pod else case.

I don't think this is a big deal as this is pretty explicit in the settings and true/false are non-string values in JSON.

pushyamig: ok, I feel something is set other than true or false are don't have that option enabled at all the default should be Prod mode. But I am ok for now since i do see this property `"PTVSD_ENABLE": false,` in prod `env.json`

jonespm: Yeah, we could flip the logic here, but we do have it documented to false by default or unset so if someone does leave it with that configuration this does work. I feel like this should be completely redone in the future when we remove the cron stuff too as I feel like the internal cron didn't do much for us other than keep a log.

Issue #1179, opened 2020-12-17
Change the title of the goal box from “My Minimum Goal” title to “My Minimum Goal (%)”


jennlove-um:
To make it clearer to students that the minimum course goal should be set as a percentage, change the wording of the title for the goal box from  “My Minimum Goal” title to “My Minimum Goal (%)”

![image](https://user-images.githubusercontent.com/39634416/102540313-d9417480-407c-11eb-9117-f07a8f8b6b73.png)

The text inside the box does not need to be changed.

jennlove-um:
@ssciolla or @chrisrrowland Would you have time to do this?

ssciolla:
Yes, @jennlove-um, you can assign me.

jennlove-um:
Testing passes: The label now reads "My Minimum Goal (%)"

PR #1180, opened 2020-12-18
Change "My Minimum Goal" to "My Minimum Goal (%)" (#1179)

ssciolla:
This PR modifies the header that accompanies the course goal input on the Assignment Planning view (redesign). It also moves the header out to the view container, to be consistent with the other headers on the page and give a better sense of the page flow from the component. The PR aims to resolve issue #1179.


pushyamig (approved):
Looks Good

PR #1181, opened 2020-12-18
Ap2 assgoal debounce

### LINKED ISSUE #1106: Setting assignment goal blocks UI while saving
In assignment planning redesign, if you change the target grade for an individual assignment the UI locks while applying the changes.

The length of the freeze seems related to how many changes you apply.  Ex. clicking the up arrow 20 times freezes the UI much longer than clicking the up arrow one time.

chrisrrowland:
Fixes #1106 

Another attempt to do this and have it jive with event logging.


assets/src/components/AssignmentTable.js
```diff
@@ -137,7 +137,7 @@ function AssignmentTable (props) {
   const [previousWeek, setPreviousWeek] = useState()
 
   const weeksPresent = useRef([])
-  const shouldScrollToCurrentWeek = useRef(false)
+  const shouldScrollToCurrentWeek = useRef(true)
```
chrisrrowland, line 140: Given the name I thought this made more sense to use this way.
Not directly related to this functionality.

pushyamig:
I will review this now

pushyamig:
I see that assignment goal grade is setting 3 decimal digits? is that intentional, since I did not notice that before

pushyamig:
Assignment Lock state needs to be preserved when user changes assignment goal, currently it is not.

https://user-images.githubusercontent.com/8579775/102639052-773d4980-4126-11eb-92b9-aa30f3fbac5d.mov

pushyamig (changes requested):
https://github.com/tl-its-umich-edu/my-learning-analytics/pull/1181#issuecomment-748197941

justin0022:
@chrisrrowland should this PR be closed? I see there's `ap2_assignment_goal_debounce` and `ap2_assgoal_debounce` (this PR is from the latter branch).

Issue #800, opened 2019-11-19
Data in the hover over on assignment bar elements in the list of assignments can't be navigated to using keyboard


ssciolla:
## Expected behavior (A clear and concise description of what you expected to happen) : 

Users should be able to navigate to or access all content in the Assignment Planning view by using the keyboard.

## Describe the bug (Tell us what happens instead of the expected behavior) :

From the 10/28/19 accessibility review:

"Progress toward Final Grade table elements are not keyboard operable. ... Some users will not be able to access the metadata available on hover (name, due, grade, points possible, etc)."

For the new Assignment Planning, there is no mouse-over in the progress bar.  Instead, these metadata are available in the individual assignment bars next to each assignment in the list of assignments.  The functional need is the same. Tabbing should allow users to access the information that is in the mouse-over data.

## Additional context (Add any other context about the problem here) :

This may be addressed as part of work on issue #691 and PR #702. This issue may also be mitigated or resolved by implementing narrative alternatives to charts, such as that discussed in issue #749.

jennlove-um:
Hold off on this one and revisit after the Assignment Planning redesign, since it may not be an issue after that.  Assignments & mouse-over won't exist in the new view.

jennlove-um:
Test keyboard accessibility for Assignment Planning redesign.

jennlove-um:
Testing indicates that while users can tab to the individual assignment bars, they cannot view the data that is available in the hover-over for the assignment.

jennlove-um:
Testing passes: Users can tab to the information bar for individual assignments and see the pop up. Tabbing off the information bar successfully closes the pop up.

PR #1182, opened 2020-12-18
Show PopUpMessage with assignment-specific progress bars when navigated to by keyboard (#800)

ssciolla:
This PR presents a functional, if imperfect, solution that allows users to view the `PopUpMessage` content when an assignment-specific progress bar is navigated to using a keyboard. It is "imperfect" in the sense that there is a visual bug where if the user scrolls or uses the up or down arrow key when a bar is focused and the pop-up is visible, the pop-up will keep its position in the entire window and move away from the bar. @jennlove-um thinks this partial solution may be acceptable for now, and we can continue to refine it going forward. These changes include a `usePopoverEl` custom hook to reduce some duplication between the `onFocus` and `onBlur` handlers and the mouse-over handlers. The PR aims to resolve issue #800.

### DIFF
--- a/assets/src/hooks/usePopoverEl.js
+++ b/assets/src/hooks/usePopoverEl.js
@@ -1,15 +1,14 @@
 import { useState } from 'react'
 
 function usePopoverEl () {
-  const emptyPopover = { popoverId: null, anchorEl: null }
-  const [popoverEl, setPopoverEl] = useState(emptyPopover)
+  const emptyPopoverEl = { popoverId: null, anchorEl: null }
+  const [popoverEl, setPopoverEl] = useState(emptyPopoverEl)
 
-  const setPopover = (key, el) => {
-    setPopoverEl({ popoverId: key, anchorEl: el })
-  }
-  const clearPopover = () => setPopoverEl(emptyPopover)
+  const setNewPopoverEl = (key, el) => setPopoverEl({ popoverId: key, anchorEl: el })
 
-  return [popoverEl, setPopover, clearPopover]
+  const clearPopoverEl = () => setPopoverEl(emptyPopoverEl)
+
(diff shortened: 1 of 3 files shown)


ssciolla:
I'm gonna tweak some of these variables names in a second, but I'm not planning to change any logic.

ssciolla:
Also, if you have any ideas on resolving the scrolling bug, let me know. I played with replacing `Popover` with `Popper`, but that component had its own set of bugs and styling to deal with.

assets/src/hooks/usePopoverEl.js
```diff
@@ -0,0 +1,14 @@
+import { useState } from 'react'
+
+function usePopoverEl () {
```
ssciolla, line 3: It's probably debatable whether the custom hook is necessary, but the ability to just use `clearPopoverEl` in several places made it seem worth it.

ssciolla: Also, I don't love `PopoverEl`, since it's really an object that also includes the key, but I didn't want to disrupt the existing language too much.

justin0022: I like that you extracted it out into a custom hook. Makes it easier to understand what's going on

ssciolla:
Something weird about GitHub is not letting me assign you as a reviewer, but @chrisrrowland, if you have the time, it'd be nice to get your feedback. If you're focused on other stuff, no worries.

pushyamig:
@ssciolla This looks Ok. With what we have discussed over before the break this is ok to merge for this release and continue to work on the refinement later

pushyamig (approved):

ssciolla:
@pushyamig, thanks for the review. I'll merge so we can get this into QA soon.

PR #1189, opened 2021-01-05
Adding a separate Dockerfile for local openshift registry

### LINKED ISSUE #1183: Create a separate Dockerfile for Openshift to avoid Docker rate limits

Because of recent Dockerrate limits, the easiest way we've found to work around this for a multi-stage build is to have a separate Dockerfile with the images in the Openshift registry hardcoded. 

In a future version of Openshift, these are possible to replace via environment variables, however the current Openshift 3 doesn't support this so will need a separate file.

This should be 99% identical to the existing Dockerfile but for the FROM it will have
`FROM docker-registry.default.svc:5000/openshift/node:14-slim AS node-webpack`

and 
`FROM docker-registry.default.svc:5000/openshift/python:3.8-slim AS app`

I believe the default is `docker.io/<dockerimagename>:version`

We should add a comment that if either file is changed the should both be updated.

jonespm:
Fixes #1183

In openshift the YAML needs to be altered to be
```
  strategy:
    dockerStrategy:
      dockerfilePath: Dockerfile.openshift
    type: Docker
```

This file is not used otherwise.


jonespm:
I tested that this builds and verified in the logs that the correct images was pulling.

In Openshift 3 this is the only way it seems with a multi-stage build. If you specify a BUILD FROM it only overrides the last FROM, you can't override all of them.

In Openshift 4 we could pass an environment variable which can override the FROM, but that isn't possible with the version of Docker in OS3.

pushyamig:
@jonespm I see that you have put this build on Beta instance, I can confirm build is working. I don't feel this needs any more review.

pushyamig (approved):

jonespm:
Thanks! This won't do anything, we just have to remember to keep it up-to-date with Dockerfile. I don't see any other way to do this unfortunately with Openshift 3. Though this file doesn't change very much, and as long as we're all aware we'll be fine till we can upgrade! 

Also, for some reason Github isn't letting me add @ssciolla to the reviewers. I think Justin noticed this for Chris earlier. On some issues it's fine, but not others. Hmm!

ssciolla (approved):
This looks okay to me! Couple minor comments. You could add a paragraph to the `README.md` with some info, or you could add another `README.md` in a `dockerfiles` directory, if you went that route.

Dockerfile.openshift
```diff
@@ -0,0 +1,66 @@
+# build react components for production mode
+FROM docker-registry.default.svc:5000/openshift/node:14-slim AS node-webpack
+WORKDIR /usr/src/app
```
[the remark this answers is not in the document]
jonespm: Yeah we could do that here. I'm good with getting more files out of the root.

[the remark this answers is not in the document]
jonespm: I updated this PR with this change.

dockerfiles/README.md
```diff
@@ -0,0 +1,16 @@
+The Dockerfile(s) for the build are in this directory
```
ssciolla, line 1:
```suggestion
The Dockerfile(s) for the build are in this directory.
```

dockerfiles/README.md
```diff
@@ -0,0 +1,16 @@
...
+
+* Dockerfile is the file used for most builds including localhost
```
ssciolla, line 3:
```suggestion
* `Dockerfile` is the file used for most builds including when using localhost.
```

dockerfiles/README.md
```diff
@@ -0,0 +1,16 @@
...
+
+* Dockerfile.openshift is a custom file with the FROM entries modified to point
+to local images in the Openshift repository.
```
ssciolla, line 6:
```suggestion
* Dockerfile.openshift is a custom file with the FROM entries modified to point
  to local images in the Openshift repository.
  (Note: we may be able to combine this with `Dockerfile` again when we have access Openshift 4).
```

dockerfiles/README.md
```diff
@@ -0,0 +1,16 @@
...
+
+If you're using Openshift you'll want to put both python and node into 
+into the openshift image stream then configure your build YAML like:
```
ssciolla, line 9:
```suggestion
into the openshift image stream then configure your build YAML to use the following `strategy` block:
```

dockerfiles/README.md
```diff
@@ -0,0 +1,16 @@
...
+
+``
```
ssciolla, line 11: You're missing a a ` here. I couldn't get the suggestion to work.

ssciolla (approved):
Restructuring looks fine. I added some suggestions to tidy up the text in `dockerfiles/README.md`.

Dockerfile.openshift
```diff
@@ -0,0 +1,66 @@
...
+# This DJANGO_SECRET_KEY is set here just so collectstatic runs with an empty key. It can be set to anything
+RUN echo yes | DJANGO_SECRET_KEY="collectstatic" python manage.py collectstatic --verbosity 0
+
+# Sets the local timezone of the docker image
+ARG TZ
+ENV TZ ${TZ:-America/Detroit}
+RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone
+
+CMD ["/code/start.sh"]
+# done!
```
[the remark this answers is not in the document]
jonespm: It doesn't look like the original Dockerfile this either. For the most part the only annoyance here is when some people add a newline and others remove it or don't have a newline. It looks like *most* of the files in this project have a newline but these don't.

```
./cypress/fixtures/example.json
./webpack-stats.json
./dashboard/migrations/0001_initial.py
./dashboard/management/commands/course.py
./dashboard/management/commands/site.py
./dashboard/graphql/mutation.py
./dashboard/graphql/loaders.py
./dashboard/graphql/view.py
./dashboard/graphql/schema.py
./dashboard/templates/frontend/index.html
./mypy.ini
./docker-compose-prod.yml
./dockerfiles/Dockerfile
./dockerfiles/Dockerfile.openshift
./demo_init.sh.sample
```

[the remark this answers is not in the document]
jonespm: I'll fix all of the ones I'm changing though!

dockerfiles/README.md
jonespm: Thanks, fixed.

jonespm:
Thanks, committed all suggestions.

dockerfiles/README.md
```diff
@@ -0,0 +1,17 @@
+The Dockerfile(s) for the build are in this directory.
+
+* `Dockerfile` is the file used for most builds including when using localhost.
+
+* `Dockerfile.openshift` is a custom file with the FROM entries modified to point
+  to local images in the Openshift repository.
+  (Note: we may be able to combine this with `Dockerfile` again when we have access Openshift 4).
```
ssciolla, line 7:
```suggestion
  (Note: we may be able to combine this with `Dockerfile` again when we have access to Openshift 4).
```

jonespm:
I'm going to test this still builds as the README suggests then I'll merge it.

dockerfiles/README.md
jonespm: Sorry, merged before I saw this one. Documentation is harder to write than code. ;)

jonespm: Hmm, is the period supposed to be in the parenthesis? This looks correct. . . But
["When a complete, independent sentence is entirely enclosed by parentheses, the period goes inside the closing parenthesis."](https://www.grammarly.com/blog/period/)

Didn't know about that edge case!

ssciolla: Yeah, I was fixing my own mistakes here. It makes sense to me from a grammar perspective. A period ends an independent clause, which in this case is entirely in the parentheses. Not a super important thing, but something I would definitely have wanted corrected when I was working as a journalist.

PR #1191, opened 2021-01-07
Add help url for Assignment Planning v2

### LINKED ISSUE #1166: Add Assignment Planning V2 help link
With the assignment planning redesign set to replace the original version, the new version needs a help url.
This will cause the info link to appear on the card.

The help url should be set in settings.py

`URL_VIEW_ASSIGNMENT_PLANNING = ENV.get("URL_VIEW_ASSIGNMENT_PLANNING", None)`

similarly as it is to the  original version

`URL_VIEW_ASSIGNMENT_PLANNING_V1 = ENV.get("URL_VIEW_ASSIGNMENT_PLANNING_V1", "https://its.umich.edu/academics-research/teaching-learning/my-learning-analytics/support/assignment-planning")`

Given that the new help content would likely contain screenshots of the content, it may be something that has to be added after the initial release.

chrisrrowland:
Close #1166


zqian (approved):
verified the help link works.

PR #1192, opened 2021-01-07
Missing dockerfile for webpack_watcher

### LINKED ISSUE #1183: Create a separate Dockerfile for Openshift to avoid Docker rate limits

jonespm:
Fixes #1183

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -43,6 +43,7 @@ services:
   webpack_watcher:
     build:
       context: .
+      dockerfile: dockerfiles/Dockerfile
       target: node-webpack
     command: sh -c "tar zxvf node_modules/all_symlinks.tgz && npm run watch"
     volumes:


pushyamig:
This looks Good

PR #1202, opened 2021-01-13
issue_1201 return term end date directly if there is no term start date

zqian:


dashboard/models.py
```diff
@@ -38,7 +38,7 @@ def __str__(self):
     # # Replace the year in the end date with start date (Hack to get around far out years)
     # # This should be replaced in the future via an API call so the terms have correct end years, or Canvas data adjusted
     def get_correct_date_end(self):
-        if (self.date_end.year - self.date_start.year) > 1:
+        if self.date_start is not None and (self.date_end.year - self.date_start.year) > 1:
```
ssciolla, line 41: I'm wondering if we should future-proof this a bit more and also check whether `self.date_end is not None`. `Course.get_course_date_range` checks for this, and we still want to do that there, but this method could also be called in other contexts. Since `date_end` is nullable, that would also break the method.

Adding type hints would also be useful.

ssciolla (commented):
Mmm, I'm sorry, when I implemented this more than a year ago, I don't think I thought this through enough (i.e. registered that `get_correct_end_date` was dependent on `Term.date_start` and that `Term.date_start` was nullable). I was less experienced back then, I guess, and this isn't something that we've had to worry about at UM.

I had one comment that I'd like to hear whether you and @jonespm think would be appropriate. Your change should fix the bug, but we should try to think through other edge cases. This is probably a decent candidate for a unit test as well (if we had a test runner set up : ( ).

dashboard/models.py
zqian: code updated. Please review.

dashboard/models.py
```diff
@@ -37,8 +37,10 @@ def __str__(self):
 
     # # Replace the year in the end date with start date (Hack to get around far out years)
     # # This should be replaced in the future via an API call so the terms have correct end years, or Canvas data adjusted
-    def get_correct_date_end(self):
-        if self.date_start is not None and (self.date_end.year - self.date_start.year) > 1:
+    def get_correct_date_end(self) -> models.DateTimeField:
```
ssciolla, line 40: I think the type hint should be `-> Optional[datetime]` or `Union[datetime, None]`. It might be possible to use the model field like you are doing, but in reality, we're going to be returning either `None` or a `datetime` object.

ssciolla (commented):
Code change looks like it will work, but I think the type hint needs to be modified; see comment.

dashboard/models.py
zqian: Thanks. Updated to Optional[datetime] now.

ssciolla (approved):
Looks okay to me. @jonespm, can you also take a look so at least you're aware of the change (and the bug I created a year ago, though the `null` term dates are not something we at UM have to deal with, I don't think)?

jonespm (approved):
Yeah, hmm, this code was originally in here so that if we had an end_date that was in a different year than the start date, it would just return the year of the start date.

I guess there are 2 possible issues.
1) The `date_start` or `date_end` is not defined. I guess in this case it's going to return None, what are the downstream effects of this. (See below, doesn't look like any)
2) The `date_start` is None but `date_end` is defined. It looks like this fix is just returning the end_date as-is then.

I think this seems fine. I only see this code used in one place and it will never return None as it's guaranteed to have a `date_end`. This feels like it all probably could be refactored to be clearer but it'll work and I can't see a better way to refactor at the moment.

```
        elif self.term is not None and self.term.date_end is not None:
            end = self.term.get_correct_date_end()
```

ssciolla:
> I think this seems fine. I only see this code used in one place and it will never return None as it's guaranteed to have a `date_end`. This feels like it all probably could be refactored to be clearer but it'll work and I can't see a better way to refactor at the moment.
> 
> ```
>         elif self.term is not None and self.term.date_end is not None:
>             end = self.term.get_correct_date_end()
> ```

1) Well, it did break for Oregon State I guess because they have terms with null dates.

2) There is a question (suggested by your code snippet) about whether to check for null in the method `Term.get_correct_date_end` or in the place where it's used `Course.get_course_date_range`. It seems more robust to me to do the checks at the `Term` level, since it's that model's data at issue and because the check at the `Course` level does not protect against other uses of the method (which is unlikely to happen here, but possible).

jonespm:
1) Yeah, right, we have "Default Term" which has both no start or end date, so that would fall into this problem if any courses had that term but no other dates defined. So this was more a problem of the start date. 
2) Yeah, I was thinking it might be better to just do all the checks and trust this, this would eliminate an elif. Could just pass a "if_none" argument to it and simplify some of the code? I wonder what the best pattern for that is. 

Like if it was
```
    def get_correct_date_end(self, if_none: datetime) -> Optional[datetime]:
```

If we actually had typehints everywhere and consistently used I'd feel safer about this code but still without it we don't know without the research.

zqian:
@jonespm and @ssciolla I guess I can merge this PR for now, and you can file another issue for the remaining problems?

zqian:
will merge it now given the approvals.

PR #1204, opened 2021-01-15
Make git check optional

andrew-gardener:
Can instead pass GIT_REPO, GIT_COMMIT, and GIT_BRANCH as environment variables instead of requiring an active git repo to pull info

### DIFF
--- a/start.sh
+++ b/start.sh
@@ -52,13 +52,14 @@ while ! nc -z "${MYSQL_HOST}" "${MYSQL_PORT}"; do
   sleep 1 # wait 1 second before check again
 done
 
-echo "Setting Git info variables"
-GIT_REPO="$(git config --local remote.origin.url)"
-export GIT_REPO
-GIT_COMMIT="$(git rev-parse HEAD)"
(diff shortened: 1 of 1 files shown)


zqian:
@andrew-gardener the PR works successfully. However, I did not find any error messages in the log file prior to the change. 

Could you please create an github issue reporting the problem you found prior to the patch, and link this PR to it?

andrew-gardener:
It may not be in the general flow of things anymore but I ran into the issue previously when deploying myla. It happened because I built the docker image in a CI using a source release download when making the image (so the image did not contain a `.git` and the git commands would throw an non blocking error on startup since the folder wasn't a git repo).

zqian:
@andrew-gardener I have created issue #1206 with those comments and associated this PR with the issue. 

For better issue tracking, we would prefer creating a git issue (with description of the bug, steps to reproduce, etc) first, and then link it to relevant PR.

zqian (approved):
Tested the PR successfully with and without .git folder.

PR #1205, opened 2021-01-15
Allow invalid/missing LTI deployment ids

andrew-gardener:
Add `LTI_CONFIG_DISABLE_DEPLOYMENT_ID_VALIDATION` (default false) which will disable deployment id validation if true. This is helpful for piloting in some environments (where a deployment id is generated for each pilot course using the tool)


dashboard/lti_new.py
```diff
@@ -35,6 +35,13 @@
 COURSE_MEMBERSHIP = 'http://purl.imsglobal.org/vocab/lis/v2/membership'
 DUMMY_CACHE = 'DummyCache'
 
+# do not require deployment ids if LTI_CONFIG_DISABLE_DEPLOYMENT_ID_VALIDATION is true
+class ExtendedDjangoMessageLaunch(DjangoMessageLaunch):
+    def validate_deployment(self):
+        if settings.LTI_CONFIG_DISABLE_DEPLOYMENT_ID_VALIDATION:
+            return self
+
+        return super(ExtendedDjangoMessageLaunch, self).validate_deployment()
```
jonespm, line 44: super() in [Python 3 should do the same thing](https://docs.python.org/3/library/functions.html#super) as this.

```suggestion
        return super().validate_deployment()
```

andrew-gardener: Fixed

jonespm (approved):

PR #1210, opened 2021-01-20
Support alternative LRS source for resource access

### LINKED ISSUE #1209: Support Additional data source for Resource Access (LRS db connection)
UBC's LRS is not setup on BigQuery. Adding (optional) support for a more generic data source connection would help us use the Resources Accessed view

Add an optional generic LRS data source (similar to DATA_WAREHOUSE)

andrew-gardener:
- Add new data source LRS (optional) which can be used instead of big query for fetching resource access info.
- Fixed `get_data_earliest_date` method to work when there are no "existing courses" (previously threw error when all `data_last_updated` where null)

Closes #1209


andrew-gardener:
Note I did not actually test the changes against BiqQuery so I'd very much appreciate testing and feedback on if the changes work properly with it!

config/env_sample.json
```diff
@@ -206,6 +206,25 @@
     "/*  DATA_WAREHOUSE_PORT=5432": "*/",
     "/*  Enable/Disable Unizin Date Warehouse specific features/data": "*/",
     "DATA_WAREHOUSE_IS_UNIZIN": true,
+    "/*  Learning Record Store configuration": "*/",
+    "/*  Uncomment these variables and fill them in if you're using a ": "*/",
+    "/*  non-bigquery LRS source. These are optional": "*/",
+    "/*  LRS database engine driver": "*/",
+    "LRS_ENGINE": "django.db.backends.postgresql",
```
jonespm, line 213: We could **probably** just make this variable `google.cloud.bigquery` rather than have the boolean "RESOURCE_USE_LRS"

dashboard/cron.py
```diff
@@ -261,39 +262,52 @@ def update_with_bq_access(self):
...
-            final_bq_query = []
+            final_query = []
             for k, query_obj in settings.RESOURCE_ACCESS_CONFIG.items():
                 # concatenate the multi-line presentation of query into one single string
                 query = " ".join(query_obj['query'])
 
                 if (data_last_updated is not None):
                     # insert the start time parameter for query
-                    query += " and event_time > @data_last_updated"
+                    query += f" {query_obj['query_data_last_updated_condition']} "
```
jonespm, line 272: Is this something not needed for Postgres or were you filling something different in here. I'd be somewhat reluctant to remove this incase someones configuration wasn't updated and maybe be something like.

```
# is_bigquery is defined somewhere earlier to tell whether the LRS is bigquery or not. 
if query_obj['query_data_last_updated_condition']:
  query += f" {query_obj['query_data_last_updated_condition']} "
elif is_bigquery:
  query += " and event_time > @data_last_updated"
```

config/env_sample.json
jonespm, line 213: Looking at this, this would also probably be better if we had this as an object similar to the settings like

```
"LRS": {
  "ENGINE": . . .
```

That probably should be a separate issue though.

andrew-gardener: I'm up for doing the change but if doing it for `LRS_*` we should also do it for `DATA_WAREHOUSE_*` and `MYSQL_*`

I can make a separate issue+PR after this is merged

dashboard/cron.py
andrew-gardener: Its because postgres uses a different param embed syntax 

postgres version looks like this `AND event_time > %(data_last_updated)s`

I can make the suggested change though. would

config/env_sample.json
andrew-gardener: See #1220

andrew-gardener: Fixed

dashboard/cron.py
```diff
@@ -235,20 +235,22 @@ def update_canvas_resource(self):
                 status += f"Row {row.id} removed as it is not available\n"
         return status
 
-    # update RESOURCE_ACCESS records from BigQuery
-    def update_with_bq_access(self):
+    # update RESOURCE_ACCESS records from BigQuery or LRS data sources
+    def update_resource_access(self):
+        is_bigquery = settings.DATABASES['LRS']['ENGINE'] == 'google.cloud.bigquery'
```
zqian, line 240: @andrew-gardener I got the follow error when the LRS->ENGINE is set to 'google.cloud.bigquery':

student_dashboard  | django.core.exceptions.ImproperlyConfigured: 'google.cloud.bigquery' isn't an available database backend.
student_dashboard  | Try using 'django.db.backends.XXX', where XXX is one of:
student_dashboard  |     'mysql', 'oracle', 'postgresql', 'sqlite3'

I suggest to use another bool setting param "LRS_BIGQUERY", outside of the LRS params. So for Unizin schools, we do not need to set LRS params at all, with LRS_BIGQUERY=true

andrew-gardener: I can instead add the LRS optionally to the list of databases (if engine != `google.cloud.bigquery`) 

Just trying to get rid of the variables as per https://github.com/tl-its-umich-edu/my-learning-analytics/pull/1210#discussion_r563967258

andrew-gardener: Also sorry about not catching the error

zqian: That should work, like in settings.py to append LRS variable only when the engine is not of BigQuery:

if LRS_ENGINE != 'google.cloud.bigquery':
    DATABASES.append(
        {
            'LRS': {
                'ENGINE': ENV.get('LRS_ENGINE', ''),
                'NAME': ENV.get('LRS_DATABASE', ''),
                'USER': ENV.get('LRS_USER', ''),
                'PASSWORD': ENV.get('LRS_PASSWORD', ''),
                'HOST': ENV.get('LRS_HOST', ''),
                'PORT': ENV.get('LRS_PORT', 5432),
                'OPTIONS': ENV.get('LRS_OPTIONS', {}),
            }
        }
    )

Could you please update this pull request, so that I can run it again?

dashboard/cron.py
```diff
@@ -261,39 +263,54 @@ def update_with_bq_access(self):
...
             for k, query_obj in settings.RESOURCE_ACCESS_CONFIG.items():
                 # concatenate the multi-line presentation of query into one single string
                 query = " ".join(query_obj['query'])
 
                 if (data_last_updated is not None):
                     # insert the start time parameter for query
-                    query += " and event_time > @data_last_updated"
+                    if query_obj['query_data_last_updated_condition']:
+                        query += f" {query_obj['query_data_last_updated_condition']} "
+                    elif is_bigquery:
```
zqian, line 275: Need to check for is_bigquery first, since query_obj['query_data_last_updated_condition'] is optional.

andrew-gardener: I'll change it to a `.get()` so it should be safe if present or not

andrew-gardener:
@zqian please check out the changes

dashboard/cron.py
zqian: That worked!

dashboard/cron.py
zqian: Your fixes worked!

zqian (approved):
I have tested the patch worked well with the following addition to env.json file. BigQuery event pull was a success, too.

    "/*  Learning Record Store configuration": "*/",
    "/*  Uncomment these variables and fill them in if you're using a ": "*/",
    "/*  non-bigquery LRS source. These are optional": "*/",
    "/*  LRS database engine driver (use `google.cloud.bigquery` for bigquery). no other LRS settings needed": "*/",
    "LRS_ENGINE": "google.cloud.bigquery",
    "/*  LRS database name": "*/",
    "LRS_DATABASE": "",
    "/*  LRS database user": "*/",
    "LRS_USER": "",
    "/*  LRS database password": "*/",
    "LRS_PASSWORD": "",
    "/*  LRS database host": "*/",
    "LRS_HOST": "",
    "/*  LRS database port": "*/",
    "LRS_PORT": 5432,
    "/*  LRS database options": "*/",
    "LRS_OPTIONS": {},

PR #1227, opened 2021-02-02
Add friendlier display name

### LINKED ISSUE #1226: Friendlier display names
Displaying the account username to logged users may not be the most friendly in all cases. In particular UBC populates usernames with a PUID which is very unfriendly 

Instead, display the first + last name (or initials for avatars). Also ensure that first and last names are always populated on each LTI launch (in case there are changes in the ISS which will eventually filter down on launch)

andrew-gardener:
Uses user's full name instead of username and initials instead of first letter of username (will fallback to username if name is blank)

Also updates first name, last name, and email on LTI launches (to sync changes over time)

Closes #1226

### DIFF
--- a/dashboard/lti_new.py
+++ b/dashboard/lti_new.py
@@ -232,6 +232,11 @@ def extract_launch_variables_for_tool_use(request, message_launch):
     # Add user to DB if not there; avoids Django redirection to login page
     try:
         user_obj = User.objects.get(username=username)
+        # update
+        user_obj.first_name = first_name
+        user_obj.last_name = last_name
+        user_obj.email = email
+        user_obj.save()
     except User.DoesNotExist:
         password = ''.join(random.sample(string.ascii_letters, settings.RANDOM_PASSWORD_DEFAULT_LENGTH))
         user_obj = User.objects.create_user(username=username, email=email, password=password, first_name=first_name,
--- a/README.md
+++ b/README.md
@@ -154,7 +154,7 @@ The application currently uses two frameworks for front-end testing: [Cypress](h
 Some front-end tests are implemented using the [Cypress framework](https://www.cypress.io/). 
 
  For running cypress tests locally, it is essential that you have Myla instance running locally. Launch Myla from the
- browser go to the admin view and add user called `donald07` with password `root`. Get the latest depersonalized datadump
+ browser go to the admin view and add user called `donald07` with password `root`, first name `donald`, and last name `07`. Get the latest depersonalized datadump
  as described from the Step 9 in [installation and setup](#installation-and-setup).
  
  Install cypress 
--- a/assets/src/__tests__/components/__snapshots__/AssignmentGradeBoxes.test.js.snap
+++ b/assets/src/__tests__/components/__snapshots__/AssignmentGradeBoxes.test.js.snap
@@ -4,11 +4,6 @@ exports[`renders correctly 1`] = `
 <div
   className="MuiGrid-root MuiGrid-item"
 >
-  <h6
-    className="MuiTypography-root MuiTypography-h6 MuiTypography-gutterBottom"
-  >
-    My Minimum Goal
-  </h6>
   <div
     className="MuiFormControl-root MuiTextField-root WithStyles(ForwardRef(TextField))-root-2 AssignmentGoalInput-goalGradeInput-1 MuiFormControl-marginNormal"
     style={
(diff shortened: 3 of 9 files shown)


andrew-gardener:
Originally forgot to run the frontend tests. 

Was successful updating the Jest tests (though I didn't touch anything that it needed to update so it might have been a bit stale)

Was not successful running the Cypress tests. The latest depersonalized data dump doesn't seem to have a matching entry in the users table for `donald07` so several tests are failing for me for that one.

pushyamig:
1. If the Full name has middle name as well then the circle looks crammed. Either Login need to include only First and last name initials
Screen Shot 2021-02-02 at 10 45 39 AM
2. if you became user from standalone or LTI tool using Django admin it is currently not displaying anything for user info
Screen Shot 2021-02-02 at 10 53 57 AM
this is not an issue when you try to became user from Canvas masquerading end. You can easily reproduce this from standalone tool by becoming the root user.

dashboard/lti_new.py
```diff
@@ -232,6 +232,11 @@ def extract_launch_variables_for_tool_use(request, message_launch):
     # Add user to DB if not there; avoids Django redirection to login page
     try:
         user_obj = User.objects.get(username=username)
+        # update
+        user_obj.first_name = first_name
+        user_obj.last_name = last_name
+        user_obj.email = email
+        user_obj.save()
```
pushyamig, line 239: What is the purpose of save this to user object since it will be populated from Cron job from respective record store. Do you see any difference in user name information from record store and LTI?

andrew-gardener: This is the `auth_user` (logged in user) info. I added this since it is currently never updated after create (and its normal to update this kind of info when LTI launching into the app).

now your absolutely right that the user's name is also in the `user` table but I figured that it would be better to focus on the `auth_user` table since its one-to-one for the current user and the `user` table is technically one-to-many (since the user table is more of an flattened `user`+`enrollment`+`overall_score` table)

andrew-gardener:
>     1. If the Full name has middle name as well then the circle looks crammed. Either Login need to include only First and last name initials
>        Screen Shot 2021-02-02 at 10 45 39 AM

Ya, I'll switch to first letter of first name and first letter of the last name (will technically be wrong for some people but I didn't realize that the avatar space only fit 2 characters).

>     2. if you became user from standalone or LTI tool using Django admin it is currently not displaying anything for user info
>        Screen Shot 2021-02-02 at 10 53 57 AM
>        this is not an issue when you try to became user from Canvas using masquerading feature. You can easily reproduce this from standalone tool by becoming the root user.

Probably the user is missing first and last name in `auth_user` table. This change does require that the user launch to create their account or that the first+last name is added for users (if added via admin console, etc). I can make some additional changes to help support that if needed.

pushyamig:
> Screen Shot 2021-02-02 at 10 45 39 AM
Ya, I'll switch to first letter of first name and first letter of the last name (will technically be wrong for some people but I didn't realize that the avatar space only fit 2 characters).

@jennlove-um @andrew-gardener made the change to have the first name and last name of the user and if the user have a middle name and more than that what is the best approach for this?

pushyamig:
Screen Shot 2021-02-02 at 10 53 57 AM
> Probably the user is missing first and last name in auth_user table. This change does require that the user launch to create their account or that the first+last name is added for users (if added via admin console, etc). I can make some additional changes to help support that if needed.

That make sense, I added the user from the Django admin console and it only ask for user name and password and not first name or last name. We could modify the Django Form when adding the user  to input username and password or we can keep it simple if we see the full name is empty we can show the 'username' in our case it is uniqname. Then the behavior won't be consistent. 

During the LTI launch or SAML the firstname and lastname is added to auth_user table. It is when manually adding the user fron django console and don't input firstname/lastname then it will display empty when become user from django console. This is affecting admins only 

@jennlove-um @jonespm Thoughts??

andrew-gardener:
The changes I made are: 
- fallback to username if full name is blank
- limit to first and last initial if there are more than 2

open to more changes based on discussion. For initials my only concern is people with multiple last names but on the flip side these people are probably used to having incorrect initials displayed (due to similar limitations) so they might not mind.

ssciolla:
> Originally forgot to run the frontend tests.
> 
> Was successful updating the Jest tests (though I didn't touch anything that it needed to update so it might have been a bit stale)
>

@andrew-gardener, thank you for updating these. At least one of the failures was on me for not updating the snapshots with earlier PRs.

andrew-gardener:
switched to first initial only

pushyamig (approved):

Issue #1168, opened 2020-12-16
Clean up the alignment of the current, max, and goal text with associated lines on the progress bar


jennlove-um:
![image](https://user-images.githubusercontent.com/39634416/102384238-78da0680-3f9a-11eb-9619-2ce5f4cc2760.png)

The text for the goal should align with the top of the goal line.
The text for the current score should align with the bottom of the current line.
The text for the max score should align with the bottom of the max line.

jennlove-um:
@ssciolla I put this in the winter release, but take a look.  Let me know if it is feasible for the winter .01 release or needs to be pushed back to the .02 release.

jennlove-um:
This passes in testing.

PR #1231, opened 2021-02-04
Refactor Line and Label components (#1168)

ssciolla:
This PR refactors the preexisting `Line` and `Label` components used in the new Assignment Planning view. The goal was to make the `Line` and `Label` positions more connected so they are easier to modify. Because the position of `Label` is or should be dependent on `Line`, I elected to make `Label` an optional child component of `Line` and consolidate the majority of the positioning code within `Line`. The PR aims to resolve the minor issue #1168.


ssciolla:
Note: I could have resolved the connected issue in a more simple way (tweaking hard-coded values in multiple places), but I wanted it to be easier to keep all this in sync. Now a couple of values in `getVerticalStyles` from `Line` can be modified to tweak the positioning.

While I've opened this up for review, I still consider this somewhat experimental, and I imagine there are still other ways this could be improved. I'm open to significant refactoring suggestions.

zqian:
@ssciolla looks like both "current" and "max possible" labels are aligned well with the bottom line. However, the "Goal" label still not aligned to the top. See the attached screenshot. Any suggestions?

assignment_goal

pushyamig:
I will review this now

pushyamig:
the Goal of the PR seem to be align the `Goal` Label as user pick a lower goal grade. This is the current behavior of it

Screen Shot 2021-02-09 at 8 46 23 AM

An alternative would be to do something like this

Screen Shot 2021-02-09 at 8 49 20 AM

I don't know if this is worth fixing this scenario is unlikely user picks a lower grade. But I see lot of code is refactored to fit this case but it still has overlap a bit

The code Look good!

assets/src/components/Line.js
```diff
@@ -1,37 +1,95 @@
...
+  if (value > maxValue) return maxValue
+  if (value < 0) return 0
+  return value / maxValue * 100
+}
+
+const getVerticalStyles = (barHeight, placement) => {
+  const verticalUnit = 25
+  const textHeight = 16
+
+  switch (placement) {
```
pushyamig, line 16: Nice use of switch statement

ssciolla:
> the Goal of the PR seem to be align the `Goal` Label as user pick a lower goal grade.

@pushyamig, this is an issue we've talked about in the past, but it's not what I'm trying to resolve here. I'm just trying to better vertically align the label text with the top or bottom of the associated line. Please see the issue, #1168.

pushyamig:
> @pushyamig, this is an issue we've talked about in the past, but it's not what I'm trying to resolve here. I'm just trying to better vertically align the label text with the top or bottom of the associated line. Please see the issue, #1168.

OK,  sorry for added confusion for lack of understanding!

ssciolla:
> OK, sorry for added confusion for lack of understanding!

No worries, I just wanted to clarify. I'm tweaking a couple things now based on @zqian's feedback.

pushyamig:
Screen Shot 2021-02-09 at 10 53 12 AM

I don't see much difference but you can change it a smidge top =-26px

ssciolla:
Okay, I've made some tweaks to the spacing, using CSS and changing hard-coded values. Here's an updated screenshot.

Screen Shot 2021-02-09 at 11 59 26 AM

Please take a look at the latest commit, @zqian and @pushyamig. Thank you for the reviews so far.

pushyamig (approved):

justin0022 (approved):
Nice @ssciolla, this looks good!

PR #1232, opened 2021-02-04
Fixes #1217 - Make the survey button more noticable

### LINKED ISSUE #1217: Make the survey button more noticeable
Experiment with button and text color to draw users' attention to the survey button.
Get feedback from the research team about the display.

jonespm:


assets/src/components/SurveyModal.js
```diff
@@ -41,6 +41,10 @@ const useStyles = makeStyles((theme) => ({
     right: 0,
     width: '100%',
     height: '100%'
+  },
+  surveyButton: {
+    color: theme.palette.primary.main,
+    background: theme.palette.getContrastText(theme.palette.primary.main),
```
ssciolla, line 47: Nit-pick: Remove the trailing comma? Standard complains, I don't love trailing commas myself, but I know some people do.

ssciolla, line 47: This feels kinda backwards (getting the `contrastText` color to use as a background?), but it works, probably not worth worrying about. I guess the alternative would be just using `white` from Material UI (`import { white } from '@material-ui/core/colors'`) or the hex value, but when I tried it seemed to affect the hover styling. I think what you have is fine.

assets/src/components/SurveyModal.js
```diff
@@ -41,6 +41,10 @@ const useStyles = makeStyles((theme) => ({
     right: 0,
     width: '100%',
     height: '100%'
+  },
+  surveyButton: {
```
ssciolla, line 45: I played around with adding `fontWeight: bold`. More noticeable, but also kinda irritating looking. I guess I'd lean towards what you have, but you could try it out and see what you think.

ssciolla (approved):
Looks fine to me. A little commentary there, but I guess the only think I'd do would be remove the unneeded comma.

assets/src/components/SurveyModal.js
ssciolla: Using `white` might make the color match the current background more, but it also might make the button less noticeable...

jonespm: Yup, don't know why I left that in there. Fixed.

jonespm: Yeah, seemed to me too but it works the way we expected.

I didn't want to import white because was somewhat concerned of what it would look like if someone had the colors different than ours, like if they had a lighter or white color for the top bar. I was hoping for some calculation that would work with custom theming, though I didn't try it with any alternative themes.

jonespm: So if I set
`    "PRIMARY_UI_COLOR" : "#ffffff"`

Then the box comes up black which seems like what we'd want here.
image

assets/src/components/SurveyModal.js
jonespm: Yeah, I don't really like it that much and it doesn't seem to add much. I think once we get the custom text in here it should look good.

assets/src/components/SurveyModal.js
ssciolla: Oh, you are correct about using `white`. I don't think anyone would use a really light color in the app bar, but it's better to handle that case.

assets/src/components/SurveyModal.js
ssciolla: Fine with me.

assets/src/components/SurveyModal.js
jonespm: Yeah, I'd first tried it with a light yellow and it was still a darker color and looked pretty good with the light yellow color. Not sure what color combos places might have.

PR #1233, opened 2021-02-07
issue_1208 changed the get submission sql in cron to set score to null when submission_dim.posted_at is of null value

zqian:


pushyamig:
I will review this today!

pushyamig:
I tested the PR with course 368581, I saw for this case `Submitted, graded, and NOT posted` when an assignment is graded but not posted I do see the corresponding row entry in the submission table and score column is null. Since we don't want bring in the score when posted_at is null. The behavior is correct and I don't see row for the graded assignment is totally missing and the UI correctly reflect that this assignment is submitted not graded.

pushyamig (approved):

PR #1237, opened 2021-02-11
add user data to survey URL (iss. #1216)

### LINKED ISSUE #1216: Pass user info in link to survey
Explore using parameters to capture more about the survey taker, course, etc.

Include information about the logged in user who is submitting the survey when they submit survey responses.

Information to collect:
- user identifier
- course identifier
- view user is on when they take the survey

This was the information about Qualtrics passing to the Survey https://www.qualtrics.com/support/survey-platform/survey-module/survey-flow/standard-elements/passing-information-through-query-strings/

The initial idea was that we would append fixed parameters generated by MyLA to the URL like 
`?userID=<user>&courseID=<course>&view=<assignment|grades|resources>`

We may also want to look at [Q_EED](https://www.qualtrics.com/support/survey-platform/survey-module/survey-flow/standard-elements/passing-information-through-query-strings/#QEED) encoding as mentioned on this page. 

Would have to see how to embed this into qualtrics and document this and test out how it works. 

* Stretch goal: This could be used to block the user from taking the survey again, or block them from taking it again in a different course. That would be a really nice feature. I'm not sure if this is possible or not. This [document on building a retake link](https://www.qualtrics.com/support/survey-platform/data-and-analysis-module/retake-survey-link/#BuildingARetakeLink) might be relevant.

lsloan:
Resolves #1216.

Collaboration with @ssciolla and @jonespm.

Adds parameters to the survey URL that will be stored with the survey data.

### DIFF
--- a/assets/src/components/SurveyModal.js
+++ b/assets/src/components/SurveyModal.js
@@ -1,4 +1,5 @@
 import React, { useState } from 'react'
+import { useLocation } from 'react-router'
 import { makeStyles } from '@material-ui/core/styles'
 import Button from '@material-ui/core/Button'
 import DialogTitle from '@material-ui/core/DialogTitle'
@@ -50,7 +51,24 @@ const useStyles = makeStyles((theme) => ({
 
 export default function SurveyModal (props) {
   const classes = useStyles()
+  const location = useLocation()
   const [open, setOpen] = useState(false)
+
+  const lastTwoSteps = location.pathname.split('/').slice(-2)
+  const viewName = lastTwoSteps.length < 2
+    ? ''
+    : lastTwoSteps[0] === 'courses'
+      ? 'courses'
+      : lastTwoSteps[1]
+
+  const params = {
+    userID: props.user.LTIlaunchID,
+    userName: props.user.username,
+    courseID: props.courseID,
+    view: viewName
+  }
+  const searchParams = new URLSearchParams()
+  Object.entries(params).map(([key, value]) => searchParams.append(key, value))
 
   const toggleOpen = () => setOpen(!open)
 
@@ -65,7 +83,7 @@ export default function SurveyModal (props) {
 
       <div id='survey-modal-description'>
         <div className={classes.iframeContainer}>
-          <iframe className={classes.iframe} src={props.surveyLink.url} height='600px' width='400px' />
+          <iframe className={classes.iframe} src={`${props.surveyLink.url}?${searchParams.toString()}`} height='600px' width='400px' />
         </div>
       </div>
     </div>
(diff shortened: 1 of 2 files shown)


jonespm:
I updated to fix a conflict here

https://github.com/tl-its-umich-edu/my-learning-analytics/compare/master...jonespm:1216-survey-user-info

It looks like it's working and passing the correct values to the iframe. You may want to also use `user.username `if `user.LTIlaunchID` is not set as we are still supporting SAML as well as local development.

assets/src/components/SurveyModal.js
```diff
@@ -50,8 +51,25 @@ const useStyles = makeStyles((theme) => ({
 
 export default function SurveyModal (props) {
   const classes = useStyles()
+  const location = useLocation()
   const [open, setOpen] = useState(false)
 
+  const lastTwoSteps = location.pathname.split('/').slice(-2)
+  const viewName = lastTwoSteps.length < 2
+    ? ''
```
ssciolla, line 59: This was just a fallback (maybe it should be `unknown`?). I don't love this approach since it could easily break if you changed the application structure, but that seems unlikely to happen for now.

I guess the alternative would be passing hardcoded values as props wherever `DashboardAppBar` is used. That could be more flexible, but would require updating a handful more files.

ssciolla: Alternatively, we could just pass the whole `pathname`.

lsloan: yeah, something like `location.pathname || 'unknown'`.

lsloan:
@zqian, @pushyamig, & @chrisrrowland: because @jonespm and @ssciolla collaborated with me on this issue, I think you three are the only developers that should review and approve this PR.  I've added your names as reviewers.  If you're unable to review this, please remove your name from the list of reviewers and tell the remaining viewers that you're leaving the task with them.

Thanks in advance!

lsloan:
@ssciolla suggested I review it so we can make progress.

GitHub doesn't think I should review it.  It won't let me select my name as a reviewer, but I'll start a review anyway.

assets/src/components/SurveyModal.js
lsloan: That's not important, though.  It'd be nice to have, but it can be changed in a future version if it's decided that's a better approach.

assets/src/components/SurveyModal.js
```diff
@@ -50,8 +51,25 @@ const useStyles = makeStyles((theme) => ({
...
+      : lastTwoSteps[1]
+
+  const params = {
+    userID: props.user.LTIlaunchID,
+    userName: props.user.username,
+    courseID: props.courseID,
+    view: viewName
+  }
+  const searchParams = new URLSearchParams()
+  Object.entries(params).map(([key, value]) => searchParams.append(key, value))
```
lsloan, line 71: Very nice.  I'd hoped there would be something like this available.  I didn't really like my initial approach, concatenating a string myself.

lsloan (commented):
Thanks for the major assist, @ssciolla!

jonespm (approved):

PR #1254, opened 2021-03-27
fix submission graded display, but grade not released (iss. #1208)

### LINKED ISSUE #1208: update job must pull null-grade submissions
(cron job needs to pull in submissions with null grade, if the submission has been graded but grades are not released)

Instructors may grade on assignment submissions, but wait till later date to release those grades.

Currently, MyLA cron job does not pull in those "graded but grades are not released yet" submissions

### Steps to Reproduce :
1. Before submitting to an assignment, student will see this assignment marked as "unsubmitted" in Assignment Planning view
2. Student submit an assignment
3. After MyLA cron job, student can view the assignment as "submitted but not graded"
4. Instructor graded the submission without releasing it
5. After another MyLA cron job, student now see the assignment as "unsubmitted" again.

zqian:
(issue_1208 address the display problem of graded but grade not released submission)

Fixes #1208.

### DIFF
--- a/assets/src/util/assignment.js
+++ b/assets/src/util/assignment.js
@@ -222,7 +222,7 @@ const createAssignmentFields = (
         ), 1
       )
       a.outOf = pointsPossible
-      a.graded = !!currentUserSubmission && !!currentUserSubmission.gradedDate
+      a.graded = !!currentUserSubmission && !!currentUserSubmission.gradedDate && !!currentUserSubmission.score
       a.submitted = !!currentUserSubmission && !!currentUserSubmission.submittedAt
       a.dueDateMonthDay = dateToMonthDay(localDate)
       a.goalGrade = ''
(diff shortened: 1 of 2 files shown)


assets/src/util/assignment.js
```diff
@@ -222,7 +222,7 @@ const createAssignmentFields = (
         ), 1
       )
       a.outOf = pointsPossible
-      a.graded = !!currentUserSubmission && !!currentUserSubmission.gradedDate
+      a.graded = !!currentUserSubmission && !!currentUserSubmission.gradedDate && !!currentUserSubmission.score
```
lsloan, line 225: I wasn't aware of the `!!` trick for using a value as Boolean.

dashboard/cron.py
```diff
@@ -508,10 +508,10 @@ def submission(self):
             submission_url = f"""with sub_fact as (select submission_id, assignment_id, course_id, user_id, global_canvas_id, published_score from submission_fact sf join user_dim u on sf.user_id = u.id where course_id = '{data_warehouse_course_id}'),
                 enrollment as (select  distinct(user_id) from enrollment_dim where course_id = '{data_warehouse_course_id}' and workflow_state='active' and type = 'StudentEnrollment'),
                 sub_with_enroll as (select sf.* from sub_fact sf join enrollment e on e.user_id = sf.user_id),
-                submission_time as (select sd.id, sd.submitted_at, sd.graded_at, sd.posted_at at time zone 'utc' at time zone '{settings.TIME_ZONE}' as grade_posted_local_date from submission_dim sd join sub_fact suf on sd.id=suf.submission_id where (posted_at is not null and posted_at < getdate()) OR (posted_at is null and grade is null)),
```
lsloan, line 511: I'm glad this is removed because the use of `getdate()` didn't look stable.  That is, assuming the function returns the current date, each time the query is run, it would get a more recent date.  The later the code is run, that condition will eventually pass.  I don't think that was desired here.

lsloan (approved):
Looks good.  I even learned a little.  😉

dashboard/cron.py
zqian: Good point!

PR #1256, opened 2021-04-01
upgrade Django and local dev. MySQL (iss. #1252)

### LINKED ISSUE #1252: Upgrade Django
Upgrade Django to a newer version to remedy security problems.  This may affect other packages related to the Django in the project.

The dependabot alerts vary, giving different versions from one to another.  Even between the title and the text of one alert the recommendations change:

* https://github.com/tl-its-umich-edu/my-learning-analytics/security/dependabot/requirements.txt/django/closed
    2021-03-18T20:33:25Z
    Title: Bump django from 3.0.8 to **_3.1.6_** (#1249)
    Remediation: Upgrade django to version **_3.0.10_** or later. 
* https://github.com/tl-its-umich-edu/my-learning-analytics/security/dependabot/requirements.txt/django/open
    2021-03-29T21:15:59Z
    Title: Bump django from 3.0.8 to **_3.0.12_** (#1255)
    Remediation: Upgrade django to version **_3.0.12_** or later.

Keep an eye on the requirements.  The remediation text seems to be the most accurate.  With the above information, it might be easiest to upgrade to 3.0.12 first, but that might not buy as much time before the next upgrade is needed.  If that version works, it might be helpful to find a 3.1.x version for the upgrade, as long as it doesn't break the app.

lsloan:
Resolves #1252.

## Django 3.1.7

After much research and testing, I decided to go with Django 3.1.7.  It appears to be the latest version available and it's one higher than highest version suggested by @dependabot.  I was cautiously thinking that we should only use the most recent version of 3.0.x, but MyLA seems to work well with 3.1.7.

## MySQL 5.7 for local development

While working on this issue, I found that MyLA's docker-compose.yml installed MySQL 5.7.22 although my DB was already configured to work with 5.7.26.  There are some changes between 22→26 that cause the older MySQL to complain about using a newer DB.  After discussion with @jonespm, we agreed it would probably be safe to specify the more generic 5.7.

At this time, the version of MySQL installed was 5.7.33.  MySQL's log contained warnings like:

```txt
InnoDB: Table mysql/innodb_table_stats has length mismatch in the column name table_name.  Please run mysql_upgrade
```

I resolved this by running the recommended upgrade in the MySQL container:

```sh
docker exec -it student_dashboard_mysql mysql_upgrade \
--user=root --password student_dashboard
```

When prompted, specify the password for the root MySQL user.  It should be found in the `MYSQL.ROOT_PASSWORD` property of `env.json`.

## Testing

I've tested this as root and as a student.  I've used both the admin interface and the user interface.  Everything seems to work well.  I expected that I may need to update some of the other packages in `requirements.txt`, but I've been surprised that it doesn't seem to be necessary.  Perhaps I've missed some features that should be tested.  I hope others can help me verify this during the code review or testing phases.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -7,10 +7,9 @@ The front end of MyLA is written in JavaScript using [React](https://reactjs.org
 
 The code for the front-end is located under `/assets/src/`. MyLA uses functional components instead of class-based components and uses [React hooks](https://reactjs.org/docs/hooks-intro.html) for state and side-effects.
 
-[Standard](https://standardjs.com/) is the JavaScript linter. Please ensure that your code is Standard-complaint by 
-running the linter manually. It is installed as a dev-dependency. 
-1. you should be able to run it by simply typing `standard` too see any problems.
-2. To fix this you should do `standard --fix`.
+[Standard](https://standardjs.com/) is the JavaScript linter. Please ensure code is standards-complaint by running the linter manually. It is installed as a dev-dependency. 
+1. Run it by simply typing `standard` too see any problems.
+2. To fix this use `standard --fix`.
 
 ### Back-end
 MyLA Backend is build using [Django](https://www.djangoproject.com/) Framework, [MySQL](https://www.mysql.com/) DataBase,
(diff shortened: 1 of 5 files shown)


zqian (approved):
MyLA app runs successfully after the upgrade. Tested the cron job and database access, all work well.

jonespm:
@lsloan Could you add this note about the database upgrade steps to the CONTRIBUTING.md? Seems like it would be useful to know in the future if a message like this ever came up and save a little time.

PR #1259, opened 2021-04-09
fix display of submission graded, not released (iss. #1208)

### LINKED ISSUE #1208: update job must pull null-grade submissions

zqian:
Fixes #1208.

### DIFF
--- a/assets/src/util/assignment.js
+++ b/assets/src/util/assignment.js
@@ -223,7 +223,7 @@ const createAssignmentFields = (
       )
       a.outOf = pointsPossible
       // filter out null values
-      a.graded = (currentUserSubmission !== null) && (currentUserSubmission.gradedDate!== null) && (currentUserSubmission.score !== null)
+      a.graded = (currentUserSubmission !== null) && (currentUserSubmission.gradedDate !== null) && (currentUserSubmission.score !== null)
       a.submitted = !!currentUserSubmission && !!currentUserSubmission.submittedAt
       a.dueDateMonthDay = dateToMonthDay(localDate)
       a.goalGrade = ''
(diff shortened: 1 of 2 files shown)


lsloan (approved):
If this is giving the results we need, then I guess it's good.  The queries were tricky to follow.  From a maintenance perspective, both for developers and code reviewers, it would be helpful if the queries didn't need to be compressed into one line each.  Better formatting would help.

lsloan:
Although I approved this a couple hours ago, I was still thinking about this because I wanted to understand it better.  I corrected the name of the variable, because I see it doesn't have anything to do with URLs at all.

I'm not sure whether the failing check is significant.  I remember this problem being flagged in other projects, like course-inventory.

Note that this PR has conflicts.  Be sure to resolve those before merging.

zqian:
Thanks for the reviews and variable name change. 

I will not fix the alerted "SQL Injection Attack" error since it is not introduced by this PR. I have also fixed the merge conflict.

lsloan:
> I will not fix the alerted "SQL Injection Attack" error since it is not introduced by this PR.

In this case, a fix for that might be more than trivial, and working on it could make us miss the chance to deploy before the production freeze starts.

I'll point out that my fellow developers wouldn't accept my "problem not introduced here" reasoning in the past.  I had to resolve problems at the time regardless, but it wasn't such a tight deadline.

PR #1264, opened 2021-04-16
Fixes #1262 - LTI login in /lti/config incorrect when backend enabled

### LINKED ISSUE #1262: LTI login in /lti/config incorrect when backend enabled
The LTI login link in the config should always be `/lti/login` but there's cases where it will be `/accounts/login/`

The LTI login in the config should always be `/lti/login/`, instead it if you have `"ENABLE_BACKEND_LOGIN": true,` it will come up as `/accounts/login/`.

This looks like it's because the **name** for `/accounts/login/` and `/lti/login` are both "login" so when it does a reverse it gets the wrong one. We need to have a separate name and also update the name in `lti_new.py`.

https://github.com/tl-its-umich-edu/my-learning-analytics/blob/f288ad3c77c3911f1831bf484a2ff275194e768b/dashboard/urls.py#L95

### Steps to Reproduce (Will require a dev) :
1. Configure MyLA for LTI based on the instructions
2. Check the URL /lti/config. The `oidc_initiation_url` should always be `/lti/login`. However it can sometimes be `/accounts/login`.
3. Switch the setting in the config ENABLE_BACKEND_LOGIN to both true and false and test that it is always `/lti/login` in both cases

The problem is this comes up as `/accounts/login/` it doesn't work and it's a little confusing to debug.

jonespm:


pushyamig:
This change make sense to me, I did not run it locally but I feel prefixing it is safer.

pushyamig (approved):

PR #1265, opened 2021-04-19
#1246 adjusted the MyLA cron query for UDP expanded table

### LINKED ISSUE #1246: rewrite resource access query to use expanded table in UDP

The resource access query as in [env_sample.json](https://github.com/tl-its-umich-edu/my-learning-analytics/blob/master/config/env_sample.json) file uses the UDP events table, and parses out attributes from the event json field.

With the newly released UDP v2.0.58, we now have access to the [expanded table](https://resources.unizin.org/display/UDP/2.0.58+Expanded+events+table), with more identifiers from Caliper event parsed out and ready to use. 

Need to rewrite the BQ query to use this new expanded table, for simple query format and reduced BQ data usage.

zqian:
Updated the env_sample.json file. 

Will update Lecture Capture and MiVideo event queries for UM deployment, too.

Fixes #1246.


config/env_sample.json
```diff
@@ -239,25 +239,25 @@
         "canvas": {
             "query": [
                 "SELECT 'canvas' AS resource_type, ",
-                "CAST(SUBSTR(JSON_EXTRACT_SCALAR(event, '$.object.id'), 35) AS STRING) AS resource_id, ",
-                "CAST(SUBSTR(JSON_EXTRACT_SCALAR(event, '$.membership.member.id'), 29) AS INT64) AS user_id, ",
-                "cast(null as string) AS user_login_name,",
-                "CAST(SUBSTR(JSON_EXTRACT_SCALAR(event, '$.group.id'), 31) AS INT64) AS course_id, ",
+                "CAST(SUBSTR(object.id, -17) AS STRING) AS resource_id, ",
```
jonespm, line 242: Looking at this, I think we might want to replace these with something more like

`REGEXP_EXTRACT(object.id, r".*[/:](.*)")`

This would return anything after the last : or / character which seems like it would more reliably capture just the id piece we want. -17 would work for Canvas, but would fail for other types. Using REGEXP_EXTRACT should work for everything.

config/env_sample.json
```diff
@@ -239,25 +239,25 @@
...
-                "UNNEST(['http://m.canvas.umich.edu/' , 'http://umich.instructure.com/' ]) ",
+                "FROM event_store.expanded ",
+                "where ",
+                "ed_app.id IN UNNEST(['http://m.canvas.umich.edu/' , 'http://umich.instructure.com/' ]) ",
                 "and type = 'NavigationEvent' ",
-                "and SUBSTR(JSON_EXTRACT_SCALAR(event, '$.object.id'), 24, 10) = 'attachment' ",
-                "and JSON_EXTRACT_SCALAR(event, '$.action') = 'NavigatedTo' ",
-                "and JSON_EXTRACT_SCALAR(event, '$.membership.member.id') is not null ",
-                "and SUBSTR(JSON_EXTRACT_SCALAR(event, '$.group.id'), 31) IN UNNEST(@course_ids) "
+                "and SUBSTR(object.id, 24, 10) = 'attachment' ",
```
jonespm, line 257: It looks like this value is the same as `object.name` now, so don't have to do this `SUBSTR` here, can do a straight and `object.name = 'attachment'`

config/env_sample.json
```diff
@@ -239,25 +239,25 @@
...
+                "and action = 'NavigatedTo' ",
+                "and membership.id is not null ",
+                "and SUBSTR(`group`.id, -17) IN UNNEST(@course_ids)"
```
jonespm, line 260: Same as earlier suggestion `REGEXP_EXTRACT(`group`.id, r".*[/:](.*)")`. Though it seems like this is available directly as `course_offering.canvas_id`

jonespm:
I was looking for something like that in BigQuery and didn't find anything.

lsloan:
I didn't have anything specific in mind.  It was a conceptual suggestion.
I'll see if I can come up with an example.

--
Mr. Lance E Sloan, Application Developer Senior
Univ. of Michigan, Info. and Tech. Services: http://its.umich.edu
ITS Teaching and Learning: https://its.umich.edu/teaching-learning

config/env_sample.json
jonespm: Yeah, I don't know what would be more efficient but I didn't find any string method for BigQuery to do that, and did see a lot of regex examples.

jonespm:
I figured I'd try to respond with email, would have been nice if it put it on the threaded conversation. :)

config/env_sample.json
lsloan: TL;DR: Using `REGEXP_EXTRACT` seems to be the way to go, unfortunately.

Yes.  I know some languages have functions that can search for one or more characters, from the beginning or end of the string.  Like (this is pseudo-code), `findChars('12:34/45', ':/', 1)` would return the index of `:` if searching from first character (`1`) or `findChars('12:34/45', ':/', -1)` would return the index of `/` if searching from last character (`-1`).  I thought Python or JavaScript have that feature.

Std. SQL's `INSTR` can't search for multiple characters that way.  `REGEXP_INSTR` could.  It would mean using a regex, but just a simple one like `[:/]`.  I see that the documentation for `REGEXP_INSTR` says it can search from position `-1`.  **SPOILER ALERT:**  The documentation is wrong.  That gives an error.

I looked into making a user-defined function (UDF) or procedure for this.  That's too convoluted.

I'm disappointed with BigQuery's standard SQL right now.

zqian: Thanks for the suggestions. I have made the changes accordingly.

jonespm (approved):

config/env_sample.json
zqian: updated with object.name

config/env_sample.json
zqian: updated.

zqian:
documented the UM-specific BQ queries for Lecture Capture and Kaltura resource accesses into the wiki page [Deploy:-Institution-Architectures](https://github.com/tl-its-umich-edu/my-learning-analytics/wiki/Deploy:-Institution-Architectures)

PR #1274, opened 2021-06-10
issue-1266 initial commit for UDW->UDP query rewrite

zqian:


jonespm:
Before doing this, I think we should either copy this file and either make the changes to the copy or just copy this class and have a new class within this file.

1) For instance if we renamed cron.py to cron_udw.py and had a new file cron_udp.py we'd just change the command that runs them to

`python manage.py runcrons "dashboard.cron_udw.DashboardCronJob"`
or
`python manage.py runscrons "dashboard.cron_udp.DashboardCronJob"`

(We should also change the classname here to be MyLACronJob)

2) If we just created a new class within this cron.py we could have something like

MyLACronJob_UDP and MyLACronJob_UDW and run them like
`python manage.py runcrons "dashboard.cron.MyLACronJob_UDW"`
or
`python manage.py runcrons "dashboard.cron.MyLACronJob_UDP"`

zqian:
@jonespm good suggestion. I have renamed the existing cron.py to cron_udw.py; and added cron_udp.py to use UDP data source.

jonespm:
Sounds good! My idea is that we could have them both *available* and after we're happy with the UDP version we can just delete the UDW code.

zqian:
The PR is ready for review. With the latest code, we will keep the cron command as it is:

`python manage.py runcrons "dashboard.cron_udw.DashboardCronJob"`

with the cron queries defined in cron.hjson for UDW and cron_udp.hjson for UDP, so it is relatively easy to switch the cron query set by renaming cron*.hjson into cron.hjson for docker build.

config/cron_udp.hjson
```diff
@@ -0,0 +1,364 @@
+## UDP (Unizin Data Platform) queries for MyLA cron job
+"CRON_QUERIES": {
+    "user" :
+        '''
+        select
+            (17700000000000000 + cast(p2.lms_ext_id as bigint)) as user_id,
```
jonespm, line 6: This won't work for other schools if we have this 17700000000000000 value hardcoded. This is also used in the submission query. 

They could customize this file but generally we've tried to use the config value `CANVAS_DATA_ID_INCREMENT`. I'd pass this into here.

config/cron_udp.hjson
```diff
@@ -0,0 +1,364 @@
...
+            COALESCE( 
+            JSON_EXTRACT_SCALAR(object.extensions, '$[\'com.instructure.canvas\'][asset_name]'), 
+            JSON_EXTRACT_SCALAR(object.extensions, '$[\'com.instructure.canvas\'][filename]'), 
+            object.name, 
+            'attachment' 
+            ) as name, 
+            datetime(EVENT_TIME) as access_time 
+            FROM event_store.expanded 
+            where 
+            JSON_EXTRACT_SCALAR(ed_app.json, '$.id') IN UNNEST(['http://m.canvas.umich.edu/', 'http://umich.instructure.com/'])
```
jonespm, line 321: We should probably look into making these configurable values

config/cron_udp.hjson
```diff
@@ -0,0 +1,364 @@
...
+            where 
+            JSON_EXTRACT_SCALAR(ed_app.json, '$.id') IN UNNEST(['http://m.canvas.umich.edu/', 'http://umich.instructure.com/']) 
+            and type = 'NavigationEvent' 
+            and STARTS_WITH(object.id, 'urn:instructure:canvas:attachment') 
+            and action = 'NavigatedTo' 
+            and membership.id is not null 
+            and REGEXP_EXTRACT(`group`.id, r'.*:(.*)') IN UNNEST(@course_ids)
+            '''
+    },
+    "leccap": {
```
jonespm, line 329: I wouldn't have lecture capture and mivideo in the default configuration, since these probably won't be used by other schools.

I'd like to move all of these configurations into the admin panel too.

config/cron.hjson
```diff
@@ -0,0 +1,239 @@
...
+            where 
+            JSON_EXTRACT_SCALAR(ed_app.json, '$.id') IN UNNEST(['http://m.canvas.umich.edu/', 'http://umich.instructure.com/']) 
+            and type = 'NavigationEvent' 
+            and STARTS_WITH(object.id, 'urn:instructure:canvas:attachment') 
+            and action = 'NavigatedTo' 
+            and membership.id is not null 
+            and REGEXP_EXTRACT(`group`.id, r'.*:(.*)') IN UNNEST(@course_ids)
+            '''
+    },
+    "leccap": {
```
jonespm, line 204: Same as with the other file, we should remove leccap and mivideo from the default config and just leave them documented in the wiki.

jonespm (changes requested):
I started reviewing through this, just noticing a few issues with the config file. I'll look more at the cron job and plan to run this PR soon.

config/cron_udp.hjson
zqian: Done. Replaced by settings.CANVAS_DATA_ID_INCREMENT

config/cron.hjson
zqian: removed

config/cron_udp.hjson
zqian: removed leccap and mivideo queries.

-I'd like to move all of these configurations into the admin panel too.
Will leave this for another issue.

config/cron_udp.hjson
zqian: That has been made as a new configuration variable "CANVAS_EVENT_URLS" into env.json file.

jonespm:
Another thing we're going to have to deal with is the unizin_metadata. As far as the latest Unizin meeting this table/field isn't available in the UDP.

`update_unizin_metadata()` is hardcoded into the cron the the method to retrieve the data `get_canvas_data_date()` is in the util. `get_canvas_data_date()` has a fallback for getting the last time the cronjob was run which is better than nothing, but we'll still want the cron to skip it.

jonespm:
I tested the hjson and json config and it looks good. I'd switch over a few more of the comments in the hjson to the new format but that's about it.

I've tested the UDW side of this and it's looking good. I'm talking to Zhen about issues with "No Term" courses not coming up in the UDP query and causing issues. Going to probably need to do something about those.

PR #1277, opened 2021-06-29
issue-1276 added the check for resource and user course_id in the resource access view

zqian:
To test: 

Install MyLA tool in two courses which has at least one enrolled student in common. Log in with as that student, and make sure no duplicated resource access event shown in the resource_access view.

### DIFF
--- a/dashboard/views.py
+++ b/dashboard/views.py
@@ -357,6 +357,7 @@ def resource_access_within_week(request, course_id=0):
                     "where a.user_id = u.user_id " \
                     "and a.resource_id = r.resource_id " \
                     "and u.sis_name=%(current_user)s " \
+                    "and a.course_id = u.course_id "
                     "group by CONCAT(r.resource_id, ';', r.name)"
     logger.debug(selfSqlString)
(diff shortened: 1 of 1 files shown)


jonespm:
That should work, though I'm wondering if we should just pass the static course_id in as a value. It seems like this query "as-is" would pull all resources accessed for that user from all sites. 

Also as possibly a separate issue, It also doesn't look like it considers the range when making this query. It *seems* like this should align better with the criteria in the earlier query in this method that builds the df.

Is the expectation that this value be the total access or just the total accesses within that range?

jonespm (approved):

lsloan:
@zqian, @jonespm:  We should change the Codacy settings to not complain about Python symbol naming.  At least for MyLA.

The problem this time is that the variable name `selfDf` isn't the style it likes.  I guess it would like `self_df` instead.  (Which I don't like. 🙄)  I tried to disable that pattern in Codacy, but it told me that only a repo admin can do that.  I'm not sure whether it refers to a GitHub admin or a Codacy admin.

PR #1279, opened 2021-07-01
issue-1278 script used to compare extended vs events query results

zqian:


README.md
```diff
@@ -205,6 +205,20 @@ To run the Jest test suite, execute the command `docker exec -it webpack_watcher
...
+Data validation scripts are in scripts/data_validation folder. To run the data validation scripts, follow the steps below:
+
+1. copy env.json file from env_sample.json file, and add configuration values:
+   ```sh
+   cp scripts/data_validation/env_sample.json scripts/data_validation/env.json
+   ```
+2. run data validation script within docker container:
+
+   ```sh
+   docker exec -it student_dashboard /bin/bash -c "python scripts/data_validation/<script file>"
```
jonespm, line 219: Rather than having a placeholder, I'd probably have this be specific like the following. I'm guessing you had it like this because additional scripts might be added? We can make these specific too. 


* Validate Expanded Table - This script will validate queries against the expanded and regular events table are identical.

```sh
   docker exec -it student_dashboard /bin/bash -c "python scripts/data_validation/events_vs_expanded.py"
```

scripts/data_validation/events_vs_expanded.py
```diff
@@ -0,0 +1,212 @@
...
+        bigquery.ArrayQueryParameter(
+            'course_ids_short', 'STRING', DATA_WAREHOUSE_SHORT_COURSE_IDS),
+        bigquery.ScalarQueryParameter(
+            'canvas_data_id_increment', 'INT64', CANVAS_DATA_ID_INCREMENT),
+        bigquery.ScalarQueryParameter(
+            'time_limit', 'STRING', TIME_LIMIT)
+    ]
+
+    # compare the Canvas query
+    canvas_expanded_query_array = [
```
jonespm, line 84: I think it's fine to have these hardcoded here. I wonder if in future validation queries we can have this include these from another file so it actually keeps them in sync with the code and doesn't repeat queries that are present somewhere else in the code.

jonespm: Or have these queries be able to be pulled on a config file, like they currently are. I think the ideal is that these were moved to the database though rather than being in files, since these are all Michigan specific queries.

README.md
jonespm: I'd add a note to this readme that the queries hard-coded in these files have Michigan specific filters so they may have to be modified to be used for other schools unless we make them more configurable.

scripts/data_validation/events_vs_expanded.py
```diff
@@ -0,0 +1,212 @@
...
+def run_canvas_query(query_array, query_params):
+
+    # Instantiates a client
+    bigquery_client = bigquery.Client()
+
+    query_string = ''.join(query_array)
+    print(query_string)
+
+    # BQ Total Bytes Billed to report to status
+    total_bytes_billed = 0
```
jonespm, line 17: It doesn't look like this variable is used

zqian (commented):
Moved query strings into the env_sample.json file, and simplified the events_vs_expanded.py file.

scripts/data_validation/events_vs_expanded.py
zqian: Moved query strings into the env_sample.json file, and simplified the events_vs_expanded.py file.

jonespm (approved):
I think this is fine. We'd wanted to switch to hjson #1056 which is used in some of our other newer projects which allows much nicer multi-line syntax. https://hjson.github.io/

I think that could be updated in the future for this fix too on there. So just going to approve this for now.

PR #1285, opened 2021-07-08
Fixes #1269 - Moving collectstatic out of Dockerfile

### LINKED ISSUE #1269: Potential Security Issue
👋 Hello, we've received a report for a potential medium severity security issue in your repository.

#### Next Steps

1️⃣  Visit **https://huntr.dev/bounties/1-other-tl-its-umich-edu/my-learning-analytics** for more advisory information.

2️⃣  **[Sign-up](https://huntr.dev/)** to validate or speak to the researcher for more assistance.

3️⃣ Propose a patch or outsource it to our community.

---

#### Confused or need more help?

- Join us on our **[Discord](https://huntr.dev/discord)** and a member of our team will be happy to help! 🤗

- Speak to a member of our team: @JamieSlome

---

*This issue was automatically generated by [huntr.dev](https://huntr.dev) - a bug bounty board for securing open source code.*

jonespm:


dashboard/settings.py
```diff
@@ -70,7 +71,7 @@
 # See https://docs.djangoproject.com/en/1.9/howto/deployment/checklist/
 
 # SECURITY WARNING: keep the secret key used in production secret!
-SECRET_KEY = ENV.get('DJANGO_SECRET_KEY')
+SECRET_KEY = ENV.get('DJANGO_SECRET_KEY', get_random_secret_key())
```
ssciolla, line 74: Cool, didn't know this function existed. Doesn't seem to be well documented. I wonder if there would be any drawback/risk to using this with deployments instead of specifying it ourselves.

ssciolla (approved):
LGTM! Based on the SO post, this seems to be the standard approach, and it makes sense that this step should be part of the build. Tested locally.

dashboard/settings.py
jonespm: I think the only problem with having it random in production would be if the server reset some tokens may not work.

And if was run with multiple pods, they'd all get a different random key, and not sure if this would cause any issues or not with the setup we have. We'd have to test if we want to try it.

PR #1292, opened 2021-08-03
Fixes #1289 Off by one error in start date calculation

### LINKED ISSUE #1289: First week of events missing from Resources Accessed view
The Resources Accessed view should show all files that have been accessed during the specified time period.  

Not all files are being included in the view.  In an example test course, students have accessed 5 files and 2 videos.  That data is present in the resources_accessed table in the database.  However, the view shows only 3 files and 1 video.  

Sample test course ID: 511607

jonespm:
Week "1" in the UI actually starts at `timedelta(0*7)` rather than `timedelta(1*7)`

### DIFF
--- a/dashboard/views.py
+++ b/dashboard/views.py
@@ -251,7 +251,7 @@ def resource_access_within_week(request, course_id=0):
 
     # read quefrom request param
     week_num_start = int(request.GET.get('week_num_start','1'))
-    week_num_end = int(request.GET.get('week_num_end','0'))
+    week_num_end = int(request.GET.get('week_num_end','1'))
     grade = request.GET.get('grade','all')
     filter_values = request.GET.get(RESOURCE_TYPE_STRING, ['files', 'videos'])
     filter_values = filter_values.split(",")
@@ -290,8 +290,8 @@ def resource_access_within_week(request, course_id=0):
 
     course_date_start = get_course_date_start(course_id)
 
-    start = course_date_start + timedelta(days=(week_num_start * 7))
-    end = course_date_start + timedelta(days=(week_num_end * 7))
+    start = course_date_start + timedelta(days=((week_num_start-1) * 7))
+    end = course_date_start + timedelta(days=((week_num_end-1) * 7))
     logger.debug("course_start=" + str(course_date_start) + " start=" + str(start) + " end=" + str(end))
 
     # get time range based on week number passed in via request
--- a/dashboard/views.py
+++ b/dashboard/views.py
@@ -291,7 +291,7 @@ def resource_access_within_week(request, course_id=0):
     course_date_start = get_course_date_start(course_id)
 
     start = course_date_start + timedelta(days=((week_num_start-1) * 7))
-    end = course_date_start + timedelta(days=((week_num_end-1) * 7))
+    end = course_date_start + timedelta(days=((week_num_end) * 7))
     logger.debug("course_start=" + str(course_date_start) + " start=" + str(start) + " end=" + str(end))
 
     # get time range based on week number passed in via request


dashboard/views.py
```diff
@@ -290,8 +290,8 @@ def resource_access_within_week(request, course_id=0):
 
     course_date_start = get_course_date_start(course_id)
 
-    start = course_date_start + timedelta(days=(week_num_start * 7))
-    end = course_date_start + timedelta(days=(week_num_end * 7))
+    start = course_date_start + timedelta(days=((week_num_start-1) * 7))
+    end = course_date_start + timedelta(days=((week_num_end-1) * 7))
```
ssciolla, line 294:
```suggestion
    start = course_date_start + timedelta(days=((week_num_start - 1) * 7))
    end = course_date_start + timedelta(days=((week_num_end - 1) * 7))
```

ssciolla: @jennlove-um, I think there is still an issue that we need to work through.
Screen Shot 2021-08-04 at 11 27 26 AM
In my opinion, resources accessed in the first week should be showing up with this setting, right? Basically, if you have "1" selected, it should show resources accessed from the first seven days; if you have "1" and "2" selected, it should show the first 14 days. I believe that's not what is happening so far here; we are only seeing the events from the weeks up to the selected number minus one.

Maybe we just need to adjust the `end` calculation so that it doesn't subtract 1?

@jonespm said this should be easy to fix, so let us know what you think.

jonespm: Yeah, currently selecting a both values on the same dot (1 to 1) only selects a specific **day**. It seems like we should either prevent this situation in the UI or treat this as a week.

1 to 1 **seems** like it's the same as saying "show me week 1" = week 1 + 7 days
1 to 2 **seems** like it's it's saying show me week 1 and week 2 (14 days)

jonespm: This is a different issue than the one identified and fixed here. The bug here was that the start and end date were both always +7 days out from where they should have been.

ssciolla: I'm not sure it's strictly a different issue. With your change, 1 to 2 becomes 0 to 1, or 0-7 days. I think maybe it was correct to keep 2 as is, so it goes up to 14 days.

jonespm: Well with my change 1 to 2 becoming 0 to 1 is still 7 days, but it's the 7 days prior to week 1 rather than the 7 days after week 1. 

Do you mean 1-2 should be 0-2? That would be 21 days the way the calculation works.

Previous code, week_num_start was like 1 to 2
```
    start = course_date_start + timedelta(days=((week_num_start) * 7))
    end = course_date_start + timedelta(days=((week_num_end) * 7))
```
course_start_date is the date defined in the term or admin UI. Assume the start date is 08-01-2021 and it's weeks 1 to 2.

`start = course_start_date + 1*7 days`, so start is 08-08-2021
`end = course_start_date + 2*7 days`, end is 08-15-2021

In the code in this change
```
    start = course_date_start + timedelta(days=((week_num_start-1) * 7))
    end = course_date_start + timedelta(days=((week_num_end-1) * 7))
```

`start = course_start_date + 0*7 days`, so start is 08-01-2021
`end = course_start_date + 1*7 days`, end is 08-08-2021

dashboard/views.py
```diff
@@ -251,7 +251,7 @@ def resource_access_within_week(request, course_id=0):
 
     # read quefrom request param
     week_num_start = int(request.GET.get('week_num_start','1'))
-    week_num_end = int(request.GET.get('week_num_end','0'))
+    week_num_end = int(request.GET.get('week_num_end','1'))
```
jonespm, line 254: I did make a change where if the UI didn't provide the dates to set the value to 1 (instead of 0) for the end date. I don't know why it was set like that and while negative days do work for this calculation it wouldn't return anything more than 0.

dashboard/views.py
ssciolla: I think we're talking past each other a bit, maybe due to differing definitions we're operating under. With your start date of 8-1, I would expect "Week 1" to mean 8-1 to 8-8. It's the first week, it's "Week 1". 

In the "Weeks 1 and 2" case, if 0 and 2 were multiplied by 7, the `timedelta`s would be 0 and 14 days (not 21 days). That's the time period I would expect for selecting 1 and 2, but that needs to be agreed upon (with @jennlove-um's input).

ssciolla: It all kinda boils down to whether the latter dot means to include that selected week, or go up to it. Since it's selected, I think it should include that week.

jennlove-um:
Okay, let's see if I followed all that.  I think if the events should include the ones in the week number selected at both sides of the slider.  So it's a very literal interpretation of the weeks selected and it will mean that we're including 21 days in many cases.

I've tried to work out the dates the way I 'm thinking about them below.  

Week 1: Day 1 - Day 7
Week 2: Day 8 - Day 14
Week 3: Day 15 - Day 21
Week 4: Day 22 - 28

Slider at week 1: include all events for week 1 --> between day 1 and day 7
Slider week 1 - week 2:  include events for weeks 1 & 2 --> between day 1 and day 14  
Slider week 1 - week 3: Include events for weeks 1, 2, & 3 --> between day 1 and day 21 
Slider week 2 - week 4: Include events for weeks 2, 3, & 4 --> between day 8 and day 28

Does that make sense?  It's a little odd and one of the reasons that I would really prefer to move to a date selector rather than try to make sense of week numbers.  

Another option would be to move to adding a week 0 to see if it visually makes more sense.  But for the moment, can we make the calculations work according to what I have above?

jonespm:
Yes, basically what it sounds like we're saying is that we should just push the start date back, not both start and end.

```
    start = course_date_start + timedelta(days=((week_num_start-1) * 7))
    end = course_date_start + timedelta(days=((week_num_end) * 7))
```

dashboard/views.py
ssciolla: Maybe these inputs should be validated someday, but I wouldn't call that high priority. The cron wouldn't pull data for dates outside of the start and end range, I don't think.

ssciolla (approved):
This looks good to me! I tested with the query below, in case that's helpful to someone. I just compared the ordering of the resources versus the UI, I didn't check the percentages.

I agree that the slider as it currently exists is a little harder to reason about, but I think this seems more intuitive to me than not including the latter dot's week.

```
SELECT r.id, r.resource_id, r.name, r.resource_type, COUNT(*) as ra_count
FROM resource r
LEFT JOIN resource_access ra
	ON r.resource_id=ra.resource_id
LEFT JOIN user ue
	ON ra.user_id=ue.user_id
WHERE ra.course_id='{course_id}'
 and ra.access_time >= '2021-XX-XX 04:00:00'
 and ra.access_time <= '2021-XX-XX 04:00:00'
 and ue.enrollment_type='StudentEnrollment'
GROUP BY r.id
ORDER by ra_count DESC
```

jonespm:
Yeah, I feel like I mentioned that validation when this was implemented. That if someone passes a non-int they're going to get an error. 

But having integers both positive and negative outside of the range will be fine, they just don't contain the dates. 

I was thinking of some ideas for improving the UI and still retaining the range slider. Maybe instead of labels 1...16 it could be the start date of the week like Aug 1, Aug 9, Aug 16, etc. I guess the only question would be if we start the week on Sunday or Monday. In the code week 1 starts on the first day of the term rather than on the first day of the week.

PR #996, opened 2020-07-28
#987 optimizing and compressing webpack assets

pushyamig:
I was able to split the single JS file to source code + vendor assets and Was able to compress it as. I see issues with compression is implemented. `django-webpack-loader` depends on `webpack-bundle-tracker`

1. `index.html` simply says 
```
{% render_bundle 'main' 'js' %}
    {% render_bundle 'vendors' 'js' %}
```
2. webpack-stats.json content is 
```
{
  "status": "done",
  "chunks": {
    "main": [
      {
        "name": "main-e5185de39171058cfd22.js",
        "path": "/usr/src/app/assets/dist/main-e5185de39171058cfd22.js"
      }
    ],
    "vendors": [
      {
        "name": "vendors-e5185de39171058cfd22.js",
        "path": "/usr/src/app/assets/dist/vendors-e5185de39171058cfd22.js"
      }
    ]
  }
}
```

1. When webpack is set on prod mode I can see that browser is loading the gzip content. 
Screen Shot 2020-07-28 at 2 48 17 PM

```
webpack_watcher    |       main-e5185de39171058cfd22.js   384 KiB     main  [emitted]  main
webpack_watcher    |    main-e5185de39171058cfd22.js.gz  47.1 KiB           [emitted]
webpack_watcher    |    vendors-e5185de39171058cfd22.js  16.1 MiB  vendors  [emitted]  vendors
webpack_watcher    | vendors-e5185de39171058cfd22.js.gz  1.49 MiB           [emitted]
```
Screen Shot 2020-07-28 at 12 04 39 PM

2. Webpack in watch mode a change on react code updates both `main` and `vendor` mode but compression of the vendor packages doesn't work. I am really not sure why this is the case

   1. With regards to `webpack-bundle-tracker` a new update is available but it has some [breaking](https://github.com/owais/webpack-bundle-tracker/issues/52#issuecomment-599245547) `webpack-stats.json ` structure that python [loader](https://github.com/owais/webpack-bundle-tracker/issues/52#issuecomment-602064536) fails to load the json. Either I have to write my own Python loader class to support this as started [here](https://github.com/owais/django-webpack-loader/#loader_class) 

With version `"webpack-bundle-tracker": "^0.4.3",` and `django-webpack-loader==0.7.0` things kind of works but upgrading to webpack-bundle-tracker": "1.0.0-alpha.1 " breaks stuff. Need to spend sometime as their documentation is very basic and i have been looking at github issues to understand what exactly going on

PR #1294, opened 2021-08-04
Fixes #1293 - Webpack doesn't seem to be reloading the frontend

### LINKED ISSUE #1293: Webpack doesn't seem to be live reloading the frontend anymore on OSX with certain options

When developing and changing the code we used to see changes in the frontend rebuilding. This is the point of webpack_watcher. However it doesn't look like this happens in OSX Docker if you have the option Use gRPC FUSE unchecked.

Unchecking this can give a disk performance boost when developing. 

It seems like the fix for this is to both switch the watchOptions poll to true and also switch `delegated` to `cached` for the mount options.

Note the `delegated` and `cached` are not used on Windows and don't appear to be used in OSX if the Docker option _"Use gRPC FUSE for file sharing"_ is checked.

### Steps to Reproduce :
1. On OSX Docker General have the option Use "gRPC FUSE" unchecked and Apply/Restart.
2. Start up MyLA for development
2. Modify some frontend javascript code
3. The changes should be updated. If you look at the `docker logs webpack_watcher` it should also show changes compiled.

This will only affect development

jonespm:
Is anyone else on any platform seeing this issue? 

Maybe it's just because I have "Use gRPC FUSE" unchecked? I'll try to restart with that checked.

Okay it's only seems like this is an issue if that option is unchecked.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -21,7 +21,7 @@ services:
         TZ: ${TZ}
     command: bash -c "./start.sh"
     volumes:
-      - .:/code:delegated
+      - .:/code:cached
       # use the container's node_modules folder (don't override)
       - /code/node_modules/
       - /code/static/
@@ -47,7 +47,7 @@ services:
       target: node-webpack
     command: sh -c "tar zxvf node_modules/all_symlinks.tgz && npm run watch"
     volumes:
-      - .:/usr/src/app:delegated
+      - .:/usr/src/app:cached
       # use the container's node_modules folder (don't override)
       - /usr/src/app/node_modules/
     container_name: webpack_watcher
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -9,6 +9,7 @@ module.exports = {
     path: path.join(__dirname, 'assets/dist'),
     filename: '[name]-[hash].js'
   },
+  watchOptions: { poll: true },
   devtool: isDevelopment ? 'inline-source-map' : 'false',
   module: {
     rules: [


jonespm:
Some of these ideas seemed to be included in #996 but that wasn't ever merged.

ssciolla:
@jonespm, why is this necessary if you just needed to check the "Use gRPC FUSE for file sharing" option? Is that not something we should be doing?

Maybe these changes are good for other reasons, but if the option change is all that's needed to fix `webpack`, I'm thinking this PR and the issue should be renamed.

jonespm:
I'd mentioned in the ticket that unchecking this option gets the builds and startup to go about twice as fast. It seems like there's still a performance issue with Docker on OSX and at least going back with these options fixes it. 

I'm also not 100% sure if this is an issue with Windows or not. I was hearing from Daniel Zhang that his webpack wasn't reloading on Windows in the Canvas App Explorer project. It might just be related to the watch options there, as I'm seeing these volume options are only applicable to OSX without gRPC.

pushyamig:
@jonespm I think in past i had problem when `Use gRPC FUSE` was checked, but not anymore(maybe after upgrade of Docker version). Now I have that checked and I didn't even applied changes from this PR and I can see webpack is creating a new `main-xx.js` bundle each time.

I will approve this request since this is fixing problem on Windows and is working for Mac.

pushyamig (approved):

jonespm:
Thanks! I was going to do a small fix and ensure the watchOptions was only set when development was enabled. I'm not **entirely** sure why I even had to add that to get it working either. 

Is it already setting watch: true in dev and watch: false in prod? Then I guess it's all set as watchOptions don't matter if it's false. 

https://webpack.js.org/configuration/watch/

jonespm:
Okay looks like the webpack_watcher container locally is running watch explicitly. So this should be fine. Thanks for checking!

PR #1297, opened 2021-08-05
Fixes #1295 - Console Error about Material-UI being renamed

### LINKED ISSUE #1295: Console error about Material-UI being renamed

Saw this error message in the browser console when using MyLA recently. Should fix this so it goes away and doesn't eventually break on us.
```
createTheme.js:94 Material-UI: the createMuiTheme function was renamed to createTheme.

You should use `import { createTheme } from '@material-ui/core/styles'`

```

### Steps to Reproduce :
1. Go to any page
2. Look for errors in the browser console

jonespm:


ssciolla (approved):
LGTM!

PR #1299, opened 2021-08-09
Remove `d3.event.preventDefault` to resolve console errors in Resources Accessed (#1298)

### LINKED ISSUE #1298: Console errors in Resources Accessed related to d3 `preventDefault` method

We should not be seeing excessive console errors in the Resource Accessed view.

When you scroll around in the Resources Accessed view, a lot of repetitive errors are logged to the console with the message "Unable to preventDefault inside passive event listener invocation." We should look at whether this line is necessary, or determine the root cause and another solution.
https://github.com/tl-its-umich-edu/my-learning-analytics/blob/f8ac513925ed13bd4eaad8bad7be28a57b91341c/assets/src/components/d3/createResourceAccessChart.js#L253-L257

### Steps to Reproduce :
1. Go to the Resources Accessed View.
2. Scroll a bit.
3. Look at the console.

ssciolla:
This PR aims to resolve #1298.

I don't know exactly the cause of this error, the scrolling stuff in this `d3` component is confusing to me. However the errors seem to go away when it's removed, and I didn't see any differences in the view. The change could use some testing though, and I think @justin0022 and @pushyamig worked on this code last.

I am still seeing a warning message in the console saying "[Violation] 'wheel' handler took 947ms", but it seems to be happening with or without this change, so it seems to be NOT related.

### DIFF
--- a/assets/src/components/d3/createResourceAccessChart.js
+++ b/assets/src/components/d3/createResourceAccessChart.js
@@ -252,7 +252,6 @@ function createResourceAccessChart ({ data, weekRange, gradeSelection, resourceT
 
     // Make sure the page doesnt scroll
     d3.event.stopPropagation()
-    d3.event.preventDefault()
     // Move the brush
     gBrush.call(brush.move, [topSection, topSection + size])
   }


assets/src/components/d3/createResourceAccessChart.js
```diff
@@ -252,7 +252,6 @@ function createResourceAccessChart ({ data, weekRange, gradeSelection, resourceT
 
     // Make sure the page doesnt scroll
     d3.event.stopPropagation()
-    d3.event.preventDefault()
```
pushyamig, line 255: This change seems to be ok. But I want to run it before I approve this.

pushyamig (approved):
Looks Good!

PR #1300, opened 2021-08-09
Update Django to 3.2 and other Python dependencies (#1286)

### LINKED ISSUE #1084: Remove the package django-settings-export
It looks like the package djang-settings-export is no longer used and can be removed as well as this SETTINGS_EXPORT variable.

#784 has moved a lot of this to the views. We only were using this to make these values easily available to the templates. 

> do we need to have this package django-settings-export? since the 
> 1. login/logout is not be using in the template and 
> 2. GA_ID is passed in via myla globals and 
> 3. I am not sure `RESOURCE_VALUES` is actually used it like that
> 4. I am not sure DEBUG is any values
> 
> If this package is not needed we need to clean up
> 
> _Originally posted by @pushyamig in https://github.com/tl-its-umich-edu/my-learning-analytics/pull/1073#discussion_r486367772_

### LINKED ISSUE #1286: Upgrade backend dependencies, including Django to 3.2
Upgrade Django to 3.2

ssciolla:
The PR aims to resolve #1084 and #1286.

I've built with these changes, and everything seems to work so far. I will be looking more closely at each updated dependency to see where it is used, what has changed, and whether I need to do anything special to test it. Any assistance with testing would be appreciated.

To do
- [x] Scan dependencies and update if possible
  - [x] Find libraries with new releases using PyPi, and update `requirements.txt`
  - [x] Update `webpack-bundle-tracker` `npm` package to ensure compatibility with `django-webpack-loader`
  - [x] Execute `npm install` in repository to update `package-lock.json`
  - [x] Remove outdated and/or minimally used dependencies, including `django-settings-export`, `django-model-utils`, and `django-macros`.
  - [x] Fix bugs associated with updates to `Django`, `pinax-eventlog`, and `pandas`.
- [x] Fix Django warnings
  - [x] Specify `DEFAULT_AUTO_FIELD`
  - [x] Replace `NullBooleanField` usage

Testing
- [x] Run with and without LTI
- [x] Review logs associated with each view and try functionality
- [x] Run cron with a few different courses
- [x] Run management commands


requirements.txt
```diff
@@ -1,51 +1,44 @@
...
 
-pinax-eventlog==2.0.3
-debugpy==1.2.0
-PyLTI1p3==1.8.6
-django-webpack-loader==0.7.0
+debugpy==1.4.1
 jsonschema==3.2.0
-django-csp==3.6
+pinax-eventlog==5.1.1
+pycryptodome==3.10.1
```
ssciolla, line 43: @lsloan and @jonespm, I'd like some feedback here. I asked @lsloan to remove `jwtcrypto` in this PR, and I think he took the initiative to see if he could also remove `pycryptodome` -- which worked at the time I guess? However, when I removed `pyOpenSSL`, it seems to have broken the `createkeys` management command (and maybe also `getjwk`). Adding `pycryptodome` back here seems to have fixed the problem.

I know the cryptographic space in Python is a little strange, but it would be nice for maintenance purposes if there was a clearer tie between library marked as a requirement, and library being imported. If `pyOpenSSL` was being used for cryptography (through some indirect way to `Crypto`), we should make note of that. It also seems like `pyOpenSSL` is unmaintained, and the recommendation is to replace it with [`cryptography`](https://cryptography.io/en/latest/). Maybe we ought to replace `pycryptodome` with `cryptography`; that would probably require some more research (see https://cryptography.io/en/latest/faq/#why-use-cryptography).

What do you think would be the appropriate course of action at this point?

jonespm: I think the management command was a nice convenience but the same could probably just be achieved with documentation or with a shell script that just runs something like

```
ssh-keygen -m PEM -t rsa -b 4096 -f jwtRS256.key 
# Don't add a passphrase

openssl rsa -in jwtRS256.key -pubout -outform PEM -out jwtRS256.key.pub
```

or even suggesting people use the IMS KeyGen utility. https://lti-ri.imsglobal.org/keygen/index. In a future version of this code, these keys will be entered into the admin UI so using something like the imsglobal page is really the easiest way. 

I think it's fine if we just make this a separate issue. I'm all for reducing requirement dependencies.

ssciolla:
I think this is ready for review and further testing. Please let me know if you have any other questions. Issue #1286 has a lot of notes and linked resources about the dependency changes.

dashboard/models.py
```diff
@@ -380,14 +378,13 @@ def get_user_in_course_id(self, user, course_id):
...
 class User(models.Model):
-    ENROLLMENT_TYPES = Choices(
-        ('StudentEnrollment', 'Student'),
-        #('StudentViewEnrollment', 'Student View'),
-        ('TaEnrollment', 'Teaching Assistant'),
-        ('TeacherEnrollment', 'Instructor'),
-        #('DesignerEnrollment', 'Designer'),
-        #('ObserverEnrollment', 'Observer'),
-    )
+    class EnrollmentType(models.TextChoices):
```
jonespm, line 381: This looks like a good change. I think we'd want to convert this over to using the roles.py in pylti but that can be on a separate issue. 

https://github.com/dmitry-viskov/pylti1.3/blob/master/pylti1p3/roles.py

dashboard/templates/base.html
```diff
@@ -1,18 +1,13 @@
...
-{% macro title %}
-    {% block title %}{% endblock %}
-{% endmacro %}
 <html lang="en" xmlns="http://www.w3.org/1999/html">
 <head>
     <meta charset="utf-8">
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
     <meta name="viewport" content="width=device-width, initial-scale=1">
-    <title>{% use_macro title %}</title>
+    <title>{% block title %}{% endblock %}</title>
```
jonespm, line 10: It looks like this was what we originally had before adding macros in 2018.
https://github.com/tl-its-umich-edu/my-learning-analytics/pull/174/files

I don't remember now why macros was used. Is it that this just isn't needed at all with React controlling the title now?

dashboard/templates/base.html
```diff
@@ -56,7 +51,9 @@
                     (commit), {{ git_version.branch }} (branch)
                 </td>
                 {% endif %}
+                {% if last_updated %}
                 <td style="text-align: left">Data last updated on {{last_updated|date:"m/d/Y P T"}}</td>
```
jonespm, line 55: I'd add an `{% else %}` message in here, like "Data has not been loaded"

jonespm (approved):
A couple of comments, I think this is good to get in and we can create new issues if needed for the createkeys and the roles change (if there isn't one for that yet)

dashboard/templates/base.html
ssciolla: Yeah, looks like we just wanted to use the title multiple times before (repeat a block), but we aren't doing that any more with React handling the frontend.

dashboard/templates/base.html
ssciolla: Okay.

dashboard/models.py
ssciolla: I'm not sure I understand why we'd want to do this. The role data is coming from Unizin/Canvas here, and I don't think it will look the exactly the same. Also, if it did work, wouldn't this result in some very tight coupling? I think choices CVs are included in migrations.

jonespm: Yeah, maybe I was thinking of how we could just use the role from the launch instead of the one in Unizin. However since we allow them to switch courses in the app this probably wouldn't be easily possible. These should semi-closely match LTI but yeah, they're not going to be exactly the same.

dashboard/templates/base.html
ssciolla: How's this? https://github.com/tl-its-umich-edu/my-learning-analytics/pull/1300/commits/b5df0747f4bc3da88d626cd1ee6877e9307cde79

requirements.txt
ssciolla: Okay, see https://github.com/tl-its-umich-edu/my-learning-analytics/issues/1313

ssciolla:
@zqian and @pushyamig, do either of you plan to or want to review this before it's merged?

pushyamig:
@ssciolla please go ahead and merge it if you like

PR #1302, opened 2021-08-10
Fixes #1301 - Remove unsed resolvecourses method

### LINKED ISSUE #1301: Remove unused resolve_courses method in GraphQL query

The method `resolvecourses` in query.py doesn't appear to be ever used and is also flagged with an SQL injection vulnerability. This could be fixed by passing a parameter to the raw SQL but since it's not used there's no way to test it. 

So it seems better just to remove it. The ideal would be if this was rewritten either using `Course.objects.extra` pattern or adding a `ForeignKey` so that the raw method wasn't even needed.

The relevant graphql that exists only queries a single course

https://github.com/tl-its-umich-edu/my-learning-analytics/blob/master/assets/src/hooks/useAssignmentData.js

jonespm:


jonespm:
@justin0022 Do you know if this is had planned to be used?

zqian (approved):
Verified there is no usage of resolve_courses() function in the code base, and the MyLA resource planning tool is functioning well with this change.

PR #1326, opened 2021-09-23
Add an event log report to the Django admin & Disable editing, removing and adding event and cron logs

### LINKED ISSUE #1253: Add an event log report to the Django admin
As schools start using MyLA where administrators don't have direct access to the database to get information about usage, reports need to be added to the Django admin screens.

One useful report that has been identified is information from the event log.

Add a report on MyLA event log for specified courses. Ideally, admins would be able to select all courses, all courses in a term, or a list of specified courses to run the report.

Users should be able to export that data into a csv file as well.

### LINKED ISSUE #1325: Disable editing, removing and adding event logs and cron logs

In Django Admin, there shouldn't be any way to remove logs from this interface. That's more of a system operator level and these logs should be considered immutable.

jonespm:
Disable editing, removing and adding event logs and cron logs

Fixes #1253 and Fixes #1325

This adds a second page where you can select the output format(s) for all events
image

It looks like we can add an export for selecting individual events but I don't know how useful that would be.

When I was in there I saw this issue #1325 that I fixed along with this to disable add/edit of events.

### DIFF
--- a/dashboard/admin.py
+++ b/dashboard/admin.py
@@ -8,6 +8,14 @@ from django.template.defaultfilters import linebreaksbr
 from django.utils.html import format_html
 from django.utils.safestring import mark_safe
 
+from pinax.eventlog.admin import LogAdmin
+from pinax.eventlog.models import Log
+from django_cron.admin import CronJobLogAdmin
+from django_cron.models import CronJobLog
+
+from import_export.admin import ExportMixin
+from import_export import resources
+
 from dashboard.common.db_util import canvas_id_to_incremented_id
 from dashboard.models import AcademicTerms, Course, CourseViewOption
 
@@ -79,5 +87,34 @@ class CourseAdmin(admin.ModelAdmin):
         obj.id = canvas_id_to_incremented_id(obj.canvas_id)
         return super(CourseAdmin, self).save_model(request, obj, form, change)
 
+class LogResource(resources.ModelResource):
+    class Meta:
+        model = Log
+
+# This is a local class for logs that adds in Export and disables some actions on the admin
+class MyLALogAdmin(ExportMixin, LogAdmin):
+    resource_class = LogResource
+    # Remove adding and editing for logs
+    def has_add_permission(self, request, obj=None):
+        return False
+    def has_change_permission(self, request, obj=None):
+        return False
+
+# This is local class for Cron that disables add
+class MyLACronJobLogAdmin(CronJobLogAdmin):
+    # Remove adding and editing for cron logs
+    def has_add_permission(self, request, obj=None):
+        return False
+    def has_change_permission(self, request, obj=None):
+        return False
+
 admin.site.register(AcademicTerms, TermAdmin)
 admin.site.register(Course, CourseAdmin)
+
+# Remove the pinax LogAdmin and add ours
+admin.site.unregister(Log)
+admin.site.register(Log, MyLALogAdmin)
+
+# Remove the pinax cron and add ours
+admin.site.unregister(CronJobLog)
+admin.site.register(CronJobLog,MyLACronJobLogAdmin)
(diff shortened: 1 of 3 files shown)


dashboard/admin.py
```diff
@@ -79,5 +87,44 @@ def save_model(self, request, obj, form, change):
...
+    # Remove adding and editing for logs
+    @staticmethod
+    def has_add_permission(request):
+        return False
+    @staticmethod
+    def has_change_permission(request):
+        return False
+    @staticmethod
+    def has_delete_permission(request):
+        return False
```
ssciolla, line 106: I think the `obj=None` is needed for change and delete in order to see the item level pages. Seeing error pages without these changes. Looks like you need to recreate the method signatures here for the overrides? https://docs.djangoproject.com/en/3.2/ref/contrib/admin/
```suggestion
    def has_change_permission(request, obj=None):
        return False
    @staticmethod
    def has_delete_permission(request, obj=None):
        return False
```

dashboard/admin.py
```diff
@@ -79,5 +87,44 @@ def save_model(self, request, obj, form, change):
...
+    # Remove adding and editing for cron logs
+    @staticmethod
+    def has_add_permission(request):
+        return False
+    @staticmethod
+    def has_change_permission(request):
+        return False
+    @staticmethod
+    def has_delete_permission(request):
+        return False
```
ssciolla, line 119: Same issue as above.
```suggestion
    @staticmethod
    def has_change_permission(request, obj=None):
        return False
    @staticmethod
    def has_delete_permission(request, obj=None):
        return False
```

ssciolla (changes requested):
I found some bugs with the method overrides, but once I fixed those, it looked good.

dashboard/admin.py
jonespm: Okay, thanks! I had it in there originally and must have removed too much converting to a staticmethod.

jonespm:
Thanks for the suggestions. I ran it with a few different exports with the changes and didn't see any issues with the logs or it running.

dashboard/admin.py
```diff
@@ -79,5 +87,44 @@ def save_model(self, request, obj, form, change):
...
+    def has_add_permission(request):
+        return False
+    @staticmethod
+    def has_change_permission(request, obj=None):
+        return False
+    @staticmethod
+    def has_delete_permission(request, obj=None):
+        return False
+
+# This is local class for Cron that disables add
```
ssciolla, line 108: Could update this to be more accurate about what's disabled.

dashboard/admin.py
```diff
@@ -79,5 +87,44 @@ def save_model(self, request, obj, form, change):
...
+        model = Log
+
+# This is a local class for logs that adds in Export and disables some actions on the admin
+class MyLALogAdmin(ExportMixin, LogAdmin):
+    resource_class = LogResource
+    # Remove adding and editing for logs
+    @staticmethod
+    def has_add_permission(request):
+        return False
+    @staticmethod
```
ssciolla, line 101: I would have thought this was flagged by Codacy, since I believe PEP8 has spaces between methods. Maybe we're not running PEP8 checkers there, IDK. Not a huge deal.

ssciolla (approved):
LGTM. Couple minor comments, but it works well.

dashboard/admin.py
jonespm: I can fix that. It looks like PyLint doesn't do this check of 

E301 expected 1 blank line, found 0

https://stackoverflow.com/questions/6879640/how-can-i-configure-pylint-to-check-all-things-pep8-checks

zqian (approved):
The export works great! One can apply filter by event type and/or timestamp.

dashboard/admin.py
jonespm: Okay fixed this

dashboard/admin.py
jonespm: Yup, fixed.

PR #1330, opened 2021-10-07
Fixes #1260 - Identified potential SQL injection attacks

### LINKED ISSUE #1260: resolve "SQL injection attack"
PR #1259 had a failing check due to a possible "SQL injection attack" on the SQL string being constructed via variable substitution.  

_Originally posted by @lsloan in https://github.com/tl-its-umich-edu/my-learning-analytics/issues/1259#issuecomment-817060097_
> I'm not sure whether the failing check is significant.  I remember this problem being flagged in other projects, like course-inventory.

It was not corrected at the time because:

1. It was a preexisting condition.  (Not a strong argument.)
1. A solution may not have been trivial and would prolong the development time, possibly missing the opportunity to deploy the application before a production freeze begins.

Unfortunately, when the PR was merged, the information about the failing check was no longer shown.  It's probably available in the checking service or could be reproduced by opening another PR involving this code.

jonespm:


jonespm:
Yeah, I didn't think it needed to be fixed either but figured it was probably better practice still than concatenating the string.

Issue #1348, opened 2021-11-08
Remove "Act As User" local feature, use masqueraded user from LMS


jonespm:
Thank you for contributing to this project!

### Describe your problem or feature you'd like added
We have a feature where you can go into Django Admin and if the user exists act as the user. I think it would make more sense to people if we just used the user that was being masqueraded from Canvas instead. We also wouldn't depend on this page anymore and could probably get closer to not having to store users in the database. (But this is future)

### Describe the solution you'd like
Remove the local "Act As User" and user the masqueraded user Id from the Canvas launch.

PR #1357, opened 2022-03-18
Fixes #1347 - Improve Github links on pages

### LINKED ISSUE #1347: Github href at bottom of page should open in new tab

When you click on the GitHub SHA link on the bottom of the page, since it's in an iframe it opens to a blank page. It should open to a page that works in Github. I think we'll have to have it open in a new tab.

### Steps to Reproduce :
1.  Click the Github SHA at the bottom of any page
2. Should be able to see the Github info
3.

image

saputran:
Hi - I have updated the code to resolve Issue #1347. Let me know if you would like me to change anything!

### DIFF
--- a/dashboard/common/utils.py
+++ b/dashboard/common/utils.py
@@ -46,6 +46,7 @@ def get_git_version_info() -> Optional[GitInfo]:
 
     return {
         "repo": format_github_url_using_https(repo),
+        "repo_abbrev": format_github_url_using_https(repo)[:-4],
         "branch": branch,
         "commit": commit,
         "commit_abbrev": commit_abbrev
--- a/dashboard/common/utils.py
+++ b/dashboard/common/utils.py
@@ -46,7 +46,6 @@ def get_git_version_info() -> Optional[GitInfo]:
 
     return {
         "repo": format_github_url_using_https(repo),
-        "repo_abbrev": format_github_url_using_https(repo)[:-4],
         "branch": branch,
         "commit": commit,
         "commit_abbrev": commit_abbrev
--- a/dashboard/templates/base.html
+++ b/dashboard/templates/base.html
@@ -47,8 +47,9 @@
                 {% if user.is_superuser and git_version %}
                 <td style="text-align: left">
                     Git version:
-                    <a href="{{ git_version.repo_abbrev }}/commit/{{ git_version.commit }}" target="_blank">{{ git_version.commit_abbrev }}</a>
-                    (commit), {{ git_version.branch }} (branch)
+                    <a href="{{ git_version.repo }}/commit/{{ git_version.commit }}" target="_blank">{{ git_version.commit_abbrev }}</a>
+                    (commit)
+                    <a href="{{ git_version.repo }}/tree/{{ git_version.branch }}" target="_blank"> {{ git_version.branch }}</a> (branch)
                 </td>
                 {% endif %}
                 <td style="text-align: left">
--- a/dashboard/templates/base.html
+++ b/dashboard/templates/base.html
@@ -1,4 +1,4 @@
-{% load static %}
+﻿{% load static %}
 {% load tz %}
 
 <!DOCTYPE html>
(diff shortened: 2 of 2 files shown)


dashboard/common/utils.py
```diff
@@ -46,6 +46,7 @@ def get_git_version_info() -> Optional[GitInfo]:
 
     return {
         "repo": format_github_url_using_https(repo),
+        "repo_abbrev": format_github_url_using_https(repo)[:-4],
```
jonespm, line 49: What's the reason for this change? 

It looks like GIT_REPO is 
`https://github.com/tl-its-umich-edu/my-learning-analytics`

And doing this
`utils.format_github_url_using_https(repo)[:-4]`

just returns
`'https://github.com/tl-its-umich-edu/my-learning-analy'`

Which wouldn't work.

dashboard/templates/base.html
```diff
@@ -47,7 +47,7 @@
                 {% if user.is_superuser and git_version %}
                 <td style="text-align: left">
                     Git version:
-                    <a href="{{ git_version.repo }}/commit/{{ git_version.commit }}">{{ git_version.commit_abbrev }}</a>
+                    <a href="{{ git_version.repo_abbrev }}/commit/{{ git_version.commit }}" target="_blank">{{ git_version.commit_abbrev }}</a>
```
jonespm, line 50: This looks good. 

While you're changing this can you make the {{git_version.branch}} on the next line 51 an anchor too?

So it would be in a link like
`<a href="{{ git_version.repo }}/tree/{{ git_version.branch }}" target="_blank">`

dashboard/common/utils.py
saputran: When I was testing it the GIT_REPO automatically adds ".git" at the end of the link so the link was broken (so I got rid of the extra ".git"):

image

dashboard/templates/base.html
saputran: Yes I can add this!

dashboard/common/utils.py
jonespm: Hmm, that sounds like something the method `format_github_url_using_https` _should_ be removing.

```
>>> repo="git@github.com:tl-its-umich-edu/my-learning-analytics.git"
>>> utils.format_github_url_using_https(repo)
'https://github.com/tl-its-umich-edu/my-learning-analytics'
```

What's your GIT_REPO from the start script? 
`git config --local remote.origin.url`

We normally clone using the[ HTTPS URL](https://docs.github.com/en/get-started/getting-started-with-git/about-remote-repositories#cloning-with-https-urls) since you can just use a token and it's generally easier to work with than SSH keypairs.

saputran: Here's what I get when running that command: https://github.com/saputran/my-learning-analytics.git (I forked it so it has my username).

Does the current way work for you when you try to use the link? I tried diffing the link it sends me to and the link it's supposed to be and get that extra ".git".

jonespm: Hmm! It does look like the git clone is adding
`https://github.com/tl-its-umich-edu/my-learning-analytics.git`

I'd say that's a different/new bug. You can file it as a new issue but don't have to fix it. I'm not sure when it was introduced. 

Mine and the one on our server has

```
origin  https://github.com/tl-its-umich-edu/my-learning-analytics (fetch)
origin  https://github.com/tl-its-umich-edu/my-learning-analytics (push)
```

I would run this to update it to the format we're using just for purposes of this bug. 

`git remote set-url origin https://github.com/saputran/my-learning-analytics`

If I was fixing it I'd fix it here
https://github.com/tl-its-umich-edu/my-learning-analytics/blob/474de8d4eb230587f826ad7d72f4a0ec80dbf687/dashboard/common/utils.py#L18

And _probably_ have it `.replace(".git", "")` on every string.

saputran: Thanks! I've updated the code (and added the branch link) and re-committed. I'm not super familiar with git - how do I update my pull request with the updated code? I think right now it just says there are 2 commits (one of which is the updated one).

jonespm: Thanks! That's fine, typically a pull request will have multiple commits against it. When we squash and merge it it'll go down to one commit. 

If you're did have a lot you could rebase and force push. There is a pretty good guide for that here: https://docs.gitlab.com/ee/topics/git/git_rebase.html

But it's a little more advanced topic that you'll probably get very familiar with someday!

saputran: Perfect - thank you! What else is needed to be done for this issue?

jonespm: I created #1348 for this new issue

jonespm (approved):

dashboard/common/utils.py
jonespm: It looks good to me. Thanks!

Issue #1359, opened 2022-03-25
change the DJANGO_DEBUG setting


zqian:
### Describe the bug (Tell us what happens instead of the expected behavior) :
The config/env_sample.hjson contains DJANGO_DEBUG setting as:

`
"DJANGO_DEBUG": true,
`

Need to change the default setting as the following:

`
"DJANGO_DEBUG": false,
`

jonespm:
This looks good and I've tested the default is false now.

PR #1360, opened 2022-03-25
issue_1359 updated the default value for DJANGO_DEBUG settings

zqian:
#1359

### DIFF
--- a/config/env_sample.hjson
+++ b/config/env_sample.hjson
@@ -59,7 +59,7 @@
     "CSRF_COOKIE_SAMESITE": "None",
     # Any logs that django outputs, turn these off for production
     "DJANGO_LOG_LEVEL": "DEBUG",
-    "DJANGO_DEBUG": true,
+    "DJANGO_DEBUG": false,
     "DJANGO_TEMPLATE_DEBUG": true,
     # Any other logs
     "ROOT_LOG_LEVEL": "INFO",
--- a/config/env_sample.hjson
+++ b/config/env_sample.hjson
@@ -57,10 +57,10 @@
     # SameSite settings for Session and CSRF (defaults in settings.py should work), if you do want non-string None set to null.
     "SESSION_COOKIE_SAMESITE": "None",
     "CSRF_COOKIE_SAMESITE": "None",
-    # Any logs that django outputs, turn these off for production
-    "DJANGO_LOG_LEVEL": "DEBUG",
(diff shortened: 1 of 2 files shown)


config/env_sample.hjson
```diff
@@ -59,7 +59,7 @@
     "CSRF_COOKIE_SAMESITE": "None",
     # Any logs that django outputs, turn these off for production
     "DJANGO_LOG_LEVEL": "DEBUG",
-    "DJANGO_DEBUG": true,
+    "DJANGO_DEBUG": false,
```
jonespm, line 62: We should add more comments around specific to these settings. Maybe like

```
# This enables debugging of Django code and should only be enabled in development. 
# Setting this to true in production may reveal internal information such as settings to a user if they encounter an error.
"DJANGO_DEBUG": false,
```

We probably should also change `DJANGO_LOG_LEVEL` to `INFO` in the line above and also set DJANGO_TEMPLATE_DEBUG to false. 
```
# Setting this to true can help with debugging Django templates. This should be false in production and enabled if needed in development.
  "DJANGO_TEMPLATE_DEBUG": true,
````

**Review by jonespm — CHANGES_REQUESTED** [review]

File: config/env_sample.hjson (hunk 1 above)
Comment (zqian) [reply config/env_sample.hjson:63]: @jonespm In order to simplify the settings and avoid confusion, I have removed the variables "DJANGO_TEMPLATE_DEBUG" and "DJANGO_LOG_LEVEL", so only "DJANGO_DEBUG" controls the log levels. Please take another look.

Comment (jonespm) [reply config/env_sample.hjson:63]: I agree tying `DJANGO_TEMPLATE_DEBUG `and `DJANGO_DEBUG `together makes sense.

I'd probably leave the log level independently configured but using this as the fallback makes sense.

File: dashboard/settings.py [hunk 2]
```diff
@@ -302,7 +302,7 @@
         'django': {
             'handlers': ['console'],
             'propagate': False,
-            'level': ENV.get('DJANGO_LOG_LEVEL', 'INFO'),
+            'level': 'DEBUG' if DEBUG else 'INFO',
```
jonespm, line 305:
```suggestion
             'level': ENV.get('DJANGO_LOG_LEVEL', 'DEBUG' if DEBUG else 'INFO'),
```

jonespm (approved):

PR #1362, opened 2022-03-29
Fixes #1198 - Increase detail for missing start/end date

### LINKED ISSUE #1198: When a course/term has no start/end date message should just be info and indicate which course/term
When a course/term doesn't have a date_start or date_end a warning is currently printed from models without any details of the problematic course term. Also this message can just be INFO.

The messages are similar to the ones below.
```

[2021-01-13 08:22:06 -0500] [WARNING] [models.py:219] No date_start value was found for course or term; setting to current date and time
[2021-01-13 08:22:06 -0500] [WARNING] [models.py:226] No date_end value was found for course or term; setting to two weeks from now
```

Reduce these to INFO and print out the actual course or term that has a problem.

### Steps to Reproduce :
1. Add a course or term without a date_start, date_end and no term. Like a brand new course that hasn't gotten into UDW yet.
1. You also can just add MyLA to a course that hasn't yet run the cron on it, and launch it. This error will also print for that course, and it shouldn't be a warning.
1. Add a term that doesn't have date_start or date_end (Might be harder to test, and wasn't the problem here but it does say term in the log message)

saputran:
Hi - here is a pull request for issue 1198 - please let me know if you would like me to update anything. Thank you!

### DIFF
--- a/dashboard/models.py
+++ b/dashboard/models.py
@@ -226,7 +226,7 @@ class Course(models.Model):
         elif self.term is not None and self.term.date_end is not None:
             end = self.term.get_correct_date_end()
         else:
-            logger.info("No date_start value was found for course " + str(self.name) + " (" + str(self.canvas_id) + ") or term; setting to current date and time")
+            logger.info(f"No date_start value was found for course {self.name} ({self.canvas_id}) or term; setting to current date and time")
             end = start + timedelta(weeks=2)
         DateRange = namedtuple("DateRange", ["start", "end"])
         return DateRange(start, end)
--- a/dashboard/models.py
+++ b/dashboard/models.py
@@ -219,7 +219,7 @@ class Course(models.Model):
         elif self.term is not None and self.term.date_start is not None:
             start = self.term.date_start
         else:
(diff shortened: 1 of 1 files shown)


dashboard/models.py
```diff
@@ -219,14 +219,14 @@ def course_date_range(self):
...
-            logger.warning("No date_start value was found for course or term; setting to current date and time")
+            logger.info("No date_start value was found for course " + str(self.name) + " (" + str(self.canvas_id) + ") or term; setting to current date and time")
             start = datetime.now(pytz.UTC)
         if self.date_end is not None:
             end = self.date_end
         elif self.term is not None and self.term.date_end is not None:
             end = self.term.get_correct_date_end()
         else:
-            logger.warning("No date_end value was found for course or term; setting to two weeks from now")
+            logger.info("No date_start value was found for course " + str(self.name) + " (" + str(self.canvas_id) + ") or term; setting to current date and time")
```
jonespm, line 229: Same as other

```suggestion
            logger.info(f"No date_start value was found for course {self.name} ({self.canvas_id}) or term; setting to current date and time")
```

jonespm (approved):

PR #1363, opened 2022-03-31
Perform front-end dependency update (#1312)

### LINKED ISSUE #1312: Upgrade frontend dependencies
This would be the logical next step after #1286, though it may not need to be done immediately.

As part of this, we should also evaluate the usage of `django-npm`. I looked into removing it (because it's unmaintained) as part of the aforementioned issue, but doing so resulted in some style changes (loss of icons and screenreader text started showing up visually). It would make sense to look at this in conjunction with the frontend dependencies. I don't think it's strictly necessary, but it would require some changes and testing to ensure nothing is broken. Also, it looks like `jquery` is being pulled in through this strategy (see `settings.py`, `NPM_FILE_PATTERNS`, but it may not be used anywhere.

ssciolla:
This PR updates, pares down, and otherwise tries to improve the usage of `npm` packages in the frontend. In the process, some changes were necessary around the use of `@apollo/client` (mostly updating imports, configuration), `react-router-dom` (adjusting `Route` usages), `fontawesome` (switching to a Python library for providing CSS styles), and `@material-ui` (getting rid of `RootRef` instances).  A lot of different areas are touched here, but the number of non-dependency file changes is hopefully not too high. The PR aims to resolve #1312.

To Do
- [x] Bump Node to 16, convert `package-lock.json` to new lockfile format.
- [x] Remove `jquery`, `node-sass` (not used)
- [x] Update
  - [x]  React (and associated dependencies)
  - [x] Update Webpack (resolving errors, warnings, etc.)
  - [x] Apollo/`graphql` (and fix imports, course query)
  - [x] `rc-slider`
  - [x] `jest`, checking to make sure tests still run
  - [x] other misc. imports (`babel`, `js-cookie`, loaders, etc.) 
  - [x] Material UI to last published release of version 4
    - [x] Remove `RootRef` usage due to deprecation
  - [x] `react-router-dom` to 5.3 (I pushed this forward a bit by adopting new version 5 patterns, but I am also waiting on updating to version 6, since they are recommending we wait until there is a backwards-compatible version. https://reactrouter.com/docs/en/v6/upgrading/v5#refactor-custom-routes).
- [x] Review un-updated dependencies
- [x] Replace `django-npm` with `fontawesomefree`
- [x] Create new issue regarding `d3` updates (see https://github.com/tl-its-umich-edu/my-learning-analytics/issues/1367).
- [ ] Test with LTI launch
- [ ] Run Cypress tests


assets/src/service/client.js
```diff
@@ -1,10 +1,12 @@
-import ApolloClient from 'apollo-boost'
+import { ApolloClient, InMemoryCache } from '@apollo/client'
 import Cookie from 'js-cookie'
 
 export default new ApolloClient({
+  uri: 'http://localhost:5001/graphql',
```
ssciolla, line 5: @jonespm, this needs to be configurable, I guess? Not sure what changed to make this required. Probably would pass the hostname through `globals`. Do we have that in the backend settings somewhere?

jonespm: I don't think so. Could you just make it relative and set it to be `/graphql`? Or maybe use `window.location.origin`?

package.json
```diff
@@ -5,25 +5,22 @@
...
-    "@fortawesome/fontawesome-svg-core": "1.2.32",
-    "apollo-boost": "^0.4.4",
+    "@apollo/client": "^3.5.10",
+    "@fortawesome/fontawesome-free": "6.1.1",
+    "@fortawesome/fontawesome-svg-core": "6.1.1",
+    "@fortawesome/react-fontawesome": "0.1.18",
     "clsx": "^1.1.1",
-    "d3": "^5.12.0",
+    "d3": "^5.16.0",
     "d3-tip": "~0.9.1",
```
ssciolla, line 14: Upgrading `d3` is a significant piece still to be done, and which I think we'll tackle separately. We're behind two major releases and the `d3-tip` companion library seems to have been archived on GitHub (https://github.com/caged/d3-tip). I found a migration guide for version 6, but no equivalent yet for version 7: https://observablehq.com/@d3/d3v6-migration-guide

package.json
```diff
@@ -91,29 +87,28 @@
...
+    "mini-css-extract-plugin": "2.6.0",
+    "react-app-polyfill": "^3.0.0",
+    "react-test-renderer": "17.0.2",
+    "standard": "^16.0.4",
+    "style-loader": "~3.3.1",
+    "typeface-roboto": "1.1.13",
+    "url-loader": "4.1.1",
+    "webpack": "5.70.x",
+    "webpack-bundle-tracker": "^1.4.0",
+    "webpack-cli": "4.9.x"
```
ssciolla, line 112: It looks like we could maybe optimize how `node_modules` is built. Some of the dependencies under `devDependencies` are more like `dependencies` (`@material-ui`, `js-cookie`, `typeface-roboto`?). Looking at the Docker artifacts, I see there is an effort to remove extraneous packages (`npm prune --production`), but I'm not sure if it's actually working. When I go to OpenShift and peak inside `node_modules`, I still see `jest`, `webpack`, `-loader`, etc. We should probably also make sure that the frontend unit test and related artifacts are left behind from the production images (currently, they are under `assets/__tests__`).

assets/src/service/client.js
ssciolla: Looks like changing it to `/graphql` was fine.

assets/src/components/AssignmentTable.js
```diff
@@ -322,190 +320,182 @@ function AssignmentTable (props) {
...
-                    'Name',
-                    'Percent of Final Grade',
-                    'Score / Out of',
-                    'Status',
-                    'Lock Goal'
-                  ].map((heading, key) => (
+      <TableContainer className={classes.container} ref={tableRef}>
+        <MTable stickyHeader>
+          <TableHead>
+            <TableRow>
```
ssciolla, line 326: All the changes here are to replace `RootRef`, which is being deprecated in future releases of Material UI and is throwing a warning/error in the console. It's mostly unindenting things, adding the `ref` to `TableContainer`, and replacing the `ConditionalWrapper` logic with the ternary `ref` work on new line 352.

assets/src/components/GoogleAnalyticsTracking.js
```diff
@@ -26,4 +25,4 @@ function GoogleAnalyticsTracking (props) {
   return null
 }
 
-export default withRouter(GoogleAnalyticsTracking)
+export default GoogleAnalyticsTracking
```
ssciolla, line 28: Not sure why `withRouter` was used here to begin with, since I'm not seeing any `react-router-dom` params (like `match` or `location`) be used in `props`...

assets/src/components/IconLabel.js
```diff
@@ -1,15 +1,30 @@
...
   return (
-    <p><i style={{ fontSize: '14px', padding: '5px 10px 5px 5px' }} className={icon} />{label}</p>
+    <Typography className={classes.typography}>
+      <i className={`${icon} ${classes.icon}`} />{label}
+    </Typography>
   )
 }
 
-export default IconLabel
+export default withStyles(styles)(IconLabel)
```
ssciolla, line 30: I refactored here to use the Material UI styling solution over inline styles, and to adjust the sizing and spacing in response to a slightly different `fontawesome` icon.

assets/src/components/TableAssignment.js
```diff
@@ -125,8 +125,8 @@ function CustomAssignmentTable (props) {
       return row
     })
 
-  const tableRow = (row, i) => (
-    <TableRow key={i}>
+  const tableRow = (row, i, isCurrentWeek) => (
+    <TableRow key={i} {...(isCurrentWeek ? { ref: currentWeekRow } : {})}>
```
ssciolla, line 129: Same deal in this table component. I'm replacing `RootRef`  by just applying the `tableRef` to `Table` and then using this logic so that the scrolling to the current week still works.

assets/src/components/d3/createHistogram.js
```diff
@@ -55,7 +55,7 @@ function createHistogram ({ data, width, height, domElement, xAxisLabel, yAxisLa
   main.append('div')
     .attr('aria-live', 'polite')
     .attr('id', 'grade-view-narrative')
-    .attr('class', 'screenreader-only sr-only')
+    .attr('class', 'fa-sr-only')
```
ssciolla, line 58: I was confused at first when playing around with `fontawesome` approaches why this was being affected, so I changed it to just use the `fa` prefix for clarity. I didn't see `screenreader-only` applying any styles.

assets/src/components/d3/createResourceAccessChart.js
```diff
@@ -3,7 +3,6 @@ import { adjustViewport } from '../../util/chart'
 import { createResourcesText } from './d3ViewsNarrative'
 import d3tip from 'd3-tip'
 import './createResourceAccessChart.css'
-import '@fortawesome/fontawesome-free'
```
ssciolla, line 6: I don't think this does anything.

assets/src/components/d3/createResourceAccessChart.js
```diff
@@ -166,7 +165,7 @@ function createResourceAccessChart ({ data, weekRange, gradeSelection, resourceT
   main.append('div')
     .attr('aria-live', 'polite')
     .attr('id', 'resource-view-narrative')
-    .attr('class', 'screenreader-only sr-only')
+    .attr('class', 'fa-sr-only')
```
ssciolla, line 168: See comment https://github.com/tl-its-umich-edu/my-learning-analytics/pull/1363/files#r846426301

dashboard/templates/base.html
```diff
@@ -8,9 +8,8 @@
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
     <meta name="viewport" content="width=device-width, initial-scale=1">
     <title>{% block title %}{% endblock %}</title>
-    <link rel="stylesheet" type="text/css" href="{% static '@fortawesome/fontawesome-free/css/all.css' %}">
-    <link rel="stylesheet" type="text/css" href="{% static '@fortawesome/fontawesome-free/css/fontawesome.css' %}">
-    <link rel="stylesheet" type="text/css" href="{% static '@fortawesome/fontawesome-free/css/fontawesome.min.css' %}">
+    <link rel="stylesheet" type="text/css" href="{% static 'fontawesomefree/css/all.min.css' %}">
+    <script src="{% static 'fontawesomefree/js/all.min.js' %}"></script>
```
ssciolla, line 12: `fontawesome` offers a Django/solution for using icons in templates called `fontawesomefree`: https://pypi.org/project/fontawesomefree/ This isn't strictly a replacement for `django-npm`, but after we removed `jquery`, `fontawesome` seemed to be the only thing we really needed `django-npm` for.

To explain a bit more, `fontawesome` CSS classes are being used for Resources Accessed and for screenreader-only content, and thus we needed a way to load the styling. Using a strictly React approach proved somewhat challenging because of 1) we currently use string classes for icons on the Resources Accessed D3 chart, where we can't really use JSX and 2) figuring out how to load the CSS file through `webpack` was a headache. If D3 is ever replaced, we could look into using this approach (https://fontawesome.com/docs/web/use-with/react/), or removing `fontawesome` all together in favor of Material UI icons. The latter though would require configuration changes, since we currently pass CSS class strings in `env.json` for each resource type.

@jonespm, let me know what you think of this solution for removing `django-npm`. Here's the commit:
https://github.com/tl-its-umich-edu/my-learning-analytics/pull/1363/commits/eb3492de43acc1752d5e587c74523855f90ddc28

ssciolla:
I'm going to do more testing, so I can't guarantee this is completely stable yet, but I think we can get the review process started.

@pushyamig, I'm going to request your review since you're knowledgeable about the frontend, but I understand you have other priorities too.

package.json
ssciolla: I'd like to discuss with @jonespm but this is probably best tackled separately at later date.

assets/src/components/GoogleAnalyticsTracking.js
ssciolla: It's probably worth testing Google Analytics in test if we can, though.

pushyamig:
> @pushyamig, I'm going to request your review since you're knowledgeable about the frontend, but I understand you have other priorities too.

is there a time I should review this by? It's been a year since I ran Myla Locally.

ssciolla:
> > @pushyamig, I'm going to request your review since you're knowledgeable about the frontend, but I understand you have other priorities too.
> 
> is there a time I should review this by? It's been a year since I ran Myla Locally.

Not sure, probably needs to be merged next week to have a chance of getting in the next release. I may be asked to scale it back if no one wants to or can review it soon.

pushyamig:
@ssciolla I will review it today, I was able to run this locally.

webpack.config.js
```diff
@@ -7,10 +7,10 @@ module.exports = {
   entry: path.join(__dirname, 'assets/src/index'),
   output: {
     path: path.join(__dirname, 'assets/dist'),
-    filename: '[name]-[hash].js'
+    filename: '[name]-[fullhash].js'
```
pushyamig, line 10: Is this change part of upgrade to new webpack version?

package.json
pushyamig: We would want to upgrade to latest since 2 of our visuals are using d3 solution. But if this not a huge risk we can handle this separately. I don't know if this is a minor, patch Myla release.

webpack.config.js
ssciolla: Yes, there was an error or deprecation warning, I forget which

dashboard/urls.py
```diff
@@ -42,7 +42,7 @@
     path('admin/', admin.site.urls),
 
     # Note the absence of a trailing slash; adding one breaks the GraphQL implementation.
-    path('graphql', DashboardGraphQLView.as_view(graphiql=True)),
+    path('graphql', DashboardGraphQLView.as_view(graphiql=settings.DEBUG)),
```
pushyamig, line 45: What is the reason for tying this debug setting?

assets/src/service/client.js
```diff
@@ -1,7 +1,9 @@
-import ApolloClient from 'apollo-boost'
+import { ApolloClient, InMemoryCache } from '@apollo/client'
 import Cookie from 'js-cookie'
 
 export default new ApolloClient({
+  uri: '/graphql',
+  cache: new InMemoryCache(),
```
pushyamig, line 6: What is the advantage of adding this?

ssciolla: It seemed to be required with the change in Apollo versions.

package.json
ssciolla: I'd say this PR has enough changes where adding `d3` stuff would put it overboard (if isn't already). We can discuss as a team whether `d3` needs to be updated in this release, but again, I don't see any vulnerabilities right now for this

assets/src/containers/Course.js
```diff
@@ -83,66 +83,50 @@ function Course (props) {
...
-                        disabled={!courseInfo.course_view_options.ra}
-                        courseInfo={courseInfo}
-                        courseId={courseId}
-                        enrollmentTypes={enrollmentTypes}
-                        isAdmin={user.admin}
-                      />}
-                  />
-                </>
+                <Switch>
+                  <Route path='/courses/:courseId/' exact>
```
pushyamig, line 87: For all the routes below, I see `{...props}` is removed. was that redundant? or props are available in some other way?

I know there is one use case that instructor given a choice to select what views they want for the course for example disabling grade distribution, I think locally we are maintaining the state using props if he doesn't launch the tool. 

I haven't tested the UI still, I will in a bit.

assets/src/components/d3/createHistogram.js
pushyamig: Can you point me to the documentation for `fa-sr-only`?

assets/src/components/GoogleAnalyticsTracking.js
pushyamig: I agree on this. https://analytics.google.com/

pushyamig: Mostly google analytics is real time update to dashboard. But Last time I used chrome extension if GA data is collected where we need to https://chrome.google.com/webstore/detail/google-analytics-debugger/jnkmfdileelhofjcijamephohjechhna?hl=en

assets/src/components/d3/createHistogram.js
pushyamig: Thanks! When I added accessibility support for d3 visuals, I added those based on Gonzalo's recommendation. I don't remember too many details now. I felt it is best I will test it now.

assets/src/components/TableAssignment.js
pushyamig: This working nicely. thanks for noticing that this was deprecated.

assets/src/components/GoogleAnalyticsTracking.js
pushyamig: This should work, I agree with you. But  see the GA action is performed using the chrome extension, But I struggled seeing data from GA official Website. But I don't feel this change has anything to do with that.

assets/src/containers/Course.js
pushyamig: This change doesn't seem to have side effect and seems to be safe

assets/src/components/d3/createHistogram.js
```diff
@@ -55,7 +55,7 @@ function createHistogram ({ data, width, height, domElement, xAxisLabel, yAxisLa
   main.append('div')
```
pushyamig, line 55: @ssciolla I see a deprecation notification in this file for `d3.histogram` and instead it is suggesting to use `bin()`? How you consider this change or you want to do this as part of #1367

ssciolla: I must have missed that. Is it a warning or an error? Likely `d3` is warning us that something will be deprecated in `v6`. If it's an easy swap, we could try it.

assets/src/components/d3/createHistogram.js
pushyamig: I don't think we are using `fontawesome` fonts for histogram. We only rely on the d3/svg fonts. why was this change needed? Especially this line is making sure that screen reader can read the text for accessibility. 

I don't see any harm done when I turn on voice over things are working fine. But Trying to understand change here.

ssciolla: Well, when I removed `fontawesome` during testing, I noticed that the screenreader text you added was being shown visually. So after analyzing the styles, I realized `fontawesome` classes were being used to make the text screenreader only. I think `sr-only` must have been an alias or alternative for `fa-sr-only`; I changed it to `fa-sr-only` because I wanted us to be able to tell later that it came from `fontawesome` (`fa-`). I don't think the `screenreader-only` class was necessary or adding any styles, so I removed it.

Does that help explain?

assets/src/components/d3/createHistogram.js
pushyamig: I think VScode is looking at future version of d3 and webpack is still referring to v5.16. it is really strange that when my local node module has v5.16 downloaded, why VSCode try to present with a warning of deprecation for future version. I think this is not an issue with that version of d3 we reference.

pushyamig (approved):
Over all this change looks good and positive. I was able to test this with few courses both as a student and instructor view. I don't see any side effect from this change. I am approving this now, but I want fellow reviewers to test and approve as well since I am coming to project after a gap.

pushyamig:
I saw a notification today that Google analytics team is sunsetting current version Universal Analytics and it is replacing with [Google analytics 4](https://support.google.com/analytics/answer/10089681?utm_campaign=2022-q1-gbl-all-gafree&utm_source=google-growth&utm_medium=email&utm_content=ga-sunset-msa-ua-only-admin). I think we are using a react-ga library which is still support [Universal analytics](https://github.com/react-ga/react-ga).

At some point i believe this library will update as well. I don't know what is difference between existing and new one. It looks like we need to configure this GA4 from our GA console as well. I think there may be some work from that in future

package.json
jonespm: It looks like @justin0022 originally added the `npm prune` afd74b5a8c654d6c0100b427210ba2751bc97f8a.

I've never checked to see if it's working either. 

It looks like it might have to be combined with [node-prune ](https://github.com/tj/node-prune) to really remove the node_modules. I saw that mentioned in this article here: https://itsopensource.com/how-to-reduce-node-docker-image-size-by-ten-times/

I think we can just address this in a separate issue but yeah having the appropriate things in `devDependencies` makes sense to me.

assets/src/components/d3/createResourceAccessChart.js
jonespm: Yeah, looks like the only use we have should be covered by IconLabel.

dashboard/templates/base.html
jonespm: I think originally pre-react we needed this for more. But that commit looks good to remove it. I'd rather have this supported package than the older unsupported one. Be great if we have time to eventually do something with D3.

jonespm:
@pushyamig I saw that today too. I created #1371 to capture that work that we'll need to do. We'll have to look at this for our other projects that use GA too and will still be running. 

I think just Student Explorer. GA in MWrite has been broken since the last switch to Google UA.

jonespm (approved):
Looks like everything has been addressed or created as a new issue.

PR #1366, opened 2022-04-07
Fixes #1353 - Remove or provide alternative for person_sourcedid to pass certification

### LINKED ISSUE #1353: Remove or provide alternative for person_sourcedid to pass certification
[http://www.imsglobal.org/spec/lti/v1p3/#learning-information-services-lis-claim](https://purl.imsglobal.org/spec/lti/claim/lis) is an optional claim value that isn't supported by the certification. There isn't any intention in supporting this optional claim.

We need to either remove it and get this value from a different source (UDW), remove it if we don't need it at all, or be able to also optionally pass it as part of a custom claim parameter.

https://github.com/tl-its-umich-edu/my-learning-analytics/blob/f4a4bdcdb1821adb09130486ac85792a2f576548/dashboard/lti_new.py#L230

jonespm:
…ass certification


jonespm:
I've tested this locally and it seems to be working as expected. Probably will need a little more testing on a more realistic course once deployed.

zqian (approved):
I've tested this PR with LTI integration using localhost running with ngrok. The LTI launch works fine

PR #1369, opened 2022-04-13
Fixes #1072 - Code cleanup for unused registration module

### LINKED ISSUE #1072: Remove django-registration and registration

I noticed in urls.py we have a block for registration but it is never possible to enable this anywhere and it's not something we're even using. We should remove it from the requirements and any other place it might be used.

jonespm:


zqian (approved):
Tested the changes with random clicks inside MyLA tool. No errors in log file; I an still do admin login with /accounts/login/

PR #1380, opened 2022-05-05
issue_1379 added the platform option for MySQL 

zqian:
Added the platform option in docker-compose.yml, since the current MySQL image does not support ARM containers (Apple M1 is an ARM-based system).

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -4,6 +4,7 @@ services:
 
   mysql:
     image: mysql:5.7
+    platform: linux/amd64
     restart: on-failure
     command: ['--character-set-server=utf8mb4', '--collation-server=utf8mb4_unicode_ci']
     env_file:


pushyamig:
I was able to run this PR both in Mac with Apple chip and Intel chip. The Application runs on both environments with out any problem.  I was able to run the Cron and in both cases and it is a success

One warning I noticed in Intel Chip version of Mac is as below, but I did not notice this in the Apple chip. I think this is probably Ok and nothing related to Mac H/W.

`student_dashboard_mysql  | 2022-05-05T19:47:32.895685Z 24 [Warning] InnoDB: Table mysql/innodb_index_stats has length mismatch in the column name table_name.  Please run mysql_upgrade`

docker-compose.yml
```diff
@@ -4,6 +4,7 @@ services:
 
   mysql:
     image: mysql:5.7
+    platform: linux/amd64
```
pushyamig, line 7: The Docker desktop [documentation](https://docs.docker.com/desktop/mac/apple-silicon/#known-issues) for Apple Chip had this under known issue and this is the possible fix.

docker-compose.yml
```diff
@@ -4,6 +4,7 @@ services:
```
pushyamig, line 4: @zqian  did you install `softwareupdate --install-rosetta` this as part of the docker desktop installation with new apple chip Mac? 

I did, but I wasn't sure if it was needed or not.

jonespm:
I can't test on ARM, I suppose it's fine. It says that allows it to run the emulated version on ARM. The also suggest that you can switch the the MariaDB (or MySQL 5.8) image.

Maybe we should see if either of those options are supported by MiDatabase now. MySQL 5.7 is going to be EOL next year.

jonespm:
So since we're going to be switching and testing on MySQL 8, I'd just change

mysql:5.7 to mysql:8

And it **should** be able to work with either platform without specifying.

zqian:
@jonespm I got the following error with the mysql:8 change:

```
zqian@m-mx67x1h656 my-learning-analytics % docker-compose up
Creating network "my-learning-analytics_default" with the default driver
Pulling mysql (mysql:8)...
8: Pulling from library/mysql
ERROR: no matching manifest for linux/arm64/v8 in the manifest list entries
```

jonespm:
Looks like only the `mysql:8-oracle` supports ARM64. I'm not sure the difference between this and MySQL. Looks like this is based on oraclelinux instead of ubuntu. Really it wouldn't matter for local development. I don't know what RDS uses. 

It also looks like

`mysql/mysql-server:8` also supports ARM.

zqian:
`image: mysql/mysql-server:8.0 ` works on Mac M1

pushyamig:
I ran it locally. Build and starting of the app is successful ( I did not do the actual launch). I was able to download the data from UDW and was able to connect from DBeaver with Mysql8 driver. The fix is successful to me

pushyamig:
I will leave the approval to @jonespm

docker-compose.yml
```diff
@@ -5,13 +5,15 @@ services:
   mysql:
     image: mysql/mysql-server:8.0
     restart: on-failure
-    command: ['--character-set-server=utf8mb4', '--collation-server=utf8mb4_unicode_ci']
+    command: ['--character-set-server=utf8mb4', '--collation-server=utf8mb4_unicode_ci', '--socket=/tmp/mysql.sock']
```
pushyamig, line 8: curious to know what is the purpose of adding the `socket` attribute.

jonespm: When I tried to start it up I got an error of 

```
2022-05-17T17:47:03.730133Z 0 [System] [MY-013602] [Server] Channel mysql_main configured to support TLS. Encrypted connections are now supported for this channel.
2022-05-17T17:47:03.745529Z 0 [ERROR] [MY-010270] [Server] Can't start server : Bind on unix socket: Operation not permitted
2022-05-17T17:47:03.745624Z 0 [ERROR] [MY-010258] [Server] Do you already have another mysqld server running on socket: /var/lib/mysql/mysql.sock ?
2022-05-17T17:47:03.746023Z 0 [ERROR] [MY-010119] [Server] Aborting
```

This information seemed like what was needed. I wasn't running on Windows though so not sure. https://dev.mysql.com/doc/refman/8.0/en/deploy-mysql-nonlinux-docker.html

pushyamig: Thanks for sharing additional details!

I did not see that error when I ran it locally on Mac with Apple chip. This should now support the Non-linux env

jonespm: Yeah, it might just be a problem with other platforms and not with M1. I'm really not sure but works fine with that there, and sounds like it would be necessary for running on Windows.

zqian:
It also works for me on the M1 environment.

PR #1383, opened 2022-06-10
Fixes #1378 Resources Accessed does not load when the course contains files with a semicolon in the filename

### LINKED ISSUE #1378: Resources Accessed does not load when the course contains files with a semicolon in the filename
The Resources Accessed should display resources in the course that have non-alphanumeric characters in the file name.

When a file name has a semicolon in it, the Resources Accessed bar chart does not load.

### Steps to Reproduce :
1.  Add MyLA to a course that has at least one file with a semicolon in the file name.
2. Click on Resources Accessed
3. The bar chart will not load.

MyLA_RA_Error_Loading_Chart

zqian:
#1378 combined resource id with resource name as index.

We can concatenate resource id with resource type, that would be unique identifier and can be used as data frame index.

### DIFF
--- a/dashboard/views.py
+++ b/dashboard/views.py
@@ -373,11 +373,11 @@ def resource_access_within_week(request, course_id=0):
                     CONCAT(r.resource_id, r.resource_type) as resource_id_type,
                     r.name as name,
                     count(*) as self_access_count,
-                    max(a.access_time) as self_access_last_time 
-                    from resource_access a, user u, resource r 
-                    where a.user_id = u.user_id 
-                    and a.resource_id = r.resource_id 
-                    and u.sis_name=%(current_user)s 
+                    max(a.access_time) as self_access_last_time
+                    from resource_access a, user u, resource r
+                    where a.user_id = u.user_id
+                    and a.resource_id = r.resource_id
+                    and u.sis_name=%(current_user)s
                     and a.course_id = %(course_id)s
                     and a.course_id = u.course_id
                     group by r.resource_id, r.resource_type, r.name"""
--- a/dashboard/views.py
+++ b/dashboard/views.py
@@ -370,6 +370,7 @@ def resource_access_within_week(request, course_id=0):
                     select
                     r.resource_id as resource_id,
                     r.resource_type as resource_type,
+                    CONCAT(r.resource_id, r.resource_type) as resource_id_type,
                     r.name as name,
                     count(*) as self_access_count,
                     max(a.access_time) as self_access_last_time 
@@ -385,7 +386,6 @@ def resource_access_within_week(request, course_id=0):
 
     selfDf= pd.read_sql(selfSqlString, conn, params={"current_user":current_user, "course_id": course_id})
 
-    selfDf['resource_id_type'] = selfDf['resource_id'].astype(str).str.cat(selfDf['resource_type'], sep='')
     selfDf.drop(columns=['resource_type'], inplace=True)
 
     output_df = output_df.join(selfDf.set_index('resource_id_type'), on=['resource_id_type'], how='left')
--- a/dashboard/views.py
+++ b/dashboard/views.py
@@ -325,7 +325,6 @@ def resource_access_within_week(request, course_id=0):
             "enrollment_type": 'StudentEnrollment'
         })
     logger.debug(df)
-
     # return if there is no data during this interval
     if (df.empty):
         return HttpResponse("{}")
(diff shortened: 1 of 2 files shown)


jonespm:
We talked on a Zoom. There is still possibility that the resource_id might not be unique across multiple types of services. Currently they are unique but seems better to at try to either keep the id+type as part of the id or use the local id to ensure it's unique.

jonespm:
I've tested this out and the view loads as expected. 

On the course 534433 ther are 42025 rows and it takes almost 10 seconds to load. This is likely a different issue and one we may want to consider as we are thinking about expanding the usage of this tool. IMHO this is kind of long to wait for this page to load and will be worse as usage increases. 

We're probably going to want to more some for the in this view into the cron and do some pre-calculations.

jonespm:
Might be slightly cleaner to do the ID CONCAT in the SQL rather than in Pandas. Maybe better for the future?

```
diff --git a/dashboard/views.py b/dashboard/views.py
index bc42ee9..196d3f3 100644
--- a/dashboard/views.py
+++ b/dashboard/views.py
@@ -301,6 +301,7 @@ def resource_access_within_week(request, course_id=0):

     sqlString = f"""SELECT a.resource_id as resource_id,
                     r.resource_type as resource_type,
+                    CONCAT(r.resource_id, r.resource_type) as resource_id_type,
                     r.name as name,
                     u.current_grade as current_grade,
                     a.user_id as user_id
@@ -329,8 +330,6 @@ def resource_access_within_week(request, course_id=0):
     if (df.empty):
         return HttpResponse("{}")

-    # group by resource_id, and resource_type
-    df['resource_id_type'] = df['resource_id'].astype(str).str.cat(df['resource_type'], sep='')
     df.set_index(['resource_id_type'])
     # drop resource records when the resource has been accessed multiple times by one user
     df.drop_duplicates(inplace=True)
@@ -370,7 +369,7 @@ def resource_access_within_week(request, course_id=0):
     selfSqlString = f"""
                     select
                     r.resource_id as resource_id,
-                    r.resource_type as resource_type,
+                    CONCAT(r.resource_id, r.resource_type) as resource_id_type,
                     r.name as name,
                     count(*) as self_access_count,
                     max(a.access_time) as self_access_last_time
@@ -386,9 +385,6 @@ def resource_access_within_week(request, course_id=0):

     selfDf= pd.read_sql(selfSqlString, conn, params={"current_user":current_user, "course_id": course_id})

-    selfDf['resource_id_type'] = selfDf['resource_id'].astype(str).str.cat(selfDf['resource_type'], sep='')
-    selfDf.drop(columns=['resource_type'], inplace=True)
-
     output_df = output_df.join(selfDf.set_index('resource_id_type'), on=['resource_id_type'], how='left')
     output_df["total_percent"] = output_df.apply(lambda row: row[GRADE_A] + row[GRADE_B] + row[GRADE_C] + row[GRADE_LOW] + row.NO_GRADE, axis=1)
```

zqian:
@jonespm I have updated the code with the suggestion "do the ID CONCAT in the SQL rather than in Pandas"

jonespm (approved):

dashboard/views.py
```diff
@@ -321,60 +326,69 @@ def resource_access_within_week(request, course_id=0):
...
+        output_df.at[row['resource_id_type'], 'r_id'] = row['resource_id']
     output_df.reset_index(inplace=True)
 
     # now insert person's own viewing records: what resources the user has viewed, and the last access timestamp
-    selfSqlString = f"""select CONCAT(r.resource_id, ';', r.name) as resource_id_name, count(*) as self_access_count, max(a.access_time) as self_access_last_time 
+    selfSqlString = f"""
+                    select
+                    r.resource_id as resource_id,
+                    r.resource_type as resource_type,
+                    r.name as name,
```
jonespm, line 373: Could also do the same id CONCAT here too instead of Pandas. Then wouldn't need to drop the resource_type later.

zqian: done!

dashboard/views.py
```diff
@@ -321,60 +326,69 @@ def resource_access_within_week(request, course_id=0):
...
                     and a.course_id = u.course_id
-                    group by CONCAT(r.resource_id, ';', r.name)"""
+                    group by r.resource_id, r.resource_type, r.name"""
     logger.debug(selfSqlString)
     logger.debug("current_user=" + current_user)
 
     selfDf= pd.read_sql(selfSqlString, conn, params={"current_user":current_user, "course_id": course_id})
 
-    output_df = output_df.join(selfDf.set_index('resource_id_name'), on='resource_id_name', how='left')
+    selfDf.drop(columns=['resource_type'], inplace=True)
```
jonespm, line 389: You could just exclude this from the query (On line 372) rather than dropping it in the dataframe. You aren't using resource_type anywhere else, only the reference to r.resource_type on 383.

Issue #1223, opened 2021-01-28
update comments for LTI config


pushyamig:
For Canvas-hosted instance the URLs below can be domain-specific or generic. Canvas supports either. It's better to keep it generic to support vanity URL changes. 

See: https://canvas.instructure.com/doc/api/file.lti_dev_key_config.html

```json
"auth_login_url": "https://canvas.test.instructure.com/api/lti/authorize_redirect",
"auth_token_url": "https://canvas.test.instructure.com/login/oauth2/token",
"key_set_url": "https://canvas.test.instructure.com/api/lti/security/jwks",
```

This issue is to update the comments in `env_sample.json`.

PR #1385, opened 2022-06-17
issue_1365 update LTIError to lti_error

### LINKED ISSUE #1365: LTIError no longer defined in lti_new.py

I as a warning from Pylance that LTIError was not defined.  `"LTIError" is not defined`

It looks like the class was removed in #1105 but used again in #1101. Not too sure what happened with this

zqian:
fixes #1365

### DIFF
--- a/dashboard/lti_new.py
+++ b/dashboard/lti_new.py
@@ -57,7 +57,7 @@ def lti_error(error_message: Any) -> JsonResponse:
 class LTIException(Exception):
     def __init__(self, config,
                  message='An error occurred processing the LTI request.'):
-        self.lti_error = LTIError(config)
+        self.lti_error = lti_error(config)
         self.message = message
         super().__init__(self.message)
 


dashboard/lti_new.py
```diff
@@ -57,7 +57,7 @@ def lti_error(error_message: Any) -> JsonResponse:
 class LTIException(Exception):
```
zqian, line 57: LTIException is used only for generate_jwks() and get_jwks().

Since #1223 , MyLA config uses the generic canvas instance urls, and jwks-related instructions have been removed from MyLA wiki, I wonder whether we should remove the above two function and hence the LTIException definition after all.

jonespm: Right, these are only used when using the LTI services (like grading service) which we aren't currently using. So these could be just cleaned up.

I was also wondering if we could use [LTIServiceException](https://github.com/dmitry-viskov/pylti1.3/blob/7474685aecc35d528b944caf05b28ea66ac6d204/pylti1p3/exception.py#L16) from the library instead of our own exception here. But I wasn't sure the type of the **config** variable. 

This seems fine now, maybe create another issue to remove those jwks methods.

jonespm (approved):

PR #1391, opened 2022-07-08
issue-1332 added cron query to retrieve Canvas page usage events

### LINKED ISSUE #1332: Potential for duplicate resource identifiers (Pages visualization)
We received feedback from a pilot instructor that Canvas wiki page views would be more useful than file views for the Resources Accessed dashboard in their course. It makes sense given that the bulk of their content is in pages. I decided that I would add an additional query to `RESOURCE_ACCESS_CONFIG`. There's a slight issue with this given that Canvas Live Events lack the attributes to generate a link to the resource (and uses a different pattern than files). I was going to workaround this by passing an empty string from `views.py` for the URL (and look at a better solution in the future).

But as I was testing, I realized that the `resource.resource_id` is populated with whatever identifier is provided in the query of the source learning tool. Not ideal when loading resources from multiple source learning tools. I suppose the odds of an identifier collision are low when the source tools are likely using different identifier formats? But when we’re loading resources from the same source that use the same pattern for identifiers (e.g. bigint in the case of Canvas) it seems like we run the risk of an identifier collision? Which seems like an issue given [the constraint on the model](https://github.com/tl-its-umich-edu/my-learning-analytics/blob/11987be1bb7073c1d57cf7af3c76a028636facc7/dashboard/models.py#L318).

So I checked our data and it looks like this will be a problem:

```
select count(*)
from file_dim
join wiki_page_dim 
on file_dim.id = wiki_page_dim.id

1330848 rows
```

I was just wondering if this has come up in any previous discussions? I searched issues but didn't come across anything about this specifically. I was wondering if it would make sense to have a generated `resource.id` and a `resource.source_id`?

zqian:
Fixes #1332

### DIFF
--- a/config/env_sample.hjson
+++ b/config/env_sample.hjson
@@ -258,7 +258,9 @@
             "query":
                 '''
                 SELECT 'canvas_pages' AS resource_type,
-                REPLACE(JSON_EXTRACT_SCALAR(extensions_json, '$[\'com.instructure.canvas\'][request_url]'), 'https://example.instructure.com/courses', '') as resource_id,
+                CONCAT(CAST(REGEXP_EXTRACT(`group`.id, r'.*:(.*)') AS INT64) - @canvas_data_id_increment, 
+        				'/pages/', 
+        				CAST(JSON_EXTRACT_SCALAR(object.extensions, '$[\'com.instructure.canvas\'][entity_id]') as INT64) - @canvas_data_id_increment) as resource_id,
                 CAST(REGEXP_EXTRACT(membership.id, r'.*:(.*)') AS INT64) AS user_id,
                 cast(null as string) AS user_login_name,
                 CAST(REGEXP_EXTRACT(`group`.id, r'.*:(.*)') AS INT64) AS course_id,
@@ -277,7 +279,7 @@
             "app_display_name": "Canvas Files",
             "canvas_course_id_format": "LONG",
             "urls": {
-                "prefix": "https://example.instructure.com/courses",
+                "prefix": "https://example.instructure.com/courses/",
                 "postfix": ""
             }
         },
--- a/config/env_sample.hjson
+++ b/config/env_sample.hjson
@@ -218,6 +218,7 @@
     # resource type for resource view add or remove per institution needs
     "RESOURCE_VALUES": {
         "files":  {"types": ["canvas"], "icon": "fas fa-file fa-lg"},
+        "pages": {"types": ["canvas_pages"],"icon": "fas fa-file-code fa-lg"},
         "videos": {"types": [""], "icon": "fas fa-video fa-lg"}
     },
     # Institute specific urls add or remove to your needs, will be used for Resources view
@@ -252,7 +253,34 @@
                 and membership.id is not null 
                 and REGEXP_EXTRACT(`group`.id, r'.*:(.*)') IN UNNEST(@course_ids)
                 '''
-        }
+        },
+        "canvas_pages": {
+            "query":
+                '''
+                SELECT 'canvas_pages' AS resource_type,
+                REPLACE(JSON_EXTRACT_SCALAR(extensions_json, '$[\'com.instructure.canvas\'][request_url]'), 'https://example.instructure.com/courses', '') as resource_id,
+                CAST(REGEXP_EXTRACT(membership.id, r'.*:(.*)') AS INT64) AS user_id,
+                cast(null as string) AS user_login_name,
+                CAST(REGEXP_EXTRACT(`group`.id, r'.*:(.*)') AS INT64) AS course_id,
+                JSON_EXTRACT_SCALAR(object.extensions, '$[\'com.instructure.canvas\'][asset_name]') as name,
+                datetime(EVENT_TIME) as access_time
+                FROM event_store.expanded
+                where
+                JSON_EXTRACT_SCALAR(ed_app.json, '$.id') IN UNNEST(['http://m.example.instructure.com/', 'http://example.instructure.com/'])
+                and type = 'NavigationEvent'
+                and object.type = 'Page'
+                and action = 'NavigatedTo'
+                and membership.id is not null
+                and not CONTAINS_SUBSTR(JSON_EXTRACT_SCALAR(extensions_json, '$[\'com.instructure.canvas\'][request_url]'), '/api/v1/')
+                and REGEXP_EXTRACT(`group`.id, r'.*:(.*)') IN UNNEST(@course_ids)
+                ''',
+            "app_display_name": "Canvas Files",
+            "canvas_course_id_format": "LONG",
+            "urls": {
+                "prefix": "https://example.instructure.com/courses",
+                "postfix": ""
+            }
+        },
     },
     # Disable/Enable courses_enabled api
     "COURSES_ENABLED": false,
--- a/config/env_sample.hjson
+++ b/config/env_sample.hjson
@@ -88,7 +88,7 @@
     # Canvas Configuration
     "CANVAS_USER": "",
     # strings for construct file download url
-    "CANVAS_FILE_PREFIX": "https://example.instructure.com/files/",
+    "CANVAS_FILE_PREFIX": "https://umich.instructure.com/files/",
     "CANVAS_FILE_POSTFIX": "/download?download_frd=1",
     # Paths to Google BigQuery credential json file
     "GOOGLE_APPLICATION_CREDENTIALS": "/secrets/bq_cred.json",
@@ -228,7 +228,7 @@
         "canvas": {
             "app_display_name": "Canvas Files",
             "canvas_course_id_format": "LONG",
-            "urls": {"prefix": "https://example.instructure.com/files/", "postfix": "/download?download_frd=1"},
+            "urls": {"prefix": "https://umich.instructure.com/files/", "postfix": "/download?download_frd=1"},
             # BigQuery event queries
             "query": 
                 '''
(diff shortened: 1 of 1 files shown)


pushyamig:
I will review this today.

config/env_sample.hjson
```diff
@@ -252,7 +253,34 @@
                 and membership.id is not null 
                 and REGEXP_EXTRACT(`group`.id, r'.*:(.*)') IN UNNEST(@course_ids)
                 '''
-        }
+        },
+        "canvas_pages": {
+            "query":
+                '''
+                SELECT 'canvas_pages' AS resource_type,
+                REPLACE(JSON_EXTRACT_SCALAR(extensions_json, '$[\'com.instructure.canvas\'][request_url]'), 'https://example.instructure.com/courses', '') as resource_id,
```
pushyamig, line 261: I am not sure why the query chose an URL as the resource id. The id contains a long page name, which can be not palatable for eyes and when doing the join this could be problematic. example: `/1233/pages/podcast-nice-white-parents-episode-2-i-still-believe-in-it?module_item_id=12333`

Each page is associated with a page id and we should use that instead. For example

```
"object":{
      "id":"urn:instructure:canvas:wikiPage:17700000000111111",
      "type":"Page",
      "extensions":{
         "com.instructure.canvas":{
            "asset_name":"Podcast: Nice White Parents, Episode 2 \u201cI Still Believe in It\u201d ",
            "asset_type":"wiki_page",
            "entity_id":"17700000000111111",
            "context_account_id":"17700000000000111",
            "http_method":"GET"
         }
```

Is there any reason for having the resource id is page URL instead of page id?

pushyamig: I tested the approach of having a page_id instead of page_URL as the resource_id it will fit the cron design. But One exception, when sending the url in the response need to construct the URL with the course id, I believe that might require a change to cron.py 

`"resource_name":"https:\/\/\/example.instructure.com\/courses\/1111111\/pages\/1770000000022222|xyg page name|fas fa-file-code fa-lg"`

So based on the resource the URL formatting might change. There might be better way of fixing this, but this is what I can think of

config/env_sample.hjson
```diff
@@ -252,7 +253,34 @@
...
+                and object.type = 'Page'
+                and action = 'NavigatedTo'
+                and membership.id is not null
+                and not CONTAINS_SUBSTR(JSON_EXTRACT_SCALAR(extensions_json, '$[\'com.instructure.canvas\'][request_url]'), '/api/v1/')
+                and REGEXP_EXTRACT(`group`.id, r'.*:(.*)') IN UNNEST(@course_ids)
+                ''',
+            "app_display_name": "Canvas Files",
+            "canvas_course_id_format": "LONG",
+            "urls": {
+                "prefix": "https://example.instructure.com/courses",
```
pushyamig, line 280: So the url will be something like `https://example.instructure.com/courses/@courseID/pages/@page_id`

zqian: Right. I think the need to update the prefix url pattern with one more parameter could be problem:

We use the following code inside views.py. to construct resource link for all resource types (file, video, and page). It is hard to add one more param just for the page type.

`    output_df['resource_name'] = output_df.apply(
        lambda row:
            (RESOURCE_ACCESS_CONFIG.get(row.resource_type).get("urls").get("prefix") +
            str(row.r_id) +
            RESOURCE_ACCESS_CONFIG.get(row.resource_type).get("urls").get("postfix") +
            CANVAS_FILE_ID_NAME_SEPARATOR +
            str(row.r_name) + CANVAS_FILE_ID_NAME_SEPARATOR +
            RESOURCE_VALUES.get(RESOURCE_VALUES_MAP.get(row.resource_type)).get('icon')
            ),
        axis=1)`

config/env_sample.hjson
zqian: I like the idea of use page id, but I think it is hard to implement alongside with other resource types. See my comments below.

config/env_sample.hjson
pushyamig: I am ok, if collectively we all agree on this. My Further thoughts on this.

1. So the URL to view page ID https://example.instructure.com/courses/@courseID/pages/@page_id is something Canvas support from Ui if people know it. From the regular Canvas User standpoint, this URL doesn't exist. So there may be a risk that we are constructing this and canvas might discontinue( I don't feel that will happen)
2.  I think having a long URL ( with a descriptive page name) as Id is not ideal. But I really don't know if this is normal elsewhere. **But I want this tested with a very long page name, page name with non-ASCII literals ( like Japanese text) to make sure this approach work**.
3.  I feel in order to support page id as resource id, we might need to re-think about the existing cron design ( with pre-fix and post-fix)

pushyamig: Resource id type is `varchar(255)`.

config/env_sample.hjson
zqian: @pushyamig please check the latest query format. I have updated it based on your suggestion.

config/env_sample.hjson
zqian: Yes, your suggestion is valid. Please check the latest change. 

I have tested the query with courses that have unicode inside page titles. Here is the query that can help locate those courses: 

`
select wpd.title,
    CASE
      WHEN regexp_instr(title, '[^[:print:][:cntrl:]]') > 0 THEN TRUE
      ELSE FALSE END AS has_non_ascii_char,
cd."name",
cd.canvas_id 
from wiki_page_dim wpd, wiki_page_fact wpf, course_dim cd 
where has_non_ascii_char 
and wpf.wiki_page_id =wpd.id
and cd.id = wpf.parent_course_id 
and cd.name like '%SU 2022%'
order by wpf.parent_course_id desc
`

config/env_sample.hjson
```diff
@@ -252,7 +253,36 @@
...
                 '''
-        }
+        },
+        "canvas_pages": {
+            "query":
+                '''
+                SELECT 'canvas_pages' AS resource_type,
+                CONCAT(CAST(REGEXP_EXTRACT(`group`.id, r'.*:(.*)') AS INT64) - @canvas_data_id_increment, 
+        				'/pages/', 
+        				CAST(JSON_EXTRACT_SCALAR(object.extensions, '$[\'com.instructure.canvas\'][entity_id]') as INT64) - @canvas_data_id_increment) as resource_id,
```
pushyamig, line 263: this is a good idea of incorporating the course id along with page id as resource id. This solves the problem not touching the view.py code.

pushyamig:
I am going to test the new changes in the afternoon.

config/env_sample.hjson
pushyamig: The change seems to work now the resource id  `course-id/pages/page-id`. I was able to test with courses that has pages with non-ASCII character as well

pushyamig:
I finished my review, and the changes adding pages seem to working. I will let other review and approve it

config/env_sample.hjson
zqian: Great! Thanks for checking!

zqian: Great! I will mark this as resolved.

config/env_sample.hjson
jlongland: Just a quick comment that I'm delighted to see y'all working on this and wish I could have made the contribution myself!

config/env_sample.hjson
```diff
@@ -252,7 +253,36 @@
...
+                CAST(REGEXP_EXTRACT(membership.id, r'.*:(.*)') AS INT64) AS user_id,
+                cast(null as string) AS user_login_name,
+                CAST(REGEXP_EXTRACT(`group`.id, r'.*:(.*)') AS INT64) AS course_id,
+                JSON_EXTRACT_SCALAR(object.extensions, '$[\'com.instructure.canvas\'][asset_name]') as name,
+                datetime(EVENT_TIME) as access_time
+                FROM event_store.expanded
+                where
+                JSON_EXTRACT_SCALAR(ed_app.json, '$.id') IN UNNEST(['http://m.example.instructure.com/', 'http://example.instructure.com/'])
```
jonespm, line 271: Maybe Nitpick: We should have these example URL's match the other query on line 248 which are actually umich.edu ones. 

We also should consider a new issue to passing these in from a configuration parameter since they're used in 2 queries now. Then the query would be able to be copied and pasted without any changes. 

We also don't need the JSON_EXTRACT_SCALAR here anymore and can just change this to

`                ed_app.id IN UNNEST(['http://m.canvas.umich.edu/', 'http://umich.instructure.com/']) 
`

config/env_sample.hjson
```diff
@@ -252,7 +253,36 @@
...
+                and type = 'NavigationEvent'
+                and object.type = 'Page'
+                and action = 'NavigatedTo'
+                and membership.id is not null
+                and not CONTAINS_SUBSTR(JSON_EXTRACT_SCALAR(extensions_json, '$[\'com.instructure.canvas\'][request_url]'), '/api/v1/')
```
jonespm, line 276: It might be nice to have a comment as to why this exclusion was in here. What were these finding?

config/env_sample.hjson
```diff
@@ -252,7 +253,36 @@
                 and membership.id is not null 
                 and REGEXP_EXTRACT(`group`.id, r'.*:(.*)') IN UNNEST(@course_ids)
                 '''
-        }
+        },
+        "canvas_pages": {
+            "query":
+                '''
+                SELECT 'canvas_pages' AS resource_type,
+                CONCAT(CAST(REGEXP_EXTRACT(`group`.id, r'.*:(.*)') AS INT64) - @canvas_data_id_increment,
```
jonespm, line 261: Interesting solution! This query would fail if it can't do the CAST though, I'm not sure the likelihood of that. 

You *could* use this pattern in these queries to make it safer and always return 0. Perhaps we want the cron to error out though. Maybe something to keep in mind if we do see a failure here. 

```
IFNULL(SAFE_CAST(REGEXP_EXTRACT(`group`.id, r'.*:(.*)') AS INT64), @canvas_data_id_increment) - @canvas_data_id_increment, 
```

jonespm (approved):
The query looks good. I left a few comments that could improve the consistency or add some commentary as to why these were needed.

config/env_sample.hjson
zqian: comment added. See [8a39627](https://github.com/tl-its-umich-edu/my-learning-analytics/commit/8a39627b1fbba7a7489335c95de1d4487306ad9e)

config/env_sample.hjson
zqian: added IFNULL(SAFE_CAST()) pattern into query.

config/env_sample.hjson
zqian: Those url values are now umich-related. Also created #1397 to parametrize those values

config/env_sample.hjson
zqian: @jlongland you are very welcome!

Issue #764, opened 2019-10-18
Limit resources access view to show only events for enrolled students in the course


jennlove-um:
##Expected behavior:
Resource access events for enrolled students only should be included in the percentage of people who have viewed an item.  Also, when anyone with any other role in the course views the page, they should not see blue bars to indicate which files they've viewed.  

## Describe the bug:
Currently, with my login, I'm seeing files I viewed in the course showing up as blue in the visualization.  I am not a student in any courses, but I am a Teacher.  My file access should not be considered in the visualization since I am not a student.

## Steps to Reproduce :
1. As someone who is not a student go to a Canvas course that is in MyLA
2.  Access a file in the course
3. View the resources view to see that the file shows up as a blue bar instead of gray

## Additional context :
This visualization is about student behavior, so it should limit the display to include access events for only students who are enrolled in the course.

jpyan:
Hey I'm trying to look into this bug yet I can't seem to see any blue file from the admin side for any courses. For example, for course id 362855, it shows blue on student's view, but no blue in admin view. Is it already being fixed? If not, would you provide a more specific example to reproduce that bug? (Screenshots would be helpful too)

jennlove-um:
Thanks for asking the question - the issue isn't really with the admin view.  That was my first pass at understanding what was going on.  I'll update the description.  But I've also put some more detail below.

For this bug, the issue is that only users with an enrollment type of "StudentEnrollment" in the user table should have their resource access events get counted and included in the visualization.

Replicating this requires going into Canvas as someone with a different role on a course (TA, Teacher, Designer, etc.) and clicking on a file.  We don't want the clicks from those enrollment types to get counted in this visualization.  So if I'm a teacher, and I click on files then go to MyLA, I shouldn't see any bars show up as blue since my file access doesn't count.

jpyan:
Thanks for your reply. So with "requires going into Canvas as someone with a different role on a course (TA, Teacher, Designer, etc.) and clicking on a file", does that mean if I'm not a teacher nor some other role instead of students, I can't reproduce this bug?

jennlove-um:
You wouldn't necessarily have to go into Canvas.  But you do need to have a user in the user table that has an enrollment_type other than StudentEnrollment.  For example, TeacherEnrollment, TaEnrollment.  

You also have to have an entry in the resource access table for that same user.

[...]

PR #1395, opened 2022-08-18
#764 filter out non-student resource access events

zqian:
filter out non-student event by merging the resource access data frame with student enrollment data frame.


dashboard/cron.py
```diff
@@ -417,6 +417,15 @@ def update_resource_access(self):
...
                          f'{resource_access_df.dtypes}')
+            
+            # only keep access events generated by students
+            student_enrollment_df = pd.read_sql(
+                'select user_id, course_id '
+                f'from user where enrollment_type=\'StudentEnrollment\'',
+                engine)
+            resource_access_df = pd.merge(
+                resource_access_df, student_enrollment_df,
+                on=['user_id', 'course_id'], how='inner')
```
pushyamig, line 429: This change seems to work well with new courses added to MyLA DB. But for the old courses, it doesn't remove the TA or TeacherEnrollment info. I think that is by design we adopted long back not deleting all the live events data for saying cost. 

To me this approach work as long we still have the filter in the view.py only sending the student access in the response.

dashboard/cron.py
```diff
@@ -417,6 +417,15 @@ def update_resource_access(self):
             logger.debug(f'resource_access_df:\n'
                          f'{resource_access_df}\n'
                          f'{resource_access_df.dtypes}')
+            
+            # only keep access events generated by students
+            student_enrollment_df = pd.read_sql(
+                'select user_id, course_id '
+                f'from user where enrollment_type=\'StudentEnrollment\'',
```
pushyamig, line 424: Optional: `StudentEnrollment` could be constant rather than buried in the code

pushyamig:
This change seems to be working, I have left come comment. But I will leave the approval to @jonespm

dashboard/cron.py
zqian: We can do `update course set data_last_updated=null` to force full refresh of resource access events. It would lead to extra BigQuery cost, but the amount should be reasonable.

pushyamig: I agree with this approach if it doesn't cost much with the 8T/byte quota limit we have. This kind of saves @jennlove-um since she often tests from old test courses. 

or we could a manual delete of all the resources access info with Teacher and TA.

dashboard/cron.py
zqian: Done with [b492ae3](https://github.com/tl-its-umich-edu/my-learning-analytics/pull/1395/commits/b492ae39078458c83f7193f051141e17ddb98c36)

dashboard/cron.py
ssciolla: This seems like it would work. The `inner` approach is a bit less readable than say a `pandas` filter, but it's less code. A comment might be appropriate.

ssciolla (commented):
Looks okay. I'll do some testing after lunch.

dashboard/cron.py
zqian: added comment for inner merge

ssciolla: Yeah, I agree with @pushyamig that clearing out the resource access events would make sense, even if it adds cost. That may need to be in the release notes too.

zqian:
@ssciolla I've replaced "StudentEnrollment" string with User.EnrollmentType.STUDENT

dashboard/cron.py
```diff
@@ -417,7 +417,17 @@ def update_resource_access(self):
...
+            student_enrollment_type= User.EnrollmentType.STUDENT
+            student_enrollment_df = pd.read_sql(
+                'select user_id, course_id from user where enrollment_type= %s',
+                engine, params={student_enrollment_type})
+            resource_access_df = pd.merge(
+                resource_access_df, student_enrollment_df,
+                on=['user_id', 'course_id'],
+                # use inner merge to keep only resource access event (left)
+                # innitiated by people with student enrollment type (right)
+                how='inner')
```
ssciolla, line 430: Because I was a bozo and forgot to update my queries, I found an edge case that I already shared with @zqian where if the query results initially fed into`resource_access_df` only contain `-1` (and thus are converted via the earlier merge to string user IDs), then a `pandas` error is thrown like this:

```
You are trying to merge on object and int64 columns. If you wish to proceed you should use pd.concat
```

This seems to be because the two `user_id` fields in the merge here have types of `object` and `int64`. The easiest way to reproduce this is to only query for MiVideo resource access events (ensuring that some data will be returned for courses). I however didn't see this in the mixed case where MiVideo, Canvas, and leccap queries are in use; maybe `pandas` is smart enough to convert to `int64` if it seems some of them fit that description.

It may be worthwhile to add a line like this around 374 to convert the user ID column back to `int64` so this merge can handle this edge case -- so long as it doesn't create any other problems 🤣 

```
resource_access_df['user_id'] = pd.to_numeric(resource_access_df['user_id'])
```

ssciolla: Looks like we found that this mismatch of types actually does cause a substantial problem whenever MiVideo data is used (alone or in conjunction with other queries). The inner join seems to remove all MiVideo data, probably because the data type of the IDs is still a string. The above line seems to fix this.

dashboard/cron.py
```diff
@@ -417,7 +417,17 @@ def update_resource_access(self):
             logger.debug(f'resource_access_df:\n'
                          f'{resource_access_df}\n'
                          f'{resource_access_df.dtypes}')
-
+            # only keep access events generated by students
+            student_enrollment_type= User.EnrollmentType.STUDENT
```
ssciolla, line 421:
```suggestion
            student_enrollment_type = User.EnrollmentType.STUDENT
```

ssciolla (changes requested):

dashboard/cron.py
zqian: added the above line change.

dashboard/cron.py
zqian: added the suggested change.

ssciolla (approved):
I'm approving. A couple things to note:
1) This could use some more real world testing. The `cron` is delicate, and there are edge cases with all these different queries and different courses.
2) The inner join is also filtering out resource access events that aren't associated with a user enrollment at all. @zqian noted that these are likely events from students who dropped, or maybe others who aren't in the enrollment table. I see this as a positive side effect, since there was no reason to store these events if they weren't being used.

zqian:
Jennifer found out this PR removed all MiVideo access data, since MiVideo Caliper events has value "-1" in user_id field for all users, thus the dataframe merge based on user_id and course_id won't keep any MiVideo data. 

We decided to revert this PR, and leave this issue open for future release.

PR #1401, opened 2022-08-25
Fixes #1396 - Check if deployment ID claim is missing and return an error even if validation is disabled

### LINKED ISSUE #1396: Check if Deployment ID claim is missing and return an error even if validation if disabled
We failed the certification because we allowed missing deployment ids. We could probably turn this on and validate it but if it's missing completely this should have an error and fail. 

It looks like from the spec that this is required to be present but it doesn't necessarily need to be validated. https://www.imsglobal.org/spec/lti/v1p3#lti-deployment-id-claim

https://github.com/tl-its-umich-edu/my-learning-analytics/blob/3d0a1ccb71e6461775ac620b642e8ecc5047fd03/dashboard/lti_new.py#L41

jonespm:


ssciolla (approved):

pushyamig:
@jonespm, So the LTI launch variable are sending 'https://purl.imsglobal.org/spec/lti/claim/deployment_id = ''` 

now with the check _get_deployment_id() it will throw a `deployment_id is not set in jwt body` is that what is expected and regardless what the chosen value for LTI_CONFIG_DISABLE_DEPLOYMENT_ID_VALIDATION?

PR #1403, opened 2022-08-25
Fixes #1402 - Add support for configuring X_FORWARDED_HOST.

### LINKED ISSUE #1402: Loophole does not work with Django Admin and returns incorrect URLs in /lti/config

If a local developer is using [loophole.site](https://loophole.cloud/) instead of ngrok for LTI testing it doesn't work correctly. This is because loophole sends the header USE_X_FORWARDED_HOST and MyLA currently doesn't use this.

### Steps to Reproduce :
1. Use Loophole in development intead of ngrok

You can launch loophole with a command like 
`loophole http 5001 --hostname myla-test-domain`

After you go through the steps to install it and add it to your path.

2. Go through the steps to setup LTI. You'll notice that the /lti/config URL in the steps has 127.0.0.1. 
3. Also saving anything in Django Admin doesn't work. 

Everything should just work the same was as with NGrok.

It's possible to configure this you may also have to add **loophole.site** to `ALLOWED_HOSTS, DEFAULT_SRC and CSRF_TRUSTED_ORIGINS.`

jonespm:
This is Needed for some proxies.

I left it at the default of False, though there should be no risk to having this True. Some proxies won't work correctly without this.


config/env_sample.hjson
```diff
@@ -54,6 +54,8 @@
     # Enable secure cookies, also set your trusted origin (example of instructure.com)
     "CSRF_COOKIE_SECURE": false,
     "CSRF_TRUSTED_ORIGINS": ["instructure.com"],
+    # If you have a proxy that sets this header then set this to true. Default is false
+    "USE_X_FORWARDED_HOST": false,
```
pushyamig, line 58: One weird thing I noticed is with or without this option, the new Assignment planning ( with GraphQL) view doesn't launch. it will say some error has occurred. I revered back to using ngrok

jonespm: Thanks, I didn't test this with GraphQL, I'll try it out and see if there's an easy fix for that.

jonespm: Do you remember what error you saw? Was it in the browser console? I'm not seeing any errors with the new assignment planning. I was testing with our test course 562259 and just populated the data. 

image

pushyamig: This is what I see 

```
CSRF verification failed. Request aborted.`
Referer checking failed - https://pu-myla.loophole.site/courses/501132/assignments does not match any trusted origins.`
```

In my .env.hjson  i have these options
```

 "CSRF_COOKIE_SECURE": true,
"SESSION_COOKIE_SAMESITE": "None",
   "CSRF_COOKIE_SAMESITE": "None",
"ENABLE_BACKEND_LOGIN": false,
 "ALLOWED_HOSTS": [
        "127.0.0.1",
        "localhost",
        ".ngrok.io",
        ".loophole.site"
    ],
```

This is my loophole version `loophole version 1.0.0-beta.15 (5cecf33)`

jonespm: Thanks, I'll check my config. What browser are you using?

jonespm: You might have to include loophole in your DEFAULT_SRC. That's the only difference I see. 

`        "DEFAULT_SRC": ["'self'","example.edu","ngrok.io","loophole.site"],
`

This one also *might* be different? What instance of Canvas were you testing on? I was testing on test. 

`    "CSRF_TRUSTED_ORIGINS": ["instructure.com", "umich.beta.instructure.com", "umich.edu"],
`

I don't have either of these set
```
"SESSION_COOKIE_SAMESITE": "None",
   "CSRF_COOKIE_SAMESITE": "None",
```

But it looks like not setting them sets them also to "None" by default.

pushyamig: Chrome is what I am using.

Let me add other values as suggested and see.

pushyamig: I am testing from canvas testa and I added the property values as per your suggestion. I don't see any difference testing from chrome/FF

jonespm: @pushyamig and I were testing this and we found it would only work in her configuration if we added ".loophole.site" to the DEFAULT_SRC and CSRF_TRUSTED_ORIGINS.

Are either of @ssciolla or @zqian able to test this? Wondering if we should include this in the default config or just on the wiki page about ngrok. 

I was able to get it to work without putting anything extra in CSRF_TRUSTED_ORIGINS and we're not sure what's different. 

It looks like this is [going to change for Django 4](https://docs.djangoproject.com/en/4.1/ref/settings/#csrf-trusted-origins) and it's doing more strict checking.

ssciolla: I'm testing with `loophole version 1.0.0-beta.15 (5cecf33)` and Chrome. I don't have `.loophole.site` in `ALLOWED_HOSTS` or `DEFAULT_SRC`, only `CSRF_TRUSTED_ORIGINS` for `graphql`. I'm able to get it to work without `X_FORWARDED_HOST`.

zqian:
@jonespm could you please provide a step-by-step test plan for this PR? Then I can follow and do the verification. Thanks!

config/env_sample.hjson
jonespm: I think I initially saw this error in the `/lti/config/` path off loophole (it had localhost). Also some areas of Django Admin seemed to have localhost. 

So you don't even have to set this at all with loophole? I guess I can try it again.

jonespm:
Yeah, I can update and try it again later. Perhaps it's not entirely clear when/if this is needed.

config/env_sample.hjson
ssciolla: How are you running `loophole`? I think I did `loophole http 5001 --hostname ssci-myla` and added `ssci-myla.loophole.site` to the relevant places in my LTI config. I'm not sure I tried the `/lti/config` route, but I don't know why that have different behavior.

jonespm: Right, just like that just with a different address. . And when I went to configure LTI and used the URL 

https://jonespm-localhost-myla.loophole.site/lti/config/

Perhaps that's minor, and somethings going on in the library to try to figure out the domain name. It has

```           
      "domain": "127.0.0.1:5001",
      "target_link_uri": "https://127.0.0.1:5001/lti/launch/"
```

and others without this setting when I went through the LTI steps which caused me some time troubleshooting.

Then when I fixed it and launched it, I added an admin user but when I went to any view in admin and tried to save or add anything I got a
```
    CSRF verification failed. Request aborted.
    Referer checking failed - https://jonespm-localhost-myla.loophole.site/admin/dashboard/course/17700000099999999/change/ does not match any trusted origins.
```

similar to Pushyami. This setting seemed to fix it completely for me.

ssciolla (approved):
I don't see any issue with making this configurable, and there does seem to be some need for it for `/lti/config/` with `loophole`. We should document this somewhere.

jonespm:
I'll merge this as there may be need for it outside of loophole. I'd like see if I can get the lti config to work without this though, but might not spend much time with it. It seems like if this is value set to true it works in most cases, ngrok, loophole or otherwise.

PR #1405, opened 2022-08-25
Fixes #1094 - STUDENT_DASHBOARD_LTI should be renamed

### LINKED ISSUE #1094: STUDENT_DASHBOARD_LTI should be renamed
The config value "`STUDENT_DASHBOARD_LTI`" should really be something like

`ENABLE_LTI `

Since the project is no longer student dashboard. We also have STUDENT_DASHBOARD_SAML but that should be deprecated.

You'd want both to work so use the STUDENT_DASHBOARD_LTI as the default value if this isn't set for backward compatibility. 

`ENV.get('ENABLE_LTI', ENV.get('STUDENT_DASHBOARD_LTI', False))
`
https://github.com/tl-its-umich-edu/my-learning-analytics/blob/a3f624fc868af2c2cfb1e5a62735506655865779/config/env_sample.json#L94

jonespm:
This is probably just the minimal we can do, getting dashboard out of the config. Almost everything else is in the database or code so harder to remove and maybe not worth it.


jonespm:
Thanks! There were 3 other places where this was referenced. I've updated these.

jonespm:
Maybe minor but do we think renaming to `ENABLE_LTI` is fine? I remember that enable/disable were no longer preferred. Maybe `ACTIVATE_LTI`?

zqian (approved):

config/env_sample.hjson
```diff
@@ -94,7 +94,7 @@
     "GOOGLE_APPLICATION_CREDENTIALS": "/secrets/bq_cred.json",
     # LTI Configuration
     # LTI is disabled by default
```
ssciolla, line 96: Could change this comment to "off by default" or "deactivated by default"

ssciolla (approved):
Seems fine. Didn't run it but don't see any issues. We may want to check if anything in the Wiki needs to be updated related to this.

config/env_sample.hjson
jonespm: Nooo! It still snuck in there. Or can change it to "not enabled" to be consistent with the name.

jonespm: So it looks like there's two other properties that are "DISABLED": `LTI_CONFIG_DISABLE_DEPLOY_ID_VALIDATION `and `VIEWS_DISABLED`, along with other comments.

I think this would have to be a bigger refactor if we wanted to change that.

pushyamig:
I used the old key STUDENT_DASHBOARD_LTI = true and I can see the test passing
I used the new key Enable_LTI=true and test passes.

Test Passes

PR #1410, opened 2022-08-26
Fixes #1399 - Add Django Admin function to clear course updated dates.

### LINKED ISSUE #1399: Add button to Django Admin to all admin to refresh all MyLA data
MyLA admins should have the ability to refresh the data in the database as a bulk refresh instead of incremental.  Add a button to the Django Admin course page to reset the course date to null. This will make the the cron pulls down all the data and refreshes it instead of just an incremental refresh of events and new course data.

jonespm:
I also added the date column to the list view.

I had a idea for adding it as a button but that required a custom template and more changes and this allows you to potentially select only specific courses so seems better. Less code and more functionality.


dashboard/admin.py
```diff
@@ -70,9 +70,15 @@ def has_add_permission(self, request):
 class CourseAdmin(admin.ModelAdmin):
     inlines = [CourseViewOptionInline, ]
     form = CourseForm
-    list_display = ('canvas_id', 'name', 'term', 'show_grade_counts', 'course_link', '_courseviewoption', 'show_grade_type')
```
ssciolla, line 73: We just removing `show_grade_type` because it was never implemented?

dashboard/models.py
```diff
@@ -205,7 +205,7 @@ class Course(models.Model):
     GRADING_CHOICES = [('Percent', 'Percent'), ('Point', 'Point'), ]
     show_grade_type = models.CharField(verbose_name="Show Grade Type", max_length=255,
                                          choices=GRADING_CHOICES, default='Percent')
-    data_last_updated = models.DateTimeField(verbose_name="Time data for this course was last updated", null=True, blank=True)
+    data_last_updated = models.DateTimeField(verbose_name="When this course was last updated", null=True, blank=True)
```
ssciolla, line 208: Maybe this doesn't need a verbose name, and we can just use `data_last_updated`? I think it would be nice if we referred to this field consistently everywhere. Above in `admin.py` you use "course updated date". If you want to include more info, we could add `help_text="When associated resource data was last updated for this course"`. That language also seems more accurate, since I think the course model data is always updated if it changes.

These model changes may also need a migration, though I still don't understand why Django requires these when they don't result in database changes.

dashboard/admin.py
zqian: Since we still have the 'show_grade_type' column in course table, I suggest we still show this value.

jonespm: Right, I switched `show_grade_type` to `data_last_updated` based on a suggestion from the meeting. `show_grade_type` will always be the same and currently doesn't do anything.

I can re-add it or not make any change here but the table has a lot of columns and we should only show the most necessary information here on the table.

dashboard/models.py
ssciolla: I'm saying we need a consistent way to refer to this field. If you use that long verbose name that doesn't match what is used in the admin action you added, it may not be clear to the user what that admin action does. Thus, I'm saying use "data last updated" everywhere, and then add help text to explain to the user what it is.

dashboard/admin.py
zqian: Given the concern of the table space, I am fine either way of adding back the show_grade_type or not.

dashboard/models.py
jonespm: Are you talking about consistency between this and 

`@admin.action(description='Clear selected Courses updated dates')`

I'm not sure how specific we need to be to an admin using the UI. This was the date that this course was last updated by the **cron** for all data, not just resource data.

dashboard/admin.py
```diff
@@ -70,9 +70,15 @@ def has_add_permission(self, request):
...
+    list_display = ('canvas_id', 'name', 'term', 'show_grade_counts', 'course_link', '_courseviewoption', 'data_last_updated')
     list_select_related = True
     readonly_fields = ('term', 'data_last_updated',)
+    actions = ['clear_course_updated_dates']
+
+    @admin.action(description='Clear selected Courses updated dates')
+    def clear_course_updated_dates(self, request, queryset):
```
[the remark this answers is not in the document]
jonespm: Okay, this sounds like a separate issue too as none of our other fields have this. We can set it as a one off or do it on a new issue to have a better display for None/Null for all the tables or specific fields. Other fields like Term can also be null and would probably want a different value if it was empty.

dashboard/models.py
ssciolla: Yes, consistency between these. You refer to this field three ways:
- `verbose_name="When this course was last updated"` (which shows up as a column header)
- `@admin.action(description='Clear selected Courses updated dates')` (which shows up in the dropdown)
- `self.message_user(request, "All selected course updated dates are cleared."` (which shows up in the success message).

Maybe these are close enough to where it doesn't matter, but if this was facing all users, I'd want it to be more standardized. Just being consistent would help people know how they are all related. If a short field name like "Data last updated" isn't clear enough, that's where help text may be beneficial.

I think you are correct, it's not just resource data. But it's not just course data either, it's data related to the course too. So "courses updated dates" or "course updated dates" isn't exactly the full picture.

dashboard/admin.py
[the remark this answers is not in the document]
zqian: I am OK with tracking this as a separate issue. Thanks!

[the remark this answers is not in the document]
zqian: I have filed this as new issue: #1415 for future consideration. Closing the comments here.

zqian (approved):
The removal of date_last_updated values works. 

Approve the PR now, with the pending discussion of verbose_name.

dashboard/models.py
jonespm: Thanks, I'll review these suggestions and try to make these more consistent. The only one I'm really adding is the `admin.action` here but the name is more exposed now if we add it to the table so might as well. I think we could create an issue to review some of these descriptions, maybe with @jennlove-um and see if any could be improved or could use additional `help_text`.

ssciolla (approved):
I didn't see any functional issues when I tested this, so I'll approve, you can merge it when you think the naming thing has been resolved (now or after some changes).

dashboard/models.py
jonespm: Okay, I'll go with your suggestion for the verbose_name and build from there

```
verbose_name="Data last updated" (which shows up as a column header)

@admin.action(description='Clear selected last updated values') (which shows up in the dropdown)

self.message_user(request, "Selected last updated values cleared." (which shows up in the success message).
```

ssciolla: You may not need the `verbose_name` value at all, it may default to that based on the field name.

jonespm: Right, It looks like the default name takes from the variable name. So right it probably isn't necessary if we just want data_last_updated -> "Data last updated"

jonespm: It looks like we have redundant verbose names for many fields already even in this table like "Term", "Name", "Show Grade Type". I think I'll just leave it explicit and consistent rather than remove. I didn't know that it expanded the field name as the verbose_name.

jonespm: I added the migration you mentioned that seemed to be needed for just this `verbose_name` changing.

Issue #1344, opened 2021-10-22
Avoid pulling some unused personal data from the cron


jonespm:
Thank you for contributing to this project!

### Describe your problem or feature you'd like added

Since we've switched over to LTI, Canvas provides some data that we were previously pulling from the UDW that's slightly different pulling from the UDP and unnecessary now. This includes things like name and email address. But it might also include UMID as person_sourcedid. 

Some of this was retained because of SAML but SAML was removed in the 2021.03.01 release. We should see what really needs to be pulled from the database and clean up unused columns and queries. 

It's good practice to also only store the data completely necessary to run the application.

jennlove-um:
Testing passes in beta

jonespm:
This is the rollback if needed for this issue

```
ALTER TABLE `user` ADD name varchar(255) NULL after user_id;
DELETE FROM django_migrations WHERE name='0024_remove_user_name';
```

jennlove-um:
Testing passes in test.

PR #1416, opened 2022-08-31
#1344 use user name value from LTI launch data

zqian:
#1344  omit the user name retrieval from cron job

Test Plan:

1. make sure the dashboard/migrations/0024_remove_user_name.py migration file is applied
2. run cron job; verify the sis_name field in User table contains only null value
4. launch the tool from Canvas; notice the User.sis_name field is updated upon LTI launch for this user; 
5. verify user can use the MyLA tool for this course
6. verify user can click on the top right profile button and can see user name in "Sign in as FULL NAME"


dashboard/lti_new.py
```diff
@@ -35,6 +38,22 @@
...
+else:
+    # else try loading settings from the json config file
+    try:
+        with open(os.getenv("ENV_FILE", "/secrets/env.hjson")) as f:
+            ENV = hjson.load(f)
+    except FileNotFoundError as fnfe:
+        print("Default config file or one defined in environment variable ENV_FILE not found. This is normal for the build, should define for operation")
+        # Set ENV so collectstatic will still run in the build
+        ENV = os.environ
+CANVAS_DATA_ID_INCREMENT = ENV.get("CANVAS_DATA_ID_INCREMENT", 17700000000000000)
```
ssciolla, line 55: I think we would want to import this from settings instead. I.e. `from django.conf import settings` then `settings.CANVAS_DATA_ID_INCREMENT`.

ssciolla: Then you don't have to double process the config file.

dashboard/lti_new.py
```diff
@@ -228,12 +246,24 @@ def extract_launch_variables_for_tool_use(request, message_launch):
...
+
+    # Add username into the MyLA User table, since the data was not pulled in from cron job
+    user_id = CANVAS_DATA_ID_INCREMENT + int(canvas_user_id)
+    try:
+        myla_user_obj = MyLAUser.objects.get(user_id=user_id)
+        # update
+        myla_user_obj.sis_name = username
+        myla_user_obj.save()
+    except MyLAUser.DoesNotExist:
+        logger.warn(f'Cannot find MyLA user {username}')
```
ssciolla, line 265: Is there a functional need for `sis_name`?

zqian: yes, it is used in db_util.get_user_courses_info(), line 119, to determine which courses the user is member of

ssciolla: I guess it's used a bunch of places. In an ideal world, we'd refactor this a bit so `auth_user` stores this data and `User` becomes `Enrollment`. That's already been discussed in the past. I guess this may work in the interim.

config/cron.hjson
```diff
@@ -9,14 +9,14 @@
...
-        user_info as (select p.unique_name,p.sis_user_id, u.name, u.id as user_id, u.global_canvas_id
+        user_info as (select p.unique_name,p.sis_user_id, '' as name, u.id as user_id, u.global_canvas_id
                     from (SELECT ROW_NUMBER() OVER (PARTITION BY user_id order by sis_user_id asc) AS row_number, * FROM pseudonym_dim) as p
                     join user_dim u on u.id = p.user_id WHERE row_number = 1),
         user_enroll as (select u.unique_name, u.sis_user_id, u.name, u.user_id, e.enroll_id,
                         u.global_canvas_id, e.type from enroll_data e join user_info u on e.user_id= u.user_id),
         course_fact as (select enrollment_id, current_score, final_score from course_score_fact
                         where course_id='{course_id}'),
-        final as (select cast(u.global_canvas_id as BIGINT) as user_id,u.name, u.unique_name as sis_name,
+        final as (select cast(u.global_canvas_id as BIGINT) as user_id,u.name, '' as sis_name,
```
ssciolla, line 19: Can we omit the name stuff completely rather than using empty strings?

dashboard/lti_new.py
zqian: Great! Code updated.

config/cron.hjson
ssciolla: Maybe they'd have to be nullable. Hm...

dashboard/lti_new.py
zqian: True. The User table name is not ideal. We can do the table renaming in the future releases. We may not need the User.name field either. So that can be removed. 

I agree this is an interim solution to get the UDP query switch happen.

config/cron.hjson
zqian: Looks like the name and sis_name fields are indeed nullable. 

I have dropped those two fields from the cron queries.

dashboard/lti_new.py
```diff
@@ -234,6 +235,18 @@ def extract_launch_variables_for_tool_use(request, message_launch):
...
+    # Add username into the MyLA User table, since the data was not pulled in from cron job
+    user_id = CANVAS_DATA_ID_INCREMENT + int(canvas_user_id)
+    try:
+        myla_user_obj = MyLAUser.objects.get(user_id=user_id)
+        # update
+        myla_user_obj.name = full_name
+        myla_user_obj.sis_name = username
+        myla_user_obj.save()
+    except MyLAUser.DoesNotExist:
+        logger.warn(f'Cannot find MyLA user {username}')
```
ssciolla, line 248: I need to test still, but I think this will fail if you have a user with multiple enrollments, maybe with a `MultipleObjectsReturned` exception?

We probably need to use `.update(...)`: https://docs.djangoproject.com/en/4.1/ref/models/querysets/#update

ssciolla: Confirmed:
```
[ERROR] [lti_new.py:56] LTI error: get() returned more than one User -- it returned 2!
```

dashboard/lti_new.py
```diff
@@ -234,6 +235,18 @@ def extract_launch_variables_for_tool_use(request, message_launch):
         password = ''.join(random.sample(string.ascii_letters, settings.RANDOM_PASSWORD_DEFAULT_LENGTH))
         user_obj = User.objects.create_user(username=username, email=email, password=password, first_name=first_name,
                                             last_name=last_name)
+
+    # Add username into the MyLA User table, since the data was not pulled in from cron job
+    user_id = CANVAS_DATA_ID_INCREMENT + int(canvas_user_id)
+    try:
+        myla_user_obj = MyLAUser.objects.get(user_id=user_id)
+        # update
+        myla_user_obj.name = full_name
```
ssciolla, line 244: I did a quick search, I don't think we're using `MyLAUser.full_name`. Thoughts on dropping this field now?

zqian: Dropped the field.

dashboard/lti_new.py
zqian: addressed the problem in [6ad1ed2](https://github.com/tl-its-umich-edu/my-learning-analytics/commit/6ad1ed26ddc4270862b2b8e8d666bc237dd7b7c2)

dashboard/lti_new.py
```diff
@@ -7,6 +7,7 @@
 import django.contrib.auth
 from django.conf import settings
 from django.contrib.auth.models import User
+from dashboard.models import User as MyLAUser
```
ssciolla, line 10: Looks like there's already a `MylaUser` alias. Can you use that one?

dashboard/lti_new.py
```diff
@@ -34,6 +35,7 @@
 TA = 'http://purl.imsglobal.org/vocab/lis/v2/membership/Instructor#TeachingAssistant'
 COURSE_MEMBERSHIP = 'http://purl.imsglobal.org/vocab/lis/v2/membership'
 DUMMY_CACHE = 'DummyCache'
+CANVAS_DATA_ID_INCREMENT = settings.CANVAS_DATA_ID_INCREMENT
```
ssciolla, line 37: Nit-pick: You could just reference `settings.CANVAS_DATA_ID_INCREMENT` if it's not used elsewhere.

dashboard/lti_new.py
```diff
@@ -181,7 +183,6 @@ def get_cache_config():
     cache_lifetime = cache_ttl if cache_ttl else 7200
     return CacheConfig(is_dummy_cache, launch_data_storage, cache_lifetime)
 
-
```
ssciolla, line 184: Nit-pick: PEP recommends two blank lines between top-level functions/classes: https://peps.python.org/pep-0008/#blank-lines

dashboard/lti_new.py
```diff
@@ -234,6 +235,14 @@ def extract_launch_variables_for_tool_use(request, message_launch):
         password = ''.join(random.sample(string.ascii_letters, settings.RANDOM_PASSWORD_DEFAULT_LENGTH))
         user_obj = User.objects.create_user(username=username, email=email, password=password, first_name=first_name,
                                             last_name=last_name)
+
+    # Add username into the MyLA User table, since the data was not pulled in from cron job
+    user_id = CANVAS_DATA_ID_INCREMENT + int(canvas_user_id)
+    try:
+        MyLAUser.objects.filter(user_id=user_id).update(sis_name = username)
```
ssciolla:
```suggestion
        MyLAUser.objects.filter(user_id=user_id).update(sis_name=username)
```

dashboard/lti_new.py
```diff
@@ -234,6 +235,14 @@ def extract_launch_variables_for_tool_use(request, message_launch):
...
+    except MyLAUser.DoesNotExist:
+        logger.warn(f'Cannot find MyLA user {username}')
```
ssciolla: Do we want to log a warning here? I think it's expected that some users will launch the tool before we have pulled cron data for them. It may just create log noise.

You could do this a little differently and not use a try/except.

```
enrollment_qs =  MyLAUser.objects.filter(user_id=user_id)
if enrollment_qs.exists():
    enrollment_qs.update(sis_name=username)
```

dashboard/lti_new.py
zqian: Removed!

ssciolla (approved):
I think this looks good. I'd like someone else to also review if possible before we merge.

dashboard/lti_new.py
zqian: this line is replaced by the following suggested code block:

`enrollment_qs =  MyLAUser.objects.filter(user_id=user_id)
if enrollment_qs.exists():
    enrollment_qs.update(sis_name=username)`

dashboard/lti_new.py
zqian: Added the empty line.

dashboard/lti_new.py
zqian: Replaced with the suggested block.

jonespm (approved):

jennlove-um:
@zqian This passes in testing.

Testing
1. launch myla from a test course (I masqueraded as a couple different users to populate the data)
2. verify the sis_name is in the user table
3. run the cron job
4. sis_name field is null for all users in the course

The only potential concern would be in pulling data at the end of the term. I _think_ it's fine this way since we have the user_id for all users in courses.  I'm trying to think ahead to identify any potential downsides.  I should be able to join that MyLA data to Canvas and SIS using the user_id instead of joining on uniqname, which is what I've done in the past.  Uniqname has always been the data point that existed in all 3 data sets.

jennlove-um:
Testing passes again in beta

jennlove-um:
Testing passes in beta

1. launch myla from a test course 
2. verify user can use the MyLA tool for this course
3. verify user can click on the top right profile button and can see user name in "Sign in as FULL NAME"
4. verify the name field is no longer in the user table in MyLA

Issue #827, opened 2019-12-12
in cron job, use batch select for artifacts across all courses, instead of doing one course at a time


zqian:
## Expected behavior (A clear and concise description of what you expected to happen) : 
There is still room for improvement for database select in cron.py 

## Describe the bug (Tell us what happens instead of the expected behavior) :

It looks like the util_function for executing database query is called for each course separately, for various objects. 

For example, https://github.com/tl-its-umich-edu/my-learning-analytics/blob/master/dashboard/cron.py line 370 is executed for every course to get the course assignment data. 

We could do a batch select for all registered courses

jonespm:
The objects that this _could_ be improved in are

However looks like the only slow process in the cron currently is the update_submission process which currently takes 13 minutes (more than half of the time of a typical 23 minute run run). That might be the only part to focus on. Though this seems like low priority.

* assignment `and co.lms_int_id = '{course_id}'`
* assignment_groups `and co.lms_int_id = '{course_id}'`
* assignment_weight `and co_km.lms_int_id = '{course_id}'`
* course `WHERE co2.lms_int_id = '{course_id}'`
* submission `co.lms_int_id = '{course_id}'`
* user `co.lms_int_id ='{course_id}'`

It's possible changing some of these would require more processing in the backend job.  

It looks like the resource query is already does this efficiently with `and co_km.lms_int_id in %(course_ids)s`

pushyamig:
I will QA this

PR #1423, opened 2022-09-06
issue #827 update the cron query to execute in batch mode

zqian:
Test Plan:

1. add MyLA tool to multiple courses
2. test the UDW query with "ENV_FILE=/secrets/env.hjson" in .env file
3. test the UDP query with "ENV_FILE=/secrets/env_udp.hjson" in .env file


dashboard/cron.py
```diff
@@ -1,6 +1,7 @@
 from datetime import datetime
 import logging
 from collections import namedtuple
+from operator import truediv
```
jonespm: It doesn't look like this truediv is being used.

zqian: also reported by Codacy. This line is removed.

jonespm (approved):
Ran both the UDP and UDW and looks like they run without error. 

The user validation is returning an expanded number for one of the queries now 

1.7700000000529132e+16
instead of
17700000000529132

I'm not sure what's changed with that. It looks fine though.

zqian:
I saw that data type change too for course id field in validation script. Don't know why either. 

I will merge in the PR, and test the execution time length of cron job.

ssciolla:
@zqian, @jonespm, probably this? https://re-thought.com/how-to-suppress-scientific-notation-in-pandas/

PR #1424, opened 2022-09-06
Fixes #1327 - Improve verify_course_ids to only consider courses with local data

### LINKED ISSUE #1327: Improve verify_course_ids to only consider courses with local data
When validating the courses, we should also check to see if the courses have any data in other tables because we consider it invalid.

There are cases where a course is too new and we haven't even pulled down data for it yet.

I think for an easier solution we could also just check the `data_last_updated` column and check to see if that has a value. It would still be an "invalid course" but one that can just be skipped if it was never processed and has no data.

This method is intended to protect removing data in the case that the data warehouse has a problem.

https://github.com/tl-its-umich-edu/my-learning-analytics/blob/719a29c69ea1c690026170fb5b347de722ff644a/dashboard/cron.py#L121

jonespm:
This is going to need to be refactored after #1423. So I'm leaving as a draft for now.

### DIFF
--- a/dashboard/cron.py
+++ b/dashboard/cron.py
@@ -143,7 +143,7 @@ class DashboardCronJob(CronJobBase):
         logger.debug("in checking course")
 
         # loop through multiple course ids
-        for course_id, data_last_updated in Course.objects.get_supported_courses(include_last_updated=True):
+        for course_id, data_last_updated in Course.objects.get_supported_courses():
             logger.debug(course_id)
             # select course based on course id
             course_sql = queries['course'].format(course_id=course_id)
@@ -677,7 +677,7 @@ class DashboardCronJob(CronJobBase):
             status += self.update_unizin_metadata()
 
         courses_added_during_cron: List[int] = list(
-            set(Course.objects.get_supported_courses()) - set(self.valid_locked_course_ids))
+            set(Course.objects.get_supported_courses().values_list('id', flat=True)) - set(self.valid_locked_course_ids))
         if courses_added_during_cron:
             logger.warning(
                 f'During the run, users added {len(courses_added_during_cron)} course(s): {courses_added_during_cron}')
--- a/dashboard/cron.py
+++ b/dashboard/cron.py
@@ -143,7 +143,7 @@ class DashboardCronJob(CronJobBase):
         logger.debug("in checking course")
 
         # loop through multiple course ids
-        for course_id in Course.objects.get_supported_courses():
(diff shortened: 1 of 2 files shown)


jonespm:
On the call we decided to open this and fix #1423 after, which may not make this release.

zqian (approved):
worked as expected. Left a suggestion for minor comment change.

dashboard/cron.py
```diff
@@ -154,8 +154,12 @@ def verify_course_ids(self):
 
             # error out when course id is invalid, otherwise add DataFrame to list
             if course_df.empty:
-                logger.error(f"""Course {course_id} don't have the entry in data warehouse yet. """)
-                invalid_course_id_list.append(course_id)
+                # Check if the course was ever locally updated. If it was updated it is invalid, otherwise don't consider this an error and skip it.
```
zqian, line 157: maybe make it clearer with " the course was ever updated by cron.", instead of "the course was ever locally updated."

jonespm, line 157:
```suggestion
                # Check if the course was ever updated by cron. If it was updated it is invalid, otherwise don't consider this an error and skip it.
```

jonespm:
Thanks, I was just trying to decide what to return from the function. I could have it always just return the date but it's then the code that only needs the ID needed to have an adjustment. I'm not sure the best pattern here. I wanted to try to avoid disrupting and complicating that code.

dashboard/models.py
```diff
@@ -139,14 +139,19 @@ class Meta:
...
         :return: [List of supported course ids]
         :rtype: [list of str (possibly incremented depending on parameter)]
         """
         try:
-            return self.values_list('id', flat=True)
+            # If we want to include the last updated date, otherwise just return a flat set
+            if include_last_updated:
+                return self.values_list('id', 'data_last_updated')
+            else:
+                return self.values_list('id', flat=True)
```
ssciolla, line 154: This might be trying to do too many things. It would be nice if the queryset always returned data with one shape. Django has a pretty good API, so I'd be included to keep this as it was, and then just use `Course.objects.values_list(`id`, `data_last_updated`)`.

I don't remember, when does this `model.DoesNotExist` error occur?

jonespm: I agree, I've simplified this back to how it was and updated it in the cron. 

The blame of this DoesNotExist goes back to me in 2018. I don't remember if it was needed or not. 793128b0f5dd2003d595781752ff18a422be1806

ssciolla (approved):
Seems to be working. I tested the code changed, no problems found.

PR #1425, opened 2022-09-06
Allow for environment variable override of `env.hjson` values (#787)

### LINKED ISSUE #787: Environment variable override
In order to deploy into a kubernetes infrastructure with more ease, I'd like the ability for any variables defined within the environment, to supersede any entries defined in the env.json file that is provided.

I'd like for [env_sample.json](https://github.com/tl-its-umich-edu/my-learning-analytics/blob/master/config/env_sample.json) to be loaded as the default configuration. Any value defined in the environment during initialization, should override what has been loaded as default.

For example, if [env_sample.json](https://github.com/tl-its-umich-edu/my-learning-analytics/blob/master/config/env_sample.json#L47-L51) was loaded as our default configuration, it contains the following entries for log levels

```
DJANGO_LOG_LEVEL: DEBUG
DJANGO_DEBUG: true
DJANGO_TEMPLATE_DEBUG: true
ROOT_LOG_LEVEL: INFO
```

If my environment were to have the following defined 

```
export DJANGO_LOG_LEVEL=WARN
export ROOT_LOG_LEVEL=DEBUG
```
I'd like the variables from my environment to override the default settings that were loaded from the json file. In the end we end up with these values being loaded.

```
DJANGO_LOG_LEVEL: WARN
DJANGO_DEBUG: true
DJANGO_TEMPLATE_DEBUG: true
ROOT_LOG_LEVEL: DEBUG
```

## Additional context
If it heps any, I was accomplishing this with this snippet 

```
if os.getenv("DJANGO_CONFIG_DEFAULTS"):
    # optionally load settings from an environment variable
    with open(os.getenv("DJANGO_CONFIG_DEFAULTS")) as f:
        CONFIG = json.load(f)

    # override defaults with variables defined by the environment
    ENV = {**CONFIG, **os.environ}
else:
    ENV = os.environ
``` 
downside of my snippent is that everything from the environment was loaded as a string. I then needed to cast variables that I knew were a bool (str_2_bool) or int when assigning from `ENV`

ssciolla:
This PR aims to resolve #787.

Test plan:
1) In your `.env` or `docker-compose.yml`, add `ROOT_LOG_LEVEL=DEBUG`, then any additional `env.hjson` keys you'd like to override.
2) Start up the server; note the log messages about overrides.
3) Make sure the new values are being properly used.

### DIFF
--- a/config/env_sample.hjson
+++ b/config/env_sample.hjson
@@ -95,8 +95,6 @@
     # strings for construct file download url
     "CANVAS_FILE_PREFIX": "https://umich.instructure.com/files/",
     "CANVAS_FILE_POSTFIX": "/download?download_frd=1",
-    # Paths to Google BigQuery credential json file
-    "GOOGLE_APPLICATION_CREDENTIALS": "/secrets/bq_cred.json",
     # LTI Configuration
     # LTI is disabled by default
     "ENABLE_LTI": false,
--- a/config/env_sample.hjson
+++ b/config/env_sample.hjson
@@ -2,8 +2,8 @@
     # Required to set this with some value
     # Run the code here to generate https://raw.githubusercontent.com/openwisp/ansible-openwisp2/master/files/generate_django_secret_key.py
     "DJANGO_SECRET_KEY": "<Generate Secret Key>",
-    # Set the Timezone for the container (must rebuild if changed)
-    "TZ": "America/Detroit",
+    # Sets the time zone for Django (defaults to environment variable TZ, which defaults to "America/Detroit")
+    # "TIME_ZONE": "America/Detroit",
     # CSV List of hosts allowed, no spaces in between commas
     "ALLOWED_HOSTS": [
         "127.0.0.1",
--- a/dashboard/settings.py
+++ b/dashboard/settings.py
@@ -10,11 +10,15 @@ For the full list of settings and their values, see
 https://docs.djangoproject.com/en/1.9/ref/settings/
 """
 
-import hjson, os
+import json, logging, os
 from typing import Tuple, Union
 
+import hjson
 from django.core.management.utils import get_random_secret_key
-from django.utils.module_loading import import_string
+
+
+logger = logging.getLogger(__name__)
+logging.basicConfig(level='INFO')
 
 # Build paths inside the project like this: os.path.join(BASE_DIR, ...)
 BASE_DIR = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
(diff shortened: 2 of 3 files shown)


config/env_sample.hjson
```diff
@@ -95,8 +95,6 @@
     # strings for construct file download url
     "CANVAS_FILE_PREFIX": "https://umich.instructure.com/files/",
     "CANVAS_FILE_POSTFIX": "/download?download_frd=1",
-    # Paths to Google BigQuery credential json file
-    "GOOGLE_APPLICATION_CREDENTIALS": "/secrets/bq_cred.json",
```
zqian, line 99: Is the "GOOGLE_APPLICATION_CREDENTIALS" param no longer needed?

dashboard/settings.py
```diff
@@ -25,6 +29,26 @@
...
+    """
+    env_copy = env.copy()
+    for key in env_copy.keys():
+        if key in environ:
+            os_value = environ[key]
+            try:
+                os_value = json.loads(os_value)
+                logger.debug('Value was valid JSON; saved parsed data in env object.')
+            except json.JSONDecodeError:
+                pass
```
zqian, line 45: is the "pass" used for non-json values only?

otherwise, I think we need to log error for malformed json value.

config/env_sample.hjson
ssciolla: We do need it, I just don't think it's being used when you define it in `env.hjson`, only in `.env`. I think the BigQuery library uses an environment variable (https://cloud.google.com/bigquery/docs/reference/libraries), and unless I'm mistaken we never add values from `env.hjson` to the environment (we're kinda doing the opposite here). For deployments, `GOOGLE_APPLICATION_CREDENTIALS` seems to be set in the environment in our Jenkins pod definitions.

I removed this here because in my testing, the needed value in `.env` was overriding the value. It was doing any harm, but it was creating some unnecessary noise. I also figured it could lead to some confusion if someone thought it needed to be updated here but it wasn't doing anything.

dashboard/settings.py
ssciolla: So the logic here is,

1) grab the value from the environment variable, store it in a variable
2) try to parse the value as JSON, if it works, reassign it to the variable
3) if it doesn't work, no sweat, it's probably just a vanilla string value.

I don't think I'd want to log an error since in many cases, it won't be valid JSON, and that's expected. I can add a debug log message if you'd like so the logic is more visible when testing. Would you like that?

zqian: Yes. If you can just add those comment lines in

config/env_sample.hjson
zqian: Sounds good. Better not to keep the redundant definition.

zqian (approved):
requested a comment change. Good for merge afterwards.

dashboard/settings.py
ssciolla: I added a debug log message. Does that work? https://github.com/tl-its-umich-edu/my-learning-analytics/pull/1425/commits/d15452f14db9a7c6efe3a4760baff0221f451d88

PR #1426, opened 2022-09-07
upgrade Python to v3.10 (iss. #1406)

### LINKED ISSUE #1406: Vulnerability: CVE-2015-20107, Python
From [vulnerabilities](https://docs.google.com/spreadsheets/d/12cyQTkIxCxwP0lAi58eFvznBcKDO3yB3mKdSenGkBOM) spreadsheet, based on Unizin analysis:
CVE ID | Effective Severity | Severity | Impacted Image | Vulnerable Package | Remediated Package | URL
-- | -- | -- | -- | -- | -- | --
CVE-2015-20107 | CRITICAL | CRITICAL | gcr.io/unizin-core/myla:2022.01.01 | python3.9/3.9.2-1 | python3.9/MAXIMUM | https://security-tracker.debian.org/tracker/CVE-2015-20107

lsloan:
To resolve vulnerability CVE-2015-20107, upgrade Python to v3.10.  Also upgrade `pandas` related modules to work with the newer version of Python.

Minor changes were also made to `start.sh` to fix a spelling mistake and clean up trailing spaces on a few lines.

Closes #1406.

## Test plan
* Run MyLA locally and verify that it works properly.
* **_Optional:_** Verify that Python 3.10 is installed.  
  ```sh
  docker exec -it student_dashboard python --version  # E.g., Python 3.10.6
  ```


requirements.txt
```diff
@@ -26,13 +26,13 @@ rules==3.0
 
 # These should be okay to update minors
 numpy==1.22.0
-pandas==1.3.1
+pandas==1.4.4
 pangres==2.3.1
```
ssciolla, line 30: Seems like bumping `numpy` is also appropriate? `pangres` has had two major releases, so we may need to change API usages, but I'd look at that as well, since we're updating `pandas`.

ssciolla: Looks like this `numpy` version is still compatible. And `pangres` is pretty far behind, but seems like it should theoretically still be compatible. Did you test the cron?

Can you rebase this with master so we can more effectively test with the latest version of cron (UDP)? Thanks

ssciolla: When testing with UDW, I'm seeing this warning I wasn't seeing before when using `pd.read_sql`:
```
UserWarning: pandas only support SQLAlchemy connectable(engine/connection) ordatabase string URI or sqlite3 DBAPI2 connectionother DBAPI2 objects are not tested, please consider using SQLAlchemy
```

Looks like we're passing a Django connection object. I guess it's probably better to create our own SQLAlchemy engine and then close it at the end of the cron?

See these:
- https://pandas.pydata.org/docs/reference/api/pandas.read_sql.html?highlight=read_sql#pandas.read_sql
- https://stackoverflow.com/a/71083448

ssciolla: Actually, this is happening on the views, too. We'll need to think about how best to open/close connections. Maybe we can form a MySQL connection string? Alternatively, we can use the Django ORM to pull data in the views, and pass it into vanilla DataFrames, but that's a bigger change.

lsloan: I didn't test the cron.

Does the warning appear in the UI or in the server log?

If it doesn't appear in the UI, is the content affected by this warning?  Is it getting the wrong information?

If the UI is unaffected, no warning or content problem, what should we do?  Do we go ahead with this change on its own, or do we need to address the warning before including it in the release?

ssciolla: I'd spend some time trying to address, and if it's a simple change, let's get it in. If not, we'll have to discuss more. Ideally, we don't have log warnings going off constantly as the tool runs, that affects our ability to inspect for other problems.

ssciolla: To be clear, it's in the logs, the UI doesn't seem to be affected. It's a warning, not an error, but it seems like `pandas` is telling us, we're not explicitly supporting this type of connection object going forward.

jonespm: It sounds like some devs are just writing code to ignore it as Pandas mentions in the Issue that they want to officially support other raw SQL connections. I think we can open this as an issue and revisit.

https://github.com/pandas-dev/pandas/issues/45660#issuecomment-1077355514

jonespm: Okay well we only use this in the cron, views and data_validation.

We could put at the top before we import pandas.

```
import warnings
# ignore warning for non-SQLAlchemy Connection
# see github.com/pandas-dev/pandas/issues/45660
warnings.simplefilter('ignore', category=UserWarning)
```

We could create and mention a new issue there too. I don't think fixing this actually needs is a blocker for this release, but I'd like to get Python updated if it's the only thing holding it back. 

Rather than using simplefilter we could use filterwarnings to be more specific to just this message.

```
import warnings
# ignore warning for non-SQLAlchemy Connection
# see github.com/pandas-dev/pandas/issues/45660
warnings.filterwarnings('ignore', category=UserWarning, message='pandas only support SQLAlchemy connectable')
```

ssciolla: I'm okay with a warning filter.

dashboard/cron.py
```diff
@@ -134,22 +138,33 @@ def __init__(self) -> None:
...
+        for course_id in Course.objects.get_supported_courses():
+            logger.debug(course_id)
+            # select course based on course id
+            course_sql = queries['course'].format(course_id=course_id)
+            logger.debug(course_sql)
+
+            connect_string = \
+                'postgresql://{user}:{password}@{host}:{port}/{database}' \
+                    .format(**(conns['DATA_WAREHOUSE'].get_connection_params()))
+            course_df = pd.read_sql(course_sql, connect_string)
```
lsloan, line 155: I'm using a different Git GUI that I'm not very familiar with yet.  I did a pull from upstream master and this is what I ended up with.  This diff appears as though I changed a lot.  I think most of this is duplicated changes from merging an upstream pull into my branch.

What I changed was to add the `connect_string` code and reference.  Before that, it had been a single line…

```python
            course_df = pd.read_sql(course_sql, conns['DATA_WAREHOUSE'])
```

So, the construction of a PostgreSQL connect string is pretty easy to do, using parameters from the Django connections.  This is a proof of concept.  Rather than construct a connect string for each call to `pd.read_sql()`, we should construct the string once or construct a SQLAlchemy engine object once.

ssciolla: Yeah, something went wrong with that merge. I'd probably revert and start over.

Just to recap the current situation, I'm okay/happy with the quick solution of a warning filter suggested by @jonespm, though that may require intervention later if this ever gets fixed. Slightly better would be the connection string and engine approach you have started. Even better would be to do Django ORM stuff in the views, but we are crunched for time and need to minimize changes to not over-complicate the release.

I would say, let's go back to only adding the warning filters to `cron.py` and `views.py` (not so worried about `data_validation`, but it could be added there too). Once that's done, Matt and I can review this more easily. Then, you can separately continue with this connection string and engine work, and if we can get it work in time, we can try to include it. What do you think? @lsloan and @jonespm

lsloan: I reset my copy of `cron.py` and made my changes again from a saved copy.  Looks correct now.

I don't mind saving this change for another time if it seems like a warning filter is the better way to go at this time.

ssciolla: I'll leave the decision up to you and @jonespm. I like the idea of using an engine here in `cron.py`.

requirements.txt
jonespm: Is this just a problem with the cron.py or were you seeing it in the application log from the views.py when using the `django.db.connection`? I guess I could run this and check.

I'm fine with either too, just seems like it's going to complicate this PR trying to implement something here for all the `read_sql` areas since we use 2 different databases.

ssciolla: I'm seeing it in the views.

ssciolla:
@jonespm, I implemented both of the quicker solutions discussed. Please let me know which you'd prefer and I'll either update this PR or create a new one:
- https://github.com/lsloan/myla/pull/3
- https://github.com/lsloan/myla/pull/4

jonespm:
Was going to talk to you about this, thanks for picking this up @ssciolla 

I'm good with #3, looks good. I'm fine with it being new after this is merged or just committed here and squashed.

Thanks! We'll probably want to improve the views in the future but we can *probably* wait to see if we have to rewrite them all anyway to use data marts.

requirements.txt
ssciolla: Should be resolved by latest commits using the engine approach.

dashboard/common/db_util.py
```diff
@@ -1,20 +1,41 @@
...
 
+class DBParams(TypedDict):
+    NAME: str
+    USER: str
+    PASSWORD: str
+    HOST: str
+    PORT: int
+
+
+def create_sqlalchemy_engine(db_params: DBParams, type: Literal['mysql', 'postgres']) -> Engine:
```
jonespm, line 28: It looks like the type parameter isn't really needed as ENGINE will always be available in DBParams if you added it to the DBParams class. There would be a value of.

`django.db.backends.postgresql `or `django.db.bacakends.mysql.`. So seems like you could just drop this parameter and use

`if 'mysql' in new_db_params['ENGINE']:`

But I guess it isn't a big deal either way.

jonespm (approved):
I'd remove the type parameter but I'm fine with this either way. Looks good!

dashboard/common/db_util.py
ssciolla: I made the requested change, see https://github.com/tl-its-umich-edu/my-learning-analytics/pull/1426/commits/d22125cc31bffa45a993edad8d95f52f4c79dab0

requirements.txt
ssciolla: I also bumped `numpy` and `pangres`, `pangres` only required a small change to respond to a parameter name change

PR #1429, opened 2022-09-10
issue_1428 replace left join with join for two tables inside submission query

zqian:
The change removed two unnecessary left joins inside the submission sub-query.

Test Plan:

1. set `ENV_FILE=/secrets/env_udp.hjson`
2. start the cron job: `docker exec -it student_dashboard /bin/bash -c "python manage.py migrate django_cron && python manage.py runcrons --force"`
3. calculate the time spend in submission query with cron log message

I have tested that for 150+ MyLA course ids, this query took about 3mins to finished

### DIFF
--- a/config/cron_udp.hjson
+++ b/config/cron_udp.hjson
@@ -235,14 +235,12 @@
             from
             course_assignment cross join enrollment
         ),
-        course_assignment_submission as
+        course_assignment_submission_sub as
         (
             select
-                cast(lar2.lms_int_id as BIGINT) as submission_id,
                 cae.assignment_id as assignment_id,
                 cae.course_id as course_id,
                 lar.published_score as published_score,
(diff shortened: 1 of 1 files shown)


jonespm (approved):
Looks good, validation passes and the number of submissions for 3 courses is the same. It's faster but still not as fast for me as the UDW. 

(UDW code ~2 secs)
`[2022-09-13 15:17:32 -0400] [DEBUG] [cron.py:69]  table: submission insert size: 2133`

(Previous UDP code ~30secs)
[2022-09-13 15:20:26 -0400] [DEBUG] [cron.py:69]  table: submission insert size: 2133

(This code ~10secs)
[2022-09-13 15:21:45 -0400] [DEBUG] [cron.py:69]  table: submission insert size: 2133

PR #1431, opened 2022-09-12
Use `course_date_range` logic for GraphQL course start date (#1430)

### LINKED ISSUE #1430: Assignment Planning week numbers are not calculated correctly for courses using Term dates
On the Assignment Planning page, the week numbers are calculated for a course based on the start/end dates of the term when a Canvas course is set up to use term dates.

For a Canvas course that use term dates rather than course dates for the course, MyLA should use the Academic Term start/end dates as the course start/end dates to determine week 1, week, 2, etc. on the Assignment Planning page.  Instead, the first week in the visualization is showing up as week 2750.

### Steps to Reproduce :
1.  Find a Canvas course that uses Term dates for the course.  Verify that the term exists in both Canvas and MyLA and has start/end dates.
2. Add MyLA to that course.
3. View the Assignment Planning.  The first week in the view shows up as week 2750 instead of week 1.

image

ssciolla:
This PR aims to resolve #1430.

### DIFF
--- a/dashboard/models.py
+++ b/dashboard/models.py
@@ -14,7 +14,6 @@ from django.conf import settings
 from django.core.exceptions import ObjectDoesNotExist
 from django.urls import reverse
 
-from collections import namedtuple
 from datetime import datetime, timedelta
 import logging
 import pytz
@@ -157,12 +156,12 @@ class CourseQuerySet(models.QuerySet):
         :return: Earliest start date of courses in the QuerySet, or None if no course dates found
         :rtype: datetime
         """
-        sorted_courses = sorted(self.all(), key=lambda course: course.course_date_range.start)
+        sorted_courses = sorted(self.all(), key=lambda course: course.determine_date_start())
 
         earliest_start = None
         if len(sorted_courses) > 0:
             earliest_course = sorted_courses[0]
-            earliest_start = earliest_course.course_date_range.start
+            earliest_start = earliest_course.determine_date_start()
             logger.info(f"Earliest start datetime for CourseQuerySet: {earliest_start.isoformat()} found in course {earliest_course.canvas_id}")
         else:
             logger.info(f"No courses in CourseQuerySet; returning None as the earliest_start_datetime")
@@ -235,14 +234,6 @@ class Course(models.Model):
             date_start = start if start else self.determine_date_start()
             end = date_start + timedelta(weeks=2)
         return end
-
-    @property
-    def course_date_range(self):
-        start = self.determine_date_start()
-        end = self.determine_date_end(start)
-        DateRange = namedtuple("DateRange", ["start", "end"])
-        date_range = DateRange(start, end)
-        return date_range
 
     @property
     def absolute_url(self):
--- a/dashboard/models.py
+++ b/dashboard/models.py
@@ -18,7 +18,7 @@ from collections import namedtuple
 from datetime import datetime, timedelta
 import logging
 import pytz
-from typing import Optional
+from typing import Optional, Union
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


dashboard/models.py
```diff
@@ -212,24 +211,29 @@ class Course(models.Model):
...
         if self.date_start is not None:
             start = self.date_start
         elif self.term is not None and self.term.date_start is not None:
             start = self.term.date_start
         else:
             logger.info(f"No date_start value was found for course {self.name} ({self.canvas_id}) or term; setting to current date and time")
             start = datetime.now(pytz.UTC)
+        return start
+
+    def determine_date_end(self, start: Union[datetime, None] = None) -> datetime:
```
ssciolla, line 224: Maybe passing the start here is too fancy, but maybe it keeps things more consistent and saves a query in some circumstances.
Scenarios to consider:
- No course dates, no term dates, current time and two weeks from now; better synchronized but probably only by milliseconds, which is basically irrelevant.
- No date end, no term dates, but a course start date. Query for course doesn't need to be re-made when getting the course start?

These may not matter much anyway, since I tried accessing a course without a term (which I manually deleted), and it said that course data was still being processed.

pushyamig:
I am reviewing it now.

pushyamig:
I am running into some setup issues, so I have a delay in reviewing. But I don't see those issues as part of this PR.

dashboard/models.py
pushyamig: I was able to test these scenarios and I was able to get correct result.

In case of no end date, Things were fine as well. I don't think Assignment planning relay on end date. But resource access view does and in that case the bar was showing 2 weeks only. The behavior is same from before. All this change was doing is name change of the  function.

dashboard/models.py
```diff
@@ -212,24 +211,29 @@ class Course(models.Model):
...
-            logger.info(f"No date_end value was found for course {self.name} ({self.canvas_id}) or term; setting to two weeks past start date.")
-            end = start + timedelta(weeks=2)
-        DateRange = namedtuple("DateRange", ["start", "end"])
-        return DateRange(start, end)
+            logger.info(
+                f"No date_end value was found for course {self.name} ({self.canvas_id}) or term; " +
+                "setting to two weeks past start date."
+            )
+            date_start = start if start else self.determine_date_start()
+            end = date_start + timedelta(weeks=2)
```
pushyamig, line 235: I felt the assignment of the end date could be done with `date_end` variable if you like as it is done for `date_start` just for consistency.

pushyamig:
One think I noticed is when `api/v1/courses/12344/info/` made during the loading for the course view, when course doesn't have start and end date the value returned is null. But the current week and total weeks value calculations is based on the `determine_start_date` and `determine_end_date` . 

It is a moot point but I feel we could populate that course start and end date in response (since we are calculating it based on our ideas and AP/RA view use that value as well )for clarity purposes. But this is nothing related to this issue

pushyamig (approved):
I see this change as an improvement and working as expected. I have some comment but are optional

jonespm (approved):

dashboard/models.py
ssciolla: Okay, I made it a bit more consistent. I was just trying to use a different name than the `start` parameter. https://github.com/tl-its-umich-edu/my-learning-analytics/pull/1431/commits/1dc178a66d426a53d6d71c7abb7b4037df5d8fad

pushyamig: Nice!

zqian (approved):
The date lookup logic looks good to me. Test was a success.

ssciolla:
> One think I noticed is when `api/v1/courses/12344/info/` made during the loading for the course view, when course doesn't have start and end date the value returned is null. But the current week and total weeks value calculations is based on the `determine_start_date` and `determine_end_date` .
> 
> It is a moot point but I feel we could populate that course start and end date in response (since we are calculating it based on our ideas and AP/RA view use that value as well )for clarity purposes. But this is nothing related to this issue

Thanks for noting this @pushyamig. I think it would be worth examining that view again to make sure we're returning consistent data that is actually used. I don't see us using these course dates or the term really from that particular endpoint in the frontend. Maybe we did at some point. I can create a separate issue to address this. I think for now we should be okay.

PR #1434, opened 2022-09-15
Add `view_on_site` method in course admin (#1411)

### LINKED ISSUE #1411: "View on Site" not working in Django Admin

Clicking "View on Site" in the top right of the course editing admin page should take you to the course's landing page.

The URL of this button is now something like this `/admin/r/5/17700000000XXXXXX/`, and doesn't seem to use `get_absolute_url` as the docs would suggest. https://docs.djangoproject.com/en/3.2/ref/models/instances/#get-absolute-url

Related issue: #936

Seems like we may need to add something like this, though I'm not sure why:
https://docs.djangoproject.com/en/3.2/ref/contrib/admin/#django.contrib.admin.ModelAdmin.view_on_site

Possibly related:
https://code.djangoproject.com/ticket/27757

ssciolla:
This PR aims to resolve #1411.

### DIFF
--- a/dashboard/admin.py
+++ b/dashboard/admin.py
@@ -75,6 +75,9 @@ class CourseAdmin(admin.ModelAdmin):
     readonly_fields = ('term', 'data_last_updated',)
     actions = ['clear_course_updated_dates']
 
+    def view_on_site(self, obj):
+        return obj.get_absolute_url()
+
     @admin.action(description='Clear selected last updated values')
     def clear_course_updated_dates(self, request, queryset):
         queryset.update(data_last_updated=None)


jonespm (approved):

ssciolla:
Ah, you are correct, @jonespm. If I change it to my ngrok URL, or add my URL to `ALLOWED_HOSTS`, then it works as expected. Is my change here worth adding? Not sure...

jonespm:
It's a small change, I'm fine merging it as it should work in all cases. 

Otherwise we'd have to document to make sure the sites table is correctly configured. It looks like it should automatically configure it if the [DOMAIN is set?](https://github.com/tl-its-umich-edu/my-learning-analytics/pull/761
) I don't actually see DOMAIN in any of the sample configuration files though and not sure if I ever set it? 

If we only ever set one site I'm not sure if we need the sites framework? Or maybe it would be useful if/when we put the LTI config in the database. Though CAE isn't using Sites.

ssciolla:
`DOMAIN` gets populated from `ALLOWED_HOSTS` in `env.hjson`. I think we may need the sites framework, but not 100 percent positive.

Issue #1443, opened 2022-09-20
Some process needs to still retrieve/persist user_id and sis_name as it is used for MiVideo access events


zqian:
### Describe the bug (Tell us what happens instead of the expected behavior) :
In [PR #1392](https://github.com/tl-its-umich-edu/my-learning-analytics/pull/1392/files), the user cron query does not pull in uniqname/sis_name from UDW. This is in preparation for UDP transition and UDP does not have non-roster user info, so that we need to get user login_name from LTI launch #1344 

However, this causes problem for MiVideo resource access events:
Since MiVideo is standalone tool, the Caliper events does not have Canvas user id information, so the user_id field has value of "-1". In [cron.py line 312](https://github.com/tl-its-umich-edu/my-learning-analytics/blob/master/dashboard/cron.py#L312), MiVideo events are joined with the MyLA user table based on sis_name, so that MiVideo events can be stored inside resource_access table with values in user_id field

With the above change of [PR #1392](https://github.com/tl-its-umich-edu/my-learning-analytics/pull/1392/files), MyLA user table no longer has sis_name field populated after cron run, and instead the sis_name value is set upon individual LTI launch and will be reset to null with next cron job. Hence the merge in [cron.py line 333](https://github.com/tl-its-umich-edu/my-learning-analytics/blob/master/dashboard/cron.py#L333) will always return empty dataframe for MiVideo access events, hence MiVideo events won't be shown in Resource Access View.

### Steps to Reproduce :
1. Enable MyLA tool for a course with MiVideo tool
2. Run MyLA cron job 
3. Go to MyLA resource access view, notice there is no MiVdeo events shown

lsloan:
To clarify, can an example of the Caliper event from MiVideo be added to the description?

zqian:
@lsloan you can find the MiVideo query inside [this wiki page](https://github.com/tl-its-umich-edu/my-learning-analytics/wiki/Deploy:-Institution-Architectures)

jonespm:
This looks good for the UDW, but this also needs to include a fix to pull what we can for UDP. I'm working on that and will test both of these out.

jennlove-um:
Expected outcomes: 

Kaltura video events will be included in Resources Accessed for users who people who exist in the UDP as students and teachers in the SIS data when they have a StudentEnrollment on a course with MyLA enabled.

Kaltura video events will NOT be available for: 

[...]

PR #1392, opened 2022-07-20
#489 enable UDP queries for MyLA cron job

zqian:
Test Plan:
- run `cp scripts/data_validation/env_validation_sample.hjson scripts/data_validation/env_validation.hjson`
- fill in `DATA_WAREHOUSE_COURSE_IDS` and `DATA_WAREHOUSE_SHORT_COURSE_IDS` with values for test courses 
- You will also need to update the DATA_WAREHOUSE inside mylasecrets/env.hjson to point to UDP context_store, and then run `docker exec -it student_dashboard /bin/bash -c "python manage.py migrate django_cron && python manage.py runcrons --force"` to start the cron job.
- run the validation script to compare UDW and UDP results: `docker exec -it student_dashboard /bin/bash -c "python scripts/data_validation/validate_udw_vs_udp.py"`

PR #1444, opened 2022-09-20
issue #1443 added the sis_name retrieval back to user query for UDW cron query

zqian:
This is partial revert of cron.hjson change in #1392: restored sis_name, and still leave out the user's name field.

Will file a UDP request for including Canvas user login_id in UDP, so that we can retrieve the sis_name field in UDP cron, too

### DIFF
--- a/config/cron.hjson
+++ b/config/cron.hjson
@@ -9,14 +9,14 @@
         with
         enroll_data as (select id as enroll_id, user_id, type, course_id from enrollment_dim where course_id in %(course_ids)s
                         and type in ('StudentEnrollment', 'TaEnrollment', 'TeacherEnrollment') and workflow_state= 'active'),
-        user_info as (select p.sis_user_id, u.id as user_id, u.global_canvas_id
+        user_info as (select p.unique_name, p.sis_user_id, u.id as user_id, u.global_canvas_id
                     from (SELECT ROW_NUMBER() OVER (PARTITION BY user_id order by sis_user_id asc) AS row_number, * FROM pseudonym_dim) as p
                     join user_dim u on u.id = p.user_id WHERE row_number = 1),
-        user_enroll as (select u.sis_user_id, u.user_id, e.enroll_id, e.course_id as course_id,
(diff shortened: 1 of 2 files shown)


pushyamig:
I am reviewing it now.

pushyamig:
I ran the cron for UDW queries and I was able to see `sis_name` populated ( but was not before unless an LTI launch happens). I saw the media events like Mivideo populated for course 562259 and I did noticed for another course `563654` leccap event also populated. 

This change looks good and the query that is revered back exactly that one in the past (i.e the working version)

jonespm:
With the current code these names are present in the auth_user->username field. 

However it looks like there's code in the cron that converts the -1 user_id to a numeric value from the _user_ table. I think the better fix for this will be just continuing along and storing the canvas user Id from the launch somewhere and just using that rather than querying from the warehouse. We may need to extend the Django User model.

`user` is really a special many to one table which still has some redundancy with the Django auth_user table. This is just legacy which isn't cleaned up.

I guess this is fine for now and probably quickest, but I don't _think_ it's ultimately necessary and can be fixed by:

1) Capturing and storing the canvas_user_id during the launch in a custom canvas_user model.
2) Using this table in the cron instead of user

The biggest negative I can see of this approach is that the canvas_user_id won't be available until the user logs in for the first time. So there will be 24 hours until the cron is able to update for this user to get MiVideo events. I think the only way that could be fixed would be to store either/or the `sis_name` or `user_id` to `resource_access` and just doing the lookup in the view rather than the cron.

jonespm:
This looks good for the UDW, but this also needs to include a fix to pull what we can for UDP. I'm working on that and will test both of these out.

jonespm (approved):

jonespm:
I tested with both UDW and UDP. You can give it a review if you'd like before merging @zqian.

For the UDP it still has the limitation of not being able to return ids for internal or staff accounts. But seems to do fine for students and friend accounts. It does put their ID in there as the name which is better than nothing and maybe all we have for now.

PR #1448, opened 2022-11-01
DRAFT: Switching mysqlclient to mysql-connector-python

jonespm:
This is done to resolve some security issues in the mariadb packages and use the pure Python MySQL driver. 

You have to update your env file in your config to use this different Django Engine. There isn't an issue related to this yet but I'll create one soon. 

Here's some [discussion of the performance of this. ](https://gist.github.com/methane/90ec97dda7fa9c7c4ef1?permalink_comment_id=4154254#gistcomment-4154254)

This is not fully tested so is a draft for now, but does startup.


jonespm:
Closing in favor of #1459

PR #1449, opened 2022-11-03
Update the d3 to version7

### LINKED ISSUE #1367: Update `d3`
Upgrading `d3` is a significant piece still to be done, and which I think we'll tackle separately. We're behind two major releases and the `d3-tip` companion library seems to have been archived on GitHub (https://github.com/caged/d3-tip). I found a migration guide for version 6, but no equivalent yet for version 7: https://observablehq.com/@d3/d3v6-migration-guide

_Originally posted by @ssciolla in https://github.com/tl-its-umich-edu/my-learning-analytics/pull/1363#discussion_r844336733_

As of now (4/8/22), I don't think `d3` version 5 poses any vulnerabilities, but updating it is something we'll have to face at some point.

pushyamig:
Fixes #1367 .

Update to d3-v7 is a breaking change and needed an update to all 3 visuals ( Resource accessed, Grade distribution, Old assignment planning)

- The main change is to the d3.event is deleted but can be accessed as part of the JS events .https://github.com/d3/d3/blob/main/CHANGES.md
- Migration guide to [v6](https://observablehq.com/@d3/d3v6-migration-guide) . I did not see a whole lot of difference to v7 affecting us 
- Updating the [d3-tip](https://github.com/caged/d3-tip) to new library [d3-v6-tip](https://github.com/bumbeishvili/d3-v6-tip#history).  
      - The [new library](https://github.com/bumbeishvili/d3-v6-tip#foreword) is based on the old library
      -  I don't see the d3-v6-tip updated to v7 but seems to be working fine.
- Assignment planning old needed some changes since it was breaking.
- New assignment planning don't use D3

### DIFF
--- a/assets/src/components/TableAssignment.js
+++ b/assets/src/components/TableAssignment.js
@@ -1,7 +1,6 @@
 // modified from https://demos.creative-tim.com/material-dashboard-react/?_ga=2.12819711.913135977.1549993496-494583875.1549993496#/table
 
 import React, { useRef, useEffect } from 'react'
-import RootRef from '@material-ui/core/RootRef'
 import withStyles from '@material-ui/core/styles/withStyles'
 import Table from '@material-ui/core/Table'
 import TableHead from '@material-ui/core/TableHead'
@@ -9,9 +8,9 @@ import TableRow from '@material-ui/core/TableRow'
 import TableBody from '@material-ui/core/TableBody'
 import TableCell from '@material-ui/core/TableCell'
 import HorizontalBar from '../components/HorizontalBar'
-import createToolTip from '../util/createToolTip'
 import { renderToString } from 'react-dom/server'
 import Paper from '@material-ui/core/Paper'
+import { tip as d3tip } from 'd3-v6-tip'
 import Typography from '@material-ui/core/Typography'
 
 const tableStyle = theme => ({
@@ -98,7 +97,7 @@ function CustomAssignmentTable (props) {
           data={[{ label: 'grade', data: percentOfFinalGrade, graded, submitted, pointsPossible, score, dropLowest, dropHighest }]}
           width={200}
           height={20}
-          tip={createToolTip(d =>
+          tip={d3tip().html((event, d) =>
             renderToString(
               <Paper className={classes.paper}>
                 <Typography component='p'>Score: {toolTipContent(d)}</Typography>
--- a/assets/src/components/Tooltip.js
+++ b/assets/src/components/Tooltip.js
@@ -4,8 +4,7 @@ import Paper from '@material-ui/core/Paper'
 import Typography from '@material-ui/core/Typography'
 import { tip as d3tip } from 'd3-v6-tip'
 
-export const AssignmentPlanningTooltip = classes => d3tip().html((event, d) => {
-  return renderToString(
+export const AssignmentPlanningTooltip = classes => d3tip().html((event, d) => { return renderToString(
   <Paper className={classes.paper}>
     <Typography>
       Assignment: <strong>{d.name}</strong><br />
--- a/assets/src/components/Tooltip.js
+++ b/assets/src/components/Tooltip.js
@@ -1,10 +1,11 @@
 import React from 'react'
 import { renderToString } from 'react-dom/server'
-import createToolTip from '../util/createToolTip'
 import Paper from '@material-ui/core/Paper'
 import Typography from '@material-ui/core/Typography'
+import { tip as d3tip } from 'd3-v6-tip'
 
-export const AssignmentPlanningTooltip = classes => createToolTip(d => renderToString(
+export const AssignmentPlanningTooltip = classes => d3tip().html((event, d) => {
+  return renderToString(
   <Paper className={classes.paper}>
     <Typography>
       Assignment: <strong>{d.name}</strong><br />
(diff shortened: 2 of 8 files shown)


assets/src/components/TableAssignment.js
```diff
@@ -1,17 +1,16 @@
 // modified from https://demos.creative-tim.com/material-dashboard-react/?_ga=2.12819711.913135977.1549993496-494583875.1549993496#/table
```
pushyamig, line 2: Old assignment planning change. I just did what worked.

assets/src/components/Tooltip.js
```diff
@@ -1,10 +1,11 @@
 import React from 'react'
```
pushyamig, line 1: Old assignment planning change. I just did what worked.

assets/src/components/d3/createProgressBar.js
```diff
@@ -56,8 +56,8 @@ function createProgressBar ({ data, width, height, domElement, tip }) {
     .style('outline-color', 'rgb(255, 255, 255)')
```
pushyamig, line 56: Old assignment planning change. I just did what worked.

assets/src/util/createToolTip.js
```diff
@@ -1,6 +0,0 @@
-import d3tip from 'd3-tip'
```
pushyamig, line 1: Old assignment planning change. I just did what worked.

assets/src/components/Tooltip.js
pushyamig: This change was made due since the event is passed as one of the arguments and cannot be access as `d3.event`

More on this https://observablehq.com/@d3/d3v6-migration-guide#events 

This change applies for all assignment and resource accessed view

assets/src/components/d3/createResourceAccessChart.js
```diff
@@ -1,7 +1,7 @@
 import * as d3 from 'd3'
```
pushyamig, line 1: This change was made due since the event is passed as one of the arguments and cannot be accessed as `d3.event`

More on this https://observablehq.com/@d3/d3v6-migration-guide#events

pushyamig: Lots of warnings were thrown during the `docker compose up ` w.r.t to `d3.event`. So fixes were based on those alerts and the migration guide help

assets/src/components/d3/createProgressBar.js
ssciolla: I'm seeing a console error sometimes that originates on line 50:
```
 Error: <rect> attribute width: Expected length, "NaN".
```

If you change the `attr('width...` part to this, you can see sometimes `NaN` is returned:
```
    .attr('width', d => {
      const result = x(d.percent_gotten) - margin.left
      console.log(result)
      return result
    })
```

assets/src/components/d3/createResourceAccessChart.js
```diff
@@ -250,17 +251,17 @@ function createResourceAccessChart ({ data, weekRange, gradeSelection, resourceT
...
+  function brushmove (event) {
     const fullRange = mainYZoom.range()
-    const selection = d3.event
-      ? d3.event.selection[1] === 0 // prevents [0, 0] from being returned, which causes bug
-        ? [0, 0.1]
-        : d3.event.selection
+    const selection = event
+      ? event.selection[1] === 0 // prevents [0, 0] from being returned, which causes bug
+          ? [0, 0.1]
+          : event.selection
```
ssciolla, line 264: Nit-pick: check whether indentation is correct.

ssciolla (commented):
This seems to be working for the most part. I'm flagging one error that I saw in old Assignment Planning, but I don't actually know if it's the result of the version change or not. I'll finish reviewing next week.

assets/src/components/d3/createProgressBar.js
pushyamig: I got a fix for this and I discussed this with @ssciolla

assets/src/components/Tooltip.js
```diff
@@ -1,10 +1,11 @@
...
 import { renderToString } from 'react-dom/server'
-import createToolTip from '../util/createToolTip'
 import Paper from '@material-ui/core/Paper'
 import Typography from '@material-ui/core/Typography'
+import { tip as d3tip } from 'd3-v6-tip'
 
-export const AssignmentPlanningTooltip = classes => createToolTip(d => renderToString(
+export const AssignmentPlanningTooltip = classes => d3tip().html((event, d) => {
+  return renderToString(
```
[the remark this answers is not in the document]
pushyamig: I think I indented as suggested, but I have a different linter so I am not sure if it same as your.

assets/src/components/d3/createResourceAccessChart.js
pushyamig: I think I indented as suggested, but I have a different linter so I am not sure if it same as your.

assets/src/components/d3/createProgressBar.js
ssciolla: No longer seeing the error 👍

assets/src/components/d3/createResourceAccessChart.js
ssciolla: Looks good.

ssciolla (approved):
I think this looks good. We'll need to do some some pretty thorough testing, hopefully with a range of courses to make sure no other edge cases cause problems and looking at the browser console.

PR #1451, opened 2022-11-04
issue_489 updated udp queries

zqian:
-  retrieve cross listed course enrollments
-  simplified submission query

Test plan:

- setup environment to run `scripts/data_validation/validate_udw_vs_upd.py`, following the instructions inside `docs/testing.md`

### DIFF
--- a/config/cron.hjson
+++ b/config/cron.hjson
@@ -165,7 +165,7 @@
             )
             select 
                 f.id, 
-                f.assignment_id, 
+                cast(f.assignment_id as bigint),
                 f.course_id, 
                 cast(f.user_id as bigint), 
                 f.score, 
--- a/config/cron_udp.hjson
+++ b/config/cron_udp.hjson
@@ -25,7 +25,7 @@
         left join entity.course_section cs
             on cse.course_section_id = cs.course_section_id
         left join keymap.course_offering co
-            on cs.course_offering_id = co.id
+            on cs.le_current_course_offering_id = co.id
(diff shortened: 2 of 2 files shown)


config/cron.hjson
```diff
@@ -165,7 +165,7 @@
             )
             select 
                 f.id, 
-                f.assignment_id, 
+                cast(f.assignment_id as bigint),
```
pushyamig, line 168: Was there any reason this change was made?

pushyamig:
I ran 2 tests for testing this 

1. 2 different courses one with SIS only section and one with NON-SIS section, I see it is pulling the correct roster information.
2. Use the validation script, but I see that for the `comparing query submission`, there was a difference found. this doesn't mean a failure, But I might not know full details of this. 

The courses I was using ["534433", "518589"],

config/cron.hjson
zqian: This is to have the same type as the assignment_id field from the udp query. 

Plus the assignment table inside MyLA database already have id field as bigint(20) type.

zqian:
Thanks for verifying, @pushyamig 

I checked the 534433 course, the submission records are diff for two users, and both got 13.95 in one submission. UDW reported the score as 13.9 vs UDP reported the score as 14. I think this is acceptable.

As for the 518589 course, the validation script did not complain about record difference.

pushyamig (approved):
I think this change seem to be working.

PR #1455, opened 2022-11-09
Fixes #1446 - Switch mysql in docker-compose.yml to just mysql:8.0

### LINKED ISSUE #1446: Switch mysql-server in docker-compose.yml to just mysql
All of our other projects are just using

` mysql:8.0`, which is a link for `mysql:8.0-oracle`. We should change over MyLA to be consistent with the other projects. I don't think it will be much effort but might require removing the local databases. 

https://github.com/tl-its-umich-edu/my-learning-analytics/blob/7e8006361d830a7a71dfe4e3cf820f5da5f7421c/docker-compose.yml#L6

jonespm:


zqian (approved):
Tested the mysql change works on localhost

PR #1459, opened 2022-11-29
Fixes #1454 - Switch MySQL Python or system library to version that has less security issues

### LINKED ISSUE #1454: Switch MySQL Python or system library to version that has less security issues

It was identified in a security report that the library libmariadb 10.5 packaged in the Debian distribution has a [number of security issues](https://security.snyk.io/package/linux/debian:11/mariadb-10.5) and has [not had an update since March.](https://tracker.debian.org/pkg/mariadb-10.5). There are fixes in the 10.6 version but this isn't due to be in until the next version of Debian next year.

Two alternatives were suggested to work around this.

1) Switch to a different python package that's pure Python and doesn't need this library. This was done in #1448 but there was some concerns about performance of this package still.

2) Switch to a different system library, either the one direct from [MariaDB](https://mariadb.com/docs/connect/clients/mariadb-client/) which is updated more often or the one from Oracle for MySQL.

Testing this will involve doing regular regression testing on MyLA as this database is used in multiple places. Option 2 would likely have less risk and less code change.

jonespm:
Opted to switch the the library from MariaDB. This doesn't report any security issues in the scan that the other one does. 

You should be able to test this with

`docker scan my-learning-analytics-web` to see that there aren't any more high issues compared to the master branch.

### DIFF
--- a/dockerfiles/Dockerfile
+++ b/dockerfiles/Dockerfile
@@ -40,8 +40,17 @@ WORKDIR /code
 COPY requirements.txt .
 RUN apt-get update && \
     apt-get install -y --no-install-recommends \
-    build-essential default-libmysqlclient-dev libpq-dev netcat jq python3-dev xmlsec1 cron git && \
-    apt-get upgrade -y && \
+        build-essential curl apt-transport-https libpq-dev netcat jq python3-dev xmlsec1 cron git && \
+    apt-get upgrade -y
+
+# Install MariaDB from the mariadb repository rather than using Debians 
+# https://mariadb.com/kb/en/mariadb-package-repository-setup-and-usage/
+RUN curl -LsS https://r.mariadb.com/downloads/mariadb_repo_setup | bash && \
+    apt install -y --no-install-recommends libmariadb-dev
+
+# Clean up anything we don't need anymore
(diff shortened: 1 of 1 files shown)


dockerfiles/Dockerfile
```diff
@@ -40,8 +40,17 @@ WORKDIR /code
...
+    apt-get upgrade -y
+
+# Install MariaDB from the mariadb repository rather than using Debians 
+# https://mariadb.com/kb/en/mariadb-package-repository-setup-and-usage/
+RUN curl -LsS https://r.mariadb.com/downloads/mariadb_repo_setup | bash && \
+    apt install -y --no-install-recommends libmariadb-dev
+
+# Clean up anything we don't need anymore
+RUN apt-get purge -y curl libcurl4 libcurl3-gnutls && \
+    apt autoremove -y && \
```
lsloan, line 53: I saw this technique used a couple weeks ago, too.  Install a newer package using other tools, then remove the tools.

I wonder whether it would be helpful to do this run command with one (or both) of the previous run commands, to get the dependencies all together in the same image layer.

lsloan (approved):
I really liked this approach.  It worked well for me.

PR #1460, opened 2022-11-29
Limit resources access view to show only events for enrolled students in the course #764

zqian:


zqian:
I will merge this PR since it has been tested during last release attempt.

PR #1461, opened 2022-12-08
Fixes #1454 (Fix 2) - Switch MySQL to library version with less security issues.

### LINKED ISSUE #1454: Switch MySQL Python or system library to version that has less security issues

jonespm:
* Don't remove libcurl3-gnutls since it's [needed for git.](https://packages.debian.org/bullseye/git)
* Apply changes to Dockerfile.openshift.

### DIFF
--- a/dockerfiles/Dockerfile
+++ b/dockerfiles/Dockerfile
@@ -49,7 +49,7 @@ RUN curl -LsS https://r.mariadb.com/downloads/mariadb_repo_setup | bash && \
     apt install -y --no-install-recommends libmariadb-dev
 
 # Clean up anything we don't need anymore
-RUN apt-get purge -y curl libcurl4 libcurl3-gnutls && \
+RUN apt-get purge -y curl libcurl4 && \
     apt autoremove -y && \
     apt-get clean -y && \
     rm -rf /var/lib/apt/lists/*
--- a/dockerfiles/Dockerfile.openshift
+++ b/dockerfiles/Dockerfile.openshift
@@ -39,8 +39,17 @@ WORKDIR /code
 COPY requirements.txt .
 RUN apt-get update && \
     apt-get install -y --no-install-recommends \
-    build-essential default-libmysqlclient-dev libpq-dev netcat jq python3-dev xmlsec1 cron git && \
-    apt-get upgrade -y && \
+        build-essential curl apt-transport-https libpq-dev netcat jq python3-dev xmlsec1 cron git && \
+    apt-get upgrade -y
+
+# Install MariaDB from the mariadb repository rather than using Debians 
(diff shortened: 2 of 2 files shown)


jonespm:
After this change

```
#31 [my-learning-analytics-web app  6/14] RUN apt-get purge -y curl libcurl4 &&     apt autoremove -y &&     apt-get clean -y &&     rm -rf /var/lib/apt/lists/*
#31 0.223 Reading package lists...
#31 0.554 Building dependency tree...
#31 0.633 Reading state information...
#31 0.724 The following packages will be REMOVED:
#31 0.724   curl* libcurl4*
#31 0.866 0 upgraded, 0 newly installed, 2 to remove and 0 not upgraded.
#31 0.866 After this operation, 1155 kB disk space will be freed.
(Reading database ... 15520 files and directories currently installed.)
#31 0.895 Removing curl (7.74.0-1.3+deb11u3) ...
#31 0.909 Removing libcurl4:arm64 (7.74.0-1.3+deb11u3) ...
#31 0.931 Processing triggers for libc-bin (2.31-13+deb11u5) ...
#31 1.008
#31 1.008 WARNING: apt does not have a stable CLI interface. Use with caution in scripts.
#31 1.008
#31 1.013 Reading package lists...
#31 1.289 Building dependency tree...
#31 1.356 Reading state information...
#31 1.445 0 upgraded, 0 newly installed, 0 to remove and 0 not upgraded.
#31 DONE 1.5s
```

ssciolla (approved):
I ran this, and Git now works as expected, and changes should now be present in OpenShift builds.

@jonespm and I talked about trying to remove `git` as a dependency in the running container, but it proved more challenging than expected (it's hard to set environment variables with command output in a Dockerfile). We could look into this at a later date.

jonespm:
Filed the git removal as an issue on #1462 and worked on a draft PR for it. Thanks for the review!

PR #1463, opened 2022-12-08
Fixes #1462 Move the git commit parsing out of the start.sh and into the build 

### LINKED ISSUE #1462: Move the git commit parsing out of the start.sh and into the build

Currently, git values are parsed in the `start.sh` script at runtime. This requires `git` be available at runtime. Because of occasional security issues with git and libcurl it would be beneficial to just remove these packages. I believe we could just do it in the build and then remove them after.

Add something like this to the Dockerfile.

```
RUN bash -l -c 'GIT_REPO="$(git config --local remote.origin.url)"; \
                GIT_COMMIT="$(git rev-parse HEAD)"; \ 
                GIT_BRANCH="$(git name-rev "$GIT_COMMIT" --name-only)"; \
                echo export GIT_REPO=$GIT_REPO > /etc/git.version; \
                echo export GIT_COMMIT=$GIT_COMMIT >> /etc/git.version; \
                echo export GIT_BRANCH=$GIT_BRANCH >> /etc/git.version'
```

Remove git and read it in the start script. 

Some hacky solutions for pasing the .git have been considered like 

```
GIT_COMMIT="$(cat .git/HEAD | cut -c 1-7)
GIT_BRANCH="$(grep $GIT_COMMIT .git/packed-refs | sort | tail -n 1 | cut -d/ -f2- | rev | cut -d/ -f1 | rev)
```

But I don't really like those a lot

### Additional context
It looks like you have to use a file and can't set these directly since every RUN only knows about it's own context.
https://stackoverflow.com/questions/34911622/dockerfile-set-env-to-result-of-command

jonespm:
It was suggested to 

- [x] Have this commands to parse git in a separate file 
- [x] Add changes to Dockerfile.openshift.
- [x] Rebase to latest code

### DIFF
--- a/dockerfiles/Dockerfile.openshift
+++ b/dockerfiles/Dockerfile.openshift
@@ -47,12 +47,6 @@ RUN apt-get update && \
 RUN curl -LsS https://r.mariadb.com/downloads/mariadb_repo_setup | bash && \
     apt install -y --no-install-recommends libmariadb-dev
 
-# Clean up anything we don't need anymore
-RUN apt-get purge -y curl libcurl4 && \
-    apt autoremove -y && \
-    apt-get clean -y && \
-    rm -rf /var/lib/apt/lists/*
-
 RUN pip install --no-cache-dir -r requirements.txt
 
 # copy built react and node libraries for production mode
@@ -63,6 +57,16 @@ COPY --from=node-prod-deps /usr/src/app/node_modules node_modules
 
 # NOTE: project files likely to change between dev builds
 COPY . .
+
+# Generate git version information
+RUN bash -l -c ./scripts/git_version_info.sh
+
+# Clean up anything we don't need anymore
+RUN apt-get purge -y git curl libcurl4 libcurl3-gnutls && \
+    apt autoremove -y && \
+    apt-get clean -y && \
+# Clean up the apt files and remove the .git directory from the container
+    rm -rf /var/lib/apt/lists/*
 
 RUN python manage.py collectstatic --verbosity 0 --noinput
 
--- a/dockerfiles/Dockerfile.openshift
+++ b/dockerfiles/Dockerfile.openshift
@@ -65,7 +65,6 @@ RUN bash -l -c ./scripts/git_version_info.sh
 RUN apt-get purge -y git curl libcurl4 libcurl3-gnutls && \
     apt autoremove -y && \
     apt-get clean -y && \
-# Clean up the apt files and remove the .git directory from the container
     rm -rf /var/lib/apt/lists/*
 
 RUN python manage.py collectstatic --verbosity 0 --noinput
--- a/scripts/git_version_info.sh
+++ b/scripts/git_version_info.sh
@@ -1,5 +1,5 @@
 #/bin/bash
-#Store the version information from the .git into a file so it can be used later at runtime after removing git
+# Store the version information from the .git into a file so it can be used later at runtime after removing git
 
 GIT_REPO=$(git config --local remote.origin.url)
 GIT_COMMIT=$(git rev-parse HEAD)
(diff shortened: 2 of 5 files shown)


jonespm:
`docker scan my-learning-analytics-web` doesn't report any critical/high issues after this, only low and medium.

dockerfiles/Dockerfile.openshift
```diff
@@ -64,6 +58,16 @@ COPY --from=node-prod-deps /usr/src/app/node_modules node_modules
...
 COPY . .
 
+# Generate git version information
+RUN bash -l -c ./scripts/git_version_info.sh
+
+# Clean up anything we don't need anymore
+RUN apt-get purge -y git curl libcurl4 libcurl3-gnutls && \
+    apt autoremove -y && \
+    apt-get clean -y && \
+# Clean up the apt files and remove the .git directory from the container
```
ssciolla, line 68: Did you mean to remove `.git` here?

jonespm: I think I originally was thinking about it, but it only saves like 4MB and it's lost me work more than once so I'd given up on the idea and just stopped trying. ;) I think I can just remove the comment.

jonespm, line 68:
```suggestion
```

scripts/git_version_info.sh
```diff
@@ -0,0 +1,10 @@
+#/bin/bash
+#Store the version information from the .git into a file so it can be used later at runtime after removing git
```
ssciolla, line 2:
```suggestion
# Store the version information from the .git into a file so it can be used later at runtime after removing git
```

scripts/git_version_info.sh
```diff
@@ -0,0 +1,10 @@
...
+
+GIT_REPO=$(git config --local remote.origin.url)
+GIT_COMMIT=$(git rev-parse HEAD)
+GIT_BRANCH=$(git name-rev "$GIT_COMMIT" --name-only)
+
+echo "export GIT_REPO=$GIT_REPO
+export GIT_COMMIT=$GIT_COMMIT
+export GIT_BRANCH=$GIT_BRANCH" > /etc/git.version
```
ssciolla, line 10: Nice to not have it at the root, but this doesn't really fit with the `data_validation` stuff already in `scripts` that much. I'm okay with this, just thinking about optimal file organization.

scripts/git_version_info.sh
```diff
@@ -0,0 +1,10 @@
+#/bin/bash
```
ssciolla, line 1: From `shellcheck`
```
In scripts/git_version_info.sh line 1:
#/bin/bash
 ^-- SC1113 (error): Use #!, not just #, for the shebang.

For more information:
  https://www.shellcheck.net/wiki/SC1113 -- Use #!, not just #, for the sheba...
```

ssciolla (approved):
This looks good, and `git` command wasn't found in the container. Couple comments to consider.

zqian (approved):

scripts/git_version_info.sh
jonespm: Thanks, I should have ran this.

scripts/git_version_info.sh
jonespm: Yeah, I saw we had a scripts directory so this *seemed* like the better place for it. I'll leave this in there for now as I'd like to have the root a little lighter. It's **possible** it'd fit better in dockerfiles since it's more just related to the build.

start.sh
```diff
@@ -3,6 +3,11 @@
 # Case insensitive match
 shopt -s nocaseglob
 
+# Load in git version information
+source /etc/git.version
+
+echo "Git tag is $GIT_BRANCH"
```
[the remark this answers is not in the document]
jonespm: Sure I can put the rest of the variables. This was really just to test that this worked before having to look at the UI.

PR #1465, opened 2022-12-19
Fixes #1464 Improve submission query speed with temp table

### LINKED ISSUE #1464: Improve performance of UDP submission query
UDP query should be able to run in a similar speed to UDW query

We've noticed when running the cron that the UDP submission query is taking the majority of time to run and taking a long time, almost an hour for many courses. There should still be improvements to be made with this query. 

We have some sample query plans in the My Learning Analytics T&L shared folder. Initial investigation is that using a temporary table might work well.

jonespm:


zqian:
Can you also address the duplicated enrollment table joins?
`            join enrollment on lar.person_id= enrollment.user_id
            join enrollment e on lar.person_id = e.user_id`

config/cron_udp.hjson
```diff
@@ -297,6 +295,6 @@
             ) as f1 
             on f.assignment_id = f1.assignment_id
         where f.id is not null
-        order by assignment_id, user_id
+        order by assignment_id, user_id;
```
zqian, line 298: can remove the ";"

zqian (approved):

config/cron_udp.hjson
jonespm:
```suggestion
        order by assignment_id, user_id
```

jonespm:
Right it looks like the enrollment joins are duplicate but for some reason when I removed one the query didn't finish. Didn't want to touch it that much right now.

PR #1468, opened 2023-01-06
Disallow changing course canvas ID (#613)

### LINKED ISSUE #613: Can't change course id in Courses admin

ssciolla:
This PR aims to resolve #613. See my comment in the issue for rationale behind the solution. I left it so you can still create courses manually in the admin UI, but you can't edit the course ID after you create the course (it'll have to be deleted and re-created).

Resource(s):
- https://docs.djangoproject.com/en/4.1/ref/contrib/admin/
- https://stackoverflow.com/a/46063862

### DIFF
--- a/dashboard/admin.py
+++ b/dashboard/admin.py
@@ -94,7 +94,7 @@ class CourseAdmin(admin.ModelAdmin):
     def course_link(self, obj):
         return format_html('<a href="{}">Link</a>', obj.absolute_url)
 
-    def change_view(self, request, object_id, extra_content=None):
+    def change_view(self, request, object_id, form_url='', extra_content=None):
         self.readonly_fields = self.readonly_fields + ('canvas_id',)
         return super(CourseAdmin, self).change_view(request, object_id)
 
--- a/dashboard/admin.py
+++ b/dashboard/admin.py
@@ -70,6 +70,7 @@ class TermAdmin(admin.ModelAdmin):
 class CourseAdmin(admin.ModelAdmin):
     inlines = [CourseViewOptionInline, ]
     form = CourseForm
+    fields = ('canvas_id', 'name', 'term', 'date_start', 'date_end', 'show_grade_counts', 'show_grade_type', 'data_last_updated')
     list_display = ('canvas_id', 'name', 'term', 'show_grade_counts', 'course_link', '_courseviewoption', 'data_last_updated')
     list_select_related = True
     readonly_fields = ('term', 'data_last_updated',)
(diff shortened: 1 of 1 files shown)


dashboard/admin.py
```diff
@@ -70,6 +70,7 @@ def has_add_permission(self, request):
 class CourseAdmin(admin.ModelAdmin):
     inlines = [CourseViewOptionInline, ]
     form = CourseForm
+    fields = ('canvas_id', 'name', 'term', 'date_start', 'date_end', 'show_grade_counts', 'show_grade_type', 'data_last_updated')
```
zqian, line 73: I do not see this newly added variable, "fields", used in the code anywhere.

dashboard/admin.py
```diff
@@ -93,6 +94,10 @@ def _courseviewoption(self, obj):
     def course_link(self, obj):
         return format_html('<a href="{}">Link</a>', obj.absolute_url)
 
+    def change_view(self, request, object_id, form_url='', extra_content=None):
+        self.readonly_fields = self.readonly_fields + ('canvas_id',)
```
zqian, line 98: maybe add the canvas_id into the `readonly_fields` definition in line 76?

dashboard/admin.py
ssciolla: This changes the order that the fields appear in in the admin form. If I don't specify this, then `canvas_id` gets pushed to the bottom when it is read only.

zqian: I see. Maybe add a comment line for this?

dashboard/admin.py
zqian: Got it!

zqian (approved):

dashboard/admin.py
ssciolla: Do you think we need a comment? In my view, it's documented well in the Django docs: https://docs.djangoproject.com/en/4.1/ref/contrib/admin/#django.contrib.admin.ModelAdmin.fields

ssciolla: I guess the readonly bit is not said explicitly. Just don't want to clutter the code too much with comments.

zqian: OK. The Django doc is enough.

PR #1470, opened 2023-01-09
Fixing the onclick on brush bug

### LINKED ISSUE #1467: On resources access view, clicking on the minimized view on the right side generates errors

You can click and navigate around on the right hand minimized view

If I click on the minimized view in resources accessed in any course when it changes to a + sign (on the area that isn't highlighted. There are a bunch of console errors thrown and the list to the left breaks. The only way I can fix this is to refresh. This doesn't seem to matter which course. Testing to see if this was introduced in  #1367.

image

image

I tested and this did work in 2022.01.03.

### Steps to Reproduce :
1. Go to the resource access view of any class with resources loaded 
2. Interact with the area on the right side
3.

We could possibly remove this area if we can't get it working. I'm not entirely sure the value it added. This view may really need to be refreshed.

pushyamig:
Fixes #1467 

https://observablehq.com/@d3/d3v6-migration-guide#pointer

### DIFF
--- a/assets/src/components/d3/createResourceAccessChart.js
+++ b/assets/src/components/d3/createResourceAccessChart.js
@@ -308,13 +308,12 @@ function createResourceAccessChart ({ data, weekRange, gradeSelection, resourceT
   }
 
   function brushcenter (event) {
-    const target = event.target
     const selection = d3.brushSelection(gBrush.node())
     const size = selection[1] - selection[0]
     const range = miniYScale.range()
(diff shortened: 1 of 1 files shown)


ssciolla (approved):
Good job finding the problem! I tested and am no longer see console or the visualization acting strange when clicking around on the mini graph.

pushyamig:
OK, I merging it as @jonespm want to test this on Beta. Thanks @ssciolla for reviewing it !

PR #1471, opened 2023-01-12
Fixes #1400 Update Django Cron to 0.6.0

### LINKED ISSUE #1400: Update Django Cron to 0.6.0
Django Cron finally had a new release after 4 years as 0.6.0 fixing a few bugs that we've had. We should update to support that version. 

https://github.com/tl-its-umich-edu/my-learning-analytics/blob/b0d05c166f1ee394a758d7e52fdcd95e14da2b4f/requirements.txt#L7

jonespm:


zqian (approved):
The cron job run was successful with the django-cron update.

PR #1472, opened 2023-01-19
Make Resources Accessed chart keyboard accessible (#1190)

### LINKED ISSUE #1190: Data in the hover over on resource bar elements in the list of resources viewed can't be navigated to using keyboard
Users should be able to navigate to or access all content in the Resources Accessed by using the keyboard.

The bars on the Resources Accessed bar chart include metadata about file views in a hover-over.  Tabbing should allow users to access the information that is in the hover-over data.

ssciolla:
This PR aims to resolve #1190.

To Do
- [x] Show tooltip on focus
- [x] Move brush/scroll when tabbing through chart
- [x] Move labels before bars in tab order
- [x] Evaluate click focus behavior (see [my comment here](https://github.com/tl-its-umich-edu/my-learning-analytics/pull/1472/files#r1086930662))

### DIFF
--- a/assets/src/components/d3/createResourceAccessChart.js
+++ b/assets/src/components/d3/createResourceAccessChart.js
@@ -474,7 +474,7 @@ function createResourceAccessChart ({ data, weekRange, gradeSelection, resourceT
     .call(truncate, resourceLabelWidth)
 
   // Add links and icons to Y axis
-  d3.selectAll('.axis--y .tick').each(function (d, i) {
+  d3.selectAll('.axis--y .tick').each(function (d) {
     // Have to use ES5 function to correctly use `this` keyword
     const link = d.split('|')[0]
     const name = d.split('|')[1]
--- a/assets/src/components/d3/createResourceAccessChart.js
+++ b/assets/src/components/d3/createResourceAccessChart.js
@@ -192,7 +192,7 @@ function createResourceAccessChart ({ data, weekRange, gradeSelection, resourceT
 
     bar.enter()
       .append('rect')
-      .attr('tabindex', d => d.index * 10 + 1)
+      .attr('tabindex', 0)
       .attr('y', d => mainYScale(d.resource_name))
       .attr('width', d => mainXScale(d.total_percent))
       .attr('height', mainYScale.bandwidth())
@@ -415,7 +415,7 @@ function createResourceAccessChart ({ data, weekRange, gradeSelection, resourceT
     .text('Percentage of All Students in the Selected Grade Range')
     .style('font-size', '14px')
 
-  mainGroup.append('g')
+  mainGroup.insert('g', ':first-child')
     .attr('class', 'axis--y black-axis-line')
     .attr('transform', 'translate(0,0)')
     .call(mainYAxis)
@@ -456,13 +456,12 @@ function createResourceAccessChart ({ data, weekRange, gradeSelection, resourceT
     // Have to use ES5 function to correctly use `this` keyword
     const link = d.split('|')[0]
     const name = d.split('|')[1]
-    const resourceDataIndex = resourceData.filter(rd => rd.resource_name === d)[0].index
     const a = d3.select(this.parentNode).append('a')
       .attr('title', name)
       .attr('target', '_blank')
       .attr('href', link)
       .attr('text-anchor', 'start')
-      .attr('tabindex', resourceDataIndex * 10)
+      .attr('tabindex', 0)
     a.node().appendChild(this)
 
     const icon = d.split('|')[2]
--- a/assets/src/components/d3/createResourceAccessChart.js
+++ b/assets/src/components/d3/createResourceAccessChart.js
@@ -173,7 +173,6 @@ function createResourceAccessChart ({ data, weekRange, gradeSelection, resourceT
     .attr('class', 'svgWrapper')
     .attr('width', availWidth)
     .attr('height', availHeight)
-    .attr('aria-hidden', 'true')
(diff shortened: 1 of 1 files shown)


assets/src/components/d3/createResourceAccessChart.js
```diff
@@ -413,7 +427,7 @@ function createResourceAccessChart ({ data, weekRange, gradeSelection, resourceT
     .text('Percentage of All Students in the Selected Grade Range')
     .style('font-size', '14px')
 
-  mainGroup.append('g')
+  mainGroup.insert('g', ':first-child')
```
ssciolla, line 430: This moves the link labels before the bars in the DOM, which makes the Tab order go from left to right (expected for our users?).

assets/src/components/d3/createResourceAccessChart.js
```diff
@@ -192,7 +192,7 @@ function createResourceAccessChart ({ data, weekRange, gradeSelection, resourceT
 
     bar.enter()
       .append('rect')
-      .attr('tabindex', d => d.index * 10 + 1)
+      .attr('tabindex', 0)
```
ssciolla, line 195: This changes see the `tabindex` strategy to use `0`, per general guidance from MDN on using `tabindex` (see Warning on https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex). The downside is that it doesn't go from link label to bar. I have asked for accessibility feedback on this, but I haven't heard back yet. See also change to old line 463.

assets/src/components/d3/createResourceAccessChart.js
```diff
@@ -201,7 +201,12 @@ function createResourceAccessChart ({ data, weekRange, gradeSelection, resourceT
         ? accessedResourceColor
         : notAccessedResourceColor
       )
+      .on('focus', (e, d) => {
+        moveBrushOnFocus(e, d.resource_name)
+        toolTip.show(e, d)
+      })
```
ssciolla, line 207: Because the handlers added use `focus`, they also trigger on mouse click as well as keyboard tabbing. There's not a simple way (I've found) to determine whether it was triggered by mouse or tabbing. The result is that if users click on bars, the brush and chart may jump a little, but I've tried to make the new location good for visibility of the rest of the chart. This will need some testing.

ssciolla:
The Codacy issues are related to using variables (`gBrush`, `moveBrushOnFocus`) before they are defined (which is allowed by JavaScript, but leads to some code that's difficult to read). I looked into to fixing this, but the entire D3 component file has this issue, and references are kinda circular. The situation seems to be partially encouraged by D3, but maybe there is a way to partially build things like the brush and then add the handlers after the fact, or use parameters. I'm not sure, I'd have to think about it more, but it will involve a lot more code changes, which may not be acceptable at this point in the release cycle. Opening this as a new issue would make sense to me.

assets/src/components/d3/createResourceAccessChart.js
```diff
@@ -450,17 +474,17 @@ function createResourceAccessChart ({ data, weekRange, gradeSelection, resourceT
     .call(truncate, resourceLabelWidth)
 
   // Add links and icons to Y axis
-  d3.selectAll('.axis--y .tick').each(function (d) {
+  d3.selectAll('.axis--y .tick').each(function (d, i) {
```
jonespm, line 477: I don't see where the this new variable `i` is used?

ssciolla: Hm, yeah, it's not, nice catch, I was using it before but my implementation of this changed a few times and this is no longer needed. I will remove.

ssciolla: Done https://github.com/tl-its-umich-edu/my-learning-analytics/pull/1472/commits/3b5e1823abb0dabba0bdab4cbd58c71c921946dd

jonespm:
I tested this and it seems like it works better for keyboard only. It allows the list to be navigated with keyboard which it wasn't already. 

I feel like adding the extra data visible in the graph about whether it was visited or not visited as well as the percentage of student to an aria-label could make this more useful, but perhaps that's a different issue. I was just testing this with voiceover and it just reads the file name when tabbing. It also gives a full overview when first visiting the page but nothing when you start tabbing. 

Another thing I noticed is that if you use the mouse and click on anything in the svg, _then_ start tabbing (on the current version) it actually _does_ tab between the name and the "Graphics Symbol". However in this code if you click on it it first goes through all of the names first then goes through the bar graphs. However I don't know how to get that svg into the tab order so it works?

jonespm (approved):

ssciolla:
> Another thing I noticed is that if you use the mouse and click on anything in the svg, _then_ start tabbing (on the current version) it actually _does_ tab between the name and the "Graphics Symbol". However in this code if you click on it it first goes through all of the names first then goes through the bar graphs. However I don't know how to get that svg into the tab order so it works?

@jonespm, I didn't understand what you were saying at first, but yes, this is what I was talking about in the meeting earlier. The current version does allow you to tab through label to bar, but it's using positive `tabindex` values. What that means is they go first in the tab order, so they appear out of order with the other Tabl-able controls in the UI. Using pattern also isn't recommended. See my comment above here: https://github.com/tl-its-umich-edu/my-learning-analytics/pull/1472#discussion_r1086807868 I'm checking with the a11y team that my change is preferred.

ssciolla:
@zqian, here's a demo of the behavior. Data shown is fabricated data from a testing site.

https://user-images.githubusercontent.com/35741256/219432545-390a6c76-7e2f-453a-9c14-1c70afe9dc92.mov

PR #1474, opened 2023-01-26
Fixes #1473 - User can't select cards on dashboard (views) via keyboard

### LINKED ISSUE #1473: User can't select cards on dashboard (views) via keyboard

The list of views should be keyboard accessible, i.e. you should be able to use Tab and Shift + Tab to navigate through them and back, and press Enter to select one.

The user can Tab through the list of view cards, but pressing Enter when the view cards is in focus does not send the user to the corresponding view.

Looking at the source code, it looks like this was partially the result of an attempt to include the help link in the card. Nested `a` tags are invalid HTML I guess, so the solution at the time was to separately link different parts of the card (title, description, image. etc.) but give them all `tabindex=-1`. The card itself still has `tabindex=0` but there is no link attached to it.

Some initial research suggests that one solution may to use CSS to layer the help link on top of card. See https://css-tricks.com/nested-links/

jonespm:


jonespm:
Close in favor of #1475

PR #840, opened 2020-01-08
#301 Adding support for function annotations (mypy)

zqian:
try to get the https://github.com/tl-its-umich-edu/my-learning-analytics/pull/816 modified and merged in

PR #1477, opened 2023-01-31
Fix use of log warning of added courses by creating separate string ID list (#1476)

### LINKED ISSUE #1476: Log warning about courses added during cron is falsely reporting all courses

The log warning near the end of the run of `cron.py` should warn users only about courses that were added to the system during the cron run.

During cron runs, all of the course IDs are reported by this log message. This seems to be because of a cast of course IDs 
to strings added in this commit: https://github.com/tl-its-umich-edu/my-learning-analytics/commit/141081a04ee92828b0c5029f7b2cb58224919692

ssciolla:
This PR aims to resolve #1476.

Note that the new `self.valid_locked_str_course_ids` is not used in all cases. There are a number of other usages of `self.valid_locked_course_ids` (with querysets and BQ event queries) where I think integers make sense.

I tested with both UDW and UDP and didn't see any issues. You can test adding a new course by launching one course, starting the cron, and then launching another course during the cron run.

### DIFF
--- a/dashboard/cron.py
+++ b/dashboard/cron.py
@@ -115,8 +115,7 @@ class DashboardCronJob(CronJobBase):
     def __init__(self) -> None:
         """Constructor to be used to declare valid_locked_course_ids instance variable."""
         super().__init__()
-        self.valid_locked_course_ids: List[int]
-        self.valid_locked_str_course_ids: List[str]
+        self.valid_locked_course_ids: List[str]
 
     # verify whether course ids are valid
     def verify_course_ids(self):
@@ -161,7 +160,7 @@ class DashboardCronJob(CronJobBase):
         status += util_function(
                                 queries['user'],
                                 'user',
-                                {'course_ids': tuple(self.valid_locked_str_course_ids),
+                                {'course_ids': tuple(self.valid_locked_course_ids),
                                 'canvas_data_id_increment': settings.CANVAS_DATA_ID_INCREMENT
                                 })
 
@@ -200,7 +199,7 @@ class DashboardCronJob(CronJobBase):
         # convert int array to str array
         df_attach = pd.read_sql(queries['resource'],
                                 data_warehouse_engine,
-                                params={'course_ids': tuple(self.valid_locked_str_course_ids)})
+                                params={'course_ids': tuple(self.valid_locked_course_ids)})
         logger.debug(df_attach)
         # Update these back again based on the dataframe
         # Remove any rows where file_state is not available!
@@ -448,7 +447,7 @@ class DashboardCronJob(CronJobBase):
         # loop through multiple course ids
         status += util_function(queries['assignment_groups'],
                                 'assignment_groups',
-                                {'course_ids': tuple(self.valid_locked_str_course_ids)})
+                                {'course_ids': tuple(self.valid_locked_course_ids)})
 
         return status
 
@@ -464,7 +463,7 @@ class DashboardCronJob(CronJobBase):
         # loop through multiple course ids
         status += util_function(queries['assignment'],
                                 'assignment',
-                                {'course_ids': tuple(self.valid_locked_str_course_ids)})
+                                {'course_ids': tuple(self.valid_locked_course_ids)})
 
         return status
 
@@ -483,7 +482,7 @@ class DashboardCronJob(CronJobBase):
         status += util_function(queries['submission'],
                                 'submission',
                                 {
-                                    'course_ids': tuple(self.valid_locked_str_course_ids),
+                                    'course_ids': tuple(self.valid_locked_course_ids),
                                     'canvas_data_id_increment': settings.CANVAS_DATA_ID_INCREMENT
                                 })
 
@@ -502,7 +501,7 @@ class DashboardCronJob(CronJobBase):
         # loop through multiple course ids
         status += util_function(queries['assignment_weight'],
                                 'assignment_weight_consideration',
-                                {'course_ids': tuple(self.valid_locked_str_course_ids)},
+                                {'course_ids': tuple(self.valid_locked_course_ids)},
                                 'weight')
 
         logger.debug(status + "\n\n")
@@ -582,7 +581,7 @@ class DashboardCronJob(CronJobBase):
                 status += f'Course {course.id}: updated {", ".join(updated_fields)}\n'
         return status
 
-    def do(self):
+    def do(self) -> str:
         logger.info("** MyLA cron tab")
 
         status = ""
@@ -596,11 +595,10 @@ class DashboardCronJob(CronJobBase):
             status += f"ERROR: Those course ids are invalid: {invalid_course_id_list}\n"
             status += "End cron: " + str(datetime.now()) + "\n"
             logger.info("************ total status=" + status + "/n/n")
-            return (status,)
+            return status
 
         # Lock in valid course IDs that data will be pulled for.
-        self.valid_locked_course_ids = [x for x in course_verification.course_data['id'].to_list()]
-        self.valid_locked_str_course_ids = [str(x) for x in self.valid_locked_course_ids]
+        self.valid_locked_course_ids = [str(x) for x in course_verification.course_data['id'].to_list()]
 
         logger.info(f'Valid locked course IDs: {self.valid_locked_course_ids}')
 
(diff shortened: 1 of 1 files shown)


jonespm:
Do you think there is value to having this separate integer array? I feel like it might cause confusion and thought this new bug was just around the specific place where the string set from the internal database was compared to this one?

If we just changed `self.valid_locked_course_ids` to just be a `List[str]` rather than adding a new variable would that break anything besides this check, which we could fix by casting the other value (supported_courses) to a string, which could probably also just be a class variable. This seems like how it was working, we probably just aren't running any type enforcement checkers. 

This is the problem of lack of typing hints in this code and why I was bummed we couldn't get https://github.com/tl-its-umich-edu/my-learning-analytics/pull/840 in a few years back.

ssciolla:
> Do you think there is value to having this separate integer array? I feel like it might cause confusion and thought this new bug was just around the specific place where the string set from the internal database was compared to this one?
> 
> If we just changed `self.valid_locked_course_ids` to just be a `List[str]` rather than adding a new variable would that break anything besides this check, which we could fix by casting the other value (supported_courses) to a string, which could probably also just be a class variable. This seems like how it was working, we probably just aren't running any type enforcement checkers.

I'm open to the idea of changing `self.valid_locked_course_ids` to a `list[str]`. When it was a list of strings, the other use cases for it (like querysets, search for `self.valid_locked_course_ids`) in the file still worked as far as I could tell.

I dunno, it just seemed like because the app's database stores them as integers, it made sense to keep them in that type for a definitive variable, rather than just change it everywhere cause it's more compatible with our external data sources.

zqian (commented):
This fix works in general, but I no longer getting the log message of "During the run, users added x course(s)", even in fact I did add some new courses in the MyLA Admin UI.

jonespm:
I think the only time you should see that log message would be if there's new courses that were added while the cron was running.

Really thinking about this I don't know how useful this check even is? This cron runs at 3AM and completes in 10 minutes. 99.9% of the time this isn't going to be a problem where a new course gets in during the run. 

Maybe the better fix is just to remove all of this code from `courses_added_during_cron` and the warnings. And since we're casting and dumping the `valid_locked_course_ids` as str into the List we should change the type on that.

ssciolla:
> This fix works in general, but I no longer getting the log message of "During the run, users added x course(s)", even in fact I did add some new courses in the MyLA Admin UI.

@zqian, I tested again, I am seeing the message even when adding the course via the admin UI. Can you provide steps to reproduce?

zqian:
@ssciolla my steps were first added the course in Admin UI; and then ran the MyLA cron.

I guess that's wrong based on the above explanation. I should start the cron and quickly add the course before the cron job finishes.

jonespm:
I think some of the problem here is that the 

```
     def __init__(self) -> None:
        self.valid_locked_course_ids: List[int]
```

doesn't seem to do anything. I'd feel like there should at least be an error from mypy like

`List item 0 has incompatible type "str"; expected "int"  [list-item]`

Do you know what's missing here? Does this have to be declared in the class body earlier than init?

ssciolla:
@jonespm,

> Really thinking about this I don't know how useful this check even is? This cron runs at 3AM and completes in 10 minutes. 99.9% of the time this isn't going to be a problem where a new course gets in during the run.

The warning message doesn't seem that useful, but also not too harmful. We run the cron at 3 a.m. most of the time, but we also run it at 8 a.m. during releases, and we don't know what other institutions will do. Having this show up in the cron would explain why a course doesn't have data loaded yet. But you're right, it's a log message for an edge case, and won't have a huge impact.

> Maybe the better fix is just to remove all of this code from `courses_added_during_cron` and the warnings. And since we're casting and dumping the `valid_locked_course_ids` as str into the List we should change the type on that.

You seem to have a strong opinion about this, so I'll do what your suggesting. I just need to to look at the cases where an integer seems to be expected.

jonespm:
Wow talk about a mypy "gotcha"

> Try adding -> None to the signatures of both methods. Unannotated functions/methods are not checked at all.
https://github.com/python/mypy/issues/2243

Adding return types to the functions in cron.py returns a lot of mypy errors. 

This would have shown this one.

```
diff --git a/dashboard/cron.py b/dashboard/cron.py
index 225eb24..9c7c0fd 100644
--- a/dashboard/cron.py
+++ b/dashboard/cron.py
@@ -581,7 +581,7 @@ def update_course(self, warehouse_courses_data: pd.DataFrame) -> str:
                 status += f'Course {course.id}: updated {", ".join(updated_fields)}\n'
         return status

-    def do(self):
+    def do(self) -> tuple[str]:
         logger.info("** MyLA cron tab")

         status = ""
```

ssciolla:
> Do you know what's missing here? Does this have to be declared in the class body earlier than init?

Not sure. It's hard to get types out of `pandas`, and that's where we're pulling the course data from. I would have expected the string cast to change it, but it doesn't really seem to be doing that.

jonespm:
> The warning message doesn't seem that useful, but also not too harmful. We run the cron at 3 a.m. most of the time, but we also run it at 8 a.m. during releases, and we don't know what other institutions will do. Having this show up in the cron would explain why a course doesn't have data loaded yet. But you're right, it's a log message for an edge case, and won't have a huge impact.

The problem with running at 8 a.m. has always been that the process currently needs to delete all the data in order to run. We could run it a few times a day if this was fixed. I think this will be something we want to figure out and improve when CD2 and the 4 hour data refresh cycle comes up this year though.

> You seem to have a strong opinion about this, so I'll do what your suggesting. I just need to to look at the cases where an integer seems to be expected.

It just seems like the easier fix for the specific bug. But finding more issues with this, like the type hint being wrong and not being checked because nothing has return types coming up. I feel like it's been working fine aside from this code at the end with the type as str (unchecked and wrong) so we could just change it to the expectation. 

I think the better fix would be to add type hints through every place this code could touch, otherwise we can't be 100% sure if it's really correct. But that's a can of worms and better left for a future fix. I think mypy has gotten better than when we first tried to implement it in 2020 with many more built in types and hints toward what needs to be fixed. So maybe it won't be such a task this time.

jonespm:
I appreciate the effort here, just offering an option to get this fix with the least changes. Having two variables seems like it could be confusing in the long term, unless we really needed it. I'm not entirely sure why we need to have str variables in here at all and it _seems_ to work better with those? Maybe there's unnecessary casting going on somewhere else. The source id's look like bigint too.

zqian:
the string value is need in the cron query, in particular UDP context_store keymap.course_offering lms_int_id field is of varchar type.

zqian (approved):

ssciolla:
Just FYI, there are three cases where `self.valid_locked_course_ids` is passed to a queryset method. Django handles this and the queries still seem to work (found this SO post: https://stackoverflow.com/questions/28888283/querying-an-integerfield-with-list-of-strings-using-in-in-django). Also, in the BQ method, it looks like these strings are passed through `db_util.incremented_id_to_canvas_id`, which casts them to `int`s before casting them back to `str`s.

dashboard/cron.py
```diff
@@ -639,12 +639,14 @@ def do(self):
...
+        all_str_course_ids = set(
+            str(x) for x in Course.objects.get_supported_courses().values_list('id', flat=True)
+        )
+        courses_added_during_cron: List[str] = list(all_str_course_ids - set(self.valid_locked_course_ids))
         if courses_added_during_cron:
-            logger.warning(
+            logger.debug(
                 f'During the run, users added {len(courses_added_during_cron)} course(s): {courses_added_during_cron}')
-            logger.warning(f'No data was pulled for these courses.')
+            logger.debug(f'No data was pulled for these courses.')
```
ssciolla, line 649: @jonespm, I kept this in, just downgraded to `debug`. Let me know if you have an objection.

dashboard/cron.py
```diff
@@ -595,7 +595,7 @@ def do(self):
             status += f"ERROR: Those course ids are invalid: {invalid_course_id_list}\n"
             status += "End cron: " + str(datetime.now()) + "\n"
             logger.info("************ total status=" + status + "/n/n")
-            return (status,)
+            return status
```
ssciolla, line 598: The "success" part of this method was returning a simple string, so I just removed the tupling here to fix the mypy error. I don't think this return value is being used anyway, but I don't know `django-cron` as well as @jonespm.

dashboard/cron.py
jonespm: Yeah that seems fine either as a debug or info, makes sense downgrading.

dashboard/cron.py
jonespm: In django-cron is just calls

`                    self.msg = self.cron_job.do()`

So a str seems more appropriate as a return. This is the value that's put into the log.

jonespm (approved):

PR #1485, opened 2023-02-27
#1418 remove cypresss

### LINKED ISSUE #1418: Fix, replace, or remove Cypress tests
Currently, the Cypress tests don't run as is following the provided instructions. It seems like there is a dependency issue with `cypress-plugin-snapshots` (which is unmaintained and not compatible with the latest `cypress` major version). See https://github.com/meinaart/cypress-plugin-snapshots/issues/215 about possible updates. But even if you pin it to the compatible version, many tests don't pass. This is likely either because the dump data changed or snapshots haven't been updated.

I'll add a note about the issues in the relevant docs, but at some point, it would be nice to get this working, replace it with something like Playwright, or just remove it if we don't plan to keep it up to date.

pushyamig:
Fixes #1418


docs/testing.md
```diff
@@ -2,8 +2,7 @@
 
 ## Testing
 
-The application currently uses two frameworks for frontend testing: [Jest](https://jestjs.io/) and
-[Cypress](https://www.cypress.io/). There are plans to implement back-end tests in the future.
+The application currently uses frameworks for frontend testing: [Jest](https://jestjs.io/). There are plans to implement back-end tests in the future.
```
ssciolla, line 5:
```suggestion
The application currently uses one framework for frontend testing: [Jest](https://jestjs.io/). There are plans to implement back-end tests in the future.
```

ssciolla (approved):
Looks good! I suggested a language tweak in the `README`.

Issue #1480, opened 2023-02-13
Optionally allow users to visually read narrative texts for charts


ssciolla:
Thank you for contributing to this project!

### Describe your problem or feature you'd like added

The accessibility team suggested that we may want to make the alternative text or narrative for the Resources Accessed chart optionally visible to users, as it may be useful to some. @jennlove-um also pointed out that this could assist in users in developing data or digital literacy.

### Describe the solution you'd like

I think some kind of button that expands or reveals an area with the text makes sense. We'd have to discuss the layout and language specifics of this. Testing would also be needed to ensure the text is still available to screenreaders whether it is hidden or not.

PR #1486, opened 2023-03-01
I1479 narrative graph text

### LINKED ISSUE #1479: Improvements to narrative text and structure
When the Resources Accessed visualization was reviewed by the accessibility team, they had some recommendations regarding the narrative alternative text for the visualization. Some of these may also apply to the Grade Distribution chart.

- Narrative should identify itself as an alternative to the visualization: This becomes more important with aria-hidden removed from the SVG. Something like "Following paragraphs provide a text description for graphical bar-chart on this page."
- Structure narrative semantically: having the resources in a list instead of together in paragraphs would probably be slightly easier for users to parse.
- Narrative should include if you have accessed a resource in the past: This information is included in the visual graph and folks reading the text description would also likely find it to be useful.

pushyamig:
Fixes #1479


jonespm:
I know this is still draft, but I'd change the `console.log` messages to `console.debug` if you wanted to leave them in there. And remove the commented out code before marking this ready for review. The changes seem pretty good, I wonder how much this feature was used. 

Thanks!

assets/src/components/ResourceChartNarrative.js
```diff
@@ -0,0 +1,43 @@
+import React from 'react'
+
+function ResourceChartNarrative (props) {
+  const { data, weekRange, gradeSelection, resourceType } = props
```
pushyamig, line 4: L3-22 is just the same code pulled out from `d3NarrativeText.js` file

assets/src/components/d3/d3ViewsNarrative.js
```diff
@@ -47,9 +24,8 @@ const createGradesText = (data, bins, gradesSummary, myGrade, firstGradeAfterBin
       }
     }
   }
-  return narrativeTextGrades.courseStats.concat(narrativeTextGrades.binnedGradeText, narrativeTextGrades.courseGrades.toString())
+  return narrativeTextGrades
```
pushyamig, line 27: returning the object instead of string

assets/src/containers/ResourcesAccessed.js
```diff
@@ -220,14 +221,17 @@ function ResourcesAccessed (props) {
...
         <ResourceAccessChart
           data={resourceData}
-          weekRange={weekRange}
-          gradeSelection={resourceGradeFilter}
-          resourceType={resourceTypeFilter}
           aspectRatio={0.3}
-          minHeight={350}
-        />
+          minHeight={350} />
+        </>
```
ssciolla, line 234: Nit-pick: Check this using `standard` linting, seeing some issues with my VSCode setup.

assets/src/components/ResourceChartNarrative.js
```diff
@@ -0,0 +1,40 @@
...
+      ? `${x.resource_name.split('|')[1]} has been accessed by ${x.total_percent}% of students.`
+      : `${x.resource_name.split('|')[1]} of type ${x.resource_type} has been accessed by ${x.total_percent}% of students.`
+  )
+  narrativeTextResources.resourceAccessList = accessedResources.map(x =>
+    resourceType.length === 1
+      ? `${x.resource_name.split('|')[1]} has been accessed by ${x.total_percent}% of students and you accessed ${x.self_access_count} times. The last time you accessed this resource was on ${new Date(x.self_access_last_time).toDateString()}`
+      : `${x.resource_name.split('|')[1]} of type ${x.resource_type} has been accessed by ${x.total_percent}% of students and you accessed ${x.self_access_count} times. The last time you accessed this resource was on ${new Date(x.self_access_last_time).toDateString()}`)
+  return (
+    <div id='resource-view-narrative' className='fa-sr-only' aria-live='polite'>
+      <p>Following paragraphs provide a text description for graphical bar-chart on this page</p>
```
ssciolla, line 24:
```suggestion
      <p>The following paragraphs provide a text description for the graphical bar-chart on this page:</p>
```

assets/src/components/ResourceChartNarrative.js
```diff
@@ -0,0 +1,40 @@
+import React from 'react'
+
+function ResourceChartNarrative ({ data, weekRange, gradeSelection, resourceType }) {
+  const resourceData = data.sort((a, b) => b.total_percent - a.total_percent)
```
ssciolla, line 4: Do you think it makes sense to sort this data higher up in the logic, maybe after it comes back from the API? Then you don't have to repeat this sorting logic in the narrative React component AND the d3 logic.

assets/src/components/d3/createHistogram.js
```diff
@@ -52,11 +52,19 @@ function createHistogram ({ data, width, height, domElement, xAxisLabel, yAxisLa
   // SVG Components
   const main = d3.select(domElement).append('div')
 
-  main.append('div')
+  const narrativeDiv = main.append('div')
     .attr('aria-live', 'polite')
     .attr('id', 'grade-view-narrative')
     .attr('class', 'fa-sr-only')
-    .text(d => narrativeTextGrades)
+    .text(d => 'Following paragraphs provide a text description for graphical bar-chart on this page')
```
ssciolla, line 59:
```suggestion
    .text(d => 'The following paragraphs provide a text description for the graphical bar-chart on this page:')
```

ssciolla (commented):
Looks good. I'll do some testing and then likely approve. Releasing a couple comments/suggestions.

ssciolla:
Looking at the issue, the accessibility team noted "Narrative should include if you have accessed a resource in the past: This information is included in the visual graph and folks reading the text description would also likely find it to be useful."

I guess you determined that was already included, right? Maybe they just missed it.

assets/src/components/ResourceChartNarrative.js
```diff
@@ -0,0 +1,40 @@
...
+  narrativeTextResources.resourcesUnaccessCount = `You have not yet accessed ${unAccessedResources.length} of ${resourceData.length} resources.  `
+  narrativeTextResources.resourcesAccessCount = accessedResources.length !== 0 ? `. You have accessed ${accessedResources.length} of ${resourceData.length} resources.  ` : ''
+  narrativeTextResources.resourcesUnAccessList = unAccessedResources.map(x =>
+    resourceType.length === 1
+      ? `${x.resource_name.split('|')[1]} has been accessed by ${x.total_percent}% of students.`
+      : `${x.resource_name.split('|')[1]} of type ${x.resource_type} has been accessed by ${x.total_percent}% of students.`
+  )
+  narrativeTextResources.resourceAccessList = accessedResources.map(x =>
+    resourceType.length === 1
+      ? `${x.resource_name.split('|')[1]} has been accessed by ${x.total_percent}% of students and you accessed ${x.self_access_count} times. The last time you accessed this resource was on ${new Date(x.self_access_last_time).toDateString()}`
```
ssciolla, line 20:
```suggestion
      ? `${x.resource_name.split('|')[1]} has been accessed by ${x.total_percent}% of students, and you accessed it ${x.self_access_count} times. The last time you accessed this resource was on ${new Date(x.self_access_last_time).toDateString()}`
```

assets/src/components/ResourceChartNarrative.js
```diff
@@ -0,0 +1,40 @@
...
+      : `${x.resource_name.split('|')[1]} of type ${x.resource_type} has been accessed by ${x.total_percent}% of students and you accessed ${x.self_access_count} times. The last time you accessed this resource was on ${new Date(x.self_access_last_time).toDateString()}`)
```
ssciolla, line 21:
```suggestion
      : `${x.resource_name.split('|')[1]} of type ${x.resource_type} has been accessed by ${x.total_percent}% of students, and you accessed it ${x.self_access_count} times. The last time you accessed this resource was on ${new Date(x.self_access_last_time).toDateString()}`)
```

assets/src/components/ResourceChartNarrative.js
```diff
@@ -0,0 +1,40 @@
...
+    <div id='resource-view-narrative' className='fa-sr-only' aria-live='polite'>
+      <p>Following paragraphs provide a text description for graphical bar-chart on this page</p>
+      <p>{narrativeTextResources.weekSelection}</p>
+      <p>{narrativeTextResources.typeOfResource}</p>
+      <p>{narrativeTextResources.gradeFilter}</p>
+      <p aria-labelledby='unaccessedResources'>'List of unaccessed resources'</p>
+      <ul id='unaccessedResources'>
+        {narrativeTextResources.resourcesUnAccessList.map((item, key) => <li key={key}>{item}</li>)}
+      </ul>
+      <p aria-labelledby='accessedResources'>'List of accessed resources'</p>
```
ssciolla, line 32: Maybe these should be "List of resources you have accessed" and "List of resources you have not accessed"?

assets/src/components/ResourceChartNarrative.js
```diff
@@ -0,0 +1,40 @@
...
+  narrativeTextResources.resourcesAccessCount = accessedResources.length !== 0 ? `. You have accessed ${accessedResources.length} of ${resourceData.length} resources.  ` : ''
+  narrativeTextResources.resourcesUnAccessList = unAccessedResources.map(x =>
+    resourceType.length === 1
+      ? `${x.resource_name.split('|')[1]} has been accessed by ${x.total_percent}% of students.`
+      : `${x.resource_name.split('|')[1]} of type ${x.resource_type} has been accessed by ${x.total_percent}% of students.`
+  )
+  narrativeTextResources.resourceAccessList = accessedResources.map(x =>
+    resourceType.length === 1
+      ? `${x.resource_name.split('|')[1]} has been accessed by ${x.total_percent}% of students, and you accessed it ${x.self_access_count} times. The last time you accessed this resource was on ${new Date(x.self_access_last_time).toDateString()}`
+      : `${x.resource_name.split('|')[1]} of type ${x.resource_type} has been accessed by ${x.total_percent}% of students, and you accessed it ${x.self_access_count} times. The last time you accessed this resource was on ${new Date(x.self_access_last_time).toDateString()}`)
```
pushyamig, line 21: Based on the comment https://github.com/tl-its-umich-edu/my-learning-analytics/pull/1486#issuecomment-1460497114, I have included here. 

They did not miss it, I just included this here after reviewing the issue description `and you accessed it ${x.self_access_count} times` This is the only change I made different from earlier narrative text. 

Does this answer your question?

assets/src/containers/ResourcesAccessed.js
pushyamig: I am configured based on Standard. So Now We need to verify why it is different.

assets/src/components/ResourceChartNarrative.js
pushyamig: May be I just need to understand the over all impact from the Graph perspective, since this was done purely for the Narrative and I just repeated this when I made narrative as react component.

assets/src/components/ResourceChartNarrative.js
ssciolla: Oh, I see you did add a little more detail here. Okay, thanks, I was wondering if anything was done specifically to address this. It did say already that "The last time you accessed this resource was ...".

assets/src/containers/ResourcesAccessed.js
ssciolla: Happy to meet with you if you want.

assets/src/components/ResourceChartNarrative.js
```diff
@@ -0,0 +1,40 @@
...
+  const narrativeTextResources = {}
+  narrativeTextResources.weekSelection = `Resources accessed from week ${weekRange[0]} to ${weekRange[1]}. `
+  narrativeTextResources.typeOfResource = `Selected resource type(s) ${resourceType.length === 1 ? 'is' : 'are'} ${resourceType}. `
+  narrativeTextResources.gradeFilter = gradeSelection.toUpperCase() !== 'ALL' ? `Filtering on grades ${gradeSelection}. ` : 'Getting resources across all grades.'
+  narrativeTextResources.resourcesUnaccessCount = `You have not yet accessed ${unAccessedResources.length} of ${resourceData.length} resources.  `
+  narrativeTextResources.resourcesAccessCount = accessedResources.length !== 0 ? `. You have accessed ${accessedResources.length} of ${resourceData.length} resources.  ` : ''
+  narrativeTextResources.resourcesUnAccessList = unAccessedResources.map(x =>
+    resourceType.length === 1
+      ? `${x.resource_name.split('|')[1]} has been accessed by ${x.total_percent}% of students.`
+      : `${x.resource_name.split('|')[1]} of type ${x.resource_type} has been accessed by ${x.total_percent}% of students.`
```
ssciolla, line 16: You could use a similar approach as I suggested for the other access list to simplify.

assets/src/components/ResourceChartNarrative.js
```diff
@@ -0,0 +1,40 @@
...
+      <p>The following paragraphs provide a text description for the graphical bar-chart on this page:</p>
+      <p>{narrativeTextResources.weekSelection}</p>
+      <p>{narrativeTextResources.typeOfResource}</p>
+      <p>{narrativeTextResources.gradeFilter}</p>
+      <p aria-labelledby='unaccessedResources'>List of resources you have not accessed</p>
+      <ul id='unaccessedResources'>
+        {narrativeTextResources.resourcesUnAccessList.map((item, key) => <li key={key}>{item}</li>)}
+      </ul>
+      <p aria-labelledby='accessedResources'>List of resources you have accessed</p>
+      <ul id='accessedResources'>
```
ssciolla, line 33: I think these `aria-labelledby` uses might need to be reversed. The list of resources is being "labeled by" the `p` tag with "List of ", so I think the `aria-labelledby` goes on the list, and the `id` goes on the `p` tag. It looks like this would also need to be changed in `createHistogram`.

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-labelledby

ssciolla: It might also work to use `aria-label` on the `ul` and remove the `p` tag. I kinda like that there's a separate `p` tag since ultimately if we make this visible (#1480), we'd want some text to be shown before the list.

assets/src/components/d3/createHistogram.js
```diff
@@ -52,11 +52,19 @@ function createHistogram ({ data, width, height, domElement, xAxisLabel, yAxisLa
...
+  const narrativeDiv = main.append('div')
     .attr('aria-live', 'polite')
     .attr('id', 'grade-view-narrative')
     .attr('class', 'fa-sr-only')
-    .text(d => narrativeTextGrades)
+    .text(d => 'The following paragraphs provide a text description for the graphical bar-chart on this page:')
+  narrativeDiv.append('span').text(d => narrativeTextGrades.courseStats)
+  narrativeDiv.append('span').text(d => narrativeTextGrades.binnedGradeText)
+  narrativeDiv.append('span').attr('aria-labelledby', 'grades-bins').text(d => 'Below is number of students in each grade range')
+  narrativeDiv.append('ul').attr('id', 'grades-bins').selectAll('li')
```
ssciolla, line 63: See comment about use of `aria-labelledby`.

assets/src/components/d3/createHistogram.js
```diff
@@ -52,11 +52,19 @@ function createHistogram ({ data, width, height, domElement, xAxisLabel, yAxisLa
...
-  main.append('div')
+  const narrativeDiv = main.append('div')
     .attr('aria-live', 'polite')
     .attr('id', 'grade-view-narrative')
     .attr('class', 'fa-sr-only')
-    .text(d => narrativeTextGrades)
+    .text(d => 'The following paragraphs provide a text description for the graphical bar-chart on this page:')
+  narrativeDiv.append('span').text(d => narrativeTextGrades.courseStats)
+  narrativeDiv.append('span').text(d => narrativeTextGrades.binnedGradeText)
+  narrativeDiv.append('span').attr('aria-labelledby', 'grades-bins').text(d => 'Below is number of students in each grade range')
```
ssciolla, line 62:
```suggestion
  narrativeDiv.append('span').attr('aria-labelledby', 'grades-bins').text(d => 'List describing the number of students in each grade range')
```

We may want to check the language here with Jennifer.

ssciolla (commented):
I found a couple of things that need to be addressed, specifically the use of `aria-labelledby` and the missing text in the RA narrative. However, I think it looks good and will approve once those addressed.

assets/src/containers/ResourcesAccessed.js
```diff
@@ -251,8 +257,8 @@ function ResourcesAccessed (props) {
                       min={minMaxWeek[0]}
                       max={minMaxWeek[1]}
                       onWeekChange={onWeekChangeHandler}
-                    />
-                  ) : ''
+                    />)
+                  : ''
```
pushyamig, line 261: This change is reflection of removing the red wiggly lines

assets/src/containers/ResourcesAccessed.js
```diff
@@ -279,8 +285,7 @@ function ResourcesAccessed (props) {
                         onChange={() => setSaveSettingClicked(!saveSettingClicked)}
                         value='checked'
                         color='secondary'
-                      />
-                    )
+                      />)
```
pushyamig, line 288: Same, This change is reflection of removing the red wiggly lines

assets/src/components/ResourceChartNarrative.js
```diff
@@ -0,0 +1,42 @@
...
+    resourceType.length === 1
+      ? `${x.resource_name.split('|')[1]} has been accessed by ${x.total_percent}% of students.`
+      : `${x.resource_name.split('|')[1]} of type ${x.resource_type} has been accessed by ${x.total_percent}% of students.`
+  )
+  narrativeTextResources.resourceAccessList = accessedResources.map(x =>
+    resourceType.length === 1
+      ? `${x.resource_name.split('|')[1]} has been accessed by ${x.total_percent}% of students, and you accessed it ${x.self_access_count} times. The last time you accessed this resource was on ${new Date(x.self_access_last_time).toDateString()}`
+      : `${x.resource_name.split('|')[1]} of type ${x.resource_type} has been accessed by ${x.total_percent}% of students, and you accessed it ${x.self_access_count} times. The last time you accessed this resource was on ${new Date(x.self_access_last_time).toDateString()}`)
+  return (
+    <div id='resource-view-narrative' className='fa-sr-only' aria-live='polite'>
```
ssciolla, line 23: We may want to set `aria-atomic=true` here since VO seems to only be announcing the changes, not the full text. I think it may be confusing if it's not announced fully, but let me know what you think.
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-atomic

assets/src/components/ResourceChartNarrative.js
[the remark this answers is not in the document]
pushyamig: ```(resourceType.length === 1 ? ` of type ${x.resource_type}` : '') ``` I did not see this logic is picked up in the sentence i.e this part is ignored for some reason don't know why. Seem logical

assets/src/components/ResourceChartNarrative.js
pushyamig: Done

assets/src/components/ResourceChartNarrative.js
pushyamig: Done.

assets/src/components/d3/createHistogram.js
pushyamig: Done

assets/src/components/ResourceChartNarrative.js
pushyamig: I do difference when adding this label. Done

assets/src/components/ResourceChartNarrative.js
pushyamig: I will leave it as is now.

assets/src/components/ResourceChartNarrative.js
ssciolla: Okay, cool, we can check this with the accessibility team as well.

assets/src/components/ResourceChartNarrative.js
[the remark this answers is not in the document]
pushyamig: Yes. when user unchecking the `Files, Pages, Video` when he is left with one option this case is handling.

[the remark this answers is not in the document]
ssciolla: Wouldn't this be if the resource had 0 or multiple types?

assets/src/containers/ResourcesAccessed.js
```diff
@@ -228,21 +229,25 @@ function ResourcesAccessed (props) {
...
   const resourceAccessChartBuilder = (resourceData, resourcesLimit) => {
+    const resourceDataSorted = resourceData.sort((a, b) => b.total_percent - a.total_percent)
     if (resourceTypeFilter.length === 0) {
       return (<AlertBanner>Please select a resource type to display data.</AlertBanner>)
-    } else if (!resourceData || Object.keys(resourceData).length === 0) {
+    } else if (!resourceDataSorted || Object.keys(resourceDataSorted).length === 0) {
       return (<AlertBanner>Resource data for your selections is not available.</AlertBanner>)
     } else {
       return (
         <>
```
ssciolla, line 239: This seems to be causing a problem when there is no data returned (I found this because I haven't run the cron recently). Basically, `resourceData` seems to be an object when no data is returned, and an array when it is. This is confusing to me, but perhaps out of scope. I would probably just move the sorting into the `else` clause, like so:

```
    if (resourceTypeFilter.length === 0) {
      return (<AlertBanner>Please select a resource type to display data.</AlertBanner>)
    } else if (!resourceData || Object.keys(resourceData).length === 0) {
      return (<AlertBanner>Resource data for your selections is not available.</AlertBanner>)
    } else {
      const resourceDataSorted = resourceData.sort((a, b) => b.total_percent - a.total_percent)
      return (
        <>
```

ssciolla (commented):
Found an issue.

assets/src/components/ResourceChartNarrative.js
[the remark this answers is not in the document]
pushyamig: If a user check on one type say 'Files' and uncheck 'Pages, Video' then the screen reader will prompt by saying `Selected resource type(s) is Files. ` in the case the list of accessed or unaccessed resources will say 'Syllabus.pdf was accessed 80% of resources`

This was introduced as of part of narrative text initially, I did not introduce this change any changes to this part through this PR. 

is this what you are trying to ask? did I answer your concern?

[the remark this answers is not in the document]
pushyamig: I will keep it for now, next time If step on this I will change it.

ssciolla (approved):
Looks good now! We will likely want to get this reviewed by accessibility folks before it is released.

pushyamig:
Thanks @ssciolla for thorough review, proposing a change narrative as react component instead in d3 and finding issue that I did not account for.

PR #1488, opened 2023-03-03
Fixes #1466 - Limit display to a configurable number of resources

### LINKED ISSUE #1466: Resources Accessed view times out and does not load the bar chart in some courses
Resources Accessed view loads a bar chart showing the percent of students who viewed resources during the selected timeframe.

For certain courses, the view does not load the bar chart.  Developer tools shows a 504 gateway timeout.

### Steps to Reproduce :
1. Go to MyLA for course 556396
2. Click on Resources Accessed
3. The bar chart does not load and a 504 shows up in the developer tools 

This is a large course with 400+ students in it.

jonespm:
Display this limit to the user

I've tested this locally, I need to test if the header works with CORS on the server.

I could have also added this metadata to the JSON but that seems like it would have been a little more work changing the format of the JSON and providing this in the header should work good. 

Okay I tested this out, looks like it works fine.

### DIFF
--- a/assets/src/containers/ResourcesAccessed.js
+++ b/assets/src/containers/ResourcesAccessed.js
@@ -67,6 +67,7 @@ function ResourcesAccessed (props) {
   // this is the filter setting last saved by the user
   const [userSavedFilterSetting, setUserSavedFilterSetting] = useState(resourceGradeFilter)
   const [resourceAccessData, setResourceAccessData] = useState('')
+  const [resourcesLimit, setResourcesLimit] = useState(0)
   const [dataControllerLoad, setDataControllerLoad] = useState(0)
 
   const [userSettingLoaded, userSetting] = useUserSetting(courseId, 'resource')
@@ -178,11 +179,14 @@ function ResourcesAccessed (props) {
   useEffect(() => {
     // Fetch data once all the setting data is fetched
     if (dataControllerLoad === 2 && !(resourceTypeFilter.length === 0)) {
-      const dataURL = `/api/v1/courses/${courseId}/resource_access_within_week?week_num_start=${weekRange[0]}&week_num_end=${weekRange[1]}&grade=${resourceGradeFilter}&resource_type=${resourceTypeFilter}`
+      const dataURL = `/api/v1/courses/${courseId}/resource_access_within_week/?week_num_start=${weekRange[0]}&week_num_end=${weekRange[1]}&grade=${resourceGradeFilter}&resource_type=${resourceTypeFilter}`
       const fetchOptions = { method: 'get', ...defaultFetchOptions }
       fetch(dataURL, fetchOptions)
         .then(handleError)
-        .then(res => res.json())
+        .then((response) => {
+          setResourcesLimit(response.headers.get("Resources-Limit"))
+          return response.json()
+        })
         .then(data => {
           setResourceAccessData(data)
           setDataLoaded(true)
@@ -213,21 +217,29 @@ function ResourcesAccessed (props) {
     }
   }
 
-  const ResourceAccessChartBuilder = (resourceData) => {
+  const ResourceAccessChartBuilder = (resourceData, resourcesLimit) => {
     if (resourceTypeFilter.length === 0) {
       return (<AlertBanner>Please select a resource type to display data.</AlertBanner>)
     } else if (!resourceData || Object.keys(resourceData).length === 0) {
       return (<AlertBanner>Resource data for your selections is not available.</AlertBanner>)
     } else {
       return (
-        <ResourceAccessChart
-          data={resourceData}
-          weekRange={weekRange}
-          gradeSelection={resourceGradeFilter}
-          resourceType={resourceTypeFilter}
-          aspectRatio={0.3}
-          minHeight={350}
-        />
+        <>
+          <div style={{ textAlign: 'center', verticalAlign: 'middle' }}>
+            {typeof resourcesLimit === 'undefined' ? <div>Displaying all available resources in this course</div> :
+              <div>Displaying top {resourcesLimit} most accessed resources in this course</div>
+            }
+          </div>
+
+          <ResourceAccessChart
+            data={resourceData}
+            weekRange={weekRange}
+            gradeSelection={resourceGradeFilter}
+            resourceType={resourceTypeFilter}
+            aspectRatio={0.3}
+            minHeight={350}
+          />
+        </>
       )
     }
   }
@@ -294,7 +306,7 @@ function ResourcesAccessed (props) {
                 successMessage='Resource filter setting saved!'
               />
               {(resourceAccessData && dataLoaded) || resourceTypeFilter.length === 0
-                ? ResourceAccessChartBuilder(resourceAccessData)
+                ? ResourceAccessChartBuilder(resourceAccessData, resourcesLimit)
                 : <Spinner />}
             </Paper>
           </Grid>
--- a/assets/src/containers/ResourcesAccessed.js
+++ b/assets/src/containers/ResourcesAccessed.js
@@ -235,12 +235,8 @@ function ResourcesAccessed (props) {
     } else {
       return (
         <>
-          <Typography style={{ textAlign: 'center', verticalAlign: 'middle' }} gutterBottom>
-            {
-              resourcesLimit === undefined
-                ? 'Displaying all available resources in this course'
-                : `Displaying top ${resourcesLimit} most accessed resources in this course`
-            }
+          <Typography style={{ textAlign: 'center' }} gutterBottom>
+            {`Displaying ${resourcesLimit ?? 'all available'} resources in this course`}
           </Typography>
           <ResourceAccessChart
             data={resourceData}
--- a/assets/src/containers/ResourcesAccessed.js
+++ b/assets/src/containers/ResourcesAccessed.js
@@ -10,6 +10,7 @@ import Grid from '@material-ui/core/Grid'
 import MenuItem from '@material-ui/core/MenuItem'
 import Paper from '@material-ui/core/Paper'
 import Select from '@material-ui/core/Select'
+import Typography from '@material-ui/core/Typography'
 import AlertBanner from '../components/AlertBanner'
 import IconLabel from '../components/IconLabel'
 import RangeSlider from '../components/RangeSlider'
@@ -67,7 +68,8 @@ function ResourcesAccessed (props) {
   // this is the filter setting last saved by the user
   const [userSavedFilterSetting, setUserSavedFilterSetting] = useState(resourceGradeFilter)
   const [resourceAccessData, setResourceAccessData] = useState('')
-  const [resourcesLimit, setResourcesLimit] = useState(0)
+  const [resourcesLimit, setResourcesLimit] = useState(undefined)
+
   const [dataControllerLoad, setDataControllerLoad] = useState(0)
 
   const [userSettingLoaded, userSetting] = useUserSetting(courseId, 'resource')
@@ -184,7 +186,9 @@ function ResourcesAccessed (props) {
       fetch(dataURL, fetchOptions)
         .then(handleError)
         .then((response) => {
-          setResourcesLimit(response.headers.get("Resources-Limit"))
+          if (response.headers.has('Resources-Limit')) {
+            setResourcesLimit(response.headers.get('Resources-Limit'))
+          }
           return response.json()
         })
         .then(data => {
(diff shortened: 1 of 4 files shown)


assets/src/containers/ResourcesAccessed.js
```diff
@@ -178,11 +179,14 @@ function ResourcesAccessed (props) {
...
-      const dataURL = `/api/v1/courses/${courseId}/resource_access_within_week?week_num_start=${weekRange[0]}&week_num_end=${weekRange[1]}&grade=${resourceGradeFilter}&resource_type=${resourceTypeFilter}`
+      const dataURL = `/api/v1/courses/${courseId}/resource_access_within_week/?week_num_start=${weekRange[0]}&week_num_end=${weekRange[1]}&grade=${resourceGradeFilter}&resource_type=${resourceTypeFilter}`
       const fetchOptions = { method: 'get', ...defaultFetchOptions }
       fetch(dataURL, fetchOptions)
         .then(handleError)
-        .then(res => res.json())
+        .then((response) => {
+          setResourcesLimit(response.headers.get("Resources-Limit"))
+          return response.json()
+        })
```
ssciolla, line 189: The `resourcesLimit` is not likely to change between sessions (or maybe ever once set), so I'm thinking this value should be added to the MyLA globals context processors, processed in `globals.js`, and passed down the component tree as a `prop`. Then we don't need the extra React state and this header. What do you think?

assets/src/containers/ResourcesAccessed.js
```diff
@@ -213,21 +217,29 @@ function ResourcesAccessed (props) {
...
-          aspectRatio={0.3}
-          minHeight={350}
-        />
+        <>
+          <div style={{ textAlign: 'center', verticalAlign: 'middle' }}>
+            {typeof resourcesLimit === 'undefined' ? <div>Displaying all available resources in this course</div> :
+              <div>Displaying top {resourcesLimit} most accessed resources in this course</div>
+            }
+          </div>
+
```
ssciolla, line 233: Nit-pick: We should probably get you set up to use `standard` linting. I think it would flag some of this.

assets/src/containers/ResourcesAccessed.js
```diff
@@ -213,21 +217,29 @@ function ResourcesAccessed (props) {
...
-          weekRange={weekRange}
-          gradeSelection={resourceGradeFilter}
-          resourceType={resourceTypeFilter}
-          aspectRatio={0.3}
-          minHeight={350}
-        />
+        <>
+          <div style={{ textAlign: 'center', verticalAlign: 'middle' }}>
+            {typeof resourcesLimit === 'undefined' ? <div>Displaying all available resources in this course</div> :
+              <div>Displaying top {resourcesLimit} most accessed resources in this course</div>
```
ssciolla, line 230: I think text should go in a different tag than `div`. You could also make use of Material UI's `Typography`.

dashboard/settings.py
```diff
@@ -449,7 +449,8 @@ def apply_env_overrides(env: Dict[str, Any], environ: os._Environ) -> Dict[str,
 
 CONSTANCE_CONFIG = {
     'SURVEY_URL': ('', 'Full URL to Qualtrics survey. If left blank no survey link will display.', str),
-    'SURVEY_TEXT': ('Take Survey', 'Custom text for Qualtrics survey link and title. If left blank will default to "Take Survey". Must also configure SURVEY_URL. For best mobile fit keep this text to under 6 words/30 characters.', str)
+    'SURVEY_TEXT': ('Take Survey', 'Custom text for Qualtrics survey link and title. If left blank will default to "Take Survey". Must also configure SURVEY_URL. For best mobile fit keep this text to under 6 words/30 characters.', str),
+    'RESOURCE_LIMIT': (100, 'Maximum number of resources shown in a course resource access.', int)
```
ssciolla, line 453:
```suggestion
    'RESOURCE_LIMIT': (100, 'Maximum number of resources shown in the Resources Accessed visualization.', int)
```

dashboard/views.py
```diff
@@ -5,9 +5,11 @@
 from datetime import timedelta
 from json import JSONDecodeError
 
+from constance import config
 import jsonschema
 import numpy as np
 import pandas as pd
+
 from django.conf import settings
```
ssciolla, line 13: Nit-pick: The new seems unnecessary. I don't remember if I set this pattern up or not, but I guess I'd stick with the separate alphabetical order for direct and `from` imports.

```suggestion
import jsonschema
import numpy as np
import pandas as pd
from constance import config
from django.conf import settings
```

dashboard/views.py
```diff
@@ -429,10 +431,22 @@ def resource_access_within_week(request, course_id=0):
...
     output_df.drop(columns=['name', 'resource_id_type'], inplace=True)
 
     logger.debug(output_df.to_json(orient='records'))
-    return HttpResponse(output_df.to_json(orient='records'),content_type='application/json')
+    response = HttpResponse(output_df.to_json(orient='records'), content_type='application/json')
+
+    # Add in this header if needed to pass that we limited this to the frontend
+    if total_rows > config.RESOURCE_LIMIT:
+        response['Resources-Limit'] = config.RESOURCE_LIMIT
+        response['Access-Control-Expose-Headers'] = 'Resources-Limit'
```
ssciolla, line 447: Is this standard practice for something like this? Maybe we need the state and other changes in the frontend then, rather than going with the `globals` approach. Let me know what you think.

ssciolla (commented):
This looks like it will work. I'll test it next. Let's have a discussion about the `globals` versus header approach. Your choice may be correct, but I'd like to hear the rationale.

assets/src/containers/ResourcesAccessed.js
```diff
@@ -213,21 +217,29 @@ function ResourcesAccessed (props) {
...
-          data={resourceData}
-          weekRange={weekRange}
-          gradeSelection={resourceGradeFilter}
-          resourceType={resourceTypeFilter}
-          aspectRatio={0.3}
-          minHeight={350}
-        />
+        <>
+          <div style={{ textAlign: 'center', verticalAlign: 'middle' }}>
+            {typeof resourcesLimit === 'undefined' ? <div>Displaying all available resources in this course</div> :
```
ssciolla, line 229: I think you can just check if `resourcesLimit === undefined`.

assets/src/containers/ResourcesAccessed.js
jonespm: I started with globals, but the backend is limiting this per course and globals is for everything. Some courses may not be limited and others might. If someone was enrolled in multiple courses and we wanted to display the specific limit to the student and it was in globals we'd have to do some calculations in the frontend rather than just passing the data from the backend. 

For instance if we passed the limit of 100, the frontend would have to see if there were 100 results and print a message that it was limited, otherwise it would print the total results. So there'd be some logic there.

There's also a small chance say if someone viewed this page before the reload and then again after that there would be more there that wouldn't be updated in the globals. Or since this is set in the admin, an admin could change the limit and get this de-synced. I figured passing it from the backend each time eliminates all these edge cases. 

I think we could also use these headers in the future, and was somewhat inspired by the Canvas pagination headers for trying this out.

assets/src/containers/ResourcesAccessed.js
```diff
@@ -178,11 +179,14 @@ function ResourcesAccessed (props) {
   useEffect(() => {
     // Fetch data once all the setting data is fetched
     if (dataControllerLoad === 2 && !(resourceTypeFilter.length === 0)) {
-      const dataURL = `/api/v1/courses/${courseId}/resource_access_within_week?week_num_start=${weekRange[0]}&week_num_end=${weekRange[1]}&grade=${resourceGradeFilter}&resource_type=${resourceTypeFilter}`
+      const dataURL = `/api/v1/courses/${courseId}/resource_access_within_week/?week_num_start=${weekRange[0]}&week_num_end=${weekRange[1]}&grade=${resourceGradeFilter}&resource_type=${resourceTypeFilter}`
```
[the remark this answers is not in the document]
jonespm: I noticed that when it called this URL without the / it was getting a 301 redirect. I thought that was causing a problem but probably not, just left it in there to avoid that redirect.

What you suggest looks like a better option.

assets/src/containers/ResourcesAccessed.js
jonespm: I had the extension installed but I checked it and looks like it needed a migration because the one I had was deprecated? I figured it wasn't reporting any problems so I was good. Updating and restarting I see it notices some issues. I can resolve that.

assets/src/containers/ResourcesAccessed.js
ssciolla: Screenshot 2023-03-09 at 11 24 33 AM

Looks okay. I'd consider adding a little margin on the bottom after the text. This may be handled for you if you use `Typography`, or you may have more utilities available to you.

assets/src/containers/ResourcesAccessed.js
ssciolla: Yeah, I think the `standard` VSCode plugin changed.

assets/src/containers/ResourcesAccessed.js
ssciolla: Okay, yeah, I thought of the edge cases you alluded to, but you are adding complexity in other ways. I'm okay with this approach if you are.

assets/src/containers/ResourcesAccessed.js
jonespm: Sure I can see if this works. I was going off [Stack Overflow advice](https://stackoverflow.com/questions/4725603/variable-undefined-vs-typeof-variable-undefined) to always use typeof since that won't fail, but I didn't try without. The way this is set it will either be set to some value or not set so easy to test.

assets/src/containers/ResourcesAccessed.js
```diff
@@ -67,6 +67,7 @@ function ResourcesAccessed (props) {
   // this is the filter setting last saved by the user
   const [userSavedFilterSetting, setUserSavedFilterSetting] = useState(resourceGradeFilter)
   const [resourceAccessData, setResourceAccessData] = useState('')
+  const [resourcesLimit, setResourcesLimit] = useState(0)
```
ssciolla, line 70: There may be a bug here, I'm looking into it. I think you meant for the initial state to be `undefined`.

assets/src/containers/ResourcesAccessed.js
ssciolla: I alluded to this in https://github.com/tl-its-umich-edu/my-learning-analytics/pull/1488/files#r1131323094, I think it may never be showing the first option because there is a default in the backend. Somehow it's just always "Displaying top most accessed resources in this course"

ssciolla: Got it. `repsonse.headers.get('Resources-Limit')` is returning `null` if it doesn't exist. Thus it's not `undefined`, and `null` gets passed into the string interpolation where it's ignored.

ssciolla: Something like this could work where you're using the setter.
```
      fetch(dataURL, fetchOptions)
        .then(handleError)
        .then((response) => {
          if (response.headers.has('Resources-Limit')) {
            setResourcesLimit(response.headers.get('Resources-Limit'))
          }
          return response.json()
        })
```

assets/src/containers/ResourcesAccessed.js
ssciolla: I do think this should be `undefined`, but I don't think it's the direct cause of the issue. See conversation elsewhere.

jonespm: That makes sense. Probably also using `null` could work out here for what I'm intending, just would have to check that.

assets/src/containers/ResourcesAccessed.js
jonespm: I can ask at the meeting!

jonespm: Jennifer said it's better to leave both in there so there's no confusion when it doesn't appear. Marking as resolved.

assets/src/containers/ResourcesAccessed.js
```diff
@@ -213,21 +227,30 @@ function ResourcesAccessed (props) {
...
-        <ResourceAccessChart
-          data={resourceData}
-          weekRange={weekRange}
-          gradeSelection={resourceGradeFilter}
-          resourceType={resourceTypeFilter}
-          aspectRatio={0.3}
-          minHeight={350}
-        />
+        <>
+          <Typography style={{ textAlign: 'center', verticalAlign: 'middle' }} gutterBottom>
```
ssciolla, line 238: `verticalAlign` may not be needed here, not sure.

assets/src/containers/ResourcesAccessed.js
```diff
@@ -213,21 +227,30 @@ function ResourcesAccessed (props) {
...
+            {
+              resourcesLimit === undefined
+                ? 'Displaying all available resources in this course'
+                : `Displaying top ${resourcesLimit} most accessed resources in this course`
```
ssciolla, line 242: Ternary could be moved into the string to reduce duplication; "Displaying ... resources in this course" is repeated.

ssciolla (approved):
LGTM! Tested and seems to be working as intended.

assets/src/containers/ResourcesAccessed.js
jonespm: Yeah interesting, looks like I could just use
`            {`Displaying ${resourcesLimit ?? 'all available'} resources in this course`}`

assets/src/containers/ResourcesAccessed.js
jonespm: Yeah doesn't look like it is

PR #1496, opened 2023-03-09
Move to Google Analytics 4 (#1371)

### LINKED ISSUE #1371: Update Google Analytics to support GA4 (By July 2023)
I saw a notification today that Google analytics team is sunsetting current version Universal Analytics and it is replacing with [Google analytics 4](https://support.google.com/analytics/answer/10089681?utm_campaign=2022-q1-gbl-all-gafree&utm_source=google-growth&utm_medium=email&utm_content=ga-sunset-msa-ua-only-admin). I think we are using a react-ga library which is still support [Universal analytics](https://github.com/react-ga/react-ga).

At some point i believe this library will update as well. I don't know what is difference between existing and new one. It looks like we need to configure this GA4 from our GA console as well. I think there may be some work from that in future

Here is the the current possible replacement package: 
 https://github.com/PriceRunner/react-ga4

_Originally posted by @pushyamig in https://github.com/tl-its-umich-edu/my-learning-analytics/issues/1363#issuecomment-1104286765_

ssciolla:
This PR aims to resolve #1371.

I'll have to run this in one of our test deployments to ensure it's working.

Resources
- https://developers.google.com/tag-platform/tag-manager/web/csp
- https://www.npmjs.com/package/react-ga4
- https://django-csp.readthedocs.io/en/latest/nonce.html

### DIFF
--- a/assets/src/components/GoogleAnalyticsTracking.js
+++ b/assets/src/components/GoogleAnalyticsTracking.js
@@ -1,5 +1,5 @@
 import { useState, useEffect } from 'react'
-import GoogleAnalytics from 'react-ga'
+import GoogleAnalytics from 'react-ga4'
 
 function GoogleAnalyticsTracking (props) {
   const {
@@ -18,7 +18,7 @@ function GoogleAnalyticsTracking (props) {
     const page = window.location.pathname + window.location.search + window.location.hash
     if (gaId && page !== previousPage) {
       setPreviousPage(page)
-      GoogleAnalytics.pageview(page)
+      GoogleAnalytics.send({ hitType: 'pageview', page })
     }
   })
 
--- a/assets/src/components/GoogleAnalyticsTracking.js
+++ b/assets/src/components/GoogleAnalyticsTracking.js
@@ -11,7 +11,7 @@ function GoogleAnalyticsTracking (props) {
     setInitialized(true)
     GoogleAnalytics.initialize([{
       trackingId: gaId,
-      gaOptions: { nonce: cspNonce }
+      gaOptions: { nonce: cspNonce, cookieFlags: 'SameSite=None; Secure' }
     }])
   }
 
--- a/assets/src/components/GoogleAnalyticsTracking.js
+++ b/assets/src/components/GoogleAnalyticsTracking.js
@@ -2,16 +2,17 @@ import { useState, useEffect } from 'react'
 import GoogleAnalytics from 'react-ga4'
 
 function GoogleAnalyticsTracking (props) {
-  const {
-    gaId
-  } = props
+  const { gaId, cspNonce } = props
 
   const [initialized, setInitialized] = useState(false)
   const [previousPage, setPreviousPage] = useState(null)
 
   if (gaId && !initialized) {
     setInitialized(true)
-    GoogleAnalytics.initialize(gaId)
+    GoogleAnalytics.initialize([{
+      trackingId: gaId,
+      gaOptions: { nonce: cspNonce }
+    }])
   }
 
   useEffect(() => {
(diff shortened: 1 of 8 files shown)


config/env_sample.hjson
```diff
@@ -27,7 +27,7 @@
         "REPORT_ONLY": true,
         # You should set your domain as the DEFAULT_SRC
         "DEFAULT_SRC": ["'self'","example.edu"],
-        "SCRIPT_SRC": ["'self'", "'unsafe-inline'", "'unsafe-eval'", "www.google-analytics.com"],
+        "SCRIPT_SRC": ["'self'", "'unsafe-inline'", "'unsafe-eval'", "www.googletagmanager.com"],
```
ssciolla, line 30: https://developers.google.com/tag-platform/tag-manager/web/csp

jonespm (approved):
I think this is fine. I feel like passing the nonce isn't super necessary here but looks like you already set it all up so should work. With us still having `unsafe-inline` for the `SCRIPT_SRC` that kind of makes passing a nonce unnecessary. 

I'm not sure if this is the only inline script, if it is and we could remove `unsafe-inline `that'd be good!

config/env_sample.hjson
```diff
@@ -27,14 +27,14 @@
...
+        "SCRIPT_SRC": ["'self'", "'unsafe-inline'", "*.googletagmanager.com"],
+        "IMG_SRC": ["'self'", "data:", "*.google-analytics.com", "*.googletagmanager.com"],
         "OBJECT_SRC": [],
         "MEDIA_SRC": [],
         # If you're embedding in Canvas you may need to include instructure.com as FRAME_SRC
         "FRAME_SRC": [],
         "FONT_SRC": ["'self'", "fonts.gstatic.com"],
-        "CONNECT_SRC": [],
+        "CONNECT_SRC": ["*.google-analytics.com", "*.analytics.google.com", "*.googletagmanager.com"],
         "STYLE_SRC": ["'self'", "'unsafe-inline'"],
```
ssciolla, line 38: We may be able to remove `unsafe-inline` from `style-src` using the nonce added here.

config/env_sample.hjson
```diff
@@ -27,14 +27,14 @@
         "REPORT_ONLY": true,
         # You should set your domain as the DEFAULT_SRC
         "DEFAULT_SRC": ["'self'","example.edu"],
-        "SCRIPT_SRC": ["'self'", "'unsafe-inline'", "'unsafe-eval'", "www.google-analytics.com"],
-        "IMG_SRC": ["'self'", "data:", "www.google-analytics.com"],
+        "SCRIPT_SRC": ["'self'", "'unsafe-inline'", "*.googletagmanager.com"],
```
ssciolla, line 30: Unfortunately, looks like the `pylti1p3`'s `script` tags are preventing us from removing `unsafe-inline` entirely here. It's no longer necessary for Google Analytics since we're passing the nonce.

assets/src/components/GoogleAnalyticsTracking.js
```diff
@@ -1,24 +1,25 @@
...
 
   const [initialized, setInitialized] = useState(false)
   const [previousPage, setPreviousPage] = useState(null)
 
   if (gaId && !initialized) {
     setInitialized(true)
-    GoogleAnalytics.initialize(gaId)
+    GoogleAnalytics.initialize([{
+      trackingId: gaId,
+      gaOptions: { nonce: cspNonce, cookieFlags: 'SameSite=None; Secure' }
```
ssciolla, line 14: The `nonce` is not strictly necessary because we are using `unsafe-inline` but using this now will set us up for the future if we are able to get the rest of the application to use this approach.

PR #1503, opened 2023-04-10
issue_1502 added comment inside cron file for the data freshness date indicator

zqian:
Added comment into the cron_udp.hjson file


pushyamig:
Test pass:
The changes are correctly updated on MylA beta and in Master branch

PR #1507, opened 2023-06-05
Removing old assignment planning view

### LINKED ISSUE #1492: Remove original assignment planning view

Currently this application has 2 assignment planning views, the original/classic view should be removed as it's not tested as much and isn't going to have any new development. It also is confusing to newer installs that see the two views and have to disable one. This view called `show_assignment_planning_v1` and was replaced by the newer assignment goals view. 

The files for the templates in React as well as the API's used should all be removed.

pushyamig:
Fixes #1492 

I found an issue where when trying to change the Course form options from Admin view, you cannot do it and it throws `Course ID None must be an integer value`.  I do see this issue in current prod. So I will treat that as separate issue

I think this is happening since Course ID become a non-editable or readonly field at some point and there is check happening to get canvas_id and it returns None. I did not see a github issue for this, I will create it and will work on this fix separately .


jonespm:
The admin issue noted is addressed in #1509

pushyamig:
This PR has the latest from the Master

ssciolla:
Looks like there might at least one mention of the old view in `.env.sample`.

dashboard/views.py
ssciolla, line 1: Check whether any imports can be removed. I see a few that are unsued, looks like definitely `numpy` can be taken out.

assets/src/containers/AssignmentPlanning.js
```diff
@@ -1,230 +0,0 @@
...
-import Spinner from '../components/Spinner'
-import TableAssignment from '../components/TableAssignment'
-import UserSettingSnackbar from '../components/UserSettingSnackbar'
-import ViewHeader from '../components/ViewHeader'
-import WarningBanner from '../components/WarningBanner'
-import useSetUserSetting from '../hooks/useSetUserSetting'
-import useUserSetting from '../hooks/useUserSetting'
-import { useAssignmentData } from '../service/api'
-import { getCurrentWeek } from '../util/data'
-import { isObjectEmpty } from '../util/object'
```
ssciolla, line 22: It looks like these two utilities may not be used elsewhere and can be removed from those files.

ssciolla (commented):
I think this looks pretty good, I ran it and I didn't see any errors. There are I think a few mentions of the view or code that's unused. In addition to what is mentioned in the comments, please take a look in the frontend at `/components/AvatarModal.js`, `/components/Tooltip.js`, and `/hooks/useFetch`.

assets/src/containers/AssignmentPlanning.js
pushyamig: These are used in tests as of now. So I will leave it. `getCurrentWeek` is used with assignment V2.

dashboard/views.py
pushyamig: I removed numpy and User objects

dashboard/common/db_util.py
```diff
@@ -85,12 +85,11 @@ def get_course_view_options (course_id):
...
         with django.db.connection.cursor() as cursor:
-            cursor.execute("SELECT show_resources_accessed, show_assignment_planning_v1, show_assignment_planning, show_grade_distribution FROM course_view_option WHERE course_id = %s", [course_id])
+            cursor.execute("SELECT show_resources_accessed, show_assignment_planning, show_grade_distribution FROM course_view_option WHERE course_id = %s", [course_id])
             row = cursor.fetchone()
             if (row != None):
                 course_view_option = {}
                 course_view_option['show_resources_accessed'] = row[0] and 'show_resources_accessed' not in settings.VIEWS_DISABLED
-                course_view_option['show_assignment_planning_v1'] = row[1] and 'show_assignment_planning' not in settings.VIEWS_DISABLED
                 course_view_option['show_assignment_planning'] = row[1] and 'show_assignment_planning' not in settings.VIEWS_DISABLED
                 course_view_option['show_grade_distribution'] = row[2] and 'show_grade_distribution' not in settings.VIEWS_DISABLED
```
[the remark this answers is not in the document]
pushyamig: I will keep this change and won't disturb the existing flow and focus on removing the assignment v1 elements. I will keep you suggestion in mind when next time I will step into it.

assets/src/containers/AssignmentPlanning.js
ssciolla: Hm, sorry, I must have missed those.

pushyamig:
> I think this looks pretty good, I ran it and I didn't see any errors. There are I think a few mentions of the view or code that's unused. In addition to what is mentioned in the comments, please take a look in the frontend at /components/AvatarModal.js, /components/Tooltip.js, and /hooks/useFetch.

 /components/AvatarModal.js: Has link to assignment V1, which I removed it
/components/Tooltip.js,: I removed the entire file as this was used for assignment purpose
 /hooks/useFetch. Removed the extra logic to support assignment planning

pushyamig:
@ssciolla I addressed all the issues raised, please review and comment.

pushyamig:
This branch is up to date with master changes #1518

pushyamig:
I have this branch latest from master

ssciolla (approved):
Looks good. I think this can move on to testing/QA.

pushyamig:
This PR is going to make some changes to model.

PR #1508, opened 2023-06-06
issue_1493 remove cron.hjson and mention of UDW

zqian:
- removed cron.hjson based on UDW data
- removed the data validation script comparing UDW vs UDP query results


docs/getting_started.md
```diff
@@ -32,9 +32,9 @@ To follow the instructions below, you will at minimum need the following:
     cp config/env_sample.hjson ~/mylasecrets/env.hjson
     ```
 
-1. Copy the `config/cron.hjson` file into `mylasecrets`.
+1. Copy the `config/cron_udp.hjson` file into `mylasecrets`.
```
jonespm, line 35: I think we could indicate that this is an optional step. If you don't do it then MyLA will use the default configuration. I believe the default should work fine for most schools on the UDP.

jonespm (approved):

PR #1510, opened 2023-06-09
#1288 limiting GD view based on course enrollment

### LINKED ISSUE #1288: Add configuration option for administrators to restrict views available based on number of students in class
For small classes, maintaining student privacy becomes a concern with data visualizations. This becomes most important for the Grade Distribution visualization. In a small class, there is a risk that students could deduce an individual student's grade based on the histogram data.

Provide a way for MyLA administrators to limit which views are available to be added to courses based on a minimum number of students in the course.  For example, allow administrators to set a threshold of 11 students, indicating that courses with 11 or fewer students would not be able to enable Grade Distribution.  

Currently, institutions are making recommendations to instructors via documentation to let them know what minimum enrollment should be considered when turning on views.  Adding an administrative configuration would allow institutions to enforce enrollment thresholds for the Grade Distribution view.

Further discussion may be needed to determine whether the threshold needs to apply to only the Grade Distribution view or if it should also apply to the Resources Accessed view.

pushyamig:
Fixes #1288

### DIFF
--- a/dashboard/views.py
+++ b/dashboard/views.py
@@ -461,8 +461,8 @@ def grade_distribution(request, course_id=0):
             'enrollment_type': 'StudentEnrollment'
         })
     if len(df) <= config.GRADE_DISTRIBUTION_MINIMUM:
-        logger.error(f"Course enrollment count {len(df)} is lower than Grade distribution minimum limit {config.GRADE_DISTRIBUTION_MINIMUM}")
-        return HttpResponse(json.dumps({'gd_disable':'true','gd_msg': f'Grade Distribution view is disabled due to minimum course enrollment less than {config.GRADE_DISTRIBUTION_MINIMUM}'}), content_type='application/json')
+        logger.error(f"Course enrollment count {len(df)} is lower than the minimum required to display the Grade Distribution visualization {config.GRADE_DISTRIBUTION_MINIMUM}")
+        return HttpResponse(json.dumps({'gd_disable':'true','gd_msg': f'Grade Distribution view is disabled because the course enrollment is less than {config.GRADE_DISTRIBUTION_MINIMUM}'}), content_type='application/json')
     if df.empty or df.count().current_grade < 6 :
         logger.info(f"Not enough students grades (only {df.count().current_grade}) in a course {course_id} to show the view")
         return HttpResponse(json.dumps({}), content_type='application/json')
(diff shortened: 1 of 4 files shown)


dashboard/views.py
```diff
@@ -467,7 +467,7 @@ def grade_distribution(request, course_id=0):
             'course_id': course_id,
             'enrollment_type': 'StudentEnrollment'
         })
-    if df.empty or df.count().current_grade < 6:
+    if df.empty or df.count().current_grade < 6 or config.GRADE_DISTRIBUTION_LIMIT > len(df):
```
jonespm, line 470: I think we should have a different warning here if it gets blocked by the limit? What do you think?

Like if he grades are less than 6 display this error message.

But if the `config.GRADE_DISTRIBUTION_LIMIT > len(df)` say something like
Course enrollments for this course are less than the configured minimum of {config.GRADE_DISTRIBUTION_LIMIT}.

Also is this logic correct? Are we blocking it if the LIMIT is greater than the length? The default limit (which describes a minimum) is set to 10. So this seem like it would return this error message if there was 11?

I feel like this should be reversed.

Also maybe the name of `GRADE_DISTRIBUTION_LIMIT` should be `GRADE_DISTRIBUTION_FLOOR` or `GRADE_DISTRIBUTION_MINiMUM` as it seems like this is a lower limit?

jonespm (changes requested):

dashboard/views.py
pushyamig: I made changes based on your comment, Please review

ssciolla:
I'm going to let @jonespm handle this one for now, but let me know if I'm needed.

zqian:
@pushyami would you please sync the PR? I am getting errors now because the code is behind since the recent drop of field show_assignment_planning_v1 in course_view_option.

`OperationalError at /admin/dashboard/course/
(1054, "Unknown column 'course_view_option.show_assignment_planning_v1' in 'field list'")`

pushyamig:
@zqian The PR should be in sync with master.

zqian:
@pushyamig Got the following error trying to checkout this branch:

```
➜  my-learning-analytics git:(master) ✗ gh pr checkout 1510
From https://github.com/tl-its-umich-edu/my-learning-analytics
 ! [rejected]        refs/pull/1510/head -> i1288_admin_configure_option_GD  (non-fast-forward)
failed to run git: exit status 1
```

PR #1512, opened 2023-06-15
Fixes #1513 - Switch base image to (Debian) bookworm

### LINKED ISSUE #1511: Switch build to reference (Debian) bullseye explicity
Build should complete without issue

This week the Python base image was switched over from (Debian) bullseye to bookworm. This is causing the builds to fail. 

The easiest fix would be to switch it to bullseye as we did with another of our apps but we'd have to upgrade this within the next year or so.

I'll investigate this and see what would be the best fix since we're a little away from a release still for MyLA.

### LINKED ISSUE #1513: Upgrade python/node Dockerfile base to use bookworm

Debian bullseye which was set explicitly in #1511 [ends security support](https://endoflife.date/debian) in July 2024. I couldn't easily switch over to it in #1511 because MariaDB did not yet have packages for it. I'm guessing it will in the next 6 months or so.

Switch the base python/node image over to bookworm.

jonespm:
This is draft for now until MariaDB supports Bookworm 
https://mariadb.com/kb/en/installing-mariadb-deb-files/

### DIFF
--- a/dockerfiles/Dockerfile
+++ b/dockerfiles/Dockerfile
@@ -30,7 +30,7 @@ RUN npm prune --production && \
     find node_modules -type d -name "examples" -print0 | xargs -0 rm -rf
 
 # FROM directive instructing base image to build upon
-FROM python:3.10-slim-bullseye AS app
+FROM python:3.10-slim-bookworm AS app
 
 # EXPOSE port 5000 to allow communication to/from server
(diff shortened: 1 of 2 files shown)


jonespm:
MariaDB got support on 2023-08-14 but going to merge this with #1545 instead

PR #1517, opened 2023-06-22
I1515 node 18 upgrade

### LINKED ISSUE #1515: Upgrade to Node 18
Node 16 is currently in maintenance mode and will reach end of life in Sept. 2023. It's time to update to 18.

https://github.com/tl-its-umich-edu/my-learning-analytics/blob/90ed091b50dd7c536d25618d464ec14b2b248eac/dockerfiles/Dockerfile#L3

Resources
- https://github.com/nodejs/release#release-schedule

pushyamig:
Fixes #1515 

I have updated to `18-bullseye-slim` upon some recommendation from Container service member Erik. 

> 18-bullseye-slim (which is currently in the registry) it's tied to a major version / specific Debian release.
> 
> We have been using 16-slim and alternative for it is 18-slim. it looks like [18-slim refers to Debian 12](https://hub.docker.com/_/node/) at the moment. If you wanted to pin it to that specific Debian release you could use 18-bookworm-slim. (Still might need to be imported to the OpenShift registry; but the idea is you avoid a situation in a couple years where 18-slim suddenly points to Debian 13, possibly breaking stuff.)

I Opted for bullseye as the current python version adaptation. Build are tested on MyLA Beta.  I am happy to change whatever version team agrees.

### DIFF
--- a/dockerfiles/Dockerfile
+++ b/dockerfiles/Dockerfile
@@ -1,6 +1,6 @@
 # NOTE: When updating this file, also update Dockerfile.openshift
 # build react components for production mode
-FROM node:16-slim AS node-webpack
+FROM node:18-slim AS node-webpack
 WORKDIR /usr/src/app
 
 # NOTE: package.json and webpack.config.js not likely to change between dev builds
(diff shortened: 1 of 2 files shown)


ssciolla (approved):
I was able to build locally and changes were picked up. I also was able to build using `docker-compose-prod.yml`, but there seemed to be some kind of configuration issue when actually trying to run it with `up`. I don't think this is an issue related to the image change.

pushyamig:
@jonespm , if you don't have any concerns with this version of Node, I can merge the change.

jonespm:
Nope, seems fine keeping them together if they have this in the repo. Thanks!

jonespm (approved):

Issue #1509, opened 2023-06-08
Courses are not editable via Django Admin


ssciolla:
After my PR #1468, it seems users are blocked from editing courses using the Django admin interface. This is because of a custom cleaning method defined on `CourseForm` that expects for `canvas_id` to always be defined.

https://github.com/tl-its-umich-edu/my-learning-analytics/blob/90ed091b50dd7c536d25618d464ec14b2b248eac/dashboard/admin.py#L53-L58

Possible solution include modifying the cleaning logic to only run on course creation, or removing the form in its entirety.

ssciolla:
There is another bug here that needs to be resolved. The previous solution for this problem seems to change the `readonly_fields` for this form for the remainder of the session. This becomes a problem because when you edit a course, and then try to create a new one, the `canvas_id` field is not editable, but it can't be null. The solution here I think is to override the`get_readonly_fields` method rather than changing the `readonly_fields` attribute.

jennlove-um:
Ready for testing.

jennlove-um:
Tested in beta.  I am able to create, update, and delete courses in the admin page.

PR #1518, opened 2023-06-28
Fix ability to edit courses via the admin and `canvas_id` readonly logic, improve validation (#1509)

ssciolla:
This PR fixes two bugs introduced by my work on PR #1468. It tries to do the following:

- Make courses editable again via the Django Admin interface by modifying the `clean` method in `CourseForm` to only validate `canvas_id` on course creation (`canvas_id` was always left out when editing since it was read only, which lead to errors being thrown).
- Fix a side-effect of the previous change where when a course was edited, subsequent creation of new courses still lableed `canvas_id` as read only. This was fixed by using the `get_readonly_fields` method rather than updating the instance variable.
- Try to improve the validation by disallowing zero and negative Canvas IDs. This wasn't in response to any bugs created by the previous change, but instead exploration of edge cases it makes sense to handle.

Test cases
- [ ] Create a new course with a valid course ID. Subsequently editing the course is possible (i.e. with the start and end times), but `canvas_id` is read only.
- [ ] Try to create a new course with the following invalid values. All should fail with appropriate error messages.
  - [ ] a decimal
  - [ ] a negative integer and/or decimal
  - [ ] zero
  - [ ] no entry
- [ ] Edit an existing course (maybe created automatically by a launch), and then try to create a new course. You should be able to edit `canvas_id` without issue.

If you think of more cases to test, please let me know.

### DIFF
--- a/dashboard/admin.py
+++ b/dashboard/admin.py
@@ -1,9 +1,10 @@
-from typing import Tuple
+from typing import Any, Tuple
 
 from django import forms
 from django.conf import settings
 from django.contrib import admin
 from django.forms.models import ModelForm
+from django.http.request import HttpRequest
 from django.template.defaultfilters import linebreaksbr
 from django.utils.html import format_html
 from django.utils.safestring import mark_safe
@@ -54,8 +55,15 @@ class CourseForm(forms.ModelForm):
         # Only validate canvas_id on initial course creation
         if self.instance.canvas_id is None:
             canvas_id = self.cleaned_data.get('canvas_id')
-            if not str(canvas_id).isdigit():
-                raise forms.ValidationError('Canvas ID must be a positive integer value.')
+            exception = forms.ValidationError(
+                f"Canvas Course ID {canvas_id} must be a positive integer value."
+            )
+            try:
+                canvas_id_int = int(canvas_id)
+            except TypeError:
+                raise exception
+            if canvas_id_int < 1:
+                raise exception
         return self.cleaned_data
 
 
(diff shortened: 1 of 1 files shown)


pushyamig:
I will review this today. Thanks for the Fix @ssciolla

dashboard/admin.py
```diff
@@ -94,9 +103,11 @@ def _courseviewoption(self, obj):
...
         return format_html('<a href="{}">Link</a>', obj.absolute_url)
 
-    def change_view(self, request, object_id, form_url='', extra_content=None):
-        self.readonly_fields = self.readonly_fields + ('canvas_id',)
-        return super(CourseAdmin, self).change_view(request, object_id)
+    def get_readonly_fields(self, request: HttpRequest, obj: Any | None = ...) -> tuple[str, ...]:
+        readonly_fields: tuple[str, ...] = self.readonly_fields
+        if obj is not None:
+            readonly_fields += ('canvas_id',)
+        return readonly_fields
```
pushyamig, line 110: This logics work. basically the `obj` contains `canvas_id`. can we make check if the obj contains 'canvas_id` or value of the canvas_id rather than just checking for None?

ssciolla: I'll verify, but I think `obj` is the model instance. I think it's `None` if we're creating a new object.

ssciolla: Verified. `obj` is `None` on create, `Course` when modifying.

pushyamig: Ok .

pushyamig (approved):
LGTM! Thanks

ssciolla:
Thanks, @pushyamig!

PR #1522, opened 2023-07-20
Fix the Openshift build support ROSA

pushyamig:
The Openshift build is failing since image registry path changed. we need to update the docker build to fix the error below

```
Trying to pull docker-registry.default.svc:5000/openshift/node:16-slim...
Warning: Pull failed, retrying in 5s
```


pushyamig:
If this #1523 get merged we don't need this PR

jonespm:
We'll plan to merge #1523 for the next release so closing

PR #1531, opened 2023-08-25
Removing docker compose prod and path change to regular on

pushyamig:
I am removing docker compose prod, looks we haven't being using it and minor path change to regular docker compose after moving out of folder (which is by missed with #1512


zqian (approved):
The change works for me.

PR #1532, opened 2023-08-25
Getting the github Repo from env

pushyamig:


pushyamig:
How to test this: We want to make sure this works for all forked braches

1. Go a pull of this repo to locally and give a gave as you prefer
2. Commit this branch as you local branch to your Myla fork
3. make changes to the branch from `master` to `<what_ever_you_named_it>`
4. This lastest commit should trigger a GH actions build workflow
5. The Build should be successful and see a nice image create from the packages tab

.github/workflows/main.yml
```diff
@@ -3,19 +3,35 @@ name: Docker image CI
...
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
+  
+      - name: Checkout code 
+        uses: actions/checkout@v3
+    
+      - name: Extract branch name
+        id: extract_branch
+        run: echo "BRANCH_NAME=$(basename ${{ github.ref }})" >> $GITHUB_ENV
```
pushyamig, line 21: This is the step the pull the branch it is built from.

.github/workflows/main.yml
```diff
@@ -3,19 +3,35 @@ name: Docker image CI
...
+      
+      - name: build Docker image
+        run: |
+          docker build . --tag ghcr.io/${{ env.REPO_URL }}:${{ env.IMAGE_TAG }}-${BRANCH_NAME}
+
```
pushyamig, line 26: Tag will like `latest-master` or `latest-2020-01.01`

.github/workflows/main.yml
pushyamig: There was an env variable called github.base_ref but some reason it did not work. So I have to adopt this workaround

.github/workflows/main.yml
```diff
@@ -2,20 +2,39 @@ name: Docker image CI
...
-  workflow_dispatch:
+  
+env:
+  IMAGE_TAG: latest
+  REPO_URL: ${{ github.repository }}
+    
 jobs:
   build:
+    # to test a feature, change the repo name to your github id
+    if: github.repository_owner == 'tl-its-umich-edu'
```
pushyamig, line 15: This way we could stop the build for forked branches. I see an action recored momentarily and it skips  https://github.com/pushyamig/my-learning-analytics/actions/runs/6000238741

The better way if we want to skip it completely, that we need to create github PAT https://github.com/peter-evans/repository-dispatch. I think i will skip that far at the moment and keep researching the best way

pushyamig:
@jonespm this is ready for review

pushyamig:
@jonespm This needs a review

.github/workflows/main.yml
```diff
@@ -2,20 +2,41 @@ name: Docker image CI
 
 on:
   push:
-    branches: ['master']
-  workflow_dispatch:
+    # takes muliple branch names
+    branches: 
+      - master
+      - '[0-9][0-9][0-9][0-9].[0-9][0-9].*' # 2021.01.01
```
pushyamig, line 8: This will support build of MyLA release branch if there is [commit](https://github.com/pushyamig/my-learning-analytics/actions/runs/6163072969)

pushyamig: Github Packages [outcomes](https://github.com/pushyamig/my-learning-analytics/pkgs/container/my-learning-analytics)

The tagging is like `latest-2023.09.x`

.github/workflows/main.yml
jonespm: I think this is good for now.

.github/workflows/main.yml
jonespm: Interesting so it auto builds the tags too. That seems convenient. I think the name is fine with the latest in it.

jonespm (approved):
I think this looks good. We'll have to try it out to see how it works with the whole process but I don't have anything I'd change here yet.

PR #1535, opened 2023-09-06
Fixes #1135 - Add a date field to the courses table to capture when the course was added to MyLA

### LINKED ISSUE #1135: Add a date field to the courses table to capture when the course was added to MyLA
Capture the date that an instructor adds MyLA to their course as an LTI.  This data is useful for administrators in troubleshooting issues.

Add a field to the courses table that captures the date and time that a course was added to MyLA.

jonespm:


dashboard/admin.py
```diff
@@ -79,10 +79,10 @@ def has_add_permission(self, request):
 class CourseAdmin(admin.ModelAdmin):
     inlines = [CourseViewOptionInline, ]
     form = CourseForm
-    fields = ('canvas_id', 'name', 'term', 'date_start', 'date_end', 'show_grade_counts', 'show_grade_type', 'data_last_updated')
-    list_display = ('canvas_id', 'name', 'term', 'show_grade_counts', 'course_link', '_courseviewoption', 'data_last_updated')
+    fields = ('canvas_id', 'name', 'term', 'date_start', 'date_end', 'show_grade_counts', 'show_grade_type', 'data_last_updated', 'created_date')
```
zqian, line 82: Should we following the naming format of other date fields? So instead of `created_date`, use `date_created` instead?

zqian (approved):
I have downloaded and tried out the PR.
The migration works fine. 

All existing MyLA courses now have the created_date of the migration time. Please check with Jennifer to see whether she is OK with this. 

I left a minor suggestion. Overall the fix works!

jonespm:
Sure, I didn't notice we had a pattern going, I can change this. 

I set the default to be now so it would back-fill anything that exists to also be now. We could have them all null or some other value. I can check. 

> I have downloaded and tried out the PR. The migration works fine.
> 
> All existing MyLA courses now have the created_date of the migration time. Please check with Jennifer to see whether she is OK with this.
> 
> I left a minor suggestion. Overall the fix works!

zqian (approved):
Thanks for the update! Now the new field is named as `date_created`. All prior MyLA courses have null values in that field; Newly added courses have current dates.

PR #1536, opened 2023-09-08
Disable introspection gql

pushyamig:
This issue [fixes disabling introspection in GQL](https://github.com/tl-its-umich-edu/my-learning-analytics/security/advisories/GHSA-xrjf-m382-gghw)


dashboard/middleware/disableintrospection.py
```diff
@@ -0,0 +1,16 @@
+from graphql import GraphQLField, GraphQLObjectType, GraphQLSchema, GraphQLString
+
+
+class IntrospectionDisabledException(Exception):
+    pass
```
jonespm, line 5: I might consider adding some more context to this exception like this but maybe isn't necessary. I haven't tried to run this but what happens now if it's run?

```
class IntrospectionDisabledException(Exception):
    def __init__(self, message="Introspection is disabled."):
        self.message = message
        super().__init__(self.message)
```

jonespm (approved):
Looks good

dashboard/middleware/disableintrospection.py
pushyamig: Locally, I am allowing introspection since we need to look at schema. But when you are in Prod Mode (Controlled via Django-debug setting), it should disallow

pushyamig: It will throw error message

pushyamig:
@jonespm I have the change on beta for you to test. Please take a look.

dashboard/middleware/disableintrospection.py
```diff
@@ -0,0 +1,21 @@
+from graphql import GraphQLField, GraphQLObjectType, GraphQLSchema, GraphQLString
+
+
+class IntrospectionDisabledException(Exception):
+     """
+     Disabling introspection in production mode
+     """
+     def __init__(self, message="Introspection is disabled."):
+        self.message = message
+        super().__init__(self.message)
```
pushyamig, line 10: @jonespm I added this change as requested

PR #1537, opened 2023-09-11
limiting GD view based on course enrollment

### LINKED ISSUE #1288: Add configuration option for administrators to restrict views available based on number of students in class

pushyamig:
Fixes #1288

### DIFF
--- a/dashboard/views.py
+++ b/dashboard/views.py
@@ -461,8 +461,9 @@ def grade_distribution(request, course_id=0):
             'enrollment_type': 'StudentEnrollment'
         })
     if len(df) <= config.GRADE_DISTRIBUTION_MINIMUM:
-        logger.error(f"Course enrollment count {len(df)} is lower than the minimum required to display the Grade Distribution visualization {config.GRADE_DISTRIBUTION_MINIMUM}")
-        return HttpResponse(json.dumps({'gd_disable':'true','gd_msg': f'Grade Distribution view is disabled because the course enrollment is less than {config.GRADE_DISTRIBUTION_MINIMUM}'}), content_type='application/json')
+        grade_distribution_limit_msg = f'Grade Distribution view is disabled because the course enrollment is less than {config.GRADE_DISTRIBUTION_MINIMUM}'
+        logger.error(f"Course enrollment count {len(df)} Hence the {grade_distribution_limit_msg}")
+        return HttpResponse(json.dumps({'gd_disable':'true','gd_msg': grade_distribution_limit_msg}), content_type='application/json')
     if df.empty or df.count().current_grade < 6 :
         logger.info(f"Not enough students grades (only {df.count().current_grade}) in a course {course_id} to show the view")
         return HttpResponse(json.dumps({}), content_type='application/json')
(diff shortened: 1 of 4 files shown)


pushyamig:
@zqian opened a new PR with same changes. I don't know why you cannot checkout the previous PR locally before.

zqian:
This works for me. Reviewing now! Thanks!

dashboard/views.py
```diff
@@ -460,7 +460,10 @@ def grade_distribution(request, course_id=0):
             'course_id': course_id,
             'enrollment_type': 'StudentEnrollment'
         })
-    if df.empty or df.count().current_grade < 6:
+    if len(df) <= config.GRADE_DISTRIBUTION_MINIMUM:
+        logger.error(f"Course enrollment count {len(df)} is lower than the minimum required to display the Grade Distribution visualization {config.GRADE_DISTRIBUTION_MINIMUM}")
```
zqian, line 464: Can define a message string to be used for both line 464 and 465.

dashboard/views.py
```diff
@@ -460,7 +460,10 @@ def grade_distribution(request, course_id=0):
...
+        return HttpResponse(json.dumps({'gd_disable':'true','gd_msg': f'Grade Distribution view is disabled because the course enrollment is less than {config.GRADE_DISTRIBUTION_MINIMUM}'}), content_type='application/json')
+    if df.empty or df.count().current_grade < 6 :
```
zqian, line 466: is the check for 6 still necessary, given the setting of `config.GRADE_DISTRIBUTION_MINIMUM`?

If so, could we make the number 6 as another admin config variable?

pushyamig: 10 is the number we set as default. But some universities set it 0 (with out knowingly) then the Grade distribution need atleast 6 grades for rest of the depersonalized logic to work. So it need to check for GD to work. 
This is not something new I introduced with this change [it was there from before](https://github.com/tl-its-umich-edu/my-learning-analytics/blob/master/dashboard/views.py#L463). 
I don't feel this needs another configuration value.

dashboard/views.py
pushyamig: OK, I have changed as requested 78eb5635a161fbca8e9143fb2c391879afb7d485

dashboard/views.py
zqian: I see. 

Can we expose the number 6 to the configuration description text? Something like `Minimum number of student enrollments required to show the Grade Distribution visualization. The number needs to be greater than 6 for depersonalized logic to work in the Grade Distribution view. `. WDYT?

pushyamig: Ok, I can do that.

zqian (approved):

PR #1538, opened 2023-09-12
Fixes #1501 - Add username to the export

### LINKED ISSUE #1501: Add the username to the csv output in the MyLA event log data
The log files that are exported include a column called "user" - this data is the id from the auth_user table.  To make this exported data more useful, it should also include the username, which would match the Canvas data.  Administrators who use this data need to be able to connect MyLA users to Canvas users.

Export the log file with 2 columns to identify the user: 
user - the id from the auth_user table
username - the username from the auth_user table

jonespm:


dashboard/admin.py
```diff
@@ -115,8 +116,14 @@ def save_model(self, request, obj, form, change):
         return super(CourseAdmin, self).save_model(request, obj, form, change)
 
 class LogResource(resources.ModelResource):
+
+    # This field is used to get the username from the user model
```
pushyamig, line 120: We have a model named literally meaning `user` and other by `auth_user`. For clarity purposes we can update the comment saying `auth_user`

dashboard/admin.py
```diff
@@ -115,8 +116,14 @@ def save_model(self, request, obj, form, change):
...
+    username = Field(attribute='user__username', column_name='username')
     class Meta:
         model = Log
+        # Show the same fields in the UI
+        fields = ('id', 'timestamp', 'username', 'action', 'extra')
```
pushyamig, line 125: Just curious, the `id` field is not being shown from Django admin console. this is not related to this change, current prod is also the say way. 

we should still pull the `id` field as part of the this? or if we remove it breaks anything?

dashboard/admin.py
```diff
@@ -115,8 +116,14 @@ def save_model(self, request, obj, form, change):
         return super(CourseAdmin, self).save_model(request, obj, form, change)
 
 class LogResource(resources.ModelResource):
+
+    # This field is used to get the username from the user model
+    username = Field(attribute='user__username', column_name='username')
```
pushyamig, line 121: Optional: is the `column_name` here giving a custom name to the data field as part of the `auth_user.username`. If yes, will it make sense to name it as `user_uniq_id`?

dashboard/admin.py
jonespm: I was debating leaving it out of the CSV and just having it match the UI but it feels like in an export having this column makes the data a little easier to deal with. I'm fine either way, maybe a @jennlove-um question.

dashboard/admin.py
jonespm: Yes that's the header in the CSV, otherwise it was `user__username`. Can change it to something else. This just matched up with the column in the other table.

dashboard/admin.py
pushyamig: Ok, I feel we can leave it, it make sense to have in the CSV export

pushyamig (approved):
Some minor suggestion on code comments and username change. The change looks good and works as expected

dashboard/admin.py
jonespm:
```suggestion
    # This field is used to get the username from the auth_user model
```

dashboard/admin.py
jonespm: I'm not sure, it does represent the username column in the auth_user table. We could call it auth_user__username maybe or left it as user__username. I think for purposes of the CSV as a column header it's fine as the shorter username or the other way. I don't think any admin using this is going to care either way.

Issue #1529, opened 2023-08-24
Upgrade to Frontend Dependencies


pushyamig:
It's time we update the Myla Major FE libraries like React 8 and Material UI 5 we are one version behind. We are going to see a possible issue when Upgrading to ApolloClient as listed here https://github.com/tl-its-umich-edu/my-learning-analytics/issues/1525

----------
Test Plan
https://docs.google.com/spreadsheets/d/18c_VDE7wdme8c36Zo5-pjad77Kit8hFz4ylbmFXmLP8/edit#gid=0

jaydonkrooss:
So this is my PR against @pushyamig's branch to address the styling regressions that we saw with the Material UI upgrade: https://github.com/pushyamig/my-learning-analytics/pull/11

Also just noting some other styling differences that Jennifer uncovered that we do **not** need to revert: 
- Assignment Planning: the "clear" and "clear filters" buttons are set to blue when enabled
- Assignment Planning: the "Set a goal" input field is now a outlined text field rather than the standard underlined input

pushyamig:
Adding the test plan https://docs.google.com/spreadsheets/d/18c_VDE7wdme8c36Zo5-pjad77Kit8hFz4ylbmFXmLP8/edit#gid=0

jennlove-um:
Testing passes in beta

PR #1545, opened 2023-10-13
#1513 updating python/node base to bookworm

### LINKED ISSUE #1513: Upgrade python/node Dockerfile base to use bookworm

pushyamig:
Fixes #1513

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # build react components for production mode
-FROM node:18-bullseye-slim AS node-webpack
+FROM node:18-bookworm-slim AS node-webpack
 WORKDIR /usr/src/app
 
 # NOTE: package.json and webpack.config.js not likely to change between dev builds
@@ -20,7 +20,7 @@ RUN apt-get update && \
     rm -rf /usr/src/app/assets/src
 
 # build node libraries for production mode
-FROM node:18-bullseye-slim AS node-prod-deps
+FROM node:18-bookworm-slim AS node-prod-deps
 
 WORKDIR /usr/src/app
 COPY --from=node-webpack /usr/src/app /usr/src/app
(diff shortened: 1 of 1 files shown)


pushyamig:
@jonespm should I update the Node base to bookworm as well?

jonespm:
Sure, I think that would have less impact since that's just being used for the build but you try to update it and if it builds great. Otherwise we might have to do it along with #1529

pushyamig:
I have upgraded both the node/python base images to bookworm. builds fine

pushyamig:
@jaydonkrooss I have added you as an observer, if you are still getting up to speed and not ready to run MyLA it's fine. @jonespm will be the approver.

jonespm:
I had a draft PR #1512 that also was fixing this. It looks about the same but it was updating the Dockerfile.openshift that doesn't exist anymore. So this seems fine. I was waiting on MariaDB.

It looks like they didn't update their webpage but MariaDB added support in 2023-08-14 so we should be all set with that. I can close the other PR.

jonespm (approved):

jaydonkrooss (approved):
builds just fine for me!

Dockerfile
```diff
@@ -40,7 +39,7 @@ WORKDIR /code
 COPY requirements.txt .
 RUN apt-get update && \
     apt-get install -y --no-install-recommends \
-        build-essential curl apt-transport-https libpq-dev netcat jq python3-dev xmlsec1 cron git && \
+        build-essential curl apt-transport-https libpq-dev netcat-traditional jq python3-dev xmlsec1 cron git && \
```
lsloan, line 42: MyLA wasn't using `netcat-traditional` until this change was added?  That seems strange to me, because I'm sure it was added a while ago.  (Maybe I'm thinking of another project.)

pushyamig: This was added this time since the build was breaking with out it.

PR #1552, opened 2023-11-09
issue_1551 added mivideo/kaltura video resources access query into en…

zqian:
…v_sample.json file


pushyamig (approved):
This looks Good! I feel it is ok to add leccap for internal U-M reference here.

Issue #1250, opened 2021-03-24
Supplement color-coding of charts with pattern


GitGators:
Location: Resources Accessed - personal views vs class views - better contrast between the blue and gray area

jennlove-um:
Add a pattern to the bars on the resources accessed.  Work with the ITS Accessibility team, especially Gonzalo Silverio for suggestions.

jaydonkrooss:
Created a branch https://github.com/tl-its-umich-edu/my-learning-analytics/pull/1553 with my initial changes.
Here's what I've proposed, using patterns that ought to contrast with the original colors of the bars:

![Screenshot 2023-11-14 at 12 13 24 PM](https://github.com/tl-its-umich-edu/my-learning-analytics/assets/147423601/94585378-83a6-4c24-a06d-ae8d85554b8f)

I've only edited the resources accessed visualization and nothing else, so let me know if we may need to think about other pages of MyLA to improve or update with this change.

jaydonkrooss:
@jennlove-um That's a fair point, I could see that the solid bar would have enough of a contrasting pattern. I wasn't given explicit design ideas from Accessibility, I was just going off of an example I saw from a blog post originating from one of the resources they gave me: https://keen.io/blog/accessibility-in-data-vis/
I'll bring the solid blue bar back, and if needed I could also try to get the accessibility team's explicit approval of either option.

jennlove-um:
Testing fails for this issue.  With the numbers pushed outside the bars, the numbers are overlapping with the brush slider on the side. 

<img width="581" alt="image" src="https://github.com/tl-its-umich-edu/my-learning-analytics/assets/39634416/e7f5cf16-c37e-40b5-9a74-5f78a75b1bd7">

We need to revisit this to avoid that overlap.

jennlove-um:
Testing passes in beta

PR #1553, opened 2023-11-14
#1250: Add patterns to Resource Access visualizations

jaydonkrooss:
Added new visual indicators on resource access page to address accessibility concerns: https://github.com/tl-its-umich-edu/my-learning-analytics/issues/1250

- "Resources Accessed" bars retain the same secondary color from palette settings (blue)~~, with the addition of small white squares as a dotted pattern~~ EDIT: not necessary, as the solid pattern will be distinguishable with the stripes
- "Resources Not Accessed" bars retain the same negative color from the palette settings (gray), with the addition of diagonal black stripes
- Patterns are visually distinct with compliant contrast ratios according to WCAG 2.1 https://www.w3.org/TR/WCAG21/#non-text-contrast

### DIFF
--- a/assets/src/components/d3/createResourceAccessChart.js
+++ b/assets/src/components/d3/createResourceAccessChart.js
@@ -393,10 +393,10 @@ function createResourceAccessChart ({ data, width, height, domElement }) {
     .on('mousedown.brush', brushcenter)
     .on('touchstart.brush', brushcenter, { passive: true })
 
-  const defs = svg.append('defs')
+  const svgDefinitions = svg.append('defs')
 
   // Clips
-  defs.append('defs')
+  svgDefinitions.append('defs')
     .append('clipPath')
     .attr('id', 'clip')
     .append('rect')
@@ -405,7 +405,7 @@ function createResourceAccessChart ({ data, width, height, domElement }) {
     .attr('height', mainHeight)
 
   // Pattern for not accessed: black stripes on gray background
-  defs.append('pattern')
+  svgDefinitions.append('pattern')
     .attr('id', 'notAccessedPattern')
     .attr('width', 10)
     .attr('height', 10)
--- a/assets/src/components/d3/createResourceAccessChart.js
+++ b/assets/src/components/d3/createResourceAccessChart.js
@@ -404,19 +404,9 @@ function createResourceAccessChart ({ data, width, height, domElement }) {
     .attr('width', mainWidth + mainMargin.left)
     .attr('height', mainHeight)
 
-  // Pattern
+  // Pattern for accessed: white dots on blue background
   defs.append('pattern')
     .attr('id', 'accessed-pattern')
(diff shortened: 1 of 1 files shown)


assets/src/components/d3/createResourceAccessChart.js
```diff
@@ -393,15 +393,34 @@ function createResourceAccessChart ({ data, width, height, domElement }) {
     .on('mousedown.brush', brushcenter)
     .on('touchstart.brush', brushcenter, { passive: true })
 
+  let defs = svg.append('defs');
```
pushyamig, line 396: I see red Squiggly line near this. Myla adapted the Standard based linter https://standardjs.com/. There is a VSCode Plugin for it. I configured VCode to  `ts-standard`. There 2 places in the code needs a change
Screenshot 2023-11-21 at 11 05 59 AM

Screenshot 2023-11-21 at 11 06 13 AM

Screenshot 2023-11-21 at 11 06 17 AM

install the linter and make the change

pushyamig:
Overall this is functional as expected

assets/src/components/d3/createResourceAccessChart.js
```diff
@@ -38,13 +38,13 @@ const toolTip = tip.attr('class', 'd3-tip')
...
-  const legendBoxTextInterval = 15
+  const legendBoxLength = 13
+  const legendBoxTextInterval = 20
   const legendInterval = 20
   const legendY = -50
 
   const legendLabels = [
-    ['Resources I haven\'t viewed', notAccessedResourceColor],
+    ['Resources I haven\'t viewed', 'url(#not-accessed-pattern)'],
     ['Resources I\'ve viewed', accessedResourceColor]
```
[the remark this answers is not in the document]
jaydonkrooss: I think it was just leftover formatting following an example where the css attributes were kebab. I went ahead and switched to camelCase based on looking at the other css assignments in this file.

assets/src/components/d3/createResourceAccessChart.js
jaydonkrooss: gotcha, added standard and fixed some of these issues it was pointing out. Thanks!

assets/src/components/d3/createResourceAccessChart.js
```diff
@@ -393,15 +393,34 @@ function createResourceAccessChart ({ data, width, height, domElement }) {
     .on('mousedown.brush', brushcenter)
     .on('touchstart.brush', brushcenter, { passive: true })
 
+  const defs = svg.append('defs')
```
pushyamig, line 396: Optional : `defs` might be a generic name. I would suggest based on the code what it is doing we can name it something relevant.

pushyamig (approved):
This Looks good to me, Functional and working as expected. Nice Work @jaydonkrooss ! I just left some optional comment feel free to commit if you don't agree.

PR #1556, opened 2023-12-11
remove local-time fields, add time zone (iss. #1490, #1491)

### LINKED ISSUE #1490: Store dates from cron "as-is", remove time conversion from cron

It was noted on some recent fixes like #1484 that MyLA is storing values in local time. We had some discussion that this probably isn't needed and the application could do the conversion in the API (either in GraphQL or in the view) to the value in the settings. This might require some minor migration or we can just require dropping the tables and re-running the cron. 

This would probably be made easier if the old assignment view API's were removed #1492 and the cron for the UDW #1493 was also removed.

### LINKED ISSUE #1491: DateTime fields should be displayed in the users local time

It was noted that times displayed are local to a specific time on the server (America/Detroit by default) and it would probably be useful to have the times match what's in Canvas for the user to their local time. This is mostly for the assignment view. 

Either get and use the custom LTI launch variable `Person.address.timezone` or use the browsers time zone to do the time conversions for the user. This would require storing the times in UTC first locally via #1490

jaydonkrooss:
Changes added:

- Drop local time conversions from data fields (on submissions & assignments) 
- Save LTI time_zone setting in SessionMiddleware to store timezone information 
- Use saved time_zone to implement a time conversion on API level when needed for frontend (assignment due_date in graphql)
- Reimplement frontend consuming due_date times, extract date without doing any UTC conversion


lsloan:
@jaydonkrooss, when I tried to run your migration (which I renamed), I got some errors…

```text
[2023-12-21 09:30:59 -0500] [DEBUG] [utils.py:123] (0.001) SHOW FULL TABLES; args=None
  Applying dashboard.0027_remove_or_change_local_date_fields...
[2023-12-21 09:30:59 -0500] [DEBUG] [schema.py:136] ALTER TABLE `assignment` DROP COLUMN `local_date`; (params ())
[2023-12-21 09:30:59 -0500] [DEBUG] [utils.py:123] (0.020) None; args=()
…
django.db.utils.OperationalError: (1091, "Can't DROP 'local_date'; check that column/key exists")
```

I'm looking into it…

lsloan:
@jaydonkrooss, the changes I've committed saves the user's Canvas time zone to the `User` model.  It looks like this in my local DB:

```text
|id  |user_id          |sis_name|course_id        |current_grade|final_grade|enrollment_type  |time_zone       |
|----|-----------------|--------|-----------------|-------------|-----------|-----------------|----------------|
|1386|17700000000151503|lsloan  |17700000000374350|             |           |TeacherEnrollment|America/New_York|
```

Now we just need to use it when displaying timestamps to the user.

lsloan:
@jaydonkrooss, my changes just save the time zone text as it was supplied by Canvas.  In my case, that is `America/New_York`.  When we use the TZ to display timestamps to the user, it might be inconvenient or inefficient to convert that string to a time zone offset value every time it's used.  So, I can change the code to store that converted value in the `User` model instead.  It can be converted to whichever format you think would be easiest to work with.

I have a couple of ideas.  Which do you prefer, if any?

* string — `'-04:00'` (it may be without the colon)
* number — `-4.0`

assets/src/util/assignment.js
```diff
@@ -207,13 +207,15 @@ const createAssignmentFields = (
...
+        dueDate,
         pointsPossible,
         assignmentGroupId,
         currentUserSubmission
       } = a
 
-      a.week = calculateWeekOffset(courseStartDate, localDate)
+      a.week = calculateWeekOffset(courseStartDate, dueDate)
+      console.log('dueDate: ' + dueDate)
+      console.log('localDate '+ new Date(dueDate).toLocaleDateString())
```
lsloan, line 218: @jaydonkrooss, don't forget to remove these `console.log()` lines as soon as you no longer need them.  Codacy complained about each of them as "Unexpected console statement."  See: https://app.codacy.com/gh/tl-its-umich-edu/my-learning-analytics/pullRequest?bid=39353410&prid=13224791

jonespm:
I just looked at this today, sorry for not keeping up with it. I'm not sure there's actually any need to persist this timezone information from the launch in the database. That would simplify this a little bit. 

This app uses [SessionMiddleware ](https://docs.djangoproject.com/en/5.0/topics/http/sessions/) but we aren't making much use of it. 

So I _think_ in the lti_new.py we should just be able to just do something like

`request.session["time_zone"] = time_zone:`

Then the views could also retrieve that. This would eliminate the model changes and migrations here. 

I haven't tried this and it doesn't look like we're using the sessions much here but I don't see any reason to persist when this will be available on every launch and we can always fall-back to the `settings.TIME_ZONE`. Maybe something to try after the break.

lsloan:
@jonespm, thanks for the feedback.  I agree that less model changes would be better, especially since the time zone is read from each launch request and updated in the model every time.  The model is only used to provide the zone for subsequent requests following the LTI launch.

Session data could be more efficient for this purpose.  I wasn't aware of the session middleware or that we used any session data in MyLA other than what Django itself may need.

So, the idea would be to get the time zone from the LTI launch parameters, but store it in the session rather than the model.

I'll give that a shot and with those changes, Jaydon could work on using the session data in the views.

dashboard/graphql/objects.py
```diff
@@ -142,7 +143,14 @@ def resolve_median_grade(parent, info):
...
             lambda submissions: AssignmentType._median_grade_lambda(parent, info, submissions)
         )
-
+    
+    def resolve_due_date(parent, info):
+        original_due_date = parent.due_date
+        if original_due_date is None:
+            return None
+        new_timezone = pytz.timezone(info.context.session.get('time_zone'))
+        return original_due_date.astimezone(new_timezone)
```
lsloan, line 152: We expect `session['time_zone']` to be set here, but there may be some rare case in which it is not.  If it's not, the call to `session.get('time_zone')` could return `None`, which will cause `pytz.timesone()` to give an error.

We could avoid the error here by returning the default MyLA time zone if it's not set in the session.  Add that to the call like this… `session.get('time_zone', default_time_zone)`.  I haven't looked to see whether the default time zone is available in this part of the code.  It should be added if it's not.

jaydonkrooss: good point, added the default time zone to that script for this case.

assets/src/__tests__/util/date.test.js
jaydonkrooss, line 1: for some reason this test was converting string datetime to a Date object even though the description says dateToMonthDay function takes "input a date time string". Changed this since this reflects the true usage for due_date on assignment view. Also added an edge case that reflects an example "local" timezone we'd see that should reflect the true local due date.

lsloan (changes requested):
Zhen suggested the two of us review each other's code, for expediency.  She wants to get this moved to testing ASAP.

I think your code looks good.  I appreciate the changes you made and your attention to detail.

The review has a red "x" because it didn't pass a Codacy test.  According to it the new complexity is…  an additional blank line in `dashboard/graphql/objects.py`.  If you correct that, we should be all set.

Unless you have any critique of the parts I contributed, you can then merge the PR.

https://app.codacy.com/gh/tl-its-umich-edu/my-learning-analytics/pullRequest?prid=13224791&bid=39353410#newIssuesView

jaydonkrooss:
> Zhen suggested the two of us review each other's code, for expediency. She wants to get this moved to testing ASAP.
> 
> I think your code looks good. I appreciate the changes you made and your attention to detail.
> 
> The review has a red "x" because it didn't pass a Codacy test. According to it the new complexity is… an additional blank line in `dashboard/graphql/objects.py`. If you correct that, we should be all set.
> 
> Unless you have any critique of the parts I contributed, you can then merge the PR.
> 
> https://app.codacy.com/gh/tl-its-umich-edu/my-learning-analytics/pullRequest?prid=13224791&bid=39353410#newIssuesView



> Zhen suggested the two of us review each other's code, for expediency. She wants to get this moved to testing ASAP.
> 
> I think your code looks good. I appreciate the changes you made and your attention to detail.
> 
> The review has a red "x" because it didn't pass a Codacy test. According to it the new complexity is… an additional blank line in `dashboard/graphql/objects.py`. If you correct that, we should be all set.
> 
> Unless you have any critique of the parts I contributed, you can then merge the PR.
> 
> https://app.codacy.com/gh/tl-its-umich-edu/my-learning-analytics/pullRequest?prid=13224791&bid=39353410#newIssuesView

Just pushed the whitespace fixes, thanks for pointing that out. Your changes look good as well, I like the descriptive logs and comments.

assets/src/util/date.js
```diff
@@ -12,13 +12,12 @@ const calculateWeekOffset = (startDateTime, targetDateTime) => {
   return Math.ceil((differenceInDays + 1) / 7)
 }
 
+// Get month day format of date string without timezone conversions
+// eg. timezone 2022-06-23T23:59:00 to 6/23
```
lsloan, line 16: I just wanted to be sure I understand the change here…

You changed it because using the `Date()` constructor uses the browser's time zone to adjust the given ISO 8601 timestamp, which we do not want.  Is that correct?

The comment is slightly confusing.

lsloan (changes requested):
I just have the one clarification question…

lsloan (approved):

Issue #1540, opened 2023-09-19
Students receive the wrong error message if MyLA isn't connected to data in Canvas


jennlove-um:
### Expected behavior (A description of what you expected to happen) : 
When a Teacher has added MyLA to a course but has not clicked on the link to enable the course in MyLA, it is not completely set up for students to use.  When a student clicks on the link before the teacher has clicked on LTI to enable the course in MyLA, they see the following message: This course has not been set up in MyLA. Contact your instructor, who can enable the visualizations by clicking on MyLA in the course navigation.   

### Describe the bug (Tell us what happens instead of the expected behavior) :
Instead, the error that displays in production and test environments is "Something went wrong; please try again later."

### Steps to Reproduce :
1. Add MyLA in the course navigation (do not click on the link once it is added)
2. As a student, click on the MyLA link in the navigation
3. The wrong error message displays

jennlove-um:
The correct error message DOES display if the student tries to go to MyLA in a course where MyLA has not been added at all by changing the course ID in the URL.

jennlove-um:
Add details about what message should display and evaluate priority.

jennlove-um:
The error message should be "This course has not been set up in MyLA. Contact your instructor, who can enable the visualizations by clicking on MyLA in the course navigation."

jennlove-um:
Testing passes in beta

PR #1561, opened 2024-01-25
issue #1540 issue_1540 empty error message was returned when student invokes MyLA before fully setup

zqian:


assets/src/containers/Course.js
```diff
@@ -55,7 +55,7 @@ function Course (props) {
     setSideDrawerState(open)
   }
 
-  if (error.message === 'Not Found') return (<WarningBanner>Course {courseId} has not been set up in MyLA. Contact your instructor, who can enable the visualizations by clicking on MyLA in the course navigation.</WarningBanner>)
+  if (error.message === 'Not Found' || (error.message === '')) return (<WarningBanner>Course {courseId} has not been set up in MyLA. Contact your instructor, who can enable the visualizations by clicking on MyLA in the course navigation.</WarningBanner>)
```
jonespm, line 58: I think Codacy just wants this to be cleaned up to be a little more formatted like this. I can test running this though as you have it. 

```
if (error.message === 'Not Found' || (error.message === '')) {
  return (
    <WarningBanner>
      Course {courseId} has not been set up in MyLA. Contact your instructor, who can enable the visualizations by clicking on MyLA in the course navigation.
    </WarningBanner>
  )
}
```

zqian: Got more warnings after code reformat...

jonespm:
This seems to work but I wonder if the error is further up in the handleError method having an exception! 

image

Checking for an empty string _seems_ like it could catch the wrong error. I have a fix for this on this commit which works to return 404 if "Not Found" isn't available in the statusText.

https://github.com/tl-its-umich-edu/my-learning-analytics/commit/61a2f9990f84df279ee83d86730691851fad0be1

zqian:
@jonespm I have tested your changes, and it is working as expected. I think yours is a better approach with both statusText and statusCode handlings. Please go ahead and merge your changes instead

jonespm (approved):

Issue #1557, opened 2023-12-15
change Postgres cron queries to BigQuery quries


zqian:
### Describe your problem or feature you'd like added

MyLA [cron.py](https://github.com/tl-its-umich-edu/my-learning-analytics/blob/master/dashboard/cron.py) queries UDP context_store for Canvas data and UDP BigQuery for Caliper events. 

### Describe the solution you'd like

The UDP context_store tables can now be queried from UDP BigQuery. We can update MyLA cron.py file, and remove the SQLAlchemy and psycopg libraries.

jonespm:
If/when we go to BIgQuery we could consider switching to access the CD2 (canvas table) directly. This could more directly fix issues like #1559 where the context_store doesn't keep pseudonyms for all users. It would also bring up the question if we should use the context_store or just go to the canvas table directly? Are there advantages for MyLA to use the context_store? 

This would put us back closer to the UDW and I think eliminate some other bugs we had to try to workaround with these UDP queries. 

I think if we don't do a full switch for some tables like the users table we *should* use the data from the Canvas tables where it's possible at least replacing things like `entity.person_email`

jonespm:
@zqian I'm not sure if you started on this but I started working on this today. 

I'm wondering if there's value in leaving the `sqlalchemy` stuff in there anymore since this is all using UDP anyway. I think I can write code so both work, at least for now, and remove that later. But I'm also think we should just remove it entirely. I feel like we're in the "Unizin 100% required" phase for this project now.

zqian:
@jonespm I took my name from the assignee list. Please go ahead update the title and description of this issue, since the current design is to use Canvas Data 2 hosted in BigQuery.

zqian:
@jonespm: Can the "DATA_WAREHOUSE" settings be removed from ConfigMap?

PR #1568, opened 2024-04-01
Fixes issue #1557 Changes to support switching to bigquery. 

jonespm:
The queries currently aren't working but it feels like it might be making progress!

- [x] Update and test query `metadata`
- [x] Update and test query `user`
- [x] Update and test query `assignment_groups`
- [x] Update and test query `assignment`
- [x] Update and test query `assignment_weight`
- [x] Update and test query `term`
- [x] Update and test query `course`
- [x] Update and test query `resource`
- [x] Update and test query `submission`
- [x] Update and test query 'submission_with_avg_score' (Combined with submission)
- [x] Clean up postgres package, DATA_WAREHOUSE setting and code using it from db_util if. this is no longer necessary. 
- [x] See if we need to do anything about `timestamp without time zone` to keep dates correct
- [x] Compare old and new databases to ensure accuracy of data
- [x] Double check that start and end date are being correctly populated into courses
- [x] Remove utility function and just bind BQ parameters directly
- [x] Remove left over DATA_WAREHOUSE references


pushyamig:
The queries are Postgres Syntax so needs some tweaking to BQ format, so that is the reason might now be working. I also feel the submission query might perform bit better with BQ may be not need to create a Temp table(Needs to play around with it)

jonespm:
I put this up as draft right now because @zqian wanted to see it. No need to review any code here as it's really just a work in progress. I'm planning on keeping them in Postgres syntax and calling them using EXTERNAL_QUERY.

I think the reason it's currently not working is the parameter replacement doesn't work for EXTERNAL_QUERY in BigQuery. I've only spent about 1 day on this. 

If we rewrote it to the tables in BigQuery using the context_store_entity and context_store_keymap it would incur a query cost. I'm not sure about what the cost would be.

pushyamig:
>If we rewrote it to the tables in BigQuery using the context_store_entity and context_store_keymap it would incur a query cost. I'm not sure about what the cost would be.

From what I understood with BQ cost for running Context Store queries shouldn't be a lot. The reason as I understood from Unizin Folks is since the Storage for Context store data is not as big as BQ Event store so we don't have query Processing cost as with event store. 

>I think the reason it's currently not working is the parameter replacement doesn't work for EXTERNAL_QUERY in BigQuery. I've only spent about 1 day on this.

IMHO, I feel if we are switching to BQ backend that I feel the queries should also be BQ Fomat otherwise we are still be dealing with 2 different system(Postgres, BQ). But I fully don't understand why parameter replacement doesn't work so I am just saying what I am thinking. Since last time I did the update I was dealing with Postgres, Bigquery connector and it was not fun updating.

jonespm:
I added some more thoughts to this on Slack. I felt this was possibly the easiest way for a first pass. If we wanted to push this forward and do a rewrite to 100% BigQuery we might decide to use the canvas (CD2) tables instead of the context_store. It just feels better to only be going to one source for everything even if BigQuery is just mostly a "proxy" for the PostGres data.

jonespm:
Thanks for the feedback here. Because of complications around named parameters and the temporary table, I've decided to take your suggestion and explore rewriting this directly into BigQuery. Currently it runs as far as the assignment table so it's getting there!

pushyamig:
yes, all the tables other than Submission should be simple, just syntax change and probably some the way the variables are queried.


With submission query, with BQ we have to check if we want to create a Temp table route, I felt BQ is might be good with handling stuff. But Only way to test is in Myla Beta since it has lot of courses.

jonespm:
Doesn't seem like temp table is necessary anymore. Can just create it using with a WITH and CTE which seems to be creating a fast temporary table now. Was only necessary back in PostGres.

pushyamig:
I will review this today or tomorrow

dashboard/common/db_util.py
```diff
@@ -195,3 +196,51 @@ def get_canvas_data_date() -> Union[datetime, None]:
     except Exception:
         logger.info("Value could not be found from metadata", exc_info = True)
     return None
+
+def map_dict_to_query_job_config(param_dict: Dict[str, Any]) -> bigquery.QueryJobConfig:
```
zqian, line 200: Cannot find the use of this function in the PR, though the cron runs successfully with this PR updates.

dashboard/cron.py
```diff
@@ -676,8 +680,14 @@ def do(self) -> str:
         else:
             logger.warn("data_last_updated not updated because of an Exception during this run")
 
-        status += "End cron: " + str(datetime.now()) + "\n"
 
-        logger.info("************ total status=" + status + "\n")
+        if settings.LRS_IS_BIGQUERY:
+            total_tbytes_billed = self.total_bytes_billed / 1024 / 1024 / 1024 / 1024
+            # $6.25 per TB as of Feb 2024 https://cloud.google.com/bigquery/pricing
```
zqian, line 686: We need to watch for the cost. I have 20 MyLA courses in localhost, and the initial run with this PR cost $18, and subsequent run cost $0.1

pushyamig: I only has 4 course and costed $24 next run cost $0.1. I would expect to cost less
```
// This is the total row count
resources_access:	2115
resources:	248
submissions:	1759
assignments:	95
user:	100
course:	5
```

pushyamig:
I feel this PR should be run from MyLA Beta and see how long it take to run. since MyLA Beta has about 50+ course

dashboard/common/db_util.py
pushyamig, line 200: Just curious, why is this function needed? with postgres this was not needed.

pushyamig:
The PR seems be working from fetching results as expected. I have some minor comment. But I feel this PR should be run against MyLA Beta course and see how it performance specially with Submission Query.

dashboard/settings.py
```diff
@@ -365,7 +352,7 @@ def apply_env_overrides(env: Dict[str, Any], environ: os._Environ) -> Dict[str,
 # IMPORT LOCAL ENV
 # =====================
 try:
-    from settings_local import *
+    from settings_local import * # type: ignore
```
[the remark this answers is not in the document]
jonespm: Yeah, without this Pylance reports that this import is missing and it is. This is in a try/catch and it's expected this file might not exist. I'm not sure if we added this feature, but I just wanted to remove the error from the log and this seems like the best suggestion to turn it off here.

config/env_sample.hjson
```diff
@@ -185,26 +185,8 @@
...
-        "PASSWORD": "",
-        # database host
-        "HOST": "",
-        # database port
-        "PORT": 5432,
-        # Enable/Disable Unizin Date Warehouse specific features/data
-        "IS_UNIZIN": true
-    },
-    # Learning Record Store configuration
+    # Data Warehoue and Learning Record Store configuration
```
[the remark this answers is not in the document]
jonespm: It's not using an external query anymore, it's retrieved directly from tables in the UDP

dashboard/common/db_util.py
```diff
@@ -181,7 +182,7 @@ def get_last_cronjob_run() -> Union[datetime, None]:
     return None
 
 
-def get_canvas_data_date() -> Union[datetime, None]:
+def get_canvas_data_date() -> Union[datetime.datetime, None]:
     if not settings.DATABASES.get('DATA_WAREHOUSE', {}).get('IS_UNIZIN'):
```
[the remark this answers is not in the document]
jonespm: Yes, this should probably also just be removed as it's useless code now and will always return the default of `{}`. I also see this in the cron.py.

dashboard/cron.py
jonespm: The cost is high likely have courses without a start date, without a term or with a really old term in your database. It might also be a bug with retrieving the dates for terms which wouldn't be good. The event store is getting huge now so doing a full query is multiple TB of data. 

I think we should have a safety case in here so it doesn't retrieve more than X months of data from the `event_store` unless explicitly directed to. Generally there won't ever be a case to populate old course data.

jonespm:
> I feel this PR should be run from MyLA Beta and see how long it take to run. since MyLA Beta has about 50+ course

Do you mean to use the course-set that's on the MyLA beta server? That could be done locally by just exporting the course table. It should run about the same local or on that server and uses the same data set.

dashboard/cron.py
```diff
@@ -23,100 +23,129 @@
...
+        try:
+            with open(CRON_QUERY_FILE) as cron_query_file:
+                self.queries = hjson.load(cron_query_file)
+        except FileNotFoundError:
+            logger.error(f'Cannot find cron queries file "{CRON_QUERY_FILE}".')
 
+    def setup_bigquery(self):
+        # Instantiates a client
+        self.bigquery_client = bigquery.Client()
+#        self.bigquery_client = bigquery.Client.from_service_account_json(settings.BIGQUERY_SERVICE_ACCOUNT_JSON)
```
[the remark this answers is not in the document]
jonespm: I remember that I had this in there because I was thinking of switching the way we configure BigQuery to use an explicit JSON value in the hjson rather than the bq_cred.json file. This is probably a separate issue but something I noticed in the newer version of the library that is supported.

dashboard/common/db_util.py
jonespm: This is needed because BigQuery has special types in it's JobConfig that have to be passed explicitly to bind parameters. 

All of the bindings could have been changed and this would have been unneeded but I wasn't sure which method would have been easier. This is more compatibility for the old code. For instance I've got a call like this which isn't changed much from the old code.

```
        query_params = {
                        'course_ids': self.valid_locked_course_ids,
                        'canvas_data_id_increment': settings.CANVAS_DATA_ID_INCREMENT,
                        }
        df = self.execute_bq_query(self.queries['submission'], query_params).to_dataframe()
```

This could have been rewritten completely in BigQuery style to pass a `QueryJobConfig `object explicitly. 

```
job_config = bigquery.QueryJobConfig(
    query_parameters=[
        bigquery.ArrayQueryParameter('course_ids', 'INT64', self.valid_locked_course_ids,),
        bigquery.ScalarQueryParameter('canvas_data_id_increment', 'INT64', settings.CANVAS_DATA_ID_INCREMENT),
    ]
)

df = self.execute_bq_query(self.queries['submission'], query_params).to_dataframe()
```

Perhaps this would have just been a better idea to do this? I could still do this and remove this helper function as there's only 5 places this was done. I thought it would have been more work to change and it would have to support PostGres syntax still, but that was before this all changed fully to BigQuery. It just requires you to learn more about the special BigQuery types rather than infer them.

pushyamig:
>Do you mean to use the course-set that's on the MyLA beta server? That could be done locally by just exporting the course table. It should run about the same local or on that server and uses the same data set.

I don't think I care if it's done locally or Beta server. If you run the courses from MyLA Beta or Test (Test has more courses) and see how the cron performance in same way as current prod then that should be good.

jonespm:
> > Do you mean to use the course-set that's on the MyLA beta server? That could be done locally by just exporting the course table. It should run about the same local or on that server and uses the same data set.
> 
> I don't think I care if it's done locally or Beta server. If you run the courses from MyLA Beta or Test (Test has more courses) and see how the cron performance in same way as current prod then that should be good.

That makes sense, I was only testing it locally on a few courses but even that ran significantly faster now. I should be able to try the full list of courses.

dashboard/common/db_util.py
jonespm: I have made this change in the latest version of the code.

jonespm:
You should only test this with a current term course for now as it will pull down too much data otherwise on an empty database.

pushyamig (approved):
Nice Work @jonespm for pulling this off in short amount of time.

jonespm:
Thanks, didn't notice the checks, they were valid about unused imports so should pass now.

jonespm:
I'm not sure if I'll have time to run on the test environment today this so planning on waiting until Monday to merge and test this out then.

zqian:
cron job runs successfully on test Jenkins.

Issue #1358, opened 2022-03-19
Remove .git from git https remote


jonespm:
It looks like when you clone (or add a remote) git is adding a .git at the end. I don't think it was always like this and not sure when it happened.

https://github.com/tl-its-umich-edu/my-learning-analytics.git

In any case our code should handle removing this. Currently it will only remove it if it's a ssh source, not https. 

It'll likely be fixed here and just replace .git everywhere.

https://github.com/tl-its-umich-edu/my-learning-analytics/blob/474de8d4eb230587f826ad7d72f4a0ec80dbf687/dashboard/common/utils.py#L18

To reproduce just clone with the current remote that has a .git extension and https scheme. 

_Originally posted by @saputran in https://github.com/tl-its-umich-edu/my-learning-analytics/pull/1357#discussion_r830399788_

niurcaq:
Hey, I'm an EECS 481 student (niurcaq@umich.edu) that is willing to solve this issue! My partner (ltheresea@umich.edu) and I will try to fix this issue. I tried to recreate the issue and was able to in https, however I was also able to recreate it in ssh as well. So, I'm not sure if I'm just not recreating the issue correctly or this is problem for ssh and https.

jonespm:
Yeah, I don't 100% remember much about this issue. It looks like my feedback was just as you say that we should be doing a replacement/cleanup no matter what the format is, so probably just need to refactor out the `.replace(".git", "")` from the if statement and do it for everything.

zqian:
Tested successfully with the four test cases specified in [PR #1571](https://github.com/tl-its-umich-edu/my-learning-analytics/pull/1571)

PR #1571, opened 2024-04-22
changes to issue #1358 regarding .git extension

niurcaq:
I added a statement to replace the .git extension at the end no matter what format it is and ensures that the ssh base still converts to https. This should remove any .git extension from any format of a url when cloning or adding a remote repo. Here is our test plan:

Our test plan is to test four different urls:

- The first test is to clone the repo with a https base url with the .git extension. The expected output is the same url but without the .git extension. This is to ensure that the https base gets the .git extension removed properly with our new code.
- input: https://github.com/tl-its-umich-edu/my-learning-analytics.git
- output: https://github.com/tl-its-umich-edu/my-learning-analytics

- The second test is to clone the repo with a ssh base url with the .git extension. The expected output will return the github url as https base and will have the .git extension removed. This is to ensure that the shh base functionality still works with the new code.
- input: git@github.com:tl-its-umich-edu/my-learning-analytics.git
- output: https://github.com/tl-its-umich-edu/my-learning-analytics

- The third test is to clone the repo with https base and no .git extension. The expected output will return the input string. This is to ensure that when an https base url is cloned without an extension, it still works properly and does not have any weird behavior.
- input: https://github.com/tl-its-umich-edu/my-learning-analytics
- output: https://github.com/tl-its-umich-edu/my-learning-analytics

- The final test is to clone the repo with ssh base and no .git extension. The expected output will return the github url as https base with no other modifications. This is to ensure that the code works properly with the new code and does not have any weird behavior. 
- input: git@github.com:tl-its-umich-edu/my-learning-analytics
- output: https://github.com/tl-its-umich-edu/my-learning-analytics

### DIFF
--- a/dashboard/common/utils.py
+++ b/dashboard/common/utils.py
@@ -13,9 +13,11 @@ logger = logging.getLogger(__name__)
 def format_github_url_using_https(github_url: str):
     ssh_base = "git@"
     https_base = "https://"
-    # If the URL is formatted for SSH, convert, otherwise, do nothing
+    # If the URL is formatted for SSH, convert, otherwise, replace .git extension with ""
     if ssh_base == github_url[:len(ssh_base)]:
         github_url = github_url.replace(":", "/").replace(".git", "").replace(ssh_base, https_base)
+    else:
+        github_url = github_url.replace(".git", "")
     return github_url
 
 


dashboard/common/utils.py
```diff
@@ -13,9 +13,11 @@
 def format_github_url_using_https(github_url: str):
     ssh_base = "git@"
     https_base = "https://"
-    # If the URL is formatted for SSH, convert, otherwise, do nothing
+    # If the URL is formatted for SSH, convert, otherwise, replace .git extension with ""
```
jonespm, line 16: This seems fine. 

I _think_ you could have avoided the else and just had it like this too since it's the same in both. But this would accomplish the same and you've already tested it out. Thanks!

```
github_url = github_url.replace(".git", "")
if ssh_base == github_url[:len(ssh_base)]:
    github_url = github_url.replace(":", "/").replace(ssh_base, https_base)
```

jonespm (approved):

PR #1577, opened 2024-04-26
changes to issue 1558 regarding failed Jest test cases

binkmywink:
I was able to make all the tests work by rewriting the CSSTransform file and updating the Jest version in package.json (this can also be done by running `npm update jest`. I used babel `npm install babel-jest --save-dev` to reconfigure some files, as I believe that was causing some of the tests to fail. The main goal was to configure Jest for ECMAScript Modules, as the issue came from trying to import d3 using ECMAScript Module syntax (import * as d3 from 'd3') in the Histogram test. Using the minified version of d3 resolved this issue. In addition to updating snapshots before running backend tests, I was able to compile all the tests successfully.

### DIFF
--- a/assets/config/jest/cssTransform.js
+++ b/assets/config/jest/cssTransform.js
@@ -1,14 +1,25 @@
-'use strict'
+// // This is a custom Jest transformer turning style imports into empty objects.
+// // http://facebook.github.io/jest/docs/en/webpack.html
 
-// This is a custom Jest transformer turning style imports into empty objects.
-// http://facebook.github.io/jest/docs/en/webpack.html
+// const cssTransform = {
+//   process() {
+//     return 'module.exports = {};';
+//   },
+//   getCacheKey() {
+//     // The output is always the same.
+//     return 'cssTransform';
+//   }
+// };
 
+const babelJest = require('babel-jest');
+
+// Export the transformer function
 module.exports = {
-  process () {
-    return 'module.exports = {};'
+  process(src, filename, config, options) {
+    return babelJest.process(src, filename, config, options);
   },
-  getCacheKey () {
-    // The output is always the same.
-    return 'cssTransform'
-  }
-}
+  'type': 'module',
+  extensionsToTreatAsEsm: ['.js', '.jsx']
+};
+
+
--- a/assets/src/components/d3/createHistogram.js
+++ b/assets/src/components/d3/createHistogram.js
@@ -1,4 +1,4 @@
-import * as d3 from 'd3'
+import * as d3 from 'd3/dist/d3.min'
 import { adjustViewport } from '../../util/chart'
 import { roundToXDecimals } from '../../util/math'
 import { siteTheme } from '../../globals'
--- a/assets/src/globals.js
+++ b/assets/src/globals.js
@@ -7,7 +7,7 @@ const mylaGlobals = mylaGlobalsEl ? JSON.parse(mylaGlobalsEl.textContent) : {}
 
 let cspNonce
 const cspMetaEl = document.querySelector('meta[name="csp-nonce"]')
-if (cspMetaEl !== null & cspMetaEl.hasAttribute('content')) {
+if (cspMetaEl !== null && cspMetaEl.hasAttribute('content')) {
   cspNonce = cspMetaEl.getAttribute('content')
 }
 
--- a/package.json
+++ b/package.json
@@ -26,8 +26,21 @@
   },
   "babel": {
     "presets": [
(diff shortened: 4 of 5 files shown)


assets/src/components/d3/createHistogram.js
```diff
@@ -1,4 +1,4 @@
-import * as d3 from 'd3'
+import * as d3 from 'd3/dist/d3.min'
```
pushyamig, line 1: I don't feel we import things from `dist` folder.  why are you doing this?

package.json
```diff
@@ -26,8 +26,21 @@
   },
   "babel": {
     "presets": [
-      ["@babel/preset-env",{"targets": {"esmodules": true}}],
-      ["@babel/preset-react",{"runtime": "automatic"}]]
+      [
+        "@babel/preset-env",
+        {
```
pushyamig, line 31: I would like to keep this format same as before. What was the reason for re-formatting this?

assets/src/globals.js
```diff
@@ -7,7 +7,7 @@ const mylaGlobals = mylaGlobalsEl ? JSON.parse(mylaGlobalsEl.textContent) : {}
 
 let cspNonce
 const cspMetaEl = document.querySelector('meta[name="csp-nonce"]')
-if (cspMetaEl !== null & cspMetaEl.hasAttribute('content')) {
+if (cspMetaEl !== null && cspMetaEl.hasAttribute('content')) {
```
pushyamig, line 10: Undo the change here as I don't feel this needed any change

assets/config/jest/cssTransform.js
```diff
@@ -1,14 +1,25 @@
...
-// http://facebook.github.io/jest/docs/en/webpack.html
+// const cssTransform = {
+//   process() {
+//     return 'module.exports = {};';
+//   },
+//   getCacheKey() {
+//     // The output is always the same.
+//     return 'cssTransform';
+//   }
+// };
```
pushyamig, line 12: You should always remove commented code , if not used

pushyamig:
Can you email me or share your Uniqname? we are keeping track of students from EECS student who are contributing to our repo

assets/src/components/d3/createHistogram.js
pushyamig: None of the imports in the project we use `dist` in path while importing. This needs to be changed.

pushyamig (changes requested):
I have commented on code and I felt needs change

assets/src/components/d3/createHistogram.js
pushyamig: As per my understanding we cannot have an import from `dist` folder. That mean something in the way we process this needs to change.

assets/src/globals.js
jxiao21: According to UM-GPT the single & causes some kind of error in a test suite where something in this line `if (cspMetaEl !== null & cspMetaEl.hasAttribute('content'))` is null

pushyamig: OK, I think I missed seeing double `&&` and concluded as space change. This change is fine

assets/src/components/d3/createHistogram.js
jxiao21: Sounds good, I'll look into what could be changed

pushyamig:
@jxiao21 I suggest if you are assigned with this issue,  you either have to take these change into you branch and have initial commit from contributor as below.
1. Create diff of this PR as https://github.com/tl-its-umich-edu/my-learning-analytics/pull/1577.diff
2. Save this diff to you local machine
3. create a local branch and apply this diff `git apply 1577.diff
4. Make as first commit from this user as author `git commit --author="John Doe <john@doe.org>" -m "Fixing the JEST`. (We want to give credit to the user)

let me know if you have any question. You might have to resolve some conflict/rebase since the code needs to in sync latest from master

jxiao21:
Forgot to comment this yesterday but right now I may start it from scratch and use the PR as a reference since the current PR is outdated and I'm failing some of the tests with the current changes. Would that still be that would be done with a diff of the current PR?

pushyamig:
>Forgot to comment this yesterday but right now I may start it from scratch and use the PR as a reference since the current PR is outdated and I'm failing some of the tests with the current changes. Would that still be that would be done with a diff of the current PR?

However you want to approach is is fine. If your code changes are adding on from this contributor change. Then have a commit with her name. But that could be added once you fixed this issue (you could do something like this `git commit --amend --author="Author Name <email@address.com>" --no-edit`)

But as of now you can work on fixing the issue and not worry committing with author name. I can help with that.

PR #1581, opened 2024-05-31
I1578 having Gunicorn log the IP address

### LINKED ISSUE #1578: Log IP addresses for MyLA access events

It is recommended that Internet Protocol (IP) addresses involved in an event are included in the logging contents. Currently, the log contents consist of timestamps, user identifiers, and actions/events. 

For security logging assistance, reverence Security Log Collection, Analysis and Retention(DS-19) and/or contact [irteam-recon@umichl.edu](mailto:irteam-recon@umichl.edu). REF: Security Log Collection, Analysis, and Retention (DS-19) - [U-M General Policies DS-19](https://it.umich.edu/information-technology-policies/general-policies/DS-19)

We are not logging anything from Gunicorn. We'd probably need to create an issue and research how we'd capture request logs with correct IP addresses. 

ROHQ does logging but doesn't capture IP addresses, there's a issue out to fix that we haven't implemented. 
https://github.com/tl-its-umich-edu/remote-office-hours-queue/issues/359

pushyamig:
Fixes #1578

Logging is happening to console
Changes from this PR are added to Myla Beta, but can also be tested from locally. Just use a Proxy like ngrok or loophole
https://docs.gunicorn.org/en/stable/settings.html#access-log-format

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -6,6 +6,7 @@ on:
     branches: 
       - master
       - '[0-9][0-9][0-9][0-9].[0-9][0-9].*' # 2021.01.x
+      - 'i1578_guin_ip_log'
     tags:
       - '[0-9][0-9][0-9][0-9].[0-9][0-9].[0-9][0-9]' # 2021.01.01
  
@@ -21,7 +22,7 @@ env:
 jobs:
   build:
     # to test a feature, change the repo name to your github id
-    if: github.repository_owner == 'tl-its-umich-edu'
+    if: github.repository_owner == 'tl-its-umich-edu' || github.repository_owner == 'pushyamig'
     runs-on: ubuntu-latest
     steps:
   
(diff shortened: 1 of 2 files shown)


jonespm:
Cool that this work. I might make a suggestion onto the format so it looks closer to what is output by Django. 

* Put the time value first, it doesn't look like it matches exactly but I don't know an easy way to do that.
* Then put the IP address in brackets
* Drop the %l and %u since those seem to just dashes. I'd think the username would display but not sure why not. 
* I don't think we need the %(b)s (response size
`%(t)s [HTTP:%(s)s] [%({x-forwarded-for}i)s] "%(r)s" %(L)s
`
This should display something like this. What do you think? 

`[31/May/2024:15:26:10 -0400] [HTTP:200] [141.213.175.192] "GET /api/v1/courses/123456/get_user_default_selection/?default_type=resource HTTP/1.1" 0.013691`

pushyamig:
>This should display something like this. What do you think?

Sure, I made changes as per you suggestion. 

>I'd think the username would display but not sure why not.
I did not notice it. Since we don't have a Shib integration but via LTI so may be that info is not passed to Gunicorn. This is what I can think of.

jonespm (approved):

PR #1586, opened 2024-06-28
Fixes #1420: clean up names in models

### LINKED ISSUE #1420: Clean up redundant and confusing names in models to improve admin interface
Two issues were noticed in the models.py file that could be improved.

1. If the verbose_name matches the name of the field just remove the verbose name as it automatically matches. This follows Django's Less Code and DRY principles.

2. If the verbose name could be improved we should do that.

3. Additionally we don't use [help_text](https://docs.djangoproject.com/en/4.1/ref/models/fields/#help-text) to provide even more additional text on forms. It's possible some of these fields could have useful text. This should be figured out with @jennlove-um what these might be.

This appears under fields on the form like 
image

I'm thinking we might want to have text like 

- Name, Start/End Date: "This field will be automatically populated by cron to match Canvas but can be adjusted if desired"
- Data last updated: "This is the last time the cron was run and can be reset on the main courses page with the dropdown"

There's probably others., especially like the course view option area. 

_Originally posted by @jonespm in https://github.com/tl-its-umich-edu/my-learning-analytics/pull/1410#discussion_r961051442_

jxiao21:
Fixes #1420
Removed verbose names that automatically match and thus don't affect UI
Added help text in Name, Start/End Date, and Data Last Updated
Could use feedback on which potential verbose names could be renamed and to what


dashboard/models.py
```diff
@@ -25,9 +25,9 @@
 class AcademicTerms(models.Model):
     id = models.BigIntegerField(primary_key=True, verbose_name="Term Id")
     canvas_id = models.BigIntegerField(verbose_name="Canvas Id")
```
jonespm, line 27: This name could be Canvas Academic Term Id

dashboard/models.py
```diff
@@ -25,9 +25,9 @@
 class AcademicTerms(models.Model):
     id = models.BigIntegerField(primary_key=True, verbose_name="Term Id")
```
jonespm, line 26: This should be "Academic Term Id"

jonespm (changes requested):
These changes to the models seem good to me.

Can you provide the migration for this?

```
docker exec -it student_dashboard /bin/bash
./manage.py  makemigrations
```

Then you should have a new file there you can add to this.

jxiao21:
> These changes to the models seem good to me.
> 
> Can you provide the migration for this?
> 
> ```
> docker exec -it student_dashboard /bin/bash
> ./manage.py  makemigrations
> ```
> 
> Then you should have a new file there you can add to this.

Is that the command I need to run in the terminal?

jxiao21:
If so, this is what comes up:
`root@27c69b6c5aff:/code# `
And I have to enter something in which I'm not sure what to enter in.

jonespm:
> If so, this is what comes up: `root@27c69b6c5aff:/code# ` And I have to enter something in which I'm not sure what to enter in.

Yeah when that comes up run

`./manage.py  makemigrations
`

jonespm (approved):

PR #1600, opened 2024-07-31
Limit BQ Query

jxiao21:
Drafting so Matt can push his fixes for configurable synthetic data to this branch


config/env_sample.hjson
```diff
@@ -90,7 +90,7 @@
         "ROOT_PASSWORD": "student_dashboard_root_pw"
     },
     # Default Canvas Data id increment for course id, user id, etc
```
zqian, line 92: maybe add a comment here to explain "1000000000000" is the value used in UDP synthetic dataset, vs 17700000000000000 is a specific value for each Canvas institution. 

Something like:
```
# Default Canvas Data id increment for course id, user id, etc
# for Unizin synthetic data, the value is 1000000000000
"CANVAS_DATA_ID_INCREMENT": <institution_canvas_id>,
```

config/env_sample.hjson
```diff
@@ -304,5 +304,11 @@
     "COURSES_ENABLED": false,
 
     # Path to the hjson file contains cron queries
-    "CRON_QUERY_FILE": "config/cron_udp.hjson"
+    "CRON_QUERY_FILE": "config/cron_udp.hjson",
+
+    # Change the default Bigquery Project ID
+    "DEFAULT_PROJECT_ID": "udp-umich-prod",
+    # Change the dataset project ID where queries are run against
+    "DATASET_PROJECT_ID": "unizin-shared"
```
zqian, line 312: I will leave this line commented by default, so developer needs to enable DATASET_PROJECT_ID in purpose, to connect to Unizin synthetic data

dashboard/cron.py
```diff
@@ -681,7 +683,7 @@ def do(self) -> str:
                     exception_in_run = True
 
         logger.info("** informational")
-        status += self.update_unizin_metadata()
+        # status += self.update_unizin_metadata()
```
zqian, line 686: make sure this line will be uncommented once the metadata query is working

dashboard/settings.py
```diff
@@ -449,6 +449,12 @@ def apply_env_overrides(env: Dict[str, Any], environ: os._Environ) -> Dict[str,
 # Only need view permission for exports
 IMPORT_EXPORT_EXPORT_PERMISSION_CODE = 'view'
 
+# Change the default project ID for BigQuery if needed (This is typically the one that quotas are run against and logged into)
+DEFAULT_PROJECT_ID = ENV.get("DEFAULT_PROJECT_ID", None)
+
+# Override the default project ID for BigQuery if needed, like to unizin-shared
+DATASET_PROJECT_ID = ENV.get("DATASET_PROJECT_ID", None)
```
zqian, line 456: Should the default value be DEFAULT_PROJECT_ID, instead of None?

config/env_sample.hjson
```diff
@@ -304,5 +304,11 @@
     "COURSES_ENABLED": false,
 
     # Path to the hjson file contains cron queries
-    "CRON_QUERY_FILE": "config/cron_udp.hjson"
+    "CRON_QUERY_FILE": "config/cron_udp.hjson",
+
+    # Change the default Bigquery Project ID
+    "DEFAULT_PROJECT_ID": "udp-umich-prod",
```
zqian, line 310: I would use a placeholder value here, instead of "udp-umich-prod", like 

`"DEFAULT_PROJECT_ID": "<UDP_institution_id>",`

zqian:
Maybe it is worthwhile to add a section into this [loading_data.md](https://github.com/tl-its-umich-edu/my-learning-analytics/blob/master/docs/loading_data.md), with the direction of using Unizin synthetic dataset.

config/env_sample.hjson
jonespm: Yeah, I don't think it would be needed otherwise. Though there could be a case where a default project is used but the data sets are in another project. Like when/if we get true shared repositories. 

Anyway, this is a temporary solution and I believe we'll need a future issue to set this value on a per course level in the admin.

dashboard/cron.py
jonespm: Added this as a task on the top. For sure it needs to be fixed.

dashboard/settings.py
jonespm: Maybe, but then we might need a new variable to indicate whether or not this is running on synthetic data or not. 

Maybe it's better to make that explicit. I'm not sure yet, this still has some more work on it.

config/env_sample.hjson
jonespm: I think this was just empty before. This should all work with the service account and the regular data if it's not set and I don't think it will work yet.

PR #1610, opened 2024-08-16
Add last_accessed_date field

### LINKED ISSUE #1609: Add last_accessed_date field to the course table

When the cron runs if it has never ran before, it goes all the way back to the earliest start_date in the database. This could potentially result in a huge query if there are old courses in here. We should probably have some limit configured that could be overridden via the cron as most of the time we don't need this old data. I'm not sure how else to avoid it without changes to BigQuery like partitions, that we are unable to do.

In addition, it will process older courses that don't actually have any activity and don't need processed. We should add a "last_accessed_date" field or similar to the course table to keep track if we actually need to update it.

Add a date either dynamic or config that limits the earliest date data will be returned. This could be 4 months, 6 months a year. Just something lower than everything. We also need a way to override this on the cron if completely necessary.

This is a two part fix, and this is the first part. The second part is PR #1574.

This part will add the `last_accessed_date` field to the course table that will be used in the second part to limit the earliest date data will be returned.

We have considered only running on active courses, but in testing and on the first run it might not be known what's active or not. Once. this has run once we shouldn't have this problem again.

jxiao21:
Fixes #1609

Adds a "last_accessed_date" field in models.py and auto creates a migration for it by logging into the container with `docker exec -it student_dashboard /bin/bash` and then running `./manage.py makemigrations`


zqian:
I can verify the `last_accessed_date` field is added into the `course` table. 

However, I did not find the code update for insert datetime value into this field.

zqian (approved):

PR #1611, opened 2024-08-19
fixes #1592 Screen reader bug - Add Unique Ids to TextFields

### LINKED ISSUE #1592: Focusing on the "My minimum goal" field reads "set a goal" 10 times.
Assignment Planning view - Focusing on the "My minimum goal" field reads "set a goal" 10 times. 

This is because all the fields have the same id and and their visually associated labels the same "for" value?

All fields should have separate ids referenced by their labels in the "for" value

MyLA Technical Review 7/24 - Row 7

jaydonkrooss:
Make unique IDs in Material UI to address the labels pointing to multiple input fields. 

Test plan: Use a screen reader to hover over all input fields ("My Minimum Goal" and multiple "Grade Goal" fields). Each input should have read aloud one singular label associated with it.

### DIFF
--- a/assets/src/components/AssignmentGoalInput.js
+++ b/assets/src/components/AssignmentGoalInput.js
@@ -52,7 +52,7 @@ function AssignmentGoalInput (props) {
     <StyledGrid item>
       <StyledTextField
         error={goalGradeInternal > 100 || mathWarning || goalGradeInternal > maxPossibleGrade}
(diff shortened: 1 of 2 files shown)


jonespm:
Tested this.

Now it says
"Set Minimum Goal, Spin Button, Editable, Blank"

jonespm (approved):

PR #1614, opened 2024-08-28
#1583 cleaning up super user references

### LINKED ISSUE #1583: Provide a way to give staff users access to other courses
In Django, the staff role allows users access to admin pages and superuser gives the user all permissions.

However there's currently no way to give staff access to view courses that they aren't explicitly in. This is because rules is checking for superuser.

https://github.com/tl-its-umich-edu/my-learning-analytics/blob/d400cce524986a9127ed037ac144009869174527/dashboard/rules.py#L14

There's also some inconsistency about the naming here like in utils it says "is_superuser" is given if the user is staff. This should just be consistent to be "is_staff". Staff and superuser have different meanings.

```
dashboard/common/utils.py
87:    is_superuser = current_user.is_staff
```

I think we'd have to do a fix here to either have it so the staff user is the one with full access to every course and make the naming in the code consistent. Or we'd have to add a new explicit permission that could be checked that would give this user access to all the courses. I feel like doing staff is enough but we might have to talk about it. We could probably include this in the next release. 

---------------------------------------

Test Plan:

Test as instructor, Student all feature should work as expected
Test with Myla staff role: Can create Myla courses and access then as well
Test as Myla full Admin ( is_superuser and is_staff): Full access to whole app

pushyamig:
Fixes #1583 

----------------------------
Test Plan:
1. Test as instructor, Student all feature should work as expected
2. Test with Myla staff role: Can create Myla courses and access then as well
3. Test as Myla full Admin ( `is_superuser and is_staff`): Full access to whole app


zqian:
Logged in as a staff user: is_staff =1, is_superuser=0. Got the following error in the /admin page:

image

jonespm:
That error is expected if user is just staff role without extra permissions.

I think we should check if app works with just super user role and not staff. Though I guess we could always just have to have super user always need staff.

jonespm:
I would also rename everything in our code to either be staff or admin rather than superuser as that's kind of confusing.

I believe this idea to just give staff access to everything (instead of superuser now) except admin will work.

assets/src/containers/Course.js
```diff
@@ -61,18 +61,16 @@ function Course (props) {
...
       </WarningBanner>
     )
-  }
-  else if (error) {
+  } else if (error) {
     return (<WarningBanner />)
   }
-  
+
   if (loaded && isObjectEmpty(courseInfo)) return (<WarningBanner>My Learning Analytics is not enabled for this course.</WarningBanner>)
```
pushyamig, line 74: These are just formatting changes as the linter alerted them.

dashboard/rules.py
```diff
@@ -11,7 +11,7 @@ def is_admin(self, user):
     if self.context.get(user.id):
         return self.context.get(user.id)
 
-    result = user.is_superuser
```
pushyamig, line 14: This was needed since all the API call after the LTI checking for if the user is super_user and that's why the user something like "You are not Part this this course 12223"

pushyamig:
>I think we should check if app works with just super user role and not staff. 
The Admin Button is disabled from FE. So access is not give due to logic. I am sure the DJango Framework might give access if the user figures out password ( but we disabled the logging feature after LTI integation). Short answer is NO

But I think a Full admin access means the user need to be give `is_superuser` and 'is_staff` permission. That will be a mandate. (I was testing with `is_superuser` = 1 and `is_staff` = 0 the Django enabled login view after you click the Admin button from Frontend. But I don't think this is any useful usecase to test)

pushyamig:
I think during the LTI launch the check is looking for `is_Staff` role of the user, don't remember clearly why we made that decision. But keeping  that workflow in mind the current PR will blend in with that idea.

I cleaned up the frontend to reflect correct name consistence with backend permission.

pushyamig:
Ready for Review

zqian (approved):
The test was successful:

Staff user can log into the Django admin interface; then depending on the assigned permission, the staff user can perform proper actions accordingly.

jonespm:
Did you test step 2 that staff user can create and fully access a course if given that permission? @zqian it's not clear from the comment.

We may also want to document this in the wiki some possible use cases for this.

jonespm:
I probably won't be able to review this until Wednesday. If you want to merge on Zhen's approval and I recheck it then I notice anything missing since I wrote up this issue.

pushyamig:
>I probably won't be able to review this until Wednesday.

I will wait until you test or approves it.

jonespm:
Okay, a few things I'd update, should be quick

- [ ] Edit the configuration doc and mention that a staff user can be setup without enabling both. https://github.com/tl-its-umich-edu/my-learning-analytics/blob/master/docs/configuration.md#setting-up-an-admin-in-the-deployed-application
  - Maybe mention there that in addition to granting them staff access, someone may want to minimally add the permission "Dashboard | Course | Can view Course" to give them access to see the other courses in the system and navigate them through the Course link in the admin. 
  - Superuser is also mentioned in getting_started but I think the documentation there is fine. 
  - We also should try to rename "dashboard" in the permission to "my learning analytics" but that's a separate issue for next release.

- [ ] I think we should change the `base.html` check in the footer from `is_superuser` to `is_staff` so that staff can see the commit and branch versions. I'm not entirely sure why we protected this from everyone but lowering it to staff makes sense
- [ ] `db_util.py` has a special case for is_superuser. It looks like for the staff user if they log init prints a WARNING in the logs _Couldn't find user <user> in enrollment info". We don't need to see this warning for the staff user so can also lower this to an is_staff check here. 

Some things I observed but probably don't need to be fixed: 
- `dashboard/management/commands/createuser.py` only supports creating super user on the command line. I think this is fine to leave as-is and not change it to include a distinct staff role.

pushyamig:
@jonespm I have implemented the change based on your feedback

PR #1616, opened 2024-08-30
fixes #1594 add label names to assignment filter dropdowns

### LINKED ISSUE #1594: Add names to Filter by Status and Filter by Type fields on Assignment Planning
Filter by status field and Filter by type do not have names

Acceptable techniques: 

aria-label="the name"
aria-labelledby="id of a container with a good name"

MyLA Technical Review 7/24 - Row 9

jaydonkrooss:
Use `id` prop on MaterialUI `InputLabel` component. Results in `aria-labelledBy` properly pointing to the component with the corresponding name:

Screenshot 2024-08-30 at 11 52 23 AM

### DIFF
--- a/assets/src/components/AssignmentTable.js
+++ b/assets/src/components/AssignmentTable.js
@@ -309,7 +309,7 @@ function AssignmentTable (props) {
       <Grid container className={classes.filterArea}>
         <Grid item xs={12} sm={4}>
           <FormControl className={classes.formControl}>
-            <InputLabel>Filter by Type</InputLabel>
+            <InputLabel id='assignment-group-checkbox-label'>Filter by Type</InputLabel>
             <Select
               labelId='assignment-group-checkbox-label'
               id='assignment-group-mutiple-checkbox'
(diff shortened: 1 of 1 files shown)


assets/src/components/AssignmentTable.js
```diff
@@ -338,7 +338,7 @@ function AssignmentTable (props) {
         </Grid>
         <Grid item xs={12} sm={3}>
           <FormControl className={classes.formControl}>
-            <InputLabel>Filter by Status</InputLabel>
+            <InputLabel id='assignment-status-checkbox-label'>Filter by Status</InputLabel>
             <Select
```
pushyamig, line 342: Should both checkboxes label differentiated by type and status? Both seems to have the same Id name.

pushyamig:
I have one comment for clarity but this seem to be straightforward change

pushyamig (approved):

assets/src/components/AssignmentTable.js
jaydonkrooss: @pushyamig I think the ids are different, `assignment-group-checkbox-label` vs `assignment-status-checkbox-label`. Kind of hard to tell, but they are originally from the id's already written for the <Select> components

pushyamig: Thanks for confirming

PR #1619, opened 2024-09-06
#1597 improving aria-label function by removing tab index on progress bar

### LINKED ISSUE #1597: Aria label on focusable div is read twice
Aria label on focusable div read twice. On rows with the spinner, it is read once per tab stop. On rows without the spinner, it is read twice, once for the spinner, and once for the focussable div.

MyLA Technical Review 7/24 - row 12

pushyamig:
Fixes #1597

### DIFF
--- a/assets/src/components/ProgressBarV2.js
+++ b/assets/src/components/ProgressBarV2.js
@@ -50,7 +50,6 @@ function ProgressBarV2 (props) {
     <Root>
       {
         <div
-          tabIndex='0'
           aria-label={description !== undefined ? description : undefined}
           className={submitted && !score ? classes.ungradedBar : classes.outOfBar} // submitted assignment, and ungraded yet : otherwise
           style={{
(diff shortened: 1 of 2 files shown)


zqian:
@pushyamig is this still in draft mode?

pushyamig:
>@pushyamig is this still in draft mode?

I was checking with Gonzalo, if the change fits the issue reported. I will open up after adding details of how to test it.

assets/src/components/ProgressBarV2.js
```diff
@@ -50,7 +50,6 @@ function ProgressBarV2 (props) {
     <Root>
       {
         <div
-          tabIndex='0'
```
pushyamig, line 53: Removing the Tabindex is the suggestion from accessibility

zqian (approved):
The PR fulfilled two requested changes in #1597

PR #1621, opened 2024-09-10
#1593 making the assignment changes screen readable

### LINKED ISSUE #1593: Communicate the number of returned results when using filters on the Assignment Planning view
Filtering  on the Assignment Planning view does not communicate the number of returned results to assistive technologies.

SOP would be to have an empty ARIA live container existing on the page onload that gets updated with "N results found, etc" whenever a filtering action takes place. Since the filtering takes place as you type, best wait till the user is done typing to update the ARIA live container.

It could be argued that this information is valuable to everyone, so maybe display it. Since real estate seems tight, you can use a [MUI screen reader only class](https://mui.com/system/screen-readers/) to visually hide it

[MyLA Technical Review 7/24 - row 8](https://docs.google.com/spreadsheets/d/186xZEvxhbh55jLR_mRXT9tMrmFUlcBlUtkN5T7BuT64/edit?gid=886470260#gid=886470260)
-----------------------------------------
Testing approach for aria-live

https://github.com/user-attachments/assets/c8214887-85a8-4d3c-b420-d12bbc25c271

pushyamig:
Fixes #1593 

The issue request if based on the filter option it list the No. of assignments listed to screen readers. The info is not presented visually, but asked for adding it to screen reader

-----
Testing
whenever any filter option changes or default filtering selected, it list the no. of assignment in the `Box` component which is rendered as `div`

Screenshot 2024-09-10 at 10 00 11 AM

https://github.com/user-attachments/assets/2b2b7cb8-1a5f-4368-976c-70d9ba5def5d


assets/src/index.js
```diff
@@ -15,7 +15,7 @@ root.render(
   <Router basename='/'>
     <ApolloProvider client={client}>
       <ThemeProvider theme={siteTheme}>
-        <App user={user} gaId={gaId} cspNonce={cspNonce} oneTrustScriptDomain={oneTrustScriptDomain}/>
+        <App user={user} gaId={gaId} cspNonce={cspNonce} oneTrustScriptDomain={oneTrustScriptDomain} />
       </ThemeProvider>
     </ApolloProvider>
   </Router>
```
pushyamig, line 21: Changes here are due to linting error due to prior commits, which i am fixing it here

assets/src/containers/App.js
```diff
@@ -9,7 +9,7 @@ import useGoogleAnalytics from '@tl-its-umich-edu/react-ga-onetrust-consent'
 
 function App (props) {
   const { user, gaId, cspNonce, oneTrustScriptDomain } = props
-  useGoogleAnalytics({ googleAnalyticsId: gaId, nonce: cspNonce, oneTrustScriptDomain})
```
pushyamig, line 12: Changes here are due to linting error due to prior commits, which i am fixing it here

assets/src/components/AssignmentTable.js
```diff
@@ -368,6 +370,7 @@ function AssignmentTable (props) {
         </Grid>
       </Grid>
       <TableContainer className={classes.container} ref={tableRef}>
+        <Box sx={visuallyHidden} aria-live='polite'>{filteredAssignments.length} assignments are listed</Box>
```
zqian, line 373: I have minor suggestion here for the singular vs plural statement, like:

0 assignment is listed
1 assignment is listed
2 assignments are listed
...

zqian (approved):
Verified the fix. Added one minor suggestion.

assets/src/components/AssignmentTable.js
pushyamig: I have implemented the change here

PR #1623, opened 2024-09-13
fixes #1589: debounce and read snackbar for assignment planning

### LINKED ISSUE #1589: Snack bar fires visually whenever a value changes.
Snack bar fires visually whenever a value changes.  If I am using the arrow keys in any input[type="number"] it will fire with every increase.

Paradoxically the firings are not voiced by assistive tech (probably bc the snackbar is wrapped in a role="presentation") which is a good thing. Maybe fire onblur instead and voice it? Whatever - retain the role="presentation" if not planning to change how the snackbar fires - otherwise it would be too distracting for screen reader users.

### Steps to Reproduce :
1. Go to Assignment Planning
2. In Set Minimum Goal field, use the arrow key to raise or lower the goal grade.  The snack bar fires with each click.
3. Once a minimum goal is set, go to an assignment and use the arrow key to raise or lower the goal grade. The snack bar fires with each click. 

[MyLA Technical Review 7/24 - row 4](https://docs.google.com/spreadsheets/d/186xZEvxhbh55jLR_mRXT9tMrmFUlcBlUtkN5T7BuT64/edit?gid=886470260#gid=886470260)

jaydonkrooss:
Use the debounce package to delay opening the snackbar, also modify the snackbar component to alert screen readers instead of being hidden previously.

### DIFF
--- a/assets/src/components/AssignmentGoalInput.js
+++ b/assets/src/components/AssignmentGoalInput.js
@@ -32,7 +32,7 @@ function AssignmentGoalInput (props) {
   } = props
 
   const [goalGradeInternal, setGoalGradeInternal] = useState(goalGrade)
-  const debouncedGoalGrade = useRef(debounce(q => setGoalGrade(q), 500)).current
+  const debouncedGoalGrade = useRef(debounce(q => setGoalGrade(q), 250)).current
   const updateGoalGradeInternal = (grade) => {
     const v = { courseGoalGrade: grade }
     if (goalGrade !== '') {
--- a/assets/src/components/UserSettingSnackbar.js
+++ b/assets/src/components/UserSettingSnackbar.js
@@ -1,8 +1,9 @@
-import React, { useState, useEffect } from 'react'
+import React, { useState, useEffect, useCallback } from 'react'
 import Snackbar from '@mui/material/Snackbar'
 import IconButton from '@mui/material/IconButton'
 import CloseIcon from '@mui/icons-material/Close'
 import Slide from '@mui/material/Slide'
+import debounce from 'lodash.debounce'
 
 function SlideTransition (props) {
   return <Slide {...props} direction='up' />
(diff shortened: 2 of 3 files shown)


pushyamig:
One usecase to test is the change of events like setting a Mini course goal, individual assignment goal, lock goal etc Is being logged in the backend with `eventlog_log`and only once. Since this change is GraphiQL mutation, so we want to make that is functional as well

I did not test this, but that is the important usecase to test.

assets/src/components/UserSettingSnackbar.js
```diff
@@ -13,22 +14,27 @@ function UserSettingSnackbar (props) {
     saved,
     response,
     successMessage = 'Setting saved successfully!',
-    failureMessage = 'Setting not saved.'
+    failureMessage = 'Setting not saved.',
+    debounceAmount = 0
```
pushyamig, line 18: Can you add a comment the purpose of this variable?

assets/src/components/AssignmentGoalInput.js
```diff
@@ -32,7 +32,7 @@ function AssignmentGoalInput (props) {
   } = props
 
   const [goalGradeInternal, setGoalGradeInternal] = useState(goalGrade)
-  const debouncedGoalGrade = useRef(debounce(q => setGoalGrade(q), 500)).current
+  const debouncedGoalGrade = useRef(debounce(q => setGoalGrade(q), 250)).current
```
pushyamig, line 35: Why was the number reduced?

assets/src/containers/AssignmentPlanningV2.js
```diff
@@ -293,6 +293,7 @@ function AssignmentPlanningV2 (props) {
               <UserSettingSnackbar
                 saved={!mutationError && !mutationLoading && settingChanged}
                 response={{ default: 'success' }}
+                debounceAmount={750}
               />
```
pushyamig, line 297: This is milli second i suppose? why is this number chosen?

assets/src/components/AssignmentGoalInput.js
jaydonkrooss: When I added a long debounce to the snackbar, the duration of this save would stack on top of the duration for the  snackbar pop up. It felt quite long compared to other inputs on the page, so I reduced this debounce amount. I didn't consider if there was any drawback to lowering this, other than the visualizations and input amounts rendering slightly faster

pushyamig (approved):
I have left minor comment for documenting. The Review is pre-approved by @gsilver  and I have tested i locally. I see the logging is working as expected

assets/src/containers/AssignmentPlanningV2.js
jaydonkrooss: Not much thought went into it other than to add a buffer approximately around the duration of the other debounced input we had (originally 500ms). I ran it by @gsilver, and the 500-1000ms option seemed to be a good range. I'm also open to other guidance on this (and we can customize it later)

pushyamig: I think 750 seems fine, I just want to understand the reason.

assets/src/components/AssignmentGoalInput.js
pushyamig: If it improve overall objective of the view that it is fine

zqian:
To Pushyami's comment "One usecase to test is the change of events like setting a Mini course goal, individual assignment goal, lock goal etc Is being logged in the backend with eventlog_logand only once. Since this change GraphiQL mutation API, so we want to make that is functional as well"

I have verified the event log count is equal to the snack bar appearance. So if I updated the goal multiple times within the 750 ms, I see only one snack bar and only one new event log record.

assets/src/components/UserSettingSnackbar.js
zqian: same question. Not sure why we need this debounceAmount param.

zqian (approved):
I have tested the PR on localhost. It passes the test cases. I left a minor comment.

assets/src/components/UserSettingSnackbar.js
jaydonkrooss: I added a new comment to clarify in the new version of this file. There are other areas of the app that use this snackbar that we don't want to add a debounce to, at least for this issue. For instance, the grade filter setting for the resources accessed page triggers the snackbar. I've left it customizable for the future in case other pages would like a debounce

PR #1627, opened 2024-12-05
Fix code scanning alert no. 1: Information exposure through an exception

jonespm:
Fixes [https://github.com/tl-its-umich-edu/my-learning-analytics/security/code-scanning/1](https://github.com/tl-its-umich-edu/my-learning-analytics/security/code-scanning/1)

To fix the problem, we need to ensure that detailed error messages and stack traces are not exposed to the end user. Instead, we should log the detailed error messages on the server and return a generic error message to the user. This can be achieved by modifying the `lti_error` function to log the detailed error message and return a generic error message in the JSON response.

_Suggested fixes powered by Copilot Autofix. Review carefully before merging._

### DIFF
--- a/dashboard/lti_new.py
+++ b/dashboard/lti_new.py
@@ -49,7 +49,7 @@ def lti_error(error_message: Any) -> JsonResponse:
     :return: JsonResponse, with status 500
     """
     logger.error(f'LTI error: {error_message}')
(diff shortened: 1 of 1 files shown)


jonespm:
It was indicated that this error will not contain sensitive information and it makes resolving this issue faster if the user can provide a more specific issue.

PR #1628, opened 2024-12-11
Update django 4.2.17

pushyamig:
DependaBot is unable to create a PR, it was failing so I created this PR

https://github.com/tl-its-umich-edu/my-learning-analytics/security/dependabot/75
Looking into log did not get much, may be we need to some congifuration around the DependaBot auto-generated PR. But as of now I went with this solution.


jonespm:
Closing for #1629

PR #1636, opened 2025-01-30
#1626 fix watchman status messages

### LINKED ISSUE #1626: Fix watchman default status check to no return an error
When you call `/status` it shoudn't have any errors by default. Currently we workaround this with the the URL

`/status/?skip=watchman.checks.storage` but it would be nice it it worked. 

There's a stacktrace in the storage section and it returns an error. It would be nice to fix this so we don't have to workaround it.

```
  "storage": {
    "ok": false,
    "error": "Could not find config for 'default' in settings.STORAGES.",
    "stacktrace": "Traceback (most recent call last):\n  File \"/usr/local/lib/python3.10/site-packages/django/core/files/storage/handler.py\", line 35, in __getitem__\n    return self._storages[alias]\nKeyError:
```

I _think_ this might be fixed by just configuring a placeholder InMemoryStorage like. The better if the watchman library checked to see if this was configured or handled the exception and just didn't print anything for storage. But this should work for us

```
STORAGES = {
    'default': {
        'BACKEND': 'django.core.files.storage.InMemoryStorage',
    },
}
```

### Steps to Reproduce :
1. Startup MyLA and go to the URL for `/status` and` /status/bare_status`. You should get a 200 response and no errors.

pushyamig:
Fixes #1626 

http://localhost:5001/status/ --> `{"databases": [{"default": {"ok": true}}], "caches": [{"default": {"ok": true}}], "storage": {"ok": true}}`
http://localhost:5001/status/bare_status --> just Returns a status 200 without any o/p response

### DIFF
--- a/dashboard/settings.py
+++ b/dashboard/settings.py
@@ -175,6 +175,9 @@ STORAGES = {
     "staticfiles": {
         "BACKEND": 'whitenoise.storage.CompressedManifestStaticFilesStorage',
     },
+    'default': {
+        'BACKEND': 'django.core.files.storage.InMemoryStorage',
+    }
 }
 CONTEXT_PROCESSORS = [
     'django.contrib.auth.context_processors.auth',


jaydonkrooss:
This works locally for me

jaydonkrooss (approved):

pushyamig:
Thanks for the review @jaydonkrooss

pushyamig:
@jonespm I changes the code now and added watchman_checks https://github.com/tl-its-umich-edu/my-learning-analytics/pull/1639 Check it out.

The behavior is same as Storage options

pushyamig:
>it seems like Django cleans up any of these files objects at the end of each request.

Can you suggest how were you testing it or seeming the cleanup by DJango

jonespm:
I was just seeming that it does based on some documentation I found but I couldn't 100% confirm it. I just had a small concern about there being a memory leak if we used memory as storage in production. I think it's better if we can just avoid this.

PR #1639, opened 2025-02-04
I1626 watchman err fix

### LINKED ISSUE #1626: Fix watchman default status check to no return an error

pushyamig:
Fixes #1626 

http://localhost:5001/status/ - > `{"databases": [{"default": {"ok": true}}], "caches": [{"default": {"ok": true}}]}`
http://localhost:5001/status/bare_status: return 200 OK

### DIFF
--- a/dashboard/settings.py
+++ b/dashboard/settings.py
@@ -111,6 +111,11 @@ WATCHMAN_TOKEN_NAME = ENV.get('DJANGO_WATCHMAN_TOKEN_NAME', 'token')
 # Only report on the default database
 WATCHMAN_DATABASES = ('default',)
 
+WATCHMAN_CHECKS = (
+    "watchman.checks.caches",
+    "watchman.checks.databases",
+)
+
 # courses_enabled api
 COURSES_ENABLED = ENV.get('COURSES_ENABLED', False)
 # Fall back to the old config name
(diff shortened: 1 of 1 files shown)


jonespm (approved):
Thanks for the quick switch, I think this is the better idea.

PR #1640, opened 2025-02-13
added the lti launch time as the course last_access_date

### LINKED ISSUE #1624: Populate last_accessed_date in the course table

The last_accessed_date column was added to the courses table.  Now that the field is there, we need to populate the data. 

### Additional context
This data is needed to implement the feature to limit the cron queries to Big Query by date as described in #1574.

zqian:
Fixes #1624 

Test plan:

open the database and locate the last_access_date field of Course table 

The last_accessed_date can be updated via two routes:

Case 1: 
1. launch MyLA LTI on Canvas
2. notice the last_access_date timestamp is updated for this course

Case 2:
1. go to <myla_url>/courses/<course_id>
2. notice the last_access_date timestamp is updated for this course

### DIFF
--- a/dashboard/lti_new.py
+++ b/dashboard/lti_new.py
@@ -261,6 +261,8 @@ def extract_launch_variables_for_tool_use(request, message_launch):
 
     try:
         Course.objects.get(canvas_id=course_id)
+        # save the timestamp as the course last_accessed_date
+        Course.objects.filter(canvas_id=course_id).update(last_accessed_date=datetime.now())
     except ObjectDoesNotExist:
         if is_instructor or user_obj.is_staff:
             Course.objects.create(id=canvas_course_long_id, canvas_id=course_id, name=course_name)
--- a/dashboard/lti_new.py
+++ b/dashboard/lti_new.py
@@ -261,8 +261,6 @@ def extract_launch_variables_for_tool_use(request, message_launch):
 
     try:
         Course.objects.get(canvas_id=course_id)
-        # save the timestamp as the course last_accessed_date
-        Course.objects.filter(canvas_id=course_id).update(last_accessed_date=datetime.now())
     except ObjectDoesNotExist:
         if is_instructor or user_obj.is_staff:
             Course.objects.create(id=canvas_course_long_id, canvas_id=course_id, name=course_name)
--- a/dashboard/views.py
+++ b/dashboard/views.py
@@ -2,7 +2,7 @@ import json
 import logging
 import math
 from collections import namedtuple
-from datetime import timedelta
+from datetime import timedelta, datetime
 from json import JSONDecodeError
 
 import jsonschema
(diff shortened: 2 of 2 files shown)


dashboard/lti_new.py
```diff
@@ -261,6 +261,8 @@ def extract_launch_variables_for_tool_use(request, message_launch):
 
     try:
         Course.objects.get(canvas_id=course_id)
+        # save the timestamp as the course last_accessed_date
+        Course.objects.filter(canvas_id=course_id).update(last_accessed_date=datetime.now())
```
jonespm, line 265: Minor but this logic should *probably* be improved as it seems the `Course.objects.get` is mostly redundant here now and this new call could accomplish the same thing without it.

It looks like the only difference is the runs a SELECT and raises an exception if it doesn't exist. The second doesn't and just returns 0 if nothing was updated. It looks like Case 2 is slightly more efficient but basically should be the same in this case. 

```
# Case 1: Using .get() (raises exception if not found)
try: 
  course = Course.objects.get(canvas_id=course_id)  # Raises Course.DoesNotExist if not found
  course.last_accessed_date = datetime.now()
  course.save()  # Saves if found
except:
  # Do the exception stuff

# Case 2: Using .filter().update() (does not raise an exception)
updated_count = Course.objects.filter(canvas_id=course_id).update(last_accessed_date=datetime.now())
if updated_count == 0:
  # Do the exception stuff
```

jonespm:
Does this work if the user navigates to another site within MyLA or only on the initial LTI launch? I think coming from Canvas will be the most likely thing but I don't know if it would be best to have it in the LTI code or in the view code. 

I think if possible we should also try to backfill the dates from the eventlog_log table even if it involves a manual migration UPDATE.

I think something like

zqian:
> I think if possible we should also try to backfill the dates from the eventlog_log table even if it involves a manual migration UPDATE.

@jonespm: Are you suggesting to run a job to update the last_access_date value periodically, instead of an automate update on doing it on every LTI launch for example?

jonespm:
No, I'm suggesting that we have a migration to backfill what it can, and then update it live as you have it. I'm just not sure if the best place for it is in the LTI launch or in the view.py. If it captures all the cases in the LTI launch it seems fine but I don't think that would capture navigation the case if a student was in multiple sites and the switched to another site within the app.

dashboard/lti_new.py
zqian: Adopted the case 1 approach, and moved the logic into views.py, instead of lti_new.py, so that the course last_accessed_date is updated whenever the MyLA course page is launched.

jonespm (approved):
The change here seems fine. Did you want to skip the migration for the backfill, add it still or include on a separate issue?

zqian:
Sure, the backfill can be done, too. 

I suggest add that step to the release process, without opening another issue.

Issue #1630, opened 2024-12-13
Upgrade react-ga-onetrust-consent module


jaydonkrooss:
With the upcoming deprecation of the OneTrust consent banner by the U-Mich Privacy team, we have to update the shared module across projects by March 21. See this issue for more details: https://github.com/tl-its-umich-edu/react-ga-onetrust-consent/issues/12

The new module version does require a config change (removing the old OneTrust domain variable, as well as editing the CSP allowed scripts)

PR #1644, opened 2025-03-05
#1630 Upgrade cookie consent react module 1.2.1

jaydonkrooss:


pushyamig:
I am reviewing this now

pushyamig:
CSP variable need to be updated with `"*.umich.edu", "umich.edu"` for `SCRIPT_SRC` and `STYLE_SRC` variables. I think the sample .env.sample.hjson should reflect this.

assets/src/containers/App.js
```diff
@@ -5,11 +5,27 @@ import Course from './Course'
...
+    gaHandlers.onConsentReject
+  ) {
+    const consentParams = {
+      developmentMode: false,
+      alwaysShow: false,
+      onConsentApprove: gaHandlers.onConsentApprove,
+      onConsentReject: gaHandlers.onConsentReject,
+    };
+    umConsentInitialize(consentParams)
+  }
```
pushyamig, line 28: I see the with Loophole or ngrok the cookie doesn't set, looks like further diagnosis on this could be local thing and not non-prod/prod issue.

pushyamig:
@jonespm I think i will defer to your approval if we want to merge this and as we put this on Non-prod the cookie issue can be resolved or not.

pushyamig (approved):
We still have the issue with cookie not being set for local run, I am ok merging it if this is solved on non-prod server. I have one comment about CSP props update with banner for documentation purpose. 

This is working as expected

assets/src/containers/App.js
pushyamig: I am fine merging it now

pushyamig: If I remember correctly with oneTrust approach last time we never tested the PR from Localhost, the PR changes were set on Openshift Non-prod and we were just testing from there. So I don't know exactly why we did that last time, so to me we can merge and test it from Openshift

Issue #1646, opened 2025-03-13
Cookie Consent banner hotfix


jaydonkrooss:
Trying a new approach to get a quick fix out before the U-M privacy team deprecates OneTrust: directly call the script in javascript with the correct parameters. This will initialize google analytics without relying on the react-ga4 library.

After the team deploys a fix to the google analytics cookie, we should see the banner respect the user decision, which was not the case when using the react package with #1630.

PR #1647, opened 2025-03-13
#1646 Cookie consent banner hotfix

jaydonkrooss:


dashboard/templates/base.html
```diff
@@ -28,6 +28,34 @@
         }
     </style>
     {{ myla_globals|json_script:"myla_globals" }}
+    <script>
+        var googleAnalyticsId = JSON.parse(document.getElementById('myla_globals').textContent).google_analytics_id;
+        window.umConsentManager = {
```
jonespm, line 33: I think this script can be minimize to just be 
```
<script>
    window.umConsentManager = {
        googleAnalyticsID: "{{ myla_globals.google_analytics_id }}", // e.g. G-XXXXXXXXXX
        googleAnalyticsCustom: {
            streamConfig: {
                cookie_flags: "SameSite=None; Secure"
            }
        }
    };
</script>
```

- myla_globals should be in the Django template context so can be accessed directly
- All of the other code appears to just be the default

jonespm: Also, add in the option to this for 

`    externalLinkBlank: true
`

dashboard/templates/base.html
```diff
@@ -28,6 +28,19 @@
...
+    <script>
+        window.umConsentManager = {
+            googleAnalyticsID: "{{ myla_globals.google_analytics_id }}",
+            googleAnalyticsCustom: 
+            {
+                streamConfig: {
+                    cookie_flags: 'SameSite=None; Secure'
+                }
+            },
+            externalLinkBlank: true
```
pushyamig, line 40: From what I noticed from this `externalLinkBlank` only applied to `U-M Privacy Statement` and there is another link on the banner `Privacy Policy` which is still opened from the same window and causing `Refused to frame 'https://umich.edu/' because an ancestor violates the following Content Security Policy directive: "frame-ancestors 'self'".`

jonespm: We can disable that one with

`privacyStatement: false`

But yeah, would make sense for both of them to follow this same option.

pushyamig:
I will review this in now

jonespm (commented):
I haven't tested this on Canvas but the changes look good. Will wait to see if @pushyamig is testing and accept her approval.

pushyamig:
"PRIVACY_REDIRECT_URL": "https://umich.edu/about/privacy/"
 
  I feel this config is not needed anymore. since the URL in the banner open in new window. So I think the referance of the shouldbe removed from the `.env.sample`

pushyamig:
I think the Banner and GA tracking is working as expected with EU and NON-EU nations, I have one minor change with config change that is needed. This seems to be functional, should be ready for testing.

I have a side note, somehow i still has problem with ConsentBanner testing with chrome, I feel something with VPN and Chrome is conflicting. So I did all my testing from Edge browser.

pushyamig (approved):

PR #1652, opened 2025-05-09
1634: create dependabot yaml file for automatic grouped updates

jaydonkrooss:
I locally tested merging these commits on my own fork. You can see the results of enabling grouped updates via these pull requests: https://github.com/jaydonkrooss/my-learning-analytics/pulls

The npm and python dependencies will separately group, and only minor and patch updates are being included.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,31 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
+# https://docs.github.com/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file
+
+version: 2
+updates:
+  # Maintain dependencies for python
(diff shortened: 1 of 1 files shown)


jonespm (approved):
This looks good, we may want to move our package.json and lock files into the assets directory (which is essentially the frontend) to clean up the root which would involve a slight change here (and in the build). But this looks fine. 

Also noticed that you pushed this from your own master rather than a branch. Please "squash and merge" when committing.

jaydonkrooss:
> This looks good, we may want to move our package.json and lock files into the assets directory (which is essentially the frontend) to clean up the root which would involve a slight change here (and in the build). But this looks fine.

Created an issue to address, I agree it'd be a simple, maybe more nice-to-have change: https://github.com/tl-its-umich-edu/my-learning-analytics/issues/1653

> Also noticed that you pushed this from your own master rather than a branch. Please "squash and merge" when committing and you'll likely have to fix your master branch with the new commits.

Didn't realize it would be off my fork's master, I just went off the github UI suggestion to do a PR, will try to properly re-pull after merging.

PR #1657, opened 2025-06-19
issue_1574 added config value of MAX_ALLOWED_UPDATE_DAYS

zqian:
MAX_ALLOWED_UPDATE_DAYS setting, defaults to 180 if not set, will limit the resource access query timeframe for cron job,

### DIFF
--- a/dashboard/cron.py
+++ b/dashboard/cron.py
@@ -1,4 +1,5 @@
-from datetime import datetime
+from datetime import datetime, timedelta
+from zoneinfo import ZoneInfo
 import logging
 from collections import namedtuple
 from typing import Any, Dict, List, Optional, Union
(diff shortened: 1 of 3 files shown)


jonespm:
Hi @zqian . On the call Jennifer suggested that this be in the admin UI config. That is done by adding a value to CONSTANCE_CONFIG here and importing that module where needed. I can review that after this change is done.

zqian:
@jonespm: Great idea! I just update the config setting accordingly. Please review!

dashboard/cron.py
```diff
@@ -268,6 +270,10 @@ def update_resource_access(self):
 
         data_last_updated = Course.objects.filter(id__in=self.valid_locked_course_ids).get_data_earliest_date()
 
+        # Maximum number of days allowed for updating course access data
+        MAX_ALLOWED_UPDATE_DATE = datetime.now(ZoneInfo(settings.TIME_ZONE)) - timedelta(days=config.MAX_ALLOWED_UPDATE_DAYS)
```
jonespm, line 274: Minor but I might log when this actually hits this value something like this.

```
# Log if we're overriding the original timestamp
if data_last_updated < MAX_ALLOWED_UPDATE_DATE:
    logger.info(
        f"Overriding data_last_updated from {data_last_updated.isoformat()} "
        f"to MAX_ALLOWED_UPDATE_DATE {MAX_ALLOWED_UPDATE_DATE.isoformat()} "
        f"due to max allowed update days limit."
    )

```

jonespm (approved):
This looks good, I'll approve. I'd probably have added some logging when this was hit and also some more clarification in the text of the message to the admin but it would work without that.

Issue #1649, opened 2025-03-31
Make use of CSP nonce for GA scripts


jaydonkrooss:
After the quick fix required for Umich consent banner with google analytics (#1646), I discovered that we had to temporarily drop the use of CSP nonce to keep GA tracking functional in production. 

The CSP nonce environment variable was added in previous versions of google analytics in this application (#1371). Ideally, for security, we should utilize CSP nonce when calling any inline scripts and re-enable it in production.

Additionally, investigate whether "unsafe-inline" is still necessary for script-src, as mentioned in [this comment](https://github.com/tl-its-umich-edu/my-learning-analytics/pull/1496#discussion_r1132918338) from that original issue.

jonespm:
Also if we changed this from inline script to an having this in an external file we could probably also avoid the nonce stuff. Since `myla_globals` should be available already it should just work fine if we had this script like.

```
<script src="init-consent-manager.js"></script>
```

I think we have the `unsafe-inline` in here for script code in PyLTI. We could possibly calculate and include the SHA for that script but I'm not sure how to test it without removing that value from the CSP header and running it.

jaydonkrooss:
For this release, I'm just going to move forward with the basic addition of CSP nonce to the existing Google Analytics script, even though it has no active effect for the time being. 

I briefly looked into removing `unsafe-inline` from CSP style and script sources. Dropping it caused failures upon LTI login flow from this line in `lti_new.py`:
```
return oidc_login.enable_check_cookies().redirect(target_link_uri)
```
In this case the PyLTI1.3 library [loads HTML](https://github.com/dmitry-viskov/pylti1.3/blob/d8fa43e1ace2885e84dfa4067f0f7379aaa86826/pylti1p3/oidc_login.py#L249-L255) from [CookiesAllowedCheckPage](https://github.com/dmitry-viskov/pylti1.3/blob/master/pylti1p3/cookies_allowed_check.py#L6), which includes various inline scripts and styles. I may be mistaken but it seems we couldn't add a nonce or calculated SHA in this code. 

I'll create an issue, but it seems to me that if we want to enable stricter CSP policies my only possible solution could be to either drop the use of this `enable_check_cookies` method or replace this redirect entirely with our own written functionality.

jaydonkrooss:
[...]

PR #1664, opened 2025-07-31
#1649: add new CSP nonce config to project

jaydonkrooss:
Test plan - 
Check meta tag on the source (see if a csp-nonce has been assigned)
test google analytics starting up without error

### DIFF
--- a/dashboard/settings.py
+++ b/dashboard/settings.py
@@ -15,7 +15,7 @@ from typing import Any, Dict, Tuple, Union
 
 import hjson
 from django.core.management.utils import get_random_secret_key
-from csp.constants import SELF
+from csp.constants import SELF, NONCE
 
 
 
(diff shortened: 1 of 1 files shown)


dashboard/settings.py
```diff
@@ -420,7 +420,7 @@ def apply_env_overrides(env: Dict[str, Any], environ: os._Environ) -> Dict[str,
     CONTENT_SECURITY_POLICY = {
         "DIRECTIVES": {
             'default-src': [SELF] + csp_sources["DEFAULT"],
-            'script-src': [SELF] + csp_sources["SCRIPT"],
```
pushyamig, line 423: 'include-nonce-in': csp_directives["INCLUDE_NONCE_IN"] This line needs to be removed

pushyamig: Please also delete this in `.env.sample`

jaydonkrooss: done.

pushyamig (approved):

jaydonkrooss:
@pushyamig for your reference, I'm adding what I saw when locally trying to test the Umich consent banner code, which enables Google Analytics for users. With a VPN enabled (eg. windscribe or TunnelBear) and set to an EU country, the `umconsentmanager.js` script occasionally fails to properly call a umich location API. The problem is that the port-forwarded ngrok URL is not a valid requester for this API, so it throws a CORS error in response:

image


As I mentioned in Slack, I don't think this type of issue would cause any trouble in canvas beta, test, or production. We ought to check using a VPN though. If it does turn out to have a CORS problem, my best guess is that we'd want to contact the privacy team or whoever runs this location API to allow for our canvas prod & nonprod instances to be set to allowed requesters.

PR #1696, opened 2026-01-08
issue_1693  converted the regex pattern to a raw string by adding the r prefix, to treat backslashes as literal characters

zqian:


jonespm (approved):
Looks like the correct fix for this. Thanks!

Looks like MyLA was the only app that had this code

PR #1708, opened 2026-02-11
issue_1707 provide MariaDB version

zqian:


jonespm:
This was the larger fix to use packages directly instead of install script that we added to Instructor Tools https://github.com/tl-its-umich-edu/instructor-tools/pull/575/changes

Was going to apply to this once we worked on it again. If what you have fixes it that seems fine, but I'd minimally use an ARG for the version.

`ARG MARIADB_VERSION=11.4` and `${MARIADB_VERSION}` at the top of the file. I also would probably use 11.4 or 11.8 rather than the 12.1 since those are LTS.

PR #1764, opened 2026-07-21
Fixes #1763: run npm audit, update jest configs + snapshots

### LINKED ISSUE #1763: Fix Jest test configs
Jest tests need a small config change in order to run locally. Also update the snapshots so that new changes are properly tracked.

jaydonkrooss:
Should fix frontend tests and [drop this dependabot PR](https://github.com/tl-its-umich-edu/my-learning-analytics/pull/1756). may possibly fix some security issues in package-lock.json as well.

test plan: 1. local app smoke test, 2. run `docker exec -it webpack_watcher npm test`

NOTE: all are now successful except `createHistogram.test.js`, which has a babel configuration problem that might be a little involved to alleviate right now.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
     "react": "18.3.1",
     "react-dom": "18.3.1",
     "react-helmet": "6.1.0",
-    "react-router-dom": "6.30.3"
+    "react-router-dom": "6.30.4"
   },
   "babel": {
     "presets": [
@@ -68,7 +68,9 @@
       "<rootDir>/assets/src/**/?(*.)(spec|test).{js,jsx,ts,tsx}"
     ],
     "testEnvironment": "jsdom",
-    "testURL": "http://localhost",
+    "testEnvironmentOptions": {
+      "url": "http://localhost"
+    },
     "transform": {
       "^.+\\.(js|jsx|ts|tsx)$": "<rootDir>/node_modules/babel-jest",
       "^.+\\.css$": "<rootDir>/assets/config/jest/cssTransform.js",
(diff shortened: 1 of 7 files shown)


pushyamig:
I am reviewing this now

package.json
```diff
@@ -21,7 +21,7 @@
     "react": "18.3.1",
     "react-dom": "18.3.1",
     "react-helmet": "6.1.0",
-    "react-router-dom": "6.30.3"
+    "react-router-dom": "6.30.4"
```
pushyamig, line 24: I see a warning in devtools since there is an upgrade to router-dom

```
 React Router Future Flag Warning: React Router will begin wrapping state updates in `React.startTransition` in v7. You can use the `v7_startTransition` future flag to opt-in early. For more information, see https://reactrouter.com/v6/upgrading/future#v7_starttransition.
 React Router Future Flag Warning: Relative route resolution within Splat routes is changing in v7. You can use the `v7_relativeSplatPath` future flag to opt-in early. For more information, see https://reactrouter.com/v6/upgrading/future#v7_relativesplatpath.

```
I think this easily solvable. https://reactrouter.com/6.30.4/upgrading/future#future-flags. 
I had to do something similar to CCM as well with Dependency update.

pushyamig:
This PR is working fine, Jest test are working for AP and still one failed as noted here. The App works fine. 
if we can remove the warning that will good!

package.json
jaydonkrooss: Good catch, thanks for sharing the link - I made the change now so let me know if it looks good to you.

pushyamig (approved):
LGTM!

